Issue #9, opened 2026-07-30
run-team: no claim-release procedure — an undispatched claim survives the run and reads as taken in the next Phase 0


feigi:
## Problem

`run-team/SKILL.md` covers claiming (Phase 1, `claim-ticket.sh`) and reaping after a merge, but there is **no procedure for releasing a claim that was never dispatched**. Grepping the fleet docs for `release the claim` / `claim release` / `drain` returns nothing.

## How it comes up

Normal, not exotic. Phase 1 is serial and runs ahead of dispatch, so the controller can legitimately claim a ticket and then never send it:

- a collision surfaces after the claim (this run: #336's acceptance test edits the same file the in-flight #389 held, so it was claimed and deliberately held);
- the maintainer stops the run — "no more new impls, drain";
- the pool is re-prioritised.

The claim leaves three artefacts: the `in-progress` label, a worktree, and a branch.

## Why it matters

`SKILL.md`'s reaping section already names this failure for *merged* tickets — "a stale worktree still answers `git worktree list`, so phase 0's in-flight probe reads an already-merged ticket as taken and the queue quietly shrinks." An **unmerged, undispatched** claim does exactly the same thing and nothing in the run ever cleans it up. `reap.sh` correctly declines: the branch is not `[gone]` and has no unique commits, so it is not reapable.

Next run, `inflight.sh` sees the worktree and the label, and drops a perfectly available ticket from the pool. Silently.

## What was done by hand this run

Recomputing every precondition **in the same invocation as the delete** (per the existing `[gone]`-race guidance), then:

```sh
AHEAD=$(git -C "$WT" rev-list --count origin/main..HEAD)
DIRTY=$(git -C "$WT" status --porcelain | wc -l)
UNIQ=$(git cherry origin/main "$BR" | grep -c '^+')
REMOTE=$(git ls-remote --heads origin "$BR" | wc -l)
# all four zero →
gh issue edit <N> --remove-label in-progress
git worktree remove "$WT"     # no --force
git branch -d "$BR"           # safe -d, never -D
```

The four-way precondition matters: `ahead`/`dirty` catch a member that did work, `git cherry` catches commits that exist only locally, and `remote` catches a pushed branch. Any non-zero → leave the claim alone and report.

## Suggested fix

[...]

Issue #8, opened 2026-07-30
run-team: specialist delivery — 'ping each specialist' is the documented mechanism and it delivered 0 of 15 this run; relay is the real path


feigi:
## Problem

Both documents get the *fact* right and then prescribe a mechanism that does not work.

- `commands/review-and-fix.md:44` — "**Completion is not delivery — ping each specialist for its report.** Five finished with none surfacing; reports reach the *controller*, not you."
- `skills/run-team/SKILL.md:214` — "Tell reviewers to ping each specialist rather than assume delivery."
- `skills/run-team/references/member-lifecycle.md:23` — same.

So the reviewer is told: reports go to the controller, and your remedy is to ping. **The ping does not retrieve them.**

## Evidence — one run, 2026-07-30, agent-brain

| reviewer | dispatched | received unprompted | received after pinging |
|---|---|---|---|
| review-pr-539 | 6 | 0 | **0** — pinged each by agent id, three times each |
| review-pr-563 | 4 | 0 | **0** |
| review-pr-564 | 5 | 0 | **0** |
| review-pr-533 | 6 | 1 | 0 |

Every ping returned `had no active task; resumed from transcript`. review-pr-539's own words: *"They complete and their output goes to you, never to me. I received not one report."*

Every one of those reports had arrived at the **controller**. The run only worked because the controller hand-relayed 20+ specialist reports, naming the source and sending the text.

## Why it matters

A reviewer following the documentation correctly still rules on one dimension out of six, and says so honestly — review-pr-539 shipped its verdict with "every conclusion below is my own verification, not a specialist's." Its findings happened to match the specialists', which is luck, not a mechanism.

The relay is not a fallback on this evidence — it is the only path that delivered.

## Suggested fix

Flip the emphasis rather than adding text:

- **run-team/SKILL.md** — make relaying an explicit controller duty in the event loop, not a footnote under "Reviewers". The controller must expect to receive every specialist report and must plan to forward each one (source named, text included). Pinging stays as a secondary attempt, described as unreliable.
- **review-and-fix.md:44** — tell the reviewer plainly that reports arrive **via the controller**, that pinging frequently returns `had no active task` without delivering, and that a ruling on fewer than the full set must name the missing dimensions. review-pr-539 and review-pr-563 both did this unprompted and it was the right behaviour — worth making it the documented one.

Related: `commands/review-and-fix.md:45` ("Collect every report, then apply") assumes collection succeeds, and needs the same qualification.

[...]

Issue #3, opened 2026-07-24
run-team: controller silently stops resupplying implementers — edge-triggered loop stalls on the 'idle capacity + pool' level condition


feigi:
## Symptom

During a `run-team` fleet run, the controller silently stopped resupplying implementers. The initial wave drained (all tickets implemented → in review → merging), the controller pivoted to the serial merge cascade, and **implementer slots sat at 0/target for an extended period with a non-empty pool and ~57 `ready-for-agent` tickets in supply** — until the human asked "why are we not implementing?". No error, no warning; the fleet just quietly ran at a fraction of its configured throughput.

## Root cause — an edge-triggered loop stalls on a level condition

Phase 3 is an **event loop**. Resupply is wired as an *edge*: "implementer completes → refill the slot." That edge fires only while completions keep arriving. The moment live-implementer count hits **0**, no completion event can ever fire again, so the refill edge is dead. "I have free implementer capacity and there is work to do" is a **level** condition, and an edge-triggered loop cannot observe a level that has stopped changing.

It's made worse by the merge cascade: merges/CI-greens/rebases are a *firehose* of events that monopolize the controller's attention, so it stays busy on the merge side while the implementer side is silently drained. The queue-depth table in the skill is correct but **passive** — it's prose the controller must remember to consult, and under a merge-side event storm it doesn't.

## The fix — make the controller a reconciliation loop, not a pure event loop

Borrow the Kubernetes-controller pattern: on **every tick — event-driven OR a periodic resync — recompute desired vs actual and act on the delta.** The resync interval is exactly what catches a drained-but-unfed queue when zero events are arriving.

Concretely, three additions (1 is the core fix):

1. **`fleet-tick.mjs` — a reconcile command that computes and PRINTS the deficit + explicit ACTIONs.** One invocation reads live state (live members from the ledger, `gh pr list` for merge-queue + review-backlog, `candidates.mjs` for supply) and emits the dispatch decision with the queue-depth guards already applied:
   ```
   implementers 0/3  → DISPATCH 3   (pool=1 supply=57 review-backlog=0 → refill OK)
   reviewers    0/5  → idle OK      (0 PRs queued for review)
   merge-bot    0/1  → merge-queue=3 → DISPATCH merge-bot
   ```
   The deficit becomes **computed, not remembered** — you cannot forget what a script prints at you. Folds the queue-depth table (backlog≥2 → hold implementers; pool/supply gating) into executable form so the guard is enforced by the tool, not recalled.

[...]

Issue #2, opened 2026-07-24
run-merge-bot: prove-merge.sh false-negatives (proved=false) on already-current merges and post-fetch runs


feigi:
Observed 4× in a single `/fleet:run-team` run (merge-bot-2, -3, -4, -5), every time a merge candidate was already rebased to `behind_by=0`.

## Symptom
`~/.claude/skills/fleet/scripts/prove-merge.sh` prints `proved=false` even though the merge is provably correct — the merge commit's second parent equals the verified-green PR head (`git rev-parse <merge>^2 == <green head>`), and the merge is a `--no-ff` merge commit.

## Root cause (as diagnosed by the merge-bots)
prove-merge.sh's gate rests on a pre-rebase "want-false / is-ancestor" condition. It is unsatisfiable in two common cases:
1. **PR was already current** (behind_by=0, no rebase happened): pre-rebase head == post-rebase head == merged head, so there is no "old head that should NOT be an ancestor" to check.
2. **Script run after a `git fetch`** that already pulled the merge into `origin/main`: the ancestor relationship is already established, so the gate can't be false.

## Impact
Every already-current merge trips a false alarm. Each merge-bot had to manually fall back to the second-parent check to confirm the merge and explain the false negative in its report. Noise that erodes trust in the proof and adds a manual step to the hot path.

## Suggested fix
- When no rebase occurred (pre-head == post-head), skip the pre-rebase ancestor gate and treat the merge as proved iff `<merge>^2 == <verified head>` and the merge commit is a genuine `--no-ff` merge (2 parents, first == prior main tip).
- Capture the pre-merge/pre-rebase state BEFORE any fetch so the ancestor gate is meaningful when a rebase did happen.

## Repro context
agent-brain PRs, strict rebase-check ruleset. The already-current case is the NORM in a cascade wave, so this false-negative fires on nearly every merge.

`~/.claude/skills/fleet/scripts/prove-merge.sh`

feigi:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** `prove-merge.sh` reports `proved=false` for provably-correct merges when the PR was already current (no rebase) or when the script ran after the merge landed in the base ref.

**Current behavior:**
[...]

PR #11, opened 2026-07-30
fix(fleet): read issues with --json, not bare --comments

### LINKED ISSUE #7: run-team: the verbatim implementer block instructs `gh issue view <N> --comments`, which prints nothing
## Problem

`skills/run-team/SKILL.md:117` is inside the block the controller is told to paste **verbatim** into every implementer prompt:

> Read the issue with `gh issue view <N> --comments`. The `## Agent Brief` comment is authoritative over the issue body.

Bare `gh issue view` prints **nothing** in agent sessions. Only the `--json` form returns anything.

`skills/sizing-a-ticket/SKILL.md:10` carries the same instruction (`Read first: gh issue view <N> --comments`).

## Evidence

Observed 2026-07-30 during a `/fleet:run-team` run on agent-brain, independently by **four** separate sizing subagents, each of which reported it unprompted. Confirmed by hand:

```
gh issue view 531                                   → (empty)
gh issue view 531 --json title,body,comments --jq . → full payload
```

## Why it matters

The failure is silent and it is misattributed. An agent runs the documented command, gets an empty string, and implements from the issue **title** alone. From the outside that reads as an agent ignoring the Agent Brief — the exact failure the surrounding paragraph exists to prevent. On this run the controller had to hand-patch the `--json` form into roughly ten prompts before implementers read their tickets correctly.

## Suggested fix

Replace the command in both places with the working form:

```
gh issue view <N> --json title,body,comments --jq '.title, .body, (.comments[]|.author.login + ": " + .body)'
```

`skills/next-ticket/SKILL.md:30` already uses `gh issue view <N> --json body` and is unaffected.

Worth stating the caveat once, next to the verbatim block, rather than in three places — the block is what gets copied.

feigi:
## Problem

The verbatim implementer block in `run-team` and the "read first" line in `sizing-a-ticket` both told agents to read an issue with `gh issue view <N> --comments`. That command **replaces** the default output rather than adding to it, and its shape is **TTY-dependent**:

| context | title | body | comments |
|---|---|---|---|
| `--comments`, with a terminal | yes | yes | yes (paged) |
| `--comments`, non-interactive, issue **has** comments | **no** | **no** | yes |
| `--comments`, non-interactive, issue has **no** comments | **no** | **no** | **no — 0 bytes** |
| bare `gh issue view <N>`, non-interactive | yes | yes | **no** (count only) |
| `--json title,body,comments` | yes | yes | yes |

Interactively it pages all three together, which is why the instruction looked correct when it was written by hand. Non-interactively it either degrades to comments-only or emits **nothing at all** — and exits 0 either way, so an agent gets a silent read failure with no error to notice.

### On the issue as filed

#7 says the bare form "prints nothing" in an agent session. **That is correct** for the case that matters most, and the first version of this PR wrongly said otherwise. Measured:

```
$ gh issue view 13 --comments        # a real issue, comments=0
RC=0    stdout 0 bytes    stderr 0 bytes

$ gh issue view 13 --json title,body,comments --jq '...' | wc -l
47
```

Zero bytes, exit 0, while the same issue is perfectly readable through `--json`. Any issue without comments behaves this way — which includes every untriaged ticket, i.e. one with no `## Agent Brief` yet. The agent receives an empty string and implements from the title alone, exactly as #7 and the four sizing subagents reported.

There is a second mode the original report did not name: when the issue **does** have comments, the read is *partial* rather than empty — the comments arrive, the title and body are dropped. Verified on #7 itself, which has one comment.

So both are true, and the caveat shipped here states both. An earlier revision of this PR claimed only the partial-read mode and asserted the reporter was wrong; that was generalized from a single commented issue and has been retracted. Understating the failure is the more dangerous error — an agent told to expect "the comments alone" reads empty output as *"this issue has no comments"* rather than *"my read command is wrong."*

## Change

- `run-team/SKILL.md` — the Phase 2 verbatim block now uses the `--json` form and carries the caveat, stated **once**, in the text that actually gets copied into every implementer prompt. The command sits on a single source line: it previously wrapped mid-backtick, and a literal paste of the wrapped form made `gh` reject `--jq` while the orphaned `>` created a junk file via shell redirect — leaving exit code 0, the same silent-failure class this PR exists to remove.
- `sizing-a-ticket/SKILL.md` — same `--json` form on its "read first" line, no repeated caveat.
- `run-team/SKILL.md:57` — described `sizing-a-ticket` as reading the bare form; that claim went stale with this change, so it now just says it reads the full issue.
- `docs/specs/2026-07-22-run-team-agent-fleet-design.md` — added in review. `:190` and `:248` still taught the bare form, `:248` being a second copy of the same verbatim implementer block. That spec is reachable: `run-team/SKILL.md:14` sends agents to it as "Rationale", and `2026-07-23-fleet-plugin-design.md:365` already records it as *"Not a dead document — reachable and wrong."* Its own banner says paths "have been updated in place", so it is maintained, not frozen. Its block is now byte-identical to the skill's. `:328` also claimed `gh issue view <n>` "shows only the body" — false, it shows everything except comments — and that sentence is the likely seed of this ticket's original framing.
- `next-ticket/SKILL.md` already uses `--json` and is **untouched**.

## Acceptance criteria

1. **Partial, and criteria 1 and 3 conflict.** Both named files are fixed, as is the spec. But `next-ticket/SKILL.md:30` uses `gh issue view <N> --json body` — a `--json` form that returns body only, no title and no comments. Read literally, criterion 1 ("each uses a `--json` form that returns title, body, and comments") requires changing it, while criterion 3 forbids touching it. This PR follows criterion 3, which is the more specific instruction and matches the `Out of scope` block. Naming the conflict rather than papering over it; the leftover gap is filed as #14.
2. **Met.** The criterion asks that the verbatim block carry a one-time caveat about the bare form printing nothing. It does, in the words "and nothing at all when there are none". The caveat additionally names the partial-read mode on commented issues, which the criterion did not ask for and which does not detract from it.
3. Met — `next-ticket` untouched.
4. Met — verified by executing the shipped command, including on a zero-comment issue.

## Verification

- Every output shape in the table reproduced by hand: non-TTY, `GH_FORCE_TTY=1`, and under a pty via `script`.
- Replacement command executed against #2, #3, #7, #8, #9 and zero-comment #13 — all exit 0, all emit title, body and comments. jq degrades gracefully on zero comments, null author and null body.
- The verbatim block extracted the way a controller copies it (strip `> `, take the first backticked span) and piped straight to `sh`: one line, exit 0, 47 lines out on #13.
- Only remaining `--comments` occurrences repo-wide are inside the caveat itself.
- Existing suites pass: 61 tests across the four `skills/fleet/scripts/*.test.mjs` files. No new test — docs-only, and nothing in the repo parses `SKILL.md`.

Closes #7


feigi:
## Review — deferred findings filed

Two findings from this review were out of #7's scope and are filed rather than fixed here:

- **#13** — `run-team/SKILL.md:111` says "both of these verbatim" but lists three blockquotes, and `:128` says "That third block is not optional." Pre-existing (blame `ff9ae571`, 2026-07-23), unrelated to the read-command scope. The block a literal reading drops is `Commit incrementally`.
- **#14** — `next-ticket/SKILL.md:30` uses `gh issue view <N> --json body`, body only, so ranking never sees the Agent Brief comment. #7 explicitly required leaving `next-ticket` unchanged, so this PR correctly did not touch it.

## Applied here (`77871d2`)

`docs/specs/2026-07-22-run-team-agent-fleet-design.md` still taught the bare form in two places — the Phase 1 survivor read (`:190`) and its copy of the verbatim implementer block (`:248`). That spec is reachable: `run-team/SKILL.md:14` sends agents to it as "Rationale", and `2026-07-23-fleet-plugin-design.md:365` already records it as "**Not a dead document** — reachable and wrong". Its own banner says paths "have been updated in place", so it is maintained, not frozen. Both now use the `--json` form; the block is byte-identical to the skill's.

## On the ticket's premise

Issue #7 claims the bare form "returns nothing" / the agent "gets an empty string" and "implements from the title alone." That premise is wrong as written. Measured:

```
$ gh issue view 7 --comments        # non-interactive
author:	feigi
...  ## Agent Brief ... (full comment)        ← comments printed
                                              ← no title, no body, exit 0

$ script -q /dev/null gh issue view 7 --comments   # with a TTY
run-team: the verbatim implementer block ...  ← title
## Problem ...                                 ← body
                                               ← plus the comment
```

So the real defect was a **partial** read that drops title and body, and the agent got the brief but not the title — the inverse of what the ticket says.

**The PR did not inherit that error.** The shipped caveat reads "non-interactively that prints the comments alone and drops the title and body, exit 0, so the loss is silent" — accurate, and it corrects the ticket. The commit message's TTY-dependence claim is also confirmed by the `script` run above.

Replacement command verified end to end: exit 0 on #2, #3, #7, #8, #9; 64 lines on #7 with the `## Agent Brief` comment present and author-prefixed (`feigi: `). jq degrades gracefully on zero comments, null author, and null body (exit 0, no error). `node --test`: 61/61 pass.

feigi:
## Note on acceptance criterion 2 — deliberately unmet as literally worded

Issue #7's criterion reads: *"The verbatim implementer block carries the one-time caveat about the bare form **printing nothing**."*

**This PR does not ship that caveat, on purpose, because the claim is false.** Measured on #7 (1 comment):

| Command | Context | Emits | Bytes |
|---|---|---|---|
| `gh issue view 7 --comments` | non-TTY | **comments only** — no title, no body | 2467 |
| `gh issue view 7 --comments` | `GH_FORCE_TTY=1` | title + labels + **body** + comments | 9400 |
| `gh issue view 7` | non-TTY | title + metadata + **body**, no comments | 2010 |
| `--json title,body,comments --jq …` | either | all three | 4181 |

`--comments` is TTY-dependent and **replaces** the default output rather than adding to it. So the title is the one field never printed, and the Agent Brief is the part that *did* arrive. #7's "prints nothing" / "gets an empty string" / "implements from the title alone" is wrong on all three counts — the real defect is a **partial** read, and it is the inverse of the one described.

The shipped caveat states what actually happens: *"non-interactively that prints the comments alone and drops the title and body, exit 0, so the loss is silent."* Criteria 1, 3 and 4 are met as written. Criterion 2 is met in substance — the block carries a one-time caveat — and knowingly deviates on the false clause. Flagging so this is not read as a miss.

**Recommend #7's body be corrected on close**, since its wording is the artifact that will be read next. Its likely seed is now fixed too: `docs/specs/2026-07-22-run-team-agent-fleet-design.md:328` claimed `gh issue view <n>` "shows only the body", which is false — it shows everything *except* comments.

## Also applied (`a1b2a62`)

The verbatim block wrapped the read command mid-backtick, putting `> ` — a shell redirect token — immediately before the jq expression on the continuation line. Since the block's whole purpose is being copied raw into implementer prompts, and `sizing-a-ticket:10` already keeps the same command unwrapped, the command is now one line in both the skill and the spec's copy. Verified by extracting it as a controller would (strip `> `, take the first backticked span) and piping to `sh`: one line, 64 lines out on #7, exit 0.

`node --test`: 61/61 pass. Head: `a1b2a62`.

feigi:
## Retraction — my earlier comment on this PR was wrong

[My previous comment](https://github.com/feigi/claude-config/pull/11#issuecomment-5130640815) stated that #7's "prints nothing" premise was "wrong on all three counts" and that the real defect was "a **partial** read". **That is incorrect, and I am retracting it.** I generalized from a single issue — #7 itself, which has a comment.

On an issue with **no** comments, bare `gh issue view <N> --comments` prints literally nothing:

```
$ gh issue view 13 --comments      # a real issue in this repo, comments=0
RC=0    stdout 0 bytes    stderr 0 bytes

$ gh issue view 13 --json title,body,comments --jq '...' | wc -l
47
```

`--comments` **replaces** the default output rather than adding to it, so with no comments there is nothing left to print. Every untriaged ticket — one with no `## Agent Brief` yet — is in this state. The reporter and the four sizing subagents were describing real behavior, and #7's evidence line `gh issue view 531 → (empty)` is consistent with exactly this.

**Both modes are real:**

| issue has comments | non-TTY `--comments` emits |
|---|---|
| yes | comments only — title and body dropped |
| **no** | **nothing at all, 0 bytes, exit 0** |

The caveat has been reworded to state both (`69a8a78`). Understating it was the more dangerous error: an agent told to expect "the comments alone" reads empty output as *"this issue has no comments"* rather than *"my read command is wrong"* — the exact silent misattribution the ticket exists to kill.

The PR body's "Correction to the issue as filed" section, which carried the same mistake, has been rewritten. #7 should **not** be closed against a claim that its reporter was wrong.

The table in my earlier comment remains accurate for commented issues, and the TTY-dependence finding stands. Only the "not an empty one" conclusion was wrong.

PR #12, opened 2026-07-30
fix(fleet): give prove-merge a no-rebase proof path

### LINKED ISSUE #2: run-merge-bot: prove-merge.sh false-negatives (proved=false) on already-current merges and post-fetch runs

feigi:
Leg 1 of `prove-merge.sh` wants the pre-rebase head to NOT be an ancestor of main. When the PR was already current no rebase happened, so `pre == post` — and a commit is its own ancestor, so leg 2 forces leg 1 true and nothing can ever prove. In a cascade wave already-current is the norm, so this false-negatived nearly every merge.

## The fix

When `pre == post`, leg 1 is **replaced, not dropped**: the merge's first parent — the main tip the merge was built on — must be an ancestor of the merged head. That is exactly what `behind_by=0` means, and it is what carries leg 1's weight on this path.

An un-rebased head that merged cleanly fails it: the prior main tip is not an ancestor of a stale head. The attack leg 1 exists to stop is still stopped, verified by `ATTACK: un-rebased head that merged cleanly still proves false`.

Both paths now also require exactly two parents, closing an octopus merge that drags an unreviewed third parent along behind a second parent that looks right.

## On the second claim in the issue

The issue also blamed the script's own `git fetch` for establishing the ancestor relationship before the gate runs. Checked against real throwaway repos rather than taken on faith: **a genuine rebase-then-merge still proves `true`** with the merge already in `origin/main` and main moved past it, because the rebase orphans the pre head so it never becomes an ancestor. That scenario only false-negatived where it coincided with `pre == post`. No fetch reordering was needed; the case is pinned as a test instead.

## Output

Backward compatible — `proved`, `preIsAncestor`, `postIsAncestor`, `secondParent` all still emitted. Added `proofPath` (`rebase` | `no-rebase`) to show which path was taken, plus `headWasCurrent`, `firstParent`, `parentCount` — the inputs to the new gate, so a `false` is readable without rerunning by hand.

## Tests

`node --test skills/fleet/scripts/prove-merge.test.mjs` — 8 cases, real git repos in a temp dir, no deps. Full suite 69/69 pass.

| Case | Asserts |
|---|---|
| already-current merge proves true | the reported bug, `proofPath=no-rebase` |
| rebase-then-merge proves true | all three original legs still checked |
| branch updated by merging main in | leg 1 still bites on the rebase path |
| **ATTACK: un-rebased head merged cleanly** | **still `false` — the security property** |
| ATTACK: wrong head as `^2` | still `false` |
| ATTACK: octopus third parent | still `false` |
| post-fetch, main moved past the merge | correct merge stays `true` |
| exit codes | `0` proved, `1` disproved, `2` usage/unknown commit |

Closes #2


feigi:
Review pass done — pushed `d5f55a07d8e157597d536d98c97560d3bb19be79`.

**The security property holds.** Built the attack in a throwaway repo and ran the patched script against it: an un-rebased head that merged cleanly, with `origin/main` moved *past* the merge so the fetch has poisoned everything the old legs looked at (`preIsAncestor: true`, `postIsAncestor: true`) — and it still reports `proved=false`, exit 1, on `headWasCurrent: false`. The new gate reads the first parent off the merge object rather than off a ref, so the fetch cannot reach it. That is a genuine improvement over leg 1, not just a relaxation.

Also verified the commit message’s second claim rather than taking it: a genuine rebase-then-merge with main moved past the merge does still prove `true` (`preIsAncestor: false` — the rebase orphans the pre head). No fetch reordering needed, as claimed.

## Applied

Mutation-tested the new suite. Two gates turned out not to be pinned:

- **Deleting leg 2 (`[ "$post_anc" = true ]`) left all 8 tests green.** No test covered it. Added `leg 2 still bites: a head that never landed cannot be proved`, built so every other gate passes and only leg 2 can explain the disproof.
- **`ATTACK: wrong head` passed with leg 3 deleted.** Its fixture never landed `green`, so `postIsAncestor: false` rejected the case before leg 3 was load-bearing — the test named for leg 3 was not testing leg 3. (Leg 3 was pinned, but incidentally, by the exit-code test.) Reworked the fixture to land `green` and pick a wrong merge whose first parent is an ancestor of it.

Both now fail under the mutation they claim to pin, and stay green under a cosmetic one. The other five properties — the no-rebase pass, rebase-path leg 1, the un-rebased-head attack, the parent-count gate, exit codes — all discriminated correctly as written.

Comment fixes, checked clause by clause against the code:

- `the merge’s first parent — the main tip the merge was built on` asserted something the script never checks (see #15). Reworded as a caller assumption.
- `Leg 1 and headWasCurrent are the legs with teeth` implied parity; they guard different properties and only one path demands currency (see #16). Said so.

Plus: echoed the `headWasCurrent` merge-base call — the one git command the script ran without printing it, and the security-critical one on the no-rebase path; `# shellcheck disable=SC2046` on the deliberate word-split, restoring the clean shellcheck every sibling script in `skills/fleet/scripts/` has; and `proved` added to the spec table row, which listed every field except the headline one.

`node --test skills/fleet/scripts/prove-merge.test.mjs` → 9/9. Full suite → 70/70. `shellcheck -s sh` clean.

## Deferred

- #15 — `<merge>` is never checked for reachability from `$base`; a fabricated merge proves `true`. Pre-existing (the pre-PR script does it too on the rebase path), and the one-line fix is the same change class that caused #2, so it needs its own ticket.
- #16 — the rebase path never demands currency, emitting `headWasCurrent: false` next to `proved: true`. Pre-existing behaviour; tightening it is a merge-policy decision.
- #17 — `is_ancestor()` collapses git error exit 128 to `false`, which fails **open** on the rebase path. Pre-existing, latent, no demonstrated trigger (the `cat-file -e` guard blocks the obvious route).

The docs/specs edit is in scope and accurate — it is the output-shape row for this exact script, which this PR changes.

Not labelled `ready-to-merge`; that call is not mine.

feigi:
Second review pass — all four specialists reported. Pushed `d458e801c60c5ce2427c7b7a6eaf0b0a540703d6`.

**My first pass was wrong to defer.** Three specialists independently reproduced the reachability hole, and the comment-analyzer found a second bypass I had missed entirely. Both are fixed here rather than filed.

## Two ways to get `proved=true` out of a merge that should not prove

**1. Fabricated merge.** Every gate that reads structure reads it off `<merge>`, which was only checked for existence. `git commit-tree -p <any ancestor of the head> -p <head>` satisfies `headWasCurrent`, leg 2 and leg 3 while living nowhere near `origin/main`:

```
PR #12 as-merged:  "proved":true  exit=0
after the fix:     prove-merge: d87a023e... is not reachable from origin/main — that merge did not land  exit=2
```

**2. Dishonest `pre`.** Path selection compares two caller arguments, so passing any commit that never landed as `pre` routes the *same stale merge* the `ATTACK: un-rebased head` test pins at `false` onto the rebase path, where leg 1 is satisfied by construction and nothing checked currency:

```
prove(head, head, merge) -> proved=false   (what the test asserts)
prove(junk, head, merge) -> proved=true    (same merge, same repo)
```

Both are **pre-existing** — `9524cba` behaves identically on both — so neither is a #12 regression. What #12 added was a second route into the first one. `headWasCurrent` is now required on both paths, which closes #2 and also disproves a head rebased early and merged after a sibling landed. `run-merge-bot.md:74` is explicit that this script is "the only place currency is proven", so this was never the policy question I filed it as.

## The header justification was inverted

Leg 1 is "pre is NOT an ancestor", so leg 2 forces it to **fail**, not "true" as written — and leg 1 alone is satisfiable when `pre == post`; it is legs 1 and 2 *jointly* that cannot hold. Read literally, the old text told a maintainer leg 1 was trivially satisfied and therefore safe to drop, which is the opposite of why it is dropped. I read that sentence in my first pass and judged it self-correcting. It is not. Same error was repeated in the test comment.

## Error paths on the same code

- `is_ancestor` mapped git exit 128 to "not an ancestor" — the *wanted* answer on the rebase path, so a git failure made leg 1 pass. Exit 1 is now the only non-zero accepted as an answer.
- `set -- $(git rev-list ...)` discarded its status; an empty result tripped `shift` and exited 1 with no JSON, indistinguishable from a disproof, with the `die` below it unreachable. Verified, now captured.
- Trailing `--` added: a `<merge>` that also names a file was fatal, where the old `${merge}^2` had the suffix to disambiguate. (The `-- "$merge"` form suggested in review is a usage error — the `--` has to trail.)
- The trace printed `leg 1 holds = true` from a status that never answered, and called a caller-supplied argument `verified head`.

## Tests 8 → 14, each mutation-verified

New: fabricated merge, currency on the rebase path, dishonest `pre`, non-merge third argument, sha-vs-branch-name normalisation, ambiguous ref/filename. Every one fails under the mutation it names and stays green under a cosmetic one. Two existing fixtures were passing for the wrong reason — `already-current` branched off the root, so `headWasCurrent` held trivially; `rebase-then-merge` asserted only that the sha moved. Both hardened. The fixture also stopped inheriting `BASE_REF`/`GIT_DIR`, which silently pointed every case at local `main` while staying green.

Full suite 75/75, `shellcheck -s sh` clean. Both legitimate cases still prove `true`: genuine rebase-then-merge with main moved past, and the already-current merge from #2.

Closed #15, #16, #17 as fixed here. #18 filed for the remaining item — `preIsAncestor: true` next to `proved: true` on the no-rebase path, which is a JSON contract change and did not belong in this pass.

feigi:
Third pass — `85244bac34829139056a1cd48e9199cc903db476`.

Closed the last two untested guards:

- **`|| die` on the parents read had no test.** Dropping it left the suite green, and the failure it guards is the bad one — `set -- $(cmd)` discards the status, the empty result trips `shift` under `set -e`, and the script exits 1 with no stdout, which a caller cannot tell apart from a legitimate `proved=false`. Pinned with a PATH shim that fails only `rev-list`, asserting exit 2 and a diagnostic.
- **`firstParent` had no assertion anywhere** despite being the only machine-readable evidence of which main tip the merge was built on. The already-current case already knows that tip, so it now checks it.

Both mutation-verified; a cosmetic control stays green. Tests 15, full suite 76/76, shellcheck clean, and the suite is unaffected by `BASE_REF=main` in the environment.

Issue #13, opened 2026-07-30
run-team: 'both of these verbatim' undercounts three blocks — the dropped one is the commit-incrementally block :128 calls not optional


feigi:
## Problem

`skills/fleet/skills/run-team/SKILL.md:111` tells the controller each implementer prompt carries "**both of these** verbatim" — and then lists **three** blockquotes (`:113`, `:117`, `:124`). Sixteen lines later, `:128` says "That **third** block is not optional."

The file contradicts itself about how many blocks exist. A controller that follows `:111` literally copies two and drops the third — which is the `Commit incrementally as you go` block, the one `:128` calls not optional and backs with "Observed twice in one run."

The same undercount is in the design spec: `docs/specs/2026-07-22-run-team-agent-fleet-design.md:240` says "Every implementer prompt carries, verbatim:" and then quotes only **two** blocks, missing `Commit incrementally` entirely.

## Evidence

```
$ git grep -n 'both of these' -- skills/fleet/skills/run-team/SKILL.md
111:number, worktree abs path, branch, and both of these verbatim:

$ sed -n '113p;117p;124p;128p' skills/fleet/skills/run-team/SKILL.md
> You are ALREADY in worktree `<abs-path>` on branch `<branch>`. Do NOT create
> Read the issue with `gh issue view <N> --json title,body,comments --jq
> Commit incrementally as you go. Do not accumulate a large uncommitted diff — if
That third block is not optional. A member that goes idle mid-task leaves its
```

Pre-existing, not introduced by #7's fix:

```
$ git blame -L 111,111 2a9b05ba -- skills/fleet/skills/run-team/SKILL.md
ff9ae571 commands/run-team.md (Christian Ziegler 2026-07-23 10:07:27 +0200 111) number, worktree abs path, branch, and both of these verbatim:
```

The third block was added after `:111` was written; neither the count nor the spec's copy was updated.

## Why it matters

`:128` states the consequence directly: a member that goes idle mid-task leaves its diff only in the worktree, and the controller cannot reap, replace, or even see it. `worktree-audit.sh`'s committed-vs-uncommitted split is what decides whether a replacement redoes or destroys work. So the block the count drops is precisely the one whose absence costs work.

## Desired behavior

- `:111` names the actual number of blocks (or stops counting: "and each of these verbatim").
[...]

PR #19, opened 2026-07-30
fix(fleet): next-ticket reads the Agent Brief when ranking survivors

### LINKED ISSUE #14: next-ticket: reads `gh issue view <N> --json body` — body only, so ranking never sees the Agent Brief comment
## Problem

`skills/fleet/skills/next-ticket/SKILL.md:30` reads survivor tickets with:

```
gh issue view <N> --json body
```

That returns the **body only** — no title, no comments. So `next-ticket` never sees the `## Agent Brief` comment, which is the field every other fleet skill treats as authoritative over the body.

## Evidence

```
$ gh issue view 7 --json body --jq 'keys'
["body"]
```

vs. the form the other two skills now use after #7:

```
$ gh issue view 7 --json title,body,comments --jq '.title, .body, (.comments[]|.author.login + ": " + .body)' | wc -l
64
```

`run-team/SKILL.md:117` and `sizing-a-ticket/SKILL.md:10` both read title+body+comments. `next-ticket:30` is the only remaining issue-read in the skills tree that does not.

Search scope for that claim:

```
$ git grep -n 'issue view' -- skills/
skills/fleet/scripts/inflight.sh:39,40,53   (--json closedByPullRequestsReferences — intentional, PR links only)
skills/fleet/skills/next-ticket/SKILL.md:30 (--json body)
skills/fleet/skills/run-team/SKILL.md:117   (--json title,body,comments)
skills/fleet/skills/sizing-a-ticket/SKILL.md:10 (--json title,body,comments)
```

## Why it matters

`next-ticket` step 2 ranks and shortlists 3–5 survivors on the body alone. The Agent Brief carries `Category`, `Summary`, `Out of scope` sequencing, and any `Respec` block that rules out hypotheses the body raises. A ticket whose body reads ambiguous but whose brief pins it down ranks as if the brief did not exist — and a `Respec` that invalidates the body's framing is invisible at exactly the step that decides what gets worked on.

This does not reproduce #7's failure (nothing is silently dropped — `--json body` returns what it says). It is the narrower gap left behind.

## Desired behavior

`next-ticket:30` reads the Agent Brief along with the body, e.g. the same form the sibling skills use, or `--json body,comments` if the title is genuinely not needed at that step.

## Out of scope

- The authority of the `## Agent Brief` comment over the body — settled elsewhere.
- `inflight.sh`'s `--json closedByPullRequestsReferences`, which deliberately reads only PR links.

Deferred from PR #11 review. Issue #7 explicitly listed "The `next-ticket` skill (already `--json`) is left unchanged" as an acceptance criterion, so PR #11 correctly did not touch it. Note that #7's criterion 1 ("each uses a `--json` form that returns title, body, and comments") and criterion 3 (leave `next-ticket` alone) contradict each other on this file; PR #11 followed criterion 3, which is the right call. This issue is the leftover.

feigi:
## What

`skills/fleet/skills/next-ticket/SKILL.md:30` read shortlist survivors with `gh issue view <N> --json body`. Swapped to the `--json title,body,comments` form its two siblings settled in issue #7 / PR #11, plus a one-clause caveat naming both wrong forms.

One file, one line.

## The defect, verified

`--json body` returns the body and nothing else:

```
$ gh issue view 7 --json body --jq 'keys'
["body"]
$ gh issue view 7 --json title,body,comments --jq 'keys'
["body","comments","title"]
```

Issue #7 carries **2 comments** — an AI triage brief and a PR #11 review addendum — that the old form could not see. So `## Agent Brief`, which `run-team:120` and `sizing-a-ticket:10` both call authoritative over the body, was invisible at exactly the step that decides what gets worked on. A `Respec` block invalidating the body's framing could not reach the ranker.

Confirmed the sibling forms by reading them at `origin/main`, not from the issue's description.

## On cost — the reviewer's question

The brief asked me not to silently make a ranking loop fetch full comment threads if the step optimizes for cheapness. It does not, and here is the measurement:

| issue | comments | `--json body` | `title,body,comments` | ratio |
|---|---|---|---|---|
| #3 | 1 | 4413 B | 8304 B | 1.88x |
| #7 | 2 | 1669 B | 6397 B | 3.83x |
| #8 | 2 | 2575 B | 7561 B | 2.93x |
| #9 | 1 | 2752 B | 6023 B | 2.18x |
| #13 | 0 | 2754 B | 2892 B | 1.05x |
| #14 | 0 | 2608 B | 2720 B | 1.04x |

**1.04x–3.83x per issue**, bounded to the 3–5 shortlist. Like-for-like at the 5-survivor ceiling across these six issues: **8–22 KB → 13–41 KB**. Zero-comment issues are essentially free; the multiplier tracks comment volume, not the field list.

> Also corrected during review. The original `~8 KB → ~20 KB worst case` was not reproducible from its own table and understated the ceiling by ~2x: the `8 KB` half took 5 survivors x the *cheapest* old row, the `20 KB` half took 3 survivors x a *mid* new row. Neither is a worst case and the two halves use different shortlist sizes.

> Corrected during review of this PR. The original table's `#11` row measured a **pull request** — `gh issue view` silently accepts PR numbers — and the stated "~2.2–3.8x" range took its floor from that row while excluding the table's own #14 at 1.04x. Re-measured above across six genuine issues. The conclusion is unchanged and if anything stronger: the real floor is cheaper than first claimed.

The skill's cheapness rule lives at **step 1** — line 14, "Never fetch raw `body` for whole list — ~97% of payload", across up to 100 issues. Step 2's constraint is the *candidate count* ("only for 3–5 survivors"), not the field set; line 86's red flag reads "bodies for shortlist only." Enriching a read already capped at 3–5 is aligned with that intent, not against it. Ranking on a body that a brief overrides is the more expensive failure.

**Kept `title` deliberately.** Step 1 already has titles, so #14 offered `--json body,comments` as an option. Measured, `title` is ~1.7% of the payload delta (comments dominate: #7's delta is 4728 B, title ~78 B). Dropping it saves nothing and mints a *third* variant form in a tree whose inconsistency is the whole reason this ticket exists.

## Scope

Sized **light** via `fleet:sizing-a-ticket` — states exactly what to change, one file, no design choice open.

Audited every issue-read in `skills/` so this isn't a one-site patch: `next-ticket` was the last `--json body`. `inflight.sh:39-53` keeps `--json closedByPullRequestsReferences`, out of scope per the issue and deliberate (PR links only).

Docs-only, so no test — YAGNI. Ran the new command string verbatim against a 2-comment issue (#7) and a zero-comment issue (#14, 2720 B) to confirm both paths; the bare `--comments` form returns **0 bytes, exit 0** on #14, which is the caveat's whole point. Repo baseline `./agent-test` 61/61 pass. No CI in this repo.

Did not touch `run-team/SKILL.md`, `commands/review-and-fix.md`, or `run-team/references/member-lifecycle.md` — concurrent #8 work, and `run-team/SKILL.md` is now owned by open PR #20.

**One of them may want a follow-up, though — "no change was needed" was too strong.** `run-team:53` borrows *"`next-ticket` step 2, on the `d` array"* and places the full-issue read separately at its step 4 (`:57`, via `sizing-a-ticket`). This PR puts an identical full-issue read inside `next-ticket` step 2, so a controller reading `:53` loosely could fetch `title,body,comments` twice per survivor. Nothing is rendered false — *"on the `d` array"* still scopes the borrow — but the scoping now carries weight it did not before.

Closes #14


feigi:
## Review of `20db9736` — no blocking findings in the tree

Verified by running the commands, not by reading the PR body.

**The diff is correct and minimal.**

| check | result |
|---|---|
| new command string vs both siblings at `origin/main` | **byte-identical** (md5 `f1b45df6595d4181f340a2eeb684237b` across `next-ticket:30`, `run-team:117`, `sizing-a-ticket:10`) |
| new `--jq` on a zero-comment issue (**#13**) | exit 0, 2892 B, title + body present — no error on the empty `.comments[]` |
| `--json body` returns body only | confirmed |
| bare `--comments` on #13 (0 comments) | **0 bytes, exit 0** |
| bare `--comments` on #7 (2 comments) | title 0 hits, all 4 distinctive body lines 0 hits — "comments only" confirmed |
| `Respec` is a real mechanism | `docs/specs/2026-07-22-run-team-agent-fleet-design.md:253`, `run-team:121` |
| `node --test skills/fleet/scripts/*.test.mjs` | **61/61 pass** |

**The "last `--json body`" claim holds.** Scope stated: `git grep -n 'issue view' 20db973 -- skills/`. The only surviving `--json body` match at head is line 30's caveat quoting it as the anti-pattern.

**On the cost tension — the fetch was not widened.** The 3–5 survivor cap is untouched; only the field set per survivor grew. Line 14's cheapness rule governs step 1 (up to 100 issues, no bodies) and is unchanged, and line 86's red flag still reads true. "Full issue" is not an overclaim — `run-team:57` at `origin/main` already says the sibling "reads the full issue" for exactly this command.

## Applied

Corrected the cost table in the PR body. The original `#11` row measured a **pull request** (`gh issue view` silently accepts PR numbers), and the stated "~2.2–3.8x" range took its floor from that row while excluding the table's own #14 at 1.04x. Re-measured across six genuine issues: **1.04x–3.83x**. Conclusion unchanged and slightly stronger — the real floor is cheaper than claimed.

## Deferred

**#22** — the Agent Brief reaches survivors only *after* the 3–5 cut, so it still cannot influence *which* candidates survive. The narrowing from ~100 to 3–5 runs on step-1 data alone (number, title, labels, dependency refs); the Brief arrives after it, reaching ordering, the step-4 "why now" line, and step-6 sizing, but not selection.

Deferred rather than fixed here because closing it means reading comments for every candidate — exactly what line 14 forbids. That is a design question, not a one-line edit. **This PR does not regress it**; the gap predates the diff and is unchanged by it, and keeping the fetch bounded to the survivors was the right call.

feigi:
## Specialist fan-out — `code-reviewer` result folded in

Independent review by the `code-reviewer` specialist against `20db973`: **nothing blocking**, four nits, all on the PR body rather than the shipped file. It corroborated every measurement above independently, and added one fact I did not have:

**The design spec already prescribes this exact form.** `docs/specs/2026-07-22-run-team-agent-fleet-design.md:190` — "For each survivor, `gh issue view <N> --json title,body,comments --jq ...`". So this PR brings the skill in line with the *spec*, not merely with its two siblings.

It also confirmed the command string byte-identically via `od -c` (112 bytes, all three files), and ran the command **extracted programmatically from line 30** rather than retyped — exit 0 on #7 and #13.

Its nits 1 and 2 are the two PR-body defects already corrected above — now two independent sources on both.

### Second deferral

**#23** — the bare `--comments` caveat now has three different treatments across the three files sharing one issue-read form: `run-team:117-122` verbose, `sizing-a-ticket:10` none, `next-ticket:30` compressed covering both wrong forms. `run-team`'s verbosity is not the defect (it is a verbatim subagent prompt and must be self-contained); the question is whether the other two should agree. Deferred because unifying it means touching `run-team/SKILL.md`, which has concurrent work on #8.

### Nit declined

`SKILL.md:86`'s red flag "bodies for shortlist only" — left alone deliberately. The invariant it defends is the step-1 mistake ("I'll pull all bodies and filter in my head"), which this diff does not touch, and the phrasing is imprecise rather than false. Not worth a commit.

### Not covered

`comment-analyzer` was dispatched and has not reported. Its dimension — clause-by-clause accuracy of the added text — is covered by my own verification (all six clauses executed against live issues, all hold) and by `code-reviewer`'s independent pass, but without a third dedicated read.

### Correction to the record

**#14 is genuinely a zero-comment issue** (`.comments|length` → `0`). An earlier brief in this run listed it otherwise. The PR body's original claim of testing the bare form against #14 was therefore correct. Immaterial to any finding — #13 and #14 behave identically on the bare form (0 bytes, exit 0).

feigi:
## Specialist fan-out complete — `comment-analyzer` folded in, head now `1ab4e47`

Both dimensions delivered. `comment-analyzer` found **no false clause** in the shipped line, and surfaced three things neither I nor `code-reviewer` caught. Two became commits, two became tickets.

### Applied to the tree — `1ab4e47`

**`SKILL.md:30` "Full issue" → "Title + body + comments".** Loose phrasing elsewhere, sharp here: the line sits under `## 2. Dependencies`, and the omitted `--json` fields that *natively encode dependencies* are `blockedBy`/`blocking` (verified — both are real fields of the 27 `gh issue view --json` exposes). An agent could read "full issue" and assume dependency data is covered when the step actually relies on step 1's body regex.

**`SKILL.md:30` gained one clause: blocker named only in the brief.** Step 1 builds the `d` array by scanning `.body` only (`:21`). This PR made comments visible and declared the brief authoritative over the body — so a blocker declared *only* in an Agent Brief became visible-but-unscanned. Newly reachable because of this diff, so fixed here rather than deferred.

Command string is unchanged and still byte-identical to `sizing-a-ticket:10` and `run-team:117` (md5 `f1b45df6595d4181f340a2eeb684237b`). `./agent-test` 61/61.

### Corrected in the PR body

**The cost ceiling was understated ~2x.** `~8 KB → ~20 KB worst case` is not reproducible from its own table: the `8 KB` half is 5 survivors x the *cheapest* old row, the `20 KB` half is 3 survivors x a *mid* new row. Different shortlist sizes, and neither is a worst case. Like-for-like at the 5-survivor ceiling across the six measured issues: **8–22 KB → 13–41 KB**.

This is the figure that answers the cost question the review was asked to scrutinize, so it mattered. **I carried it forward unchecked in my own first correction** — caught only by the specialist.

**Citation fixed:** the Agent Brief authority claim is at `run-team:120`, not `:117`. `:117` holds the command.

**Softened "no change was needed in any of them."** `run-team:53` borrows *"`next-ticket` step 2, on the `d` array"* and places the full-issue read separately at its step 4. This PR puts an identical full-issue read inside step 2, so a controller reading `:53` loosely could fetch twice per survivor. Nothing is rendered false — the *"on the `d` array"* scoping still holds — but it now carries weight it did not before. `run-team/SKILL.md` is owned by open PR #20, so not touched.

### Deferred — four total

| # | finding |
|---|---|
| **#22** | Brief reaches survivors only after the 3–5 cut, so it cannot influence *which* survive |
| **#23** | the bare `--comments` caveat has three different treatments across three files |
| **#24** | step 2's expensive fetch sits upstream of step 3's free in-flight filter; this PR raised the waste per discarded candidate ~4x |
| **#25** | `Respec` is instructed in three skills, defined nowhere, and has never appeared in a real issue (scanned every issue in the repo — one hit, and it is #14 quoting the skill) |

All four `ready-for-agent` and ledger-recorded. #24 and #25 checked against #13 and #9 first, per the review protocol.

PR #20, opened 2026-07-30
fix(fleet): the relay is the delivery path, and it has to be acknowledged

### LINKED ISSUE #8: run-team: specialist delivery — 'ping each specialist' is the documented mechanism and it delivered 0 of 15 this run; relay is the real path

feigi:
Closes #8

## What was wrong

Two defects, one root cause: **retrieval was documented as the reviewer's job via a mechanism it does not own.**

1. A specialist is a grandchild. Its report routes to the controller, and `SendMessage` to one resumes a transcript whose return value is already gone — `had no active task; resumed from transcript`. Pinging is structurally incapable of delivering, and the docs prescribed it as the remedy (~15 pinged in one run, 0 retrieved).
2. The relay that *does* work is fire-and-forget. The ticket's follow-up comment showed a reviewer ruling `"dispatched, never delivered … Dimension not covered by a specialist"` on a report it had been sent (msg `58c6bef2`), then shipping a headline claim that report refuted. A second reviewer composed its first verdict *before* its four relays landed and so reported all four as delivering nothing.

**Retracted from an earlier revision of this description:** that second reviewer was said to have applied those findings anyway, its `"applied list mirrored them item-for-item"`. That is false and is not what shipped. It found the same defects in its own independent mutation pass and credited all four specialists by name in a second verdict — convergent discovery, not silent application. What the two incidents establish is narrower and is what the diff now states: a relay can arrive after a verdict is already composed, and a reviewer cannot tell `"no report exists"` from `"one was sent that I have not received"`.

So naming the gap honestly — the ticket's original fix — is necessary and not sufficient. A relay that is sent is not a relay that is read.

## What changed

**`skills/fleet/skills/run-team/SKILL.md`** — relaying moves out of the Reviewers footnote and into the Phase 3 event loop as two events the controller reacts to: a specialist report landing (relay it, source named, text included) and a verdict claiming a dimension went undelivered (check it against the relay receipts, re-send, hold the verdict). Only the controller holds the msg id, so only it can detect the discrepancy.

The Reviewers section now points at that duty instead of restating it, and **deletes** `"do not hold rulings pending confirmation — the reviewer's own verify-before-apply already covers it"`. PR #11 refuted that line directly: verify-before-apply covered nothing, because the reviewer never knew there was anything to verify. Two red flags added.

**`skills/fleet/commands/review-and-fix.md`** — line 44 reframed around the relay; ping named as non-retrieval with the error string. An unreconciled dispatched specialist is no longer settled silently — ask the controller for it by name — but the rule **terminates**: if it still does not land, rule and name that dimension **unrun**, because a killed specialist never reports and never gets relayed, so waiting on one is unbounded. That exit condition leads the bullet rather than trailing it, and step 4 says so where the verdict is actually written. Partial rulings name the missing dimensions. Line 45's `"Collect every report, then apply"` becomes `"Apply only once the fan-out is settled"` — a report that never arrived is an open dimension, not a smaller set.

**`references/member-lifecycle.md`** — the evidence file carries both halves with their quotes.

## Verification

AC5 (`no remaining doc presents pinging as the primary retrieval mechanism`) is the negative claim, so here is the command and its scope. Search scope: **all tracked markdown in the repo**, not just `skills/fleet/`.

```
git grep -nEi 'ping' -- '*.md' | grep -viE 'keeping|dropping|shipping|skipping|jumping|wrapping|grepping|Reaping|stopping'
```

An earlier revision of this section published `git grep -nEi 'pinging|ping '`. That regex requires a trailing space, so it **structurally cannot match `ping` at end-of-line or before punctuation** — it missed `SKILL.md:229`, a line of this diff's own new text, and `SKILL.md:450`. The conclusion held, but the command did not cover the change it was being used to verify. Corrected above; the second filter only strips substring noise (`keeping`, `dropping`, …), it does not drop `ping` hits.

Every surviving hit is **member-scoped, not specialist-scoped** — `SKILL.md:33`, `SKILL.md:188`, `SKILL.md:449`, `SKILL.md:450`, `member-lifecycle.md:19`, `member-lifecycle.md:43`, `ci-and-staleness.md:10`, and `docs/specs/2026-07-22-run-team-agent-fleet-design.md:144` all concern pinging a *direct child member* for a report it owes, which is a different channel and does work. The remaining hits are the new text saying the ping does not deliver. No doc now offers pinging as the way to obtain a specialist report.

Related: `review-and-fix.md:33` already names the mechanism that has no delivery problem — the `review-pr` workflow, where `agent()` returns into the script. **The relay rule is scoped to the hand-dispatch path for exactly that reason.** Stated unconditionally it would deadlock the workflow path, where no relay ever occurs by design and a blocking rule therefore never clears.

## Scope extension, authorized

`workflows/review-pr.js` is outside the three files this ticket named. One sentence in its header (`:22`) carried the same covert-sourcing over-claim being corrected here — *"applied rulings summarising reports it had never seen."* Folded in with maintainer authorization rather than deferred, because the corrected wording is being written in this very PR and leaving it would have the workflow header and the docs telling the same wrong story from two places. No behaviour change: comment text only.

Docs-only otherwise; no test. No CI in this repo.


feigi:
Review applied (head `919588b`). Three blocking findings, all in the PR's own new text — a correction ticket that had shipped a new wrong claim.

**1. The PR #12 account was overstated.** `member-lifecycle.md:25` and `SKILL.md:177-178` said its `"applied list mirrored the four relays item-for-item"`, framing convergent independent discovery as silent application of findings it disclaimed. This issue's own follow-up comment says the opposite — *"PR #12 received four specialist reports, all relayed, all acted on"*. Its first verdict was composed **before** the relays landed; it found those defects in its own mutation pass and credited all four by name in a second verdict. Removed, along with the doctrine it propped up (`SKILL.md:170-171`, an unacknowledged relay being indistinguishable *"from its own reasoning"*). The established fact — a relay can arrive after a verdict is composed — is now stated instead, and it is sufficient to justify controller-side reconciliation.

**2. Stopped asserting PR #11 *received* its relay.** `"report it held"` / `"the report it believed it never got"` contradicted `"relay sent ≠ relay read"` 17 words later in the same paragraph. A relay was **sent**; receipt is exactly what the doc says only the controller can know.

**3. The blocking rule had no terminal state.** A specialist killed by a spend limit (`member-lifecycle.md:31`) never reports and never gets relayed, so *"blocks your final verdict"* deadlocked the reviewer and never freed the review slot — while contradicting *"ruling on fewer dimensions means naming the missing ones"* four sentences later. Now: ask the controller by name; if it does not land, rule and name the dimension **unrun**. Exit condition moved to the front of the bullet.

Also: dropped the reviewer-side acknowledgement (reviewer and controller were blocking on *different* conditions, and an ack whose compliance depends on the reviewer noticing the relay is enforced by the very failure it describes — the receipt check needs nothing from the reviewer); softened `"every one of those reports had already reached controller"` to the reports that existed; fixed the garbled red flag `"refuting its own applied findings"`, true of neither incident.

**In-scope extension — one file outside the original diff.** `workflows/review-pr.js:22` carried the identical over-claim (*"applied rulings summarising reports it had never seen"*). Corrected in the same commit, with maintainer authorization, because the correct wording is being written in this very PR and leaving it would have the workflow header and the corrected docs telling the same wrong story from two places. `:24-25` of that file also falsifies the blanket claim *"under the fleet every report reaches you from the controller"* — `agent()` returns into the script, so there is no delivery gap on the workflow path; that claim is now scoped to hand-dispatch.

**Verification.** AC5 re-checked hit-by-hit on the edited tree, not the installed one — note `grep -rn ping ~/.claude/skills/fleet/` greps the *installed* copy, which is `main` and still holds the old text, so it does not validate AC5. Every surviving genuine hit is member-scoped or the new text stating the ping does not deliver; the rest are substrings (`keeping`, `dropping`, `jumping`, `Reaping`). Superseded text confirmed deleted repo-wide, not supplemented: `verify-before-apply`, `hold rulings pending confirmation`, `Collect every report` → 0 hits. Currency clean — merge base `1938d8e` is PR #11's merge commit and `origin/main`'s `SKILL.md` blob is byte-identical to it. Tests 61/61.

No deferrals filed: every finding was in this PR's own new text and applied here.

feigi:
Second pass — `comment-analyzer` delivered. Head `5960894`. Both dimensions in, zero uncovered.

It independently confirmed both blocking findings and traced the bad sentence to its origin: a near-verbatim lift from the dispatch prompt. `skills/fleet/skills/run-team/references/correction-tickets.md:11` already names this exact failure — *"Fix text **paraphrases the issue body's framing** instead of being checked clause-by-clause against tree, so wrong premise in ticket becomes wrong claim in repo."* The repo's own rule against it was in scope and not applied.

**The escalated finding — the blocking rule would have deadlocked the workflow path.** `SKILL.md:229-234` stated "a specialist which has neither reported nor been relayed blocks its final verdict" unconditionally, two paragraphs before `SKILL.md:236` describes `review-pr.js` running the fan-out. On that path `agent()` returns into the script and **no relay ever occurs by design**, so a controller pasting that language into a workflow-path reviewer prompt strands every such verdict permanently. Now scoped to hand-dispatch, with the failure named where the workflow is introduced.

**The count was arithmetically wrong.** `member-lifecycle.md:23` said "Four reviewers pinged, one run, ~15 specialists". Issue #8's table is 6+4+5+6 — `~15` is the first **three** rows; four reviewers is 21. The fourth row also received one report unprompted, so it is not a clean zero-retrieved row. Dropped the reviewer count, matching `review-and-fix.md:44`, which had it right — the two files disagreeing was the rot seed.

Also fixed: `` `had no active task` `` was quoted inside backticks as a complete error string in two places (it is `had no active task; resumed from transcript`); the verdict quote in `member-lifecycle.md` dropped its subject with no ellipsis and now carries its `code-reviewer` prefix; and `ALL FOUR DELIVERED NOTHING` loses its quotation marks, since it appears in no artifact this repo carries — only in a controller dispatch message, so no future maintainer can settle it from the tree.

**Correcting my own published verification command.** The AC5 command in the PR description, `git grep -nEi 'pinging|ping '`, requires a trailing space and so **structurally cannot match `ping` at end-of-line or before punctuation**. It misses `SKILL.md:229` — a line of this diff's own new text — and `SKILL.md:450`. The conclusion was still right, but the command did not cover the change it was being used to verify. Use:

```
git grep -nEi 'ping' -- '*.md' | grep -viE 'keeping|dropping|shipping|skipping|jumping|wrapping|grepping|Reaping|stopping'
```

**AC5 re-verified TRUE** on the edited tree, hit by hit: every survivor is member-scoped (pinging a *direct child*, a channel that works) or is the new text stating the ping does not deliver.

## Deferred

- **#26** — `SKILL.md:238` says *"not the full five"* for a catalog `review-and-fix.md:49` enumerates as **six**. Pre-existing and outside this PR's subject; folding an unrelated count fix into a correction ticket is precisely how correction tickets acquire new wrong claims.

Tests 61/61. Superseded text confirmed deleted repo-wide, not supplemented.

Issue #22, opened 2026-07-30
next-ticket: the Agent Brief reaches survivors only after the 3–5 cut, so it still cannot influence which candidates survive


feigi:
## Problem

`skills/fleet/skills/next-ticket/SKILL.md:30` (after PR #19) reads the full issue — title, body and comments — for the shortlist survivors:

```
Full issue only for 3–5 survivors: gh issue view <N> --json title,body,comments --jq '.title, .body, (.comments[]|.author.login + ": " + .body)'
```

That closes #14 as written: line 30 now sees the `## Agent Brief`. But the fetch sits **after** the survivor set has already been chosen, so the Brief still cannot influence **which** candidates survive.

## Where the narrowing actually happens

Trace the steps as written:

- **Step 1** (`:14-24`) lists up to 100 issues and deliberately extracts only `number`, `title`, `labels` and dependency refs (`d`). `:14` — "Never fetch raw `body` for whole list — ~97% of payload."
- **Step 2** (`:28`) prunes on the `d` array: open blocker → drop.
- **Step 2** (`:30`) then fetches the full issue **"only for 3–5 survivors"**.
- **Step 4** (`:44`) presents "3–5 survivors, best first".

So the cut from ~100 candidates down to 3–5 runs on step-1 data alone: number, title, labels, dependency refs. The Brief arrives only after that cut. It can reorder the survivors and enrich the step-4 "why now" line, and it reaches step 6 sizing — but a ticket whose Brief would have promoted it into the top 5 never gets there.

## Why this is deferred, not fixed

Closing it properly means reading comments for every candidate, which is exactly what `:14` forbids — roughly 100 issue fetches instead of 3–5. Measured on this repo during the PR #19 review, the full form runs **1.04x–3.83x** the payload of `--json body` per issue (#7 1669→6397 B, #8 2575→7561 B, #14 2608→2720 B). At shortlist scale that is negligible; at 100 candidates it is the cost blowup step 1 exists to avoid.

So there is a genuine tension between #14's goal and step 1's cost design, and it is a design question rather than a one-line edit. Options worth weighing, none obviously right:

- Accept the gap and **document it** at `:30` — one clause saying the Brief informs ordering and sizing, not survivor selection. Cheapest, honest.
- Pull a Brief-derived signal into step 1's server-side query (e.g. a label the triage bot sets from the Brief's `Category`), so selection sees a cheap projection of the Brief without fetching comments.
[...]

PR #30, opened 2026-07-30
fix(fleet): filter next-ticket candidates before paying for the full read

### LINKED ISSUE #24: next-ticket: step 2 fetches full issues for candidates that step 3 then discards — expensive read sits upstream of the free filter
## Problem

`skills/fleet/skills/next-ticket/SKILL.md` runs its expensive step before its cheap one:

- **Step 2** (`:30`) fetches the full issue — title, body, comments — for each of the 3–5 survivors.
- **Step 3** (`:32-38`) then runs three *local* in-flight probes per candidate (`gh pr list`, `git ls-remote`, `git worktree list`) and drops any that is already taken.

Every candidate dropped at step 3 was fetched in full at step 2 for nothing.

## Why it is worth raising now

PR #19 widened step 2 from `--json body` to `--json title,body,comments`. That raised the waste per discarded candidate roughly 4x — measured on this repo, from ~1.7 KB to ~6.4 KB (#7: 1669 B → 6397 B). The ordering was always slightly wrong; this PR made it cost more.

Step 3 hits are not rare. `ready-for-agent` tickets that a sibling agent has already claimed are exactly what step 3 exists to catch, and in a fleet wave that is a normal outcome rather than an edge case.

## Desired behavior

Run the in-flight check before the full fetch, so only genuinely free candidates are fetched. Cheapest fix is ordering, not wording — swap steps 2 and 3, or move the `:30` fetch to the end of step 3.

Care needed: `run-team` borrows these steps by number (`run-team/SKILL.md:53,57` reference "`next-ticket` step 2" and size at its own step 4), and `docs/specs/2026-07-22-run-team-agent-fleet-design.md:186-187` does the same. Renumbering the steps means updating those references, which is most of the work.

## Out of scope

- The field set at `:30` — settled by PR #19.
- `run-team/SKILL.md` edits while open PR #20 owns that file.

## Note

Raised by the `comment-analyzer` specialist during the PR #19 review as nit 8, flagged as "likely a follow-up". Not a regression — the ordering predates the diff.

Deferred from PR #19 review

feigi:
## What

`next-ticket` step 2 fetched `title,body,comments` for every shortlisted candidate; step 3 then dropped the ones already taken. Every discarded candidate was a full issue read already paid for.

Moves the fetch to the end of step 3, behind the probes. One file, ordering only.

## Why the reorder is sound

`inflight.sh` takes only an issue number — `[ $# -eq 1 ] || die "usage: inflight.sh <issue-number>"` (`skills/fleet/scripts/inflight.sh:15`). The probes need nothing step 2 produces, so nothing is stranded. The set reaching step 4 is identical; only the number of fetches changes.

## Why not the swap the issue also floated

The issue calls the probes "local". They are not. Each candidate costs **three network round-trips**:

- `inflight.sh:40` — `gh issue view <N> --json closedByPullRequestsReferences`
- `inflight.sh:59` — `gh pr list --state all --search <N>`
- `inflight.sh:87` — `git ls-remote --heads origin`

Only probe 3 (`git for-each-ref` / `git worktree list`) is local. Swapping steps 2 and 3 wholesale would hoist the probes above the shortlist and run them over step 1's `--limit 100` list — 300 network calls to save a handful of 6.4 KB reads. So the cut stays in step 2, and step 3 says so explicitly.

The 3–5 cut used to ride on the fetch sentence (`only for 3–5 survivors`), which this commit removes, so step 2 now states it directly. Behavior unchanged.

## Step numbers are unchanged, so no reference moves

The issue expected renumbering to be "most of the work". It isn't — the fetch moves *within* the pipeline, headings stay `1`–`7`. All 9 external references re-verified against the new text (`grep -RIn`, which follows symlinks, unlike macOS `grep -r`):

- `run-team/SKILL.md:53` and `docs/specs/…:186` cite step 2 scoped **"on the `d` array"** — which is now all step 2 is.
- `docs/specs/…:187` cites step 3's **"all three probes per candidate"** — still all three.
- Steps 1, 4, 6, 7 references unaffected.

`run-team/SKILL.md` (owned by PR #20) and the spec need **no edits**. Notably both already ran the full fetch *after* the in-flight check at their own step 4 — `next-ticket` was the outlier and now agrees with them.

## Test

`node --test skills/fleet/scripts/*.test.mjs` → 76 pass, 0 fail, before and after rebase. Docs-only change; no test added.

## Adjacent, deliberately not fixed here

- **#22** (Agent Brief reaches survivors only after the 3–5 cut) — untouched. The cut stays at step 2 and the brief still arrives after it, so #22's problem is unchanged in kind and its fix is unblocked either way.
- Step 3's inline `gh pr list --search "<N>"` is the noisy full-text probe that `inflight.sh:26-29` documents as unusable on its own, and the skill never points at `inflight.sh`. That is a probe-correctness bug, not an ordering one — out of scope, worth its own ticket.

Closes #24


feigi:
Review complete — `code-reviewer` + `comment-analyzer` (docs-only, so the two-dimension subset of the catalog). Both delivered; no dimension unrun. Fixes applied at `27046b6`.

**The ordering change is sound and stays as-is.** The implementer was told it could conclude the reorder was unsound and stop; it engaged honestly instead. It refused the wholesale step-2/step-3 swap #24 floated, and refused it on a real ground — the probes are *not* local, `inflight.sh` makes three network round-trips per candidate (`:40`, `:59`, `:87`), so hoisting them above the shortlist would run them over the `--limit 100` list. It also invented **no refill loop**, leaving the below-floor case unresolved, which was the right call.

Every defect was in the added prose:

- **`:42` / `:90` — "three network round-trips" and "they are network calls, not local".** False for probe 3: `git worktree list; git branch -vv` is local, as is `inflight.sh`'s own probe 3 (`:97-106`). The block as written costs **two** round-trips; three is the count for `inflight.sh`, where they split 2/1/0 across the probes. The conclusion survives, the stated reason did not.
- **`:42` — "a taken `ready-for-agent` ticket is the normal fleet case".** Contradicted by this skill's own step 1, which excludes `-label:in-progress` server-side (`:18`) while step 5 adds that label as the first claim action (`:56`). A properly claimed ticket never reaches step 3. Measured live mid-wave: **0 of 10** step-1 candidates would have been dropped. The claim came in verbatim from #24's prose and was restated as fact. Step 3 catches the claim that never reached the label; the reorder is cheaper either way, which is the actual argument.
- **`:42` — "`scripts/inflight.sh` is the real implementation".** Wrong twice. The path resolves to `skills/fleet/skills/next-ticket/scripts/inflight.sh`, which does not exist (`run-team:54` uses `~/.claude/skills/fleet/scripts/inflight.sh`, the only form in the tree). And it is not the same implementation — see #33.
- **`:28` — "everything below runs per candidate".** False past step 3: step 4 presents the set once and blocks, steps 5–7 act on the single picked ticket.
- **`:28` — the moved cut leaked into a delegation.** `run-team:53` and `docs/specs/2026-07-22…:186` delegate to "step 2, on the `d` array", and that same phase presents **~10** (`spec:199`). Previously the 3–5 rode on the fetch sentence, which neither delegates to. Now scoped to next-ticket.
- **Line drift falsified two live citations.** Step *numbers* are unchanged, but lines moved +2 between steps 3 and 4 and +3 after the red-flag insertion. `docs/specs/2026-07-23-fleet-plugin-design.md:371-372` cites `next-ticket` SKILL.md`:63`/`:90` as bare-backticked-name sites; at the PR head `:63` was the `## 6.` heading and `:90` was this diff's own new bullet. Repointed to `:65`/`:93`. The same spec at `:367` complains about exactly this drift class.

**Verified and left alone:** `inflight.sh:15` takes only `<issue-number>`, so the probes need nothing step 2 produces; step 4 still gets titles and step 6's `sizing-a-ticket` re-reads the issue itself, so nothing is stranded; the moved paragraph is byte-identical but for `only for 3–5 survivors:` → `survivors only:`, both the `--json body` and bare-`--comments` warnings intact; the commit's `inflight.sh` `:40`/`:59`/`:87` citations are **exact**; `~6.4 KB` reproduces at 6397 B (n=1, on #7).

**Deferred:** #33 (step 3 hand-rolls the probes and diverges from `inflight.sh` three ways — pre-existing, and the reason the "real implementation" phrasing had to go), #34 (survivor count is 3–5 / ~10 / unstated across three files, `spec:36` contradicts `spec:199`, and step 4 is undefined below the floor).

`node --test skills/fleet/scripts/*.test.mjs` → **76 pass, 0 fail**. Not merged.

**Finisher pass** (independent, at `27046b6`): tests re-run → **76 pass, 0 fail**. All six corrections confirmed landed; no "local" claim about the probe set survives anywhere in the file; the `~/.claude/skills/fleet/scripts/inflight.sh` pointer is the runnable form; the 3–5 cut is scoped to next-ticket (`run-team:71` presents its own set, delegating only step 2's `d`-array logic). The repointed spec citations `:65`/`:93` were opened and both resolve to bare-backticked `sizing-a-ticket` references, which is the claim the spec makes about them. The moved paragraph carries both the `--json body` and the bare-`--comments` warnings verbatim, so PRs #11/#19 are not regressed. `ready-to-merge` added.

**Also cross-referenced:** #22 (the Agent Brief still arrives after the 3–5 cut) is unchanged in kind by this reorder — the brief lands after the cut in both the old and the new order. Detail in [the comment on #22](https://github.com/feigi/claude-config/issues/22#issuecomment-5132592234): the cut is now stated explicitly at `:28` instead of riding on the fetch sentence, and the brief now arrives after the probes as well as after the cut, which makes the reorder a trade rather than strictly dominant.

Issue #27, opened 2026-07-30
claim-ticket.sh: the dependency-free exception never fires in a clean checkout, and is wrong in both directions


feigi:
## Problem

`skills/fleet/scripts/claim-ticket.sh:40` (commit `182306b`, carried into PR #21) added a "no declared dependencies" exception to the lockfile refusal, so the fleet could claim `claude-config` itself. It does not work outside the tree it was written in, and it errs in both directions.

### 1. It never fires in a clean checkout

```
$ gh repo clone feigi/claude-config ghclone && cd ghclone
$ git checkout setup/agent-skills-config
$ ls package.json                                  → absent
$ sh skills/fleet/scripts/claim-ticket.sh 42 slug fix
claim-ticket: no recognised lockfile — refusing to guess an install command   (rc=2)
```

The probe is `require(process.cwd()+"/package.json")`, but **`.gitignore:42` ignores `package.json`** — it is not tracked, so no clone or worktree has it. The `require` throws, `2>/dev/null` swallows it, the branch falls through to the `die`.

It only appeared to work because an untracked, gitignored `{"type":"commonjs"}` happened to sit in one working copy. `git clean -xdn -- package.json` → `Would remove package.json`. One `git clean -xdf` re-breaks every future claim.

### 2. `"dependencies": {}` still refuses

`p.dependencies||p.devDependencies` is a truthiness test and `{}` is truthy, so a manifest declaring empty dependency objects selects the refusal — the exact false positive `182306b` set out to remove.

### 3. Three cases silently under-provision instead of refusing

Only `dependencies`/`devDependencies` are consulted. Measured `install="true"` (no install, claim proceeds) for:

- `peerDependencies`-only manifest
- `optionalDependencies`-only manifest
- npm workspaces root whose children declare the deps

The old code refused loudly. The new code hands back a worktree with no `node_modules`; the failure surfaces later as module-not-found inside `./agent-test` and reads as a broken worktree — the misattribution the sibling comment at `:47` says it exists to prevent.

**Not** a lockfile-corruption risk: the branch only ever assigns the literal no-op `true`, so no mutating install can be selected. Every error path is fail-closed — missing, malformed, and empty `package.json`, and `node` absent from `PATH`, all refuse.

## Fix (verified, patch below)

```diff
 # Derive the frozen install from the lockfile. No match is a refusal, not a
 # default — guessing here is what corrupts the tree. The one safe exception is
[...]

PR #21, opened 2026-07-30
chore(setup): scaffold matt-pocock engineering-skills config

feigi:
Runs `/setup-matt-pocock-skills` for this repo so `/triage` and the other engineering skills read a recorded tracker/label/domain contract instead of re-inferring it every run.

## What's added
- **`docs/agents/issue-tracker.md`** — GitHub tracker via `gh`; PRs-as-request-surface flag **off**
- **`docs/agents/triage-labels.md`** — default canonical labels (label string == role name: `needs-triage`, `needs-info`, `ready-for-agent`, `ready-for-human`, `wontfix`)
- **`docs/agents/domain.md`** — single-context `CONTEXT.md` + `docs/adr/` consumer rules
- **`CLAUDE.md`** — `## Agent skills` block pointing at the three files
- **`.gitignore`** — ignore `.ponytail-active` runtime marker (sibling to `.caveman-active`)

## Context
Prior triage sessions had no config to read, so the tracker (GitHub), label vocabulary (canonical defaults), and PR scope (off) were inferred each time. This records them. Choices confirmed interactively: GitHub tracker, default labels, single-context (repo is not a monorepo).

Note: this repo's root is `~/.claude`, so the `## Agent skills` block lives in the **global** `CLAUDE.md` — three pointer lines that only resolve to `docs/agents/*` when cwd is this repo.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #31, opened 2026-07-30
feat(fleet): release-ticket.sh — give back a claim that was never dispatched

### LINKED ISSUE #9: run-team: no claim-release procedure — an undispatched claim survives the run and reads as taken in the next Phase 0

feigi:
Closes #9

Phase 1 is serial and runs ahead of dispatch, so the controller can claim a ticket and then never send it — a collision surfaces after the claim, the maintainer drains the run, the pool is re-prioritised. That leaves the `in-progress` label, a worktree and a branch, and nothing in the run takes them back: `reap.sh` correctly declines, since the branch is not `[gone]` and has no unique commits. Next run, `inflight.sh` sees the worktree and a free ticket drops out of the pool, silently.

## What this adds

**`skills/fleet/scripts/release-ticket.sh <N> <slug> <type> [--apply]`** — the inverse of `claim-ticket.sh`. Dry-run by default. All four preconditions are recomputed inside the same invocation as the delete, the way `reap.sh` does it: 0 commits ahead of `origin/main`, clean worktree, no unique commits (`git cherry`), no branch on `origin`. All clear → drops the label, removes the worktree **without** `--force`, deletes the branch with `-d`. Any one failing → nothing is touched and the blocker is named. Exit 0 released, 1 blocked, 2 unanswerable, matching `inflight.sh` and `no-undo-audit.sh`.

**The drain step** in `run-team/SKILL.md`, next to the reaping section, plus an event-loop trigger and a red flag for the assumption that hides the bug ("the reap at the end will pick it up" — it declines). The heavy-row bail already said "reap the worktree and branch", which `reap.sh` will not do for the same reason; it now points here. Evidence in `references/reaping.md`.

Two calls it deliberately does not make, pinned by a test that greps the script: `worktree remove --force` and `branch -D` are the two that make `commit-commands:clean_gone` unusable here, and the two a future edit would reach for the first time a precondition refuses.

## Verification

`node --test skills/fleet/scripts/*.test.mjs` → **92 pass, 0 fail** (76 before this branch; the new file adds 16). `sh -n` clean.

The tests build real throwaway git repos in a temp dir and stub only `gh` on `PATH`, so they measure the script rather than this checkout. Each precondition has a case proving it blocks **on its own** — asserting the blocker it emits, not merely a non-zero exit, which any other precondition would satisfy. Mutation-verified: disabling the ahead check, the cherry check, the dirty check, the remote check, the fail-closed `ls-remote` guard, the `gh` read, the mistyped-slug guard, the main-worktree exclusion, or adding `--force`, each fails the case named after it and nothing else was needed to catch it.

Including the case that shows `ahead` and `git cherry` are not one check written twice: a commit already cherry-picked upstream is *equivalent* to `git cherry` (a `-` line) and only the ahead count still sees it.

`skills/fleet/scripts/release-ticket.test.mjs:23` — "after a release the in-flight probe no longer reads the ticket as taken" runs the real `inflight.sh` before and after, so the acceptance criterion is measured against the actual consumer.

## Two bugs found by not testing only in the authoring worktree

`claim-ticket.sh`'s live defect (#27) exists because it was verified in the one checkout where it happened to work, so this was also run from a fresh `git clone` with no `package.json` and no `node_modules`. That found both:

1. A claim blocked by its preconditions reported the `gh` failure (exit 2, unanswerable) instead of the blockers (exit 1), because the tracker read came first. The blockers already settle the answer; the read now happens only on the path that is about to delete something.
2. The worktree is located by the branch it has checked out — and in a clone, that branch is the **main checkout's** HEAD, so the probe returned the clone root as the thing to delete. `git worktree remove` refuses a main working tree, but only *after* the label was dropped: a half-release, the exact state this script exists to prevent. Only linked worktrees are candidates now, and a branch checked out in the main checkout is a blocker in its own right.

This script reads no repo files at all, so the `package.json` trap that broke `claim-ticket.sh` has no analogue here.

## Scope

`reap.sh`'s post-merge behavior is untouched, and a claim carrying commits or a pushed branch is never auto-released — that refusal is the finding. `claim-ticket.sh` is left alone; its dependency-probe bug is #27, owned by PR #21.


feigi:
Review complete. Two defects found and fixed on this branch (`5877fe4`), two deferred.

**Fixed here**
- The worktree is located only by the branch it has checked out, and it does not stay there — an interrupted rebase detaches it, a member can switch it. The lookup then found nothing, which read as "no worktree of ours", so the dirty check was skipped entirely: the script dropped the label, deleted the branch and reported `released:true` while the worktree stood with uncommitted work in it. Running the real `inflight.sh` before and after returns `1` both times — the ticket still reads as taken, i.e. the failure this script exists to fix, reported as success. Now its own blocker, matched on `claim-ticket.sh`'s `.worktrees/<issue>-<slug>` by exact suffix.
- The worktree path was read as awk's `$2`. `worktree list --porcelain` prints it raw, so any checkout under a directory with a space in its name truncated at the first space and died out of the dirty check — the script was unusable for every claim in such a repo. Now `substr($0,10)`.

Three cases added, each mutation-verified to fail alone; a cosmetic reword leaves all 19 green. `node --test skills/fleet/scripts/*.test.mjs` → **95 pass, 0 fail**, also from a fresh clone with no `package.json` and no `node_modules`.

**Deferred:** #37 (a prunable worktree makes the release permanently unanswerable — fails closed, pre-existing) and #38 (`<slug>`/`<type>` unvalidated, so a quote emits unparseable JSON — needs hand-typed input, harmless direction).

**Verified sound, no change needed:** all four preconditions block independently on their own fixtures, asserted on the blocker string rather than the exit code — including `ahead` firing alone where `git cherry` reports `-`. `git worktree remove` carries no `--force` and `git branch -d` is never `-D`. The delete-time recompute is real: dirtying the worktree *between* the last precondition and the delete leaves the late file intact, `worktree remove` refuses, and the script dies with the branch kept. The main-checkout guard holds with zero `gh` calls, so the label is never touched. Stashed work survives a release (`refs/stash` is in the common dir).

feigi:
Controller hold: `ready-to-merge` removed pending investigation. The local worktree `.worktrees/9-release-ticket` is at `5877fe4` while this PR's head is `61a19d6` — diverged — and it holds 243 uncommitted lines. Merge bot held off until it is established whether those lines are already in the PR head or exist nowhere else. No files were touched.

feigi:
Hold released. Evidence: the worktree's working tree contains **zero** lines absent from this PR's head. Its only unique content was an older `jstr()` (`tr '\n\r\t'`) which head supersedes at 61a19d6 with the full C0 range; the test file had 0 working-tree-only lines. So the 243-line uncommitted diff is a stale earlier draft of work already pushed — nothing at risk.

Both criticals verified closed at head 61a19d6, not in the working tree:
- `git ls-remote` at :168 no longer captures stderr into the value; the `gh issue edit` at :252 is checked by exit status into `halt()`, so a gh failure can no longer read as success while artefacts are deleted.
- `git worktree prune` at :260 warns instead of exiting 1, so a prune failure no longer fails a completed release.

`ready-to-merge` restored. Note for the merge bot: the local worktree is at 5877fe4 and dirty — it must not be rebased in place.

feigi:
Correction to my earlier hold comment — two mechanism details in it were wrong, both verified against the tree just now.

**1. I wrote "substring-match". It was comma-delimited.** Pre-fix at `fca1464:147`:
```sh
if ! labels=$(gh issue view "$issue" --json labels --jq '[.labels[].name]|join(",")' 2>&1); then   # :144
case ",$labels," in *,in-progress,*) has_label=true;; *) has_label=false;; esac                      # :147
```
The mechanism still holds and is arguably sharper than I described: folded-in stderr prepends a line, so `in-progress` arrives preceded by a **newline** rather than a comma, the match **misses**, `has_label=false`, and the label edit is skipped while both deletes still run — exit 0, `"label":false`, `"released":true`. Head cuts three legs: no `2>&1`, one name per line, `grep -qx` whole-line.

**2. I wrote that a prune failure made the release "exit 1".** It propagated prune's *own* exit code — the regression test observes **3**. Pre-fix `fca1464:176` was a bare `git worktree prune` as the final statement under `set -e`. Same wrong-direction failure, wider than I said.

Both corrections came from `finisher-pr-31`, and I re-derived them from `git show` before posting rather than relaying. Verdict is unchanged: both criticals closed at head `61a19d6`, pins at `release-ticket.test.mjs:422` and `:480` proven to fail against the pre-fix script, 125/125 green. `ready-to-merge` stands.

PR #32, opened 2026-07-30
fix(fleet): correction tickets also MINT false claims, not just inherit them

feigi:
Found live during a `/fleet:run-team` run on agent-brain, and fixed mid-run per the skill's own "fix the tooling mid-run" rule.

## The gap

`references/correction-tickets.md` documented exactly one mechanism: fix text **paraphrases the issue body's framing**, so a wrong premise in the ticket becomes a wrong claim in the repo. I passed that warning to an implementer verbatim in its dispatch prompt. It did not prevent either defect below — because neither came from the ticket.

## What happened

agent-brain PR #574 (issue #559, *"correct four stale or wrong comment clauses"*) shipped **two brand-new false claims**, both CI-green by construction:

| site | claim | reality |
|---|---|---|
| `:34` | "the **closing** `report-file verification` block" | 4th of 5; `cache-verify key` (1150-1216) closes the file |
| `:983` | "The **second** assertion is the one with teeth" | it is the **third** — read literally it names the assertion the same sentence calls toothless |

Three independent specialists converged on both (code-reviewer, comment-analyzer, pr-test-analyzer, the last with a mutation matrix).

**The issue asked for one clause. The PR wrote five lines.** Both errors entered in that expansion — which the inherited-claim check structurally cannot catch, there being no issue clause to compare against.

## Two things worth naming

**Positional references are the exact rot these tickets exist to retire**, and they re-rot the moment anyone inserts ahead of them. The file already had the right pattern one hunk away: `the path-list assertion above`.

**Ordinals fake their own verification.** The wrong ordinal named an assertion *earlier* than the real one, and the test runner stops at the first failed `expect` — so the obvious mutation (`if: ${{ always() }}`) reds the named assertion and never reaches the true one, appearing to confirm the wrong claim. Measured: `always()` reds #2; `always() && steps.mutate.outcome == 'success'` reds #3.

## The change

- `references/correction-tickets.md` +10 lines — the minted-vs-inherited mechanism, the ordinal trap, and two cheap implementer rules: **match the ticket's stated size** (added prose is where minted claims enter) and **never write a positional reference**, name the thing semantically.
- `SKILL.md` — its one-line assertion said correction tickets "inherit them from the ticket", which had become incomplete. Corrected in place, +8/-5.

Scope is deliberately lean per the skill's own "cut before you append": no speculative polish, nothing outside the defect this run produced.


feigi:
Review of `0bfd09dc`. Docs-only, so the fan-out was code-reviewer + comment-analyzer; comment-analyzer never delivered, so the claim-accuracy dimension rests on code-reviewer plus my own pass. Nothing pushed — this is your branch and both fixes below are authorial wording.

**The headline: the PR commits the defect it documents, in the sentence that documents it.**

### Blocking 1 — the positive example does not exist

`references/correction-tickets.md:21` holds up a good semantic reference: "name the thing semantically, as `the path-list assertion above` already does one hunk away."

```
git grep -n "path-list" 0bfd09dc    → 1 hit: line 21 itself, the citing sentence
git grep -n "path-list" origin/main → 0 hits
git log --all -S"path-list"         → 1 commit, 0bfd09d — this PR
```

That string has never existed anywhere in this repository. It lives in agent-brain's file. The present tense ("already does") asserts an example in this tree that is not there — a minted false claim in added prose, which is precisely the mechanism the new section exists to warn about.

### Blocking 2 — "one hunk away" is a positional reference, in the PR's own voice

Two clauses after banning `the closing/second/last X`, the same sentence locates its example by distance-in-diff. Not a quoted specimen — narrative voice. And a hunk is a coordinate of a diff, not of a file at rest, so it points at nothing even in principle once the diff is merged.

Both clear with one cut:

```
Two rules for the implementer, both cheap: **match the ticket's stated size** —
added prose is where minted claims enter — and **never write a positional
reference** (`the closing/second/last X`); name what the thing *is*, not where
it sits.
```

(code-reviewer offered `review-and-fix.md:58` as a substitute example. I checked it — that is the "Comments are findings" bullet, and it cites `"N of 3" counts` as a defect pattern. Not a model of semantic naming. Recommend just dropping the example clause.)

### Should-fix

- **`:15` — `#574` reads as this repo's PR 574.** It does not resolve (`Could not resolve to a PullRequest`; our ceiling is #32), and `#559` 404s. The base file avoids this by using counts, never numbers (`:7` "Four for four in one run", `:11` "Two of four"). Suggest `owner/repo#574` or "Second run, on a downstream repo:". The fix is provenance, not deletion — the specifics are what make the section useful. Filed the wider convention as #35.
- **`SKILL.md:266-269` — dangling antecedent.** "every factual claim it restates must have a settling command run … **and it must match the ticket's stated size**". Read clause by clause — which the next sentence orders reviewers to do — the subject of "it must match" is *every factual claim*. A claim has no size; the diff does. The paragraph's own instruction catches its own sentence.
- **`:17` — "the exact rot these tickets exist to retire" overstates 4×.** `:7` enumerates the four classes; exactly one (wrong-line citation) is positional. Settled by the adjacent paragraph.

### Nits

`(3rd)` has no denominator where `(4th of 5)` bounds itself. And `:19` says "stops at the first failed **expect**" — `git grep "expect(" 0bfd09dc` → zero hits; this repo is `node:test` + `assert.*` throughout. Foreign vocabulary inherited from the source repo; fixing the provenance line fixes this too.

### Verified clean — do not re-derive

- **The `SKILL.md` rewrite drops nothing.** All six load-bearing clauses of `origin/main:264-268` survive, most verbatim. Net +3 for two new assertions — proportionate, and the diff does match its own stated size rule.
- **Trailing newline preserved** (base and head both end `2e`, no `\n`); insertion is mid-file.
- **No collision with #31.** Both branch from SKILL.md blob `f70a6ac`. #32 occupies base 261-271; #31's nearest hunks end at base 183 and start at base 315 — 78 and 44 lines clear of git's 3-line context. Whichever rebases second applies cleanly.
- **Tests unaffected:** 76 pass / 0 fail on both `origin/main` and `0bfd09dc`, run from `git archive` copies.

Not adding `ready-to-merge`.

feigi:
Addendum — `comment-analyzer` delivered after my first comment. Both dimensions now in, none unrun. Blocking findings unchanged; three additions and one correction to my own framing.

**Correction to my irony framing.** I called it "a section warning about unverifiable claims, resting on one." That's loose — `:25` governs claims *about the tree*, while `:15` is testimony about an external run, and testimony may legitimately be uncheckable. The version that needs no qualification: **`:21` makes a locative claim about this very document, of exactly the class the section bans, and it is false.** That's the finding; the `#574` provenance is a separate, milder one.

**New should-fix — `:19`, `the obvious mutation` is load-bearing and never specified.** The reader can't reproduce the ordinal trap from the file alone. The measurement exists — your PR body has `always()` reds #2, `always() && steps.mutate.outcome == 'success'` reds #3 — it just didn't travel into the file. Same lossy transfer as the dropped `cache-verify key (1150-1216)` datum.

**New should-fix — `:15`, `the inherited-claim check above` has an ambiguous referent.** Settled from source, because the two specialists split on it and neither was right: `:11` does contain "instead of being checked clause-by-clause against tree", so a check *is* named above. But the prescribed duty is `## The clause-by-clause duty` at `:23-25`, **below**. A reader taking "the check above" to mean the duty is sent the wrong way. Not a false reference — an ambiguous one. `the inheritance mechanism above` disambiguates.

**Nit — `:17`, "fake their own verification" overstates slightly.** The mutation genuinely establishes that #2 has teeth. What it fails to falsify is the *exclusivity* in "the second assertion is **the one** with teeth".

**Nit — `:13`, the heading ordinal `## Second mechanism`** rots if anything is inserted at `:9-11`, and its predecessor isn't numbered "First". Structural and defensible; noting only for completeness.

**One thing worth calling out in your favour:** `SKILL.md` is *cleaner than the reference file on the very rule this PR introduces* — it states semantic naming without citing the absent exemplar and without a positional reference in its own voice. The fix for `correction-tickets.md:21` is to make it behave the way your `SKILL.md` edit already does.

---

**On applying these myself:** I did not push. `fix/correction-tickets-minted-claims` is currently checked out in a live worktree at `/private/tmp/claude-501/-Users-chris-dev-agent-brain/9bca232d-…/scratchpad/cc-main` (clean, at `0bfd09d`). Pushing into a branch another session holds risks clobbering it with no local marker on their side. Happy to apply the one-sentence fix if you confirm that session is done.

feigi:
Second review, of `70acb5c6`. Docs-only, so the fan-out was code-reviewer + comment-analyzer; both delivered, neither dimension unrun. Nothing pushed — **both prior blocking findings are genuinely closed** and nothing new rises to blocking. Suite: **76 pass / 0 fail** (`./agent-test skills/fleet/scripts/*.test.mjs`; the bare-directory form dies under Node 26, which is #42, not this PR).

### Both blocking findings closed — verified, not inherited

- **The absent exemplar.** `git log --all -S"path-list"` → exactly two commits, `0bfd09d` (introduced) and `324b630` (removed). `git grep "path-list" 70acb5c6` → exit 1. `grep -RIn "path-list" .` from repo root → 0 hits (tree has **zero** symlinks: `git ls-tree -r 70acb5c6 | awk '$1=="120000"'` → empty, so the macOS non-following caveat does not bite). The string never existed here and is gone.
- **The positional reference in the PR's own voice.** `git grep "one hunk away" 70acb5c6` → exit 1. The replacement clause carries no locative.

Two commits, two changed lines, two real findings closed. Elegant on those two — but they did not re-read the rest of the hunk, and `SKILL.md` was not re-read at all. That residue is filed, not dropped.

### `#574` — the fix disambiguates rather than relocating

Dropping the `#` sigil is the right call, and for a better reason than the commit body gives. Repo markdown files do not autolink at all:
```
gh api /markdown -X POST -f mode=markdown ... "see #32 and CoCo/agent-brain#574"
  -> <p>see #32 and CoCo/agent-brain#574</p>          # both literal
gh api /markdown -X POST -f mode=gfm -f context=feigi/claude-config ...
  -> #32 links; CoCo/agent-brain#574 stays literal    # cross-host form never linked
```
So `#574` never *rendered* as anything — the defect was reader ambiguity, which is real and which the fix removes. The autolink rationale in `70acb5c`'s body is wrong in both directions, but it argues for the correct conclusion and it is immutable history; not worth a force-push of your branch.

Same for **"Our ceiling is #36"** in that body. Issues and PRs share one number sequence, and #37/#38/#39 predate the commit by 3–23 minutes, so the repo ceiling was #39. Under the narrower "highest PR" reading — which the surrounding `gh pr view` context supports — it was true, by 23 seconds (PR #40 was created at `15:43:14Z`, the commit at `15:42:51Z`). Ambiguous rather than false, and the argument survives either number.

### `:19` "expect" — left, but validly scoped

`git grep "expect(" 70acb5c6` → exit 1; `grep -RIn --exclude-dir=.git "expect(" .` → 0 hits. This repo is `node:test` + `assert.*` across five files. The sentence is unchanged, but `:15`'s new "test vocabulary … not ours" clause does reach it — the `##` section runs `:13`–`:22`, next heading at `:23`. Not a false local claim. The residual issue (the *mechanism* transfers to `assert` and shouldn't read as foreign) is in #35.

### Counts

`:7` "Four for four" lists exactly four classes and matches `review-and-fix.md:51` item-for-item. `:11` "Two of four" is unchanged from base. `:17`'s "the exact rot these tickets exist to retire" is the one that does not hold — 1 of the 4 classes is positional; filed.

### No collision with #31

`git diff -U0 origin/main...` — #32 occupies base `264-268`; #31's nearest hunks end at base `181` and start at base `317`. 83 lines above, 49 below, against git's 3-line context. Whichever rebases second applies cleanly.

### Deferred

- **#35** (commented) — cross-repo citation residue: `:15`'s "files" names a category with zero instances below it; `:15` never names the source repo; `:17`'s `(4th of 5)`/`(3rd)` are unfalsifiable from this repo; `:19` scopes a runner *behavior* as if it were vocabulary.
- **#47** (new) — four clauses that do not survive the section's own clause-by-clause rule: `:15` "the inherited-claim check above" (better referent is `:25`, below), `:17` "the exact rot" (1 of 4), `:19` "the obvious mutation" (no reachable antecedent), `SKILL.md:268` "and it must match the ticket's stated size" (no named subject).
- **#48** (new) — `docs/specs/2026-07-23-fleet-plugin-design.md`: `:208`'s three-section row under-reports now that this PR adds a fourth (caused here); `:225` "both callers" is one caller and `:428` "all 33 patterns resolve" has 3 misses (both pre-existing, identical on `origin/main`).

### One thing in the PR's favour

It passes its own new size rule exactly: `git diff --numstat origin/main...70acb5c6` → `8 5` SKILL.md, `10 0` correction-tickets.md, matching the PR body's stated `+8/-5` and `+10`. Both fix commits net zero, so the head still matches. That is the rule's only live demonstration in the repo.

Not adding `ready-to-merge` — a finisher owns that. **No release label is set**; on the precedent of #20 and #40 (`fix(fleet):`, docs/guidance only) I would pick `patch`, but it is your call.

feigi:
Pushed `56625cc` — two blocking fixes, +2/−2, `correction-tickets.md` only. **Zero lines changed in `run-team/SKILL.md`**, so #31 sequencing is unaffected.

Revising my earlier call: I graded `:15` and `:17` should-fix. They are blocking. Both are factual claims about *this document's own contents* — the class this PR exists to police — and three independent passes (`fix-pr-32`, code-reviewer, comment-analyzer) each measured them wrong by a different route. A should-fix that is a wrong claim is a blocking finding.

**`:17`** — "the exact rot these tickets exist to retire" → "one of the four classes in the four-for-four finding". Of `:7`'s four, exactly one (commit body citing wrong line) is positional; the other three are attribution, polarity and self-refutation errors. Settled against `review-and-fix.md:51`, which expands the same four.

**`:15`** — "the inherited-claim check above" → "comparing fix text against the issue clause". Above `:15` no check is stated as a procedure (`:3` abstract pointer, `:11` by negation); the actionable one is `:25`, **below**. The locative sent readers the wrong way, and `above` is a positional reference inside the paragraph that bans them.

Two more defects closed in that same sentence while it was open:
- **`files` dropped.** `sed -n '16,25p' | grep -oE '[A-Za-z0-9_./-]+\.(md|mjs|js|sh|json|ts)'` → exit 1. Nothing below it cited a file, so the noun asserted an existence it could not support.
- **`below` re-scoped.** It was unbounded and ran to EOF, so it swallowed `:21`'s "Two rules for the implementer" — ours, not theirs. Now scoped by property ("describing it") rather than by region.

That last one I got wrong once: my first attempt scoped it to "in this section", which has the identical extent and swallows `:21` just the same. Caught before push.

`sed -n '13,21p' | grep -oE '\babove\b|\bbelow\b'` → exit 1. No positional references remain in the section that bans them.

**Left standing, deliberately:**
- `SKILL.md:268`'s dangling antecedent — real, but open PR #31 owns that file and this is not worth a collision. #47.
- `:19`'s "the obvious mutation" and the `(3rd)` denominator. #47.
- The commit-body claims in `70acb5c` ("Our ceiling is #36"; the autolink rationale). Both reach sound conclusions from wrong reasoning. Fixing them means rewriting your branch — not worth it, and not my call. For the record, the autolink mechanism is wrong in both renderers: repo `.md` files render in `mode=markdown`, which does not autolink at all, and even in `gfm` a cross-host `owner/repo#574` stays literal. The `#` was still worth dropping; the reason is reader ambiguity, not rendering.

**Verification:** `./agent-test skills/fleet/scripts/*.test.mjs` → `tests 76 / pass 76 / fail 0`. Trailing-newline state unchanged (still absent, as on `origin/main`). `git merge-tree --merge-base=origin/main` against #31 → exit 0 in **both** orders, at the new head.

Deferrals: **#35** (commented), **#47**, **#48**. Not adding `ready-to-merge`. Still no release label — I would pick `patch` on the precedent of #20 and #40.

PR #36, opened 2026-07-30
fix(fleet): claim-ticket.sh — derive install and test command from the ref it builds from

feigi:
Lifts the `claim-ticket.sh` work out of PR #21, where it was sitting alongside unrelated agent-skills config scaffolding. Both commits are cherry-picked from that branch; nothing new is authored here.

## Why this was blocking

`claim-ticket.sh` refused to claim a ticket in this repo at all — `no recognised lockfile — refusing to guess an install command`, exit 2. That gate exists to stop a lockfile-mutating install corrupting a shared worktree, but this repo declares no dependencies, so there was nothing to corrupt and nothing to install.

## The first attempt was wrong, and the second commit is the fix

The initial fix (`1ae4d4c`) probed `package.json` with `require(process.cwd() + "/package.json")`. Two defects, both found in review:

1. **It never fired in a clean checkout.** `package.json` is gitignored (`.gitignore:42`) and untracked, so a fresh clone has none, the probe throws, `2>/dev/null` swallows it, and the script falls through to the original refusal. It worked only in the checkout that authored it.
2. **It derived from `$PWD` while the worktree is built from `origin/main`.** Nothing reconciled them. Measured: with a lockfile and `scripts.test` on `origin/main` but stripped locally, the script printed `install: true` / `node --test` and then built a worktree *containing the lockfile it had just reported absent*, while printing `lockfile clean after install`.

`86aec92` derives everything from `origin/main` — the ref the worktree is actually built from — via `git show`/`git cat-file`, which also removes the dependency on an untracked file. It additionally:

- fixes `"dependencies": {}` being refused (`{}` is truthy in JS, so the empty-object spelling hit the very false positive the change was written to kill);
- turns `peerDependencies`-only, `optionalDependencies`-only and workspace roots back into loud refusals instead of silently under-provisioned worktrees;
- refuses to emit a runner when there is no `scripts.test` **and** no test file, rather than falling back to bare `node --test` — which exits **0 having run nothing**, and is fed to the review fan-out via `workflows/review-pr.js:113`.

## Verification

- `node --test skills/fleet/scripts/*.test.mjs` → **97 pass, 0 fail**
- `claim-ticket.test.mjs` is new: **21 cases** over the `package.json` shape matrix, mutation-verified — 13 fail against the pre-fix script, 8 stay green, so the suite discriminates rather than failing under everything
- Checked in a **fresh full clone**, not the authoring checkout — the omission that shipped defect 1 in the first place

## Note on overlap

PR #21 contains these same two commits. Whichever merges first, the other's copies become no-ops on rebase. Splitting them out means the fleet tooling fix can land without waiting on the config scaffolding, which is held pending unrelated decisions.


feigi:
Superseded by #21, which merged as `5c63634` and carries both of these commits.

Verified redundant, not abandoned:
- `skills/fleet/scripts/claim-ticket.sh` is **byte-identical** between this branch and `main`
- `skills/fleet/scripts/claim-ticket.test.mjs` (21 cases) landed with #21

This PR was split out while #21 was held pending unrelated decisions. #21 merged instead, so both commits are already on `main` and nothing here is lost.

Not rebasing it: the branch predates #19, #20, #21 and #30, so `git diff main..HEAD` is `7 files changed, 5 insertions(+), 143 deletions(-)` — it would revert the `## Agent skills` block in `CLAUDE.md`, all three `docs/agents/*.md` files, and PR #30's `next-ticket/SKILL.md` reorder. The rebase conflicts are git refusing to let a stale branch undo `main`.

Issue #23, opened 2026-07-30
fleet skills: the bare `--comments` caveat has three different treatments across the three files sharing one issue-read form


feigi:
## Problem

Three skill files now read issues with the same `gh issue view <N> --json title,body,comments --jq …` form, but carry **three different treatments** of the caveat explaining why the wrong forms are wrong:

| file | caveat |
|---|---|
| `skills/fleet/skills/run-team/SKILL.md:117-122` | verbose — 5 lines on bare `--comments` |
| `skills/fleet/skills/sizing-a-ticket/SKILL.md:10` | **none** |
| `skills/fleet/skills/next-ticket/SKILL.md:30` | compressed, covers **both** wrong forms |

The command string itself is byte-identical across all three (verified during the PR #19 review: `od -c` on the backticked substring, 112 bytes each, no divergence). Only the surrounding caveat diverges.

## Why it matters

The bare-`--comments` warning is now stated twice in two different wordings, and omitted entirely from the one file (`sizing-a-ticket`) that is the nearest structural sibling to `next-ticket`. Two wordings of one fact drift independently — that is how the tree got inconsistent in the first place, which is what #7, #14 and PR #11/#19 have all been cleaning up.

`run-team`'s verbosity is **not** the defect: `:117-122` is a verbatim subagent prompt, so it has to carry its own context and cannot reference anything. The question is only whether the other two should agree with each other.

## Desired behavior

Pick one and apply it to `next-ticket:30` and `sizing-a-ticket:10`: either both carry the compressed both-forms caveat, or both carry none and the rationale lives in one place. `run-team:117-122` stays verbose regardless.

## Out of scope

- The command string itself — settled in #7/#11, applied to `next-ticket` by PR #19.
- `run-team/SKILL.md`'s block-count contradiction (#13) and claim-release gap (#9) — different defects in the same file.

## Note

Raised by the `code-reviewer` specialist during the PR #19 review as nit 4 of 4, explicitly as "a follow-up ticket, not a change here". PR #19 deliberately declined to touch `run-team/SKILL.md` because of concurrent work on #8, which is a legitimate reason to defer unification rather than attempt it mid-wave.

Deferred from PR #19 review

feigi:
PR #21 adds a **fourth** site of the bare `--comments` form, and it carries the weakest treatment of the four: **no caveat at all**.

`docs/agents/issue-tracker.md:8`
[...]

Issue #34, opened 2026-07-30
fleet: survivor count is 3-5, ~10, and unstated across three files, and nothing says what step 4 does below the floor


feigi:
## Problem 1 — three files disagree on the count, and the spec contradicts itself

- `skills/fleet/skills/next-ticket/SKILL.md:44` — step 4 presents **3–5 survivors**.
- `docs/specs/2026-07-22-run-team-agent-fleet-design.md:199` — Phase 0 item 6 presents **~10 admissible survivors**.
- `skills/fleet/skills/run-team/SKILL.md:71` — Phase 0 item 7 presents **admissible survivors**, no number at all.

The spec contradicts itself directly. `:36` says "**the controller runs `next-ticket` steps 1–4 exactly once**" — and `next-ticket` step 4 is the 3–5 step — while the spec's own item 6 then presents ~10. A controller following the spec literally cannot satisfy both. `run-team/SKILL.md:76` separately reasons about waves "five wide", which fits neither number cleanly.

The PR #30 review scoped the cap sentence at `:28` so the number no longer travels through the `run-team:53` / `spec:186` delegation ("Count is next-ticket's own; fleet Phase 0 sets its own"). That stops the leak but does not settle what Phase 0's number actually is, nor reconcile `spec:36` with `spec:199`.

## Problem 2 — the survivor set can fall below the floor, and nothing says what happens

Step 2 cuts to 3–5. Step 3's probes then drop every candidate already taken. Nothing states the behavior when the set lands below 3 — refill from step 1's list, or proceed with fewer. Step 4 still reads "3–5 survivors, best first", which is unreachable once probes have dropped two of four. Zero survivors is reachable too, and step 4 has no wording for it.

This is pre-existing: the same arithmetic held before PR #30, which only moved the full issue read behind the probes. PR #30 deliberately did **not** invent a refill loop — that was the right call, and the gap is named here rather than smuggled into an ordering fix.

Worth recording how reachable this actually is. Measured live mid-wave during the PR #30 review, the step-1 query returned 10 candidates and **0 of 10** would have been dropped by step 3, because step 1 already excludes `-label:in-progress` (`:18`) and step 5 adds that label as the first claim action (`:56`). So step 3 fires on the claim that never reached the label, not on the common case — which makes the below-floor path rarer than the prose suggested, but it is exactly the unattended path nobody is watching when it does fire.

## Desired behavior

[...]

PR #40, opened 2026-07-30
fix(fleet): next-ticket step 3 taught the probe inflight.sh calls unusable

### LINKED ISSUE #33: next-ticket: step 3 hand-rolls the three probes and diverges from inflight.sh, which documents the inline PR probe as unusable
## Problem

`skills/fleet/skills/next-ticket/SKILL.md:33` instructs the bare full-text PR probe that `skills/fleet/scripts/inflight.sh` documents, at length, as unusable on its own:

```bash
gh pr list --state all --search "<N>"
```

`inflight.sh:26-29` says exactly why not:

> NOT `gh pr list --search "<N>"` on its own. That is a full-text search and is
> uselessly noisy: measured here, searching 41 returned six PRs, five of which
> merely contained the digits somewhere. Trusting it makes nearly every ticket
> read as taken, so the fleet skips free work silently and permanently.

The script instead asks GitHub for `closedByPullRequestsReferences`, then adds branch-segment matching (`(^|[/-])N([-/]|$)`) and an `OPEN`-state filter, so a merged or abandoned PR does not make a finished ticket read as taken forever.

The skill's inline block has none of that. An agent following `next-ticket` step 3 literally gets the noisy probe and the documented failure mode: free tickets read as taken, silently and permanently.

The skill's other two probes diverge as well:

- `git ls-remote --heads origin | grep -E "[/-]<N>[-/]"` requires a delimiter on **both** sides, so it misses a branch whose name ends in the number — `fix/24` does not match. The script anchors both ends as alternatives instead: `(^|[/-])$n([-/]|$)` (`inflight.sh:88`).
- `git worktree list; git branch -vv` has **no matching logic at all** — it dumps both listings for the agent to eyeball. The script matches `refs/heads` with the same anchored pattern and matches worktrees on the **basename only** (`inflight.sh:97-105`), deliberately, because grepping the full absolute path false-hits every ticket whenever a parent directory contains the digits.

## Why this is a ticket and not a PR #30 fix

Pre-existing, and orthogonal to what PR #30 changed. #30 reordered *when* the probes run relative to the full issue read; it did not touch the block's contents. Fixing probe correctness is a different change with a different risk profile.

PR #30 does make it more visible: it adds a sentence at `:42` pointing at `inflight.sh` as "the real implementation" one paragraph below a block that contradicts it. The file now names the authority and disagrees with it in the same step.

## Desired behavior

`next-ticket` step 3 points at `~/.claude/skills/fleet/scripts/inflight.sh <N>` — the way `run-team/SKILL.md:54` already does — rather than carrying a divergent hand-rolled copy of the three probes. That deletes the divergence instead of re-syncing it, and is the smaller diff.

If the inline block is kept for readability, it must at minimum match the script's semantics: the `closedByPullRequestsReferences` query, the `OPEN`-only filter, the anchored branch-segment regex, and basename-only worktree matching.

## Out of scope

- The step ordering — settled by PR #30.
- `run-team/SKILL.md`, which already calls `inflight.sh` correctly.

Deferred from PR #30 review

feigi:
`next-ticket` step 3 carried a hand-rolled copy of the three in-flight probes, and all three diverged from `skills/fleet/scripts/inflight.sh` — the script the same step names, one paragraph below, as the real implementation.

## What was wrong

| inline probe | divergence |
|---|---|
| `gh pr list --state all --search "<N>"` | the bare full-text search `inflight.sh:26-29` documents at length as unusable alone. No `closedByPullRequestsReferences`, no `OPEN` filter — so a merged or abandoned PR makes a finished ticket read as taken forever. |
| `git ls-remote --heads origin \| grep -E "[/-]<N>[-/]"` | requires a delimiter on **both** sides, so it misses a branch ending in the number (`fix/24`). The script anchors both ends as alternatives: `(^\|[/-])$n([-/]\|$)`. |
| `git worktree list; git branch -vv` | no matching logic at all — it dumps both listings to eyeball. The script matches worktrees on **basename only**, deliberately, because grepping the absolute path false-hits every ticket when a parent directory holds the digits. |

The documented failure mode is that free tickets read as taken, so the fleet skips free work silently and permanently.

## Measured again on this ticket

Running the script on #33 reproduces the premise exactly:

```
$ gh pr list --state all --search 33 …
    no PR is about #33 (3 full-text match(es) were all incidental)
    local branches: fix/33-next-ticket-probes
inflight: #33 taken=true
```

Three PRs match the digits; **all three are incidental**. The old inline probe would have reported #33 as taken on that basis alone. The script discards them and hits only on the real local branch and worktree.

## The change

Step 3 now calls `~/.claude/skills/fleet/scripts/inflight.sh <N>`, the way `run-team/SKILL.md:54` already does. That deletes the divergence instead of re-syncing it, per the issue's preferred option, and is the smaller diff.

Two supporting edits:

- Exit `2` is called out as **not free** — an unanswered probe (`gh` failed, not a repo) must not read as available work. Verified: `inflight.sh abc` → rc 2, `inflight.sh 33` → rc 1.
- The step-ordering paragraph said "three round-trips *if you run* `inflight.sh`", phrasing the script as the alternative to the inline block. It is now the instruction, so that reads as the cost of the step.
- One red flag against re-inlining the probes, using the file's existing mechanism for "you will be tempted; don't".

## Verification

- Tests: `./agent-test skills/fleet/scripts/*.test.mjs` → **97 pass, 0 fail**, before the change and after the rebase onto `origin/main`. (The `skills/fleet/scripts/` directory form fails on node v26 — it resolves the path as a module — so the glob is the working invocation.) This repo has no CI; local tests are the gate.
- Divergence sweep: `grep -rn -e 'gh pr list --state all --search' -e 'ls-remote --heads origin' -e 'git worktree list; git branch' .` over the repo, excluding `.git` and `node_modules`. The only hand-rolled copy was `next-ticket/SKILL.md:33-35`, now gone. Remaining hits are `inflight.sh` itself and one prose line in `docs/specs/2026-07-22-run-team-agent-fleet-design.md:188`, a dated point-in-time design record left alone deliberately — it is already stale w.r.t. `run-team`, which calls the script.
- `grep -rln "SKILL.md" --include="*.test.mjs" .` returns nothing: no test asserts skill-doc content, so none needed extending and none was added for a markdown-only change.

## Scope

Untouched, deliberately: the step ordering (settled by #30), `run-team/SKILL.md` (correct already, and PRs #31/#32 edit it), and the bare-`--comments` caveat wording at `:40` that #23 owns — it appears in the diff only as unchanged context.

Closes #33


feigi:
Review deferrals filed — three issues, all labelled `ready-for-agent`, none fixed in this PR:

- **#44** — `inflight.sh`: the `closedByPullRequestsReferences` probe is not state-filtered, so a merged PR makes a finished ticket read as taken (measured: `inflight.sh 8` → `taken=true`, rc=1, via merged #20). Also refutes this PR's commit-message claim that the `OPEN` filter prevents it — that filter covers only the branch-matched path. Includes the unguarded command substitution at `inflight.sh:76`, which can exit 1 and now collides with the `0/1/2` contract this PR documents.
- **#45** — both fleet design specs describe `inflight.sh` with a payload and exit contract it does not have (`{taken, probe, evidence}` / "non-zero: never", against actual `{issue, taken, hits, evidence}` and exits 1 and 2).
- **#46** — `run-team/SKILL.md:55` calls probe 1 "open/closed PRs", wrong in both directions. Deferred rather than fixed because open PRs #31 and #32 both edit that file.

Searched `gh issue list --search` over `inflight`, `closedByPullRequestsReferences`, `spec design out`, and `state filter merged` before filing; no existing issue covered any of the three. Nothing found for #23 (bare-`--comments` caveat) or #34 (survivor count) — this diff neither moves nor restates either, so no deferral was added to them.

Fixes applied here in `48ab2cd`: the `one-sided` regex inversion, the unconditional worktree false-hit claim, the exit-2 cause list, and the probe-count bridge this PR had deleted.

Issue #28, opened 2026-07-30
triage-labels.md records needs-triage and needs-info as this repo's labels; neither exists (HTTP 404)


feigi:
## Problem

`docs/agents/triage-labels.md` (added by PR #21) opens with:

> This file maps those roles to the actual label strings used in **this repo's issue tracker**.

Two of the five rows name labels that do not exist here. Measured:

```
GET repos/feigi/claude-config/labels/needs-triage   → HTTP 404
GET repos/feigi/claude-config/labels/needs-info     → HTTP 404
GET repos/feigi/claude-config/labels/ready-for-agent → ready-for-agent
GET repos/feigi/claude-config/labels/ready-for-human → ready-for-human
GET repos/feigi/claude-config/labels/wontfix         → wontfix
```

Full actual label set:
```
bug  documentation  duplicate  enhancement  good first issue  help wanted
invalid  question  wontfix  ready-for-agent  in-progress  ready-to-merge
ready-for-human  minor  patch  major
```

`CLAUDE.md:140` repeats the claim to every session, without the file's own "edit this" caveat:

> Default canonical vocabulary — label string equals role name (`needs-triage`, `needs-info`, `ready-for-agent`, `ready-for-human`, `wontfix`).

## Why it matters

The **read** direction is harmless today: `skills/fleet/scripts/candidates.mjs:43` only ever *excludes* both (`-label:needs-triage -label:needs-info`), and excluding a nonexistent label is a no-op in GitHub search. `skills/fleet/skills/next-ticket/SKILL.md:18` carries the same exclusion.

The **write** direction is the breakage. A skill told "apply the needs-triage triage label" runs `gh issue edit <n> --add-label needs-triage`, which cannot succeed against a label that does not exist. These files exist precisely so agents stop re-inferring the contract, so the wrong value is trusted rather than checked.

## Root cause

`383e6e4` scaffolded these files from `/setup-matt-pocock-skills` and its own commit body calls them "**default** canonical labels (role name == label string)". `triage-labels.md` closes with:

> Edit the right-hand column to match whatever vocabulary you actually use.

The right-hand column was never edited. The file is an unadapted template whose header sentence nonetheless asserts it records reality, and `CLAUDE.md` propagates the defaults as settled fact.

## Desired behavior

Maintainer's call, which is why this is a ticket and not a patch — the two roles have no existing equivalent here, so the resolution is a decision, not a correction:

- create `needs-triage` / `needs-info` and keep the mapping, **or**
[...]

PR #41, opened 2026-07-30
docs(agents): drop the two references to tooling this repo doesn't have

### LINKED ISSUE #29: docs/agents/*.md are the unedited vendor templates — one impossible command, one nonexistent skill, two 404 label namespaces
## Root cause

All three files added by PR #21 are **byte-identical to the seed templates** they were generated from:

```
docs/agents/domain.md         == ~/.agents/skills/setup-matt-pocock-skills/domain.md
docs/agents/triage-labels.md  == ~/.agents/skills/setup-matt-pocock-skills/triage-labels.md
docs/agents/issue-tracker.md  == ~/.agents/skills/setup-matt-pocock-skills/issue-tracker-github.md
```

(`git show 8155cd19:docs/agents/<f>.md | diff - <seed>` → no output for all three.)

Zero customization. `triage-labels.md:15` even says "Edit the right-hand column to match whatever vocabulary you actually use" — that step never happened. Every defect below is one instance of that single root cause, which is why they are filed together.

`CLAUDE.md` now points every session at these files, so unedited defaults are being served as this repo's contract.

## The label vocabulary is tracked separately in #28

`needs-triage` / `needs-info` returning 404 is #28. Not repeated here.

## 1. `issue-tracker.md:8` describes an impossible command

> - **Read an issue**: `gh issue view <number> --comments`, filtering comments by `jq` and also fetching labels.

Two independent defects in one line:

```
$ gh issue view 13 --repo feigi/claude-config --comments --jq '.comments'
cannot use `--jq` without specifying `--json`
```

`--jq` requires `--json`, which is mutually exclusive with the `--comments` form named. So the filtering the sentence instructs cannot be performed at all. Separately, the bare `--comments` form drops title and body (0 bytes, exit 0 on a comment-less issue) — that half is tracked in #23, where I have commented.

## 2. `domain.md:11,45` names a skill that does not exist

Both lines point at `/domain-modeling` as the mechanism that lazily creates `CONTEXT.md` and `docs/adr/`:

> The `/domain-modeling` skill (reached via `/grill-with-docs` and `/improve-codebase-architecture`) creates them lazily when terms or decisions actually get resolved.

```
skills/domain-modeling                 → ABSENT
skills/grill-with-docs                 → symlink → ../../.agents/skills/grill-with-docs
skills/improve-codebase-architecture   → symlink → ../../.agents/skills/improve-codebase-architecture
```

The two routes exist; the destination does not. The file's "if these files don't exist, proceed silently" escape hatch covers the missing *files*, but not the assertion that a named mechanism will eventually create them — an agent that believes it will defers the work forever.

**Method note:** `grep -r` over `skills/` gives a clean but wrong negative here — macOS `grep -r` does not descend into symlinks, and several entries under `skills/` are symlinks into `~/.agents/skills/`. Use `-R` or test the resolved target.

## 3. `issue-tracker.md:39-40` uses a label namespace that does not exist

`wayfinder:map` and `wayfinder:<type>` both 404, and `gh issue create --label` does **not** create a missing label — it fails. So the documented wayfinding flow cannot run as written.

Related: `/wayfinder` and `/triage` exist in the environment but are absent from `git ls-tree -r 8155cd19` (`skills/*` is gitignored bar `caveman-compress` and `fleet`). The section reads as repo-resident contract while describing tooling this repo does not version.

## 4. Minor — `CLAUDE.md:144` present tense

> Single-context — one `CONTEXT.md` + `docs/adr/` at the repo root.

Neither exists. Note this one is **by design** rather than an oversight: the setup skill (`SKILL.md:59`) writes single-context "without asking" as the default layout, so this is a layout *choice*, not a claim that the files are present. Wording like "Single-context layout (no `CONTEXT.md` yet)" would remove the ambiguity. Lowest priority of the four.

## Verified correct — do not "fix" these

- **`issue-tracker.md:26`** — "resolve with `gh pr view 42` and fall back to `gh issue view 42`" is **right, and right for the correct reason**. The PR probe runs first and is strict (`gh pr view <issue>` exits 1), so `gh issue view` is only reached once the number is known not to be a PR. Reversing it would introduce the exact confusion that produced four wrong claims in one fleet run, because `gh issue view` silently accepts a PR number.
- **The GitHub API claims** — `issue_dependencies_summary.blocked_by`, the `dependencies/blocked_by` endpoint, `sub_issues`, and `.id` being the numeric database id distinct from `number`/`node_id` all verified live against this repo.
- **`triage-labels.md` omitting `in-progress`/`ready-to-merge`/`patch`/`minor`/`major`** — correct. It scopes itself to five triage *roles*; those are lifecycle and release labels, and it makes no completeness claim.

## Coupling worth recording

`ready-for-agent`, `ready-for-human` and `wontfix` are not triage-local — they drive automated dispatch (`candidates.mjs:43`, `board.mjs:173`, `run-team/SKILL.md:49`). A `/triage` run moving an issue to `ready-for-agent` feeds a live fleet queue, and neither doc says so.

Deferred from PR #21 review

feigi:
Closes #29.

Finishes what PR #21 left. Verified against `origin/main` first: the impossible `--jq`-without-`--json` command (`issue-tracker.md:8`), the bare `--comments` form (`:34`) and the two 404 triage labels (tracked as #28) are already fixed there. Two defects remained, and both point agents at things that do not exist in this environment.

## 1. `domain.md:11,45` — `/domain-modeling` is not installed

Both lines named it as the mechanism that lazily creates `CONTEXT.md` and `docs/adr/`. It exists nowhere:

```
$ ls -d skills/domain-modeling ~/.agents/skills/domain-modeling
ls: skills/domain-modeling: No such file or directory
ls: /Users/chris/.agents/skills/domain-modeling: No such file or directory
```

The two *routes* it named (`/grill-with-docs`, `/improve-codebase-architecture`) do exist, which is what made the sentence dangerous — it reads as a live promise, so an agent defers the glossary edit to a skill that will never run.

`/improve-codebase-architecture` already does the lazy creation itself, so the fix names the real mechanism rather than deleting the guarantee:

- `SKILL.md:68` — "Add the term to `CONTEXT.md`. Create the file lazily if it doesn't exist."
- `SKILL.md:70` — offers an ADR when a rejected candidate has a reason worth recording.

A second paragraph states plainly that no `/domain-modeling` exists and that an installed skill deferring to it should do the edit inline. That last clause is load-bearing: three installed skills still delegate to it — `improve-codebase-architecture/SKILL.md:66`, `triage/SKILL.md:76`, `grill-with-docs/SKILL.md:7`, plus `skills/wayfinder/SKILL.md:113`. Those live outside this repo (`skills/*` is gitignored bar `caveman-compress` and `fleet`), so they are out of scope here — flagging them rather than fixing them.

## 2. `issue-tracker.md` — the `wayfinder:*` labels 404

`gh issue create --label wayfinder:map` and `--label wayfinder:<type>` were documented as runnable. Neither label exists:

```
$ gh api repos/feigi/claude-config/labels --jq '.[].name'
bug documentation duplicate enhancement good first issue help wanted in-progress
invalid major minor patch question ready-for-agent ready-for-human ready-to-merge wontfix
```

Per the second comment on #29, `gh` resolves label names to ids *before* the write and fails the whole command with `labels not found: %s` — a hard failure, not a silent one. The live command was **not** run here either, since it would create an issue or a label in this repo; absence was verified read-only with the `gh api` call above.

Follows the in-tree precedent in `docs/agents/triage-labels.md`: state plainly that this tracker has none and tell the reading skill to skip the step rather than invent a string. No new convention is minted.

Skipping is safe rather than lossy, which is why the label is dropped instead of demanding the maintainer create it:

- `skills/wayfinder/SKILL.md:23` explicitly defers to *this file's* "Wayfinding operations" section for tracker specifics, so this doc is the authority the skill consults.
- The map reaches `/wayfinder` as a URL or number (`SKILL.md:118`), not by label query, so `wayfinder:map` is not load-bearing for discovery.
- Children are found as sub-issues, which `:46` here already specifies.

Only the ticket *type* (`research`/`prototype`/`grilling`/`task`) needed a new home, since nothing else carried it. It moves into the child body, reusing the `Part of #<map>` body-fallback convention already established in this file.

## Not changed

`CLAUDE.md:144` — "Neither exists yet; they get created lazily" names no nonexistent skill and stays true under the corrected `domain.md`. Left alone.

## Testing

Docs only.

```
$ ./agent-test skills/fleet/scripts/*.test.mjs
ℹ tests 97
ℹ pass 97
ℹ fail 0
```

Note for the fleet: `./agent-test skills/fleet/scripts/` (directory argument) fails with `MODULE_NOT_FOUND` on node v26.5.0; the glob form above is the one that runs.


feigi:
## Review complete — one fix applied, four findings deferred

**Fix pushed: `6b950ee`.** `docs/agents/issue-tracker.md:47` quoted the wrong `gh` error string. This is the correction-ticket failure mode: a PR that removes two false claims shipped a third.

The doc said `gh issue create --label <unknown>` fails with `labels not found: <name>`. That string is real in the binary but belongs to `gh discussion` (`pkg/cmd/discussion/shared/labels.go:33`). `gh issue create` takes a different path — `create.go:404` → `pkg/cmd/pr/shared/params.go:106-109` (`could not add label: %w`) wrapping `api/queries_repo.go:812-814` (`'%s' not found`) — so the real message is `could not add label: 'wayfinder:map' not found`. Pinned to tag `v2.96.0`, the installed version.

The claim came in from the second comment on #29, whose `strings "$(which gh)"` probe searched the whole binary and matched the wrong command family. Corrected at the source in #29 so it does not propagate a third time.

The **ordering** half of the claim is correct and was left intact: `AddMetadataToIssueParams` (`create.go:404`) runs before `api.IssueCreate` (`create.go:410`).

### Verified correct — please don't "fix" these

Everything else in the diff holds up:

- **Label absence** — `gh api repos/feigi/claude-config/labels` returns 16 labels, no `wayfinder:*`.
- **`/domain-modeling` absent, but `/improve-codebase-architecture`, `/grill-with-docs` and `/wayfinder` all exist.** A `grep -RIn` inside the worktree finds none of them and that clean negative is **wrong** — `skills/*` is gitignored bar `caveman-compress` and `fleet`; the real ones are at `~/.agents/skills/`. Naming the reachable skill instead of deleting the guarantee was the right call.
- **The `/improve-codebase-architecture` side-effect clauses** — both check out against `SKILL.md` §3: "Add the term to `CONTEXT.md`. Create the file lazily if it doesn't exist." and the ADR offer on a rejected candidate with a load-bearing reason.
- **`domain.md:13`'s conditional is non-vacuous** — three installed skills really do defer to `/domain-modeling` (listed in #29).
- **The `/wayfinder` consolation claim** — verified rather than assumed: `skills/wayfinder/SKILL.md:120` takes a map "as URL or number", and `wayfinder:map` appears only at `:21` and `:113`, always asserted or created, never queried. So dropping the label genuinely costs nothing for discovery.
- **Precedent conformance** — `docs/agents/triage-labels.md`'s shape is followed, down to "rather than inventing a string". No second convention minted.
- **`CLAUDE.md:144` correctly left alone** — it names no skill, so it stays true under the corrected `domain.md`.

### Deferred → #29 ([comment](https://github.com/feigi/claude-config/issues/29#issuecomment-5133268036))

1. Three installed skills still delegate to the absent `/domain-modeling` (`improve-codebase-architecture:66`, `grill-with-docs:7`, `triage:76`) — all outside this repo.
2. `skills/wayfinder/SKILL.md:113` still instructs `label wayfinder:map`; `:25` delegates only *where things live* to this tracker doc, not labelling, so a `/wayfinder` run still hits the failure.
3. `domain.md:11` "creates them lazily" over-reaches on `CONTEXT-MAP.md` — inherited from the vendor template, not introduced here.
4. `/grill-with-docs` dropped from `domain.md:11` though it is a real second route.

### Testing

`./agent-test skills/fleet/scripts/*.test.mjs` → **97 pass, 0 fail**, before and after the fix. This repo has no `.github/workflows/`, so there are no checks to bind.

Note the runner needs the glob form: `./agent-test skills/fleet/scripts/` dies with `MODULE_NOT_FOUND` under Node 26 — already tracked as #42.

feigi:
## Round 2 — fan-out settled, three follow-ups applied in `6c639fa`

Both specialists (`code-reviewer`, `comment-analyzer`) reported. They reached the CRITICAL error-string finding **independently of each other and of me**, with the same minimal fix already shipped in `6b950ee`. Three further findings warranted edits:

1. **`issue-tracker.md` — the override was left implicit.** The "little is lost" paragraph was true but silently contradicted three live instructions in the skill it names: `wayfinder/SKILL.md:21` ("labelled `wayfinder:map` — the canonical artifact"), `:65` ("Each ticket carries a `wayfinder:<type>` label"), `:113` ("Create the map (label `wayfinder:map`)"). `triage-labels.md` closes exactly this gap in its final line; now this file does too.
2. **`domain.md:13` — the one absence claim with no scope behind it.** `triage-labels.md` cites the 404s, `issue-tracker.md:44` cites the labels call, this cited nothing. Now names the two paths checked. `in this environment` deliberately kept — labels are repo-scoped, skills are not, and both specialists flagged that distinction as correct.
3. **`domain.md:11` — antecedent overshoot.** "creates them lazily" took its referent from the bullets above, which include `CONTEXT-MAP.md`. `grep -c CONTEXT-MAP ~/.agents/skills/improve-codebase-architecture/SKILL.md` → `0`. Narrowed to what the skill does.

### Specialist disagreement, adjudicated

They split on `/grill-with-docs`, dropped from `domain.md:11`. `code-reviewer` rated it IMPORTANT: three references were dropped, not two, and this one **exists**. `comment-analyzer` rated it clean: its entire body is *"Run a `/grilling` session, using the `/domain-modeling` skill"* and **both** referents are absent, so it creates nothing on its own.

Ruling with `comment-analyzer`, and `code-reviewer` concedes the substance ("dropping it is defensible — but not for the reason the title gives"). Verified: `/grilling` is absent from both skill roots, so the shim is inert. The shipped doc is correct either way and `:13`'s generic clause already covers it; naming an inert shim would add a line and change no reader behaviour. What is imprecise is `cdb9c93c`'s **title** — "tooling this repo doesn't have" undersells it, since the shim does exist. Not worth rewriting a pushed commit; recorded on #29 instead.

Also noted, not fixed: `cdb9c93c`'s body cites `skills/wayfinder/SKILL.md:118` for the URL-or-number claim; the sentence is at `:120` (`:118` is the `### Work through the map` heading). Commit-body only, already pushed.

### Deferred → #29 ([comment](https://github.com/feigi/claude-config/issues/29#issuecomment-5133308060))

**The strongest finding of the review is out of this PR's scope:** `~/.agents/skills/setup-matt-pocock-skills/` is the generator for `docs/agents/*.md`, and its templates still hold the **byte-identical pre-PR text** — `git show origin/main:docs/agents/domain.md | diff - ~/.agents/skills/setup-matt-pocock-skills/domain.md` exits 0. A re-run silently overwrites every fix in #21 and #41, and `SKILL.md:116` calls re-running routine. Plus: `/domain-modeling` was one of **five** absent referents, and `issue-tracker.md:53` documents `gh api` where installed gh has `--parent`.

### Testing

`./agent-test skills/fleet/scripts/*.test.mjs` → **97 pass, 0 fail**, after each of the two commits. No `.github/workflows/`, so no checks to bind.

feigi:
## Finisher pass — deferrals filed, `ready-to-merge` added

Head verified `6c639fa4490002197037a3824ce9561c1f2a0adc`, reachable on `origin/docs/29-agents-vendor-templates` (branch tip). Suite from the worktree: **97 pass, 0 fail, 0 skipped** (`./agent-test skills/fleet/scripts/*.test.mjs`). Release labels: `patch` only.

### Deferrals were recorded but never filed

The seven deferrals from this review existed only as comments on #29 — and this PR body says `Closes #29`, so merging would have closed the issue and buried them. `review-and-fix.md:10` requires each deferred finding to be its own `gh issue create`; commenting is permitted only on "an existing follow-up", and #29 is this PR's own source issue, not a follow-up. Filed now:

| Deferral | Issue |
|---|---|
| 1 + K — five absent skill referents, three delegating call sites | #49 |
| 2 — `wayfinder/SKILL.md` still instructs `wayfinder:*` labels | #50 |
| J — generator templates revert every `docs/agents` fix | #51 |
| L — `issue-tracker.md:55` documents `gh api` over native `--parent` | #52 |

1 and K were merged into #49: K explicitly supersedes 1 ("Defect 2 in the opening comment scoped this to `/domain-modeling`; the real surface is five"), and filing both would duplicate.

### Two deferrals deliberately not filed

Each claim was re-verified against the tree before filing, per `run-team/SKILL.md:265` — the issue body is a lead, never a citation. Two did not survive:

- **Deferral 3** (`domain.md:11` "creates them lazily" over-reaches on `CONTEXT-MAP.md`) — already fixed at this head. `:11` now ends "Nothing creates `CONTEXT-MAP.md`." Filing it would file a non-defect.
- **Deferral 4** (`/grill-with-docs` dropped from `domain.md:11`) — withdrawn by the review itself in the later comment K: "Dropping it from `domain.md:11` was correct."

With #49–#52 filed, #29's remaining surface is tracked independently, so `Closes #29` no longer loses work on merge.

Issue #46, opened 2026-07-30
run-team/SKILL.md:55 describes inflight.sh's PR probe as open/closed, which is wrong in both directions


feigi:
## Problem

`skills/fleet/skills/run-team/SKILL.md:55-56`:

> It runs all three probes (open/closed PRs, remote heads, local worktrees + branches) so a partial one cannot read as free.

`open/closed PRs` misdescribes probe 1 in both directions:

- **Too broad** for the branch-matched signal, which filters `state == "OPEN"` (`inflight.sh:73`) precisely so a closed or merged PR cannot make a ticket read as taken forever.
- **Too narrow** as a description of the whole probe, because probe 1 also queries `closedByPullRequestsReferences`, which is not full-text search over PR state at all — it is GitHub's own resolution of which PRs close the issue.

A reader who takes `open/closed PRs` literally would expect a plain `--state all` search, which is the exact probe `inflight.sh:26-29` documents as unusable.

Note the parenthetical is separately entangled with #44: `closedByPullRequestsReferences` is currently **not** state-filtered, so today a closed PR does hit. The wording should be settled after #44 decides the behavior, not before.

## Why deferred rather than fixed in PR #40

Open PRs #31 and #32 both edit `run-team/SKILL.md`. PR #40 does not touch that file and must not, so this was left alone rather than creating a conflict.

Deferred from PR #40 review

feigi:
Unblocked by #53 (closes #44). The behavior this ticket was waiting on is now settled, so the wording can be written against it:

- **A closed or merged PR no longer hits.** Both signals in probe 1 are state-filtered now, not just the branch-matched one. The parenthetical `open/closed PRs` is wrong in the direction this ticket already identified, and there is no longer a "but today a closed PR does hit" caveat to preserve.
- **There is a third state to describe, `?`.** A PR that `closedByPullRequestsReferences` returns but the `gh pr list --search` window does not contain has no knowable state, so it stays a hit and prints `?`. Measured: `closedByPullRequestsReferences` nodes are `{id, number, repository, url}` — no `state` field — which is why the state has to come from the search window at all.

So probe 1 is two signals, not one full-text search: GitHub own resolution of which PRs close the issue, plus branch-segment matching, both filtered to OPEN-or-unknown.

Line numbers moved in #53 — re-measure before citing. `run-team/SKILL.md:55-56` is untouched and still says what this ticket quotes.

PR #53, opened 2026-07-30
fix(fleet): inflight.sh state-filtered only one of its two PR signals

### LINKED ISSUE #44: inflight.sh: the linked-PR probe is not state-filtered, so a merged PR makes a finished ticket read as taken
## Problem

`skills/fleet/scripts/inflight.sh` applies its `state == "OPEN"` filter to only **one** of the two signals in probe 1.

- `by_branch` (`inflight.sh:73`) filters `p.get("state") == "OPEN"` — correct.
- `linked`, from `closedByPullRequestsReferences` (`inflight.sh:40-41`), has **no state filter at all**.

The script's own comment at `:69-72` states the intent for both:

> Only an OPEN PR is in-flight. A merged PR means the work is done; a closed, unmerged PR means it was abandoned. Either would otherwise make a finished or dead ticket read as taken forever.

`linked` does not implement that.

## Measured

Issue #8 is CLOSED, closed by MERGED PR #20:

```
$ ~/.claude/skills/fleet/scripts/inflight.sh 8
    PRs for #8: #20 (linked)   (11 full-text match(es) considered)
    HIT: pr
inflight: #8 taken=true
{"issue":8,"taken":true,"hits":["pr"],"evidence":{"pr":"#20 (linked)",...}}
exit 1
```

Exit 1 = taken, on a ticket that is finished.

The stderr line makes it undiagnosable: `#20 (linked)` prints **no state**, and is indistinguishable from an open linked PR. Contrast `inflight.sh 33`, which prints `#40 (linked), #40 OPEN (branch)`.

## Blast radius

Inert for `next-ticket`, whose step 1 lists open issues only, so a closed issue never reaches probe 1. It is **not** inert for any caller that probes an arbitrary number, and `next-ticket/SKILL.md:36` still tells the reader `open PR means unmerged` — which exit 1 does not guarantee.

Note PR #40's own commit message asserts the opposite, that the script "filters to OPEN, so a merged or abandoned PR cannot make a finished ticket read as taken forever". That claim is wrong for the `linked` half.

## Also, same file, low severity

`inflight.sh:76` (`raw=$(... python3 -c ...)`) is the only command substitution in the file without `|| die`. Under `set -eu` a failure there exits **1**, which PR #40 now documents as a contract meaning "taken". Near-unreachable — the guarded `python3` at `:63` parses the same `$pr_json` first and would `die 2` — but the `0/1/2` contract is now written down, so an unguarded path that can emit 1 is worth closing.

## Suggested

Filter `linked` by state the way `by_branch` is filtered, and print the state in the stderr evidence line so a hit is diagnosable. Add `|| die` at `:76`.

Deferred from PR #40 review

feigi:
Probe 1 of `inflight.sh` carries two signals. `by_branch` filtered `state == "OPEN"`; `linked` had no state filter at all, so any PR GitHub reports as closing the issue read as in-flight forever.

## Measured, against the live repo, before and after

```
issue #8   BEFORE:  PRs for #8: #20 (linked)     exit 1     (#20 MERGED)
issue #8    AFTER:  no PR is about #8            exit 0
issue #33  BEFORE:  PRs for #33: #40 (linked)    exit 1     (#40 MERGED)
issue #33   AFTER:  no PR is about #33           exit 0
```

Both are finished tickets. #33 is the newer instance and was not in the ticket: PR #40 merged after the ticket was filed, which turned the file own worked example into a second live case.

## Why the suggested fix could not be applied literally

The ticket suggests filtering `linked` "the way `by_branch` is filtered" — inside the jq. The data is not there. Measured:

```
$ gh issue view 8 --json closedByPullRequestsReferences
{"closedByPullRequestsReferences":[{"id":"PR_kwDOTgQPfs74h739","number":20,
  "repository":{...},"url":"https://github.com/feigi/claude-config/pull/20"}]}
```

No `state` field — the nodes are `{id, number, repository, url}`. So the state comes instead from the `gh pr list` window already fetched at `:59` and already parsed at `:63`, which does carry it. The two signals merge into that one python filter, which also absorbs the shell-side join that used to stitch them together. **No new network call.**

The comment the ticket cites already stated the rule for both signals; it just could not be implemented for `linked` as the code stood.

## The unknown case

A linked PR the search window cannot supply a state for reports `?` and stays a hit. That case is real rather than defensive: a PR linked through the Development sidebar has no `#N` text for a full-text search to match, and the search is capped at `--limit 100`. A wrong "taken" costs one skipped ticket; a wrong "free" puts two agents on the same ticket.

The stderr evidence line now prints the state for both signals, so a `linked` hit is diagnosable instead of indistinguishable from a merged one — the second thing the ticket asked for.

## Also

`|| die` on the `raw=` substitution, the file only unguarded one. Under `set -eu` a failure there exited **1**, which the exit contract now reads as "taken". No *input* can reach it: the guarded `python3` at `:63` parses the same `$pr_json` first, so anything that would break the count already died 2 there. It is reachable as a *process* though — see the follow-up commit.

## Tests

New `skills/fleet/scripts/inflight.test.mjs` — the file had no suite, which is why this shipped. 10 cases at first, 14 after the follow-up commit; `gh` stubbed on PATH, `git`/`python3`/`jq` real. 5 were red before the fix (every `linked` row) and 5 green throughout as guards on what the change moves past them: the branch half OPEN filter, the considered-count, and both `gh issue view` failure paths, which share the jq expression this edits.

The stub pipes canned JSON through the real `jq` rather than baking the reduced string into the fixture, so the script own `--jq` is under test. It hard-fails when `jq` is absent instead of letting every case report exit 2, which reads as a real red and proves nothing.

```
./agent-test skills/fleet/scripts/*.test.mjs   ->  139 pass, 0 fail   (at d827034)
```

## Corrections to the ticket body

This is a correction ticket, so its own citations were re-measured against the tree rather than repeated:

| Ticket says | Settled |
|---|---|
| comment at `:69-72` | `:67-69` |
| `next-ticket/SKILL.md:36` | correct at `:36` — this row was itself wrong; `:38` is an unrelated paragraph |
| `:73`, `:40-41`, `:76` | correct |
| PR #40 commit asserts the opposite | correct — commit `425b155`. But that PR own follow-up `48ab2cd` already recorded the refutation and filed it, so the PR did not leave the claim standing |

## Scope

`inflight.sh` and its new test only. Nothing else consumes the `(linked)` string or `evidence.pr` — searched the whole worktree excluding `.git/` and `.worktrees/` for `linked`, `closedByPullRequests`, and `evidence`. The one doc reference to the contract, `docs/specs/2026-07-23-fleet-plugin-design.md:165`, is a `{taken, probe, evidence}` shape table this does not change.

`run-team/SKILL.md` and `next-ticket/SKILL.md` are deliberately untouched (PRs #31, #32, #40, #41 own them). #46 already covers the doc wording and explicitly waits on this ticket to settle the behavior; commenting there rather than filing a duplicate.

Closes #44

---

## Follow-up commit `d827034` — review fixes

The state filter above keyed the lookup on a bare PR number. A closing reference can live in another repository (`Closes owner/repo#N`; the node carries `repository` for exactly this reason), where that number means nothing here:

```
linked other/fork#5, local #5 MERGED
BEFORE  taken=false exit=0   (the fork's PR silently inherited MERGED)
AFTER   taken=true  exit=1   "other/fork#5 ? (linked)"
```

A regression this PR introduced, in the dangerous direction — an open PR in a fork closing the ticket, reported free. Now keyed on `url`, which both sides already carry, so it is a field swap and not a new call. URLs are globally unique, so a foreign PR falls out of the window into `?`, which already means "stays a hit".

Also measured and fixed: `state.get(num, "?")` returned `None` on a present-but-null state, contradicting its own comment; the comment blamed `closedByPullRequestsReferences` for a missing `state` field that is really `gh issue view --json`'s projection (the GraphQL nodes do carry it, at the cost of a third round-trip); the error classifier reported repository-level failures as "issue #N does not exist"; and `gh pr list` discarded its stderr, making rate-limited, unauthenticated and offline indistinguishable.

Tests 10 → 14. Three are red against `c237184` (cross-repo, null state, repository failure). The fixture also stops inheriting `GIT_DIR`/`GIT_WORK_TREE`/`GH_ISSUE_ERR`: an ambient `GIT_DIR` outranks `cwd` and retargets probe 3, and if its directory is named for the ticket the failures look exactly like the real bug.

`#8` and `#33` still exit 0; `#44` still exits 1 on `#53 OPEN (linked), #53 OPEN (branch)`.


feigi:
Controller hold: `ready-to-merge` removed. The label was added while head is still `c237184` — no fix commit was pushed — and a blocking finding is outstanding.

**Blocker (silent-failure-hunter, verified):** this PR introduces a new wrong-"free". `inflight.sh:77` keys the state map on a bare PR number and `:85` looks a *linked* PR's number up in it. Linked PRs can be cross-repo (the node carries `repository`; `Closes owner/repo#N` is legal), so a cross-repo linked PR colliding with a local MERGED/CLOSED number inherits the wrong state and is filtered out of the hits.

```
GH_ISSUE_JSON='{"closedByPullRequestsReferences":[{"number":5,"repository":{"nameWithOwner":"other/fork"}}]}'
GH_PR_JSON='[{"number":5,"state":"MERGED","headRefName":"unrelated"}]'

BEFORE  taken=true  exit=1   "#5 (linked)"
AFTER   taken=false exit=0   ""
```

An open PR in a fork is closing the ticket and the fleet is told it is free — two agents on one ticket, the failure the 0/1/2 contract exists to prevent. No marker distinguishes it from a genuine merged-PR filter.

The rest of the review is strongly positive: 4 specialists ran, the suite mutation-tests clean (each mutation caught by exactly one correct test, comment-only control stays green), 135 pass at head, and the new suite goes 5 red against the pre-fix script. This one finding is what blocks.

Stopgap: key the map on `owner/repo#number` and treat a linked PR from another repo as an unconditional hit. Not the full `gh api graphql` rewrite — that is a design change.

Issue #42, opened 2026-07-30
claim-ticket.sh: the emitted agent-test runner dies on a directory argument under Node 26


feigi:
## Problem

`skills/fleet/scripts/claim-ticket.sh` emits the per-worktree isolation runner as:

```sh
#!/bin/sh
export TEST_COMPOSE_PROJECT=ab-<N> TEST_POSTGRES_PORT=16<N> TEST_OLLAMA_PORT=22<N>
exec node --test "$@"
```

Passing a **directory** — the natural way to say "run this suite" — fails on Node 26, before any test runs:

```
$ node --version
v26.5.0

$ ./agent-test skills/fleet/scripts/
test at skills/fleet/scripts:1:1
✖ skills/fleet/scripts (32.740541ms)
  'test failed'
    Error: Cannot find module '.../skills/fleet/scripts'
    code: 'MODULE_NOT_FOUND'
$ echo $?
1
```

The glob form works:

```
$ ./agent-test skills/fleet/scripts/*.test.mjs
ℹ tests 97
ℹ pass 97
ℹ fail 0
$ echo $?
0
```

Reproduced on a pristine `git archive origin/main` copy, so it is the invocation, not any branch's regression.

## Why it matters

It **fails in the safe direction** — exit 1, loud, unmistakable — so this is not the silent-green class of defect that #27 closed. The cost is misattribution, and it is not hypothetical: during a fleet run a member passed the directory form, got a red, and spent a detour establishing that the red belonged to the invocation rather than to the diff it was reviewing. A reviewer that does *not* establish that files the red as a finding against an innocent PR.

The surface is wider than one member's habit. `workflows/review-pr.js:113` defaults `testCmd` to `./agent-test` and hands it to the whole review fan-out, and `:241` instructs specialists to use it and "never a bare test command". Specialists inherit no environment and no worked example, so the argument they choose is guesswork.

`skills/fleet/skills/run-team/SKILL.md` does say to brief members with `./agent-test <file>` — singular, and correct. But `<file>` reads naturally as "a path", the directory form is the ergonomic one, and nothing in the emitted runner or its surrounding text says a directory is rejected.

## Suggested fix

Make the runner accept what callers actually pass. A directory argument should expand to the test files under it rather than reaching `--test` as a module specifier — the same `\.(test|spec)\.[cm]?[jt]sx?$` shape `claim-ticket.sh:63` already greps for when it decides whether a runner may be emitted at all.

[...]

PR #57, opened 2026-07-30
fix(fleet): the workflow path reviewers are told to prefer is unreachable from a member

### LINKED ISSUE #43: fleet reviewers are told to prefer the review-pr workflow, which is unreachable from inside a dispatched member
## Problem

`skills/fleet/commands/review-and-fix.md`, **Specialists**, opens with:

> **Prefer the workflow.** `Workflow({name: "review-pr", args: {pr, branch, worktree, testCmd, scratch}})` runs the fan-out with `agent()` returning **into the script**, so no report can go undelivered — the failure that cost one fleet five reports on one PR and four on another.

That file is read by a **dispatched fleet member**, and `Workflow` is not in a member's tool environment. Reported by a reviewer this run:

> `Workflow` is not a tool in my environment and `ToolSearch("select:Workflow")` returned no match.

So the preferred path is unreachable in the only context where the instruction is read. Every fleet reviewer spends turns discovering this before falling back.

## Why the existing caveat does not cover it

The paragraph continues: *"Needs explicit opt-in to multi-agent orchestration; without it, dispatch manually under the rules below."* That frames the blocker as **authorization** — something a controller can grant in the prompt. It is actually **availability**, which no amount of authorization fixes. A controller reading this reasonably writes "this is explicit authorization for multi-agent orchestration" into the reviewer prompt and believes it has unblocked the workflow path. That is exactly what happened this run, in three separate reviewer prompts.

## The consequence is not just a wasted turn

The two paths have **opposite delivery semantics**, and `run-team/SKILL.md` is emphatic that the wrong one must not be briefed:

- **Workflow path:** `agent()` returns into the script; no relay occurs; a reviewer told to block on an unrelayed report would strand its verdict permanently.
- **Hand-dispatch path:** reports surface to the controller; the relay is the controller's obligation; a reviewer that settles a dimension without asking ships a claim its own specialist may refute.

`run-team/SKILL.md` instructs the controller to put the relay/blocking language in hand-dispatch prompts and to keep it **out** of workflow-path prompts. Since the workflow path is unreachable from a member, a controller that believes it authorized the workflow writes the **wrong** brief — omitting the relay obligation on the only path that actually runs. This run produced exactly that: two reviewer prompts told a member "there is no relay step, nothing to wait on, and nobody to ask for a missing report", on the path where the relay is the sole delivery mechanism. Both had to be corrected mid-flight.

That is the failure family already recorded against this repo — a reviewer ruling a relayed dimension "not covered by a specialist" and shipping a headline claim the report refuted.

## Desired behavior

State **availability**, not just opt-in, at the point of the recommendation, and say which caller each form is for. Concretely, the workflow form is usable by a controller or a standalone invocation; a dispatched member's path is hand-dispatch, and the relay obligation therefore always applies to fleet reviewers.

Worth settling while there: whether `run-team/SKILL.md`'s guidance to the controller — which currently spends a paragraph on keeping the blocking rule out of `review-pr.js`-path prompts — should say instead that member-dispatched reviews are always hand-dispatch, which would remove the branch rather than document it.

## Note

Found by the `review-pr-40` reviewer during a fleet run; it fell back to hand-dispatch immediately and correctly. Controller confirmed the mis-briefing in its own three reviewer prompts. Filed rather than fixed in-tree because the run's controller must not edit skill files in the maintainer's main checkout.

feigi:
Closes #43.

`review-and-fix.md:33` opened with **"Prefer the workflow"** and handed every dispatched reviewer a `Workflow({name: "review-pr", ...})` call it cannot make.

## Settled in this repo, this session

A `general-purpose` subagent, dispatched exactly the way a reviewer is:

```
ToolSearch({query: "select:Workflow", max_results: 1})
-> No matching deferred tools found
```

`Workflow` is absent from the member's **direct** tool list and absent from its **deferred** list — so it can neither be called outright nor loaded.

**Control:** the same probe for `Agent` returned a full `<functions>` definition block, and `Agent` was already directly available. Both lookup mechanisms work, so the `Workflow` absence is a real absence and not a broken probe.

The instruction was not merely dead — it **cost a turn**. A reviewer reads "prefer", tries to resolve the tool, fails, and only then falls back to the hand-dispatch rules that were the real instruction all along.

## What changed

**Hunk 1 — ownership of the two paths is now explicit.** A controller running this file itself should prefer `Workflow({name: "review-pr", ...})`; a dispatched reviewer skips it and hand-dispatches under the rules below. The *reason* to prefer it survives in the controller's half — `agent()` returns into the script, so no report can go undelivered — and that is also stated as why the relay rules exist on the hand-dispatch path.

That gap is not theoretical. In the run that produced this PR, `review-pr-53` hand-dispatched four specialists and **all four reported to the controller instead**; the reviewer declared every dimension unrun. Two of those reports carried findings that changed the outcome, including a blocking one.

**Hunk 2 — the specialist-runner bullet carried a war story from another repo with no runnable command for this one.** On `main` it read *"Hand them the worktree's `./agent-test` runner"*, which is the collision this repo's own runner cannot escape. The variant that named `npx vitest run --config vitest.ci.config.ts <file>` — *"the config CI's `check` job runs"* — exists only in the maintainer's uncommitted working copy and is **not** this PR's base; an earlier revision of this description mis-attributed it to `main`. Either way the escape it points at is absent here. Measured: no compose file, no `globalSetup`, no vitest, no `.github/workflows/`. A specialist handed that config gets `Cannot find module`, which reads as a broken tree rather than a bad instruction.

The collision that bullet warns about is inert here, for a reason worth recording: `claim-ticket.sh:106` writes `TEST_COMPOSE_PROJECT`, `TEST_POSTGRES_PORT` and `TEST_OLLAMA_PORT` into the emitted runner, and **nothing in this repo reads any of them** — the suite is `node --test` throughout.

```
$ grep -rn 'TEST_COMPOSE_PROJECT' --include='*.sh' --include='*.mjs' --include='*.js' .
skills/fleet/scripts/claim-ticket.sh:106:export TEST_COMPOSE_PROJECT=ab-$issue …
```

Written once, read nowhere. Local instruction is now `node --test skills/fleet/scripts/*.test.mjs` — glob form, per the open directory-argument bug #42. The **rule** is kept for repos that do have a stack (N concurrent specialists sharing one runner still collide); only the **command** is made true here.

## Provenance

The two hunks were found uncommitted in the maintainer's main checkout after this run's fleet had closed — no branch, no PR, no worktree, so `inflight.sh 43` reported the ticket `taken=false` while its implementation already existed. Committing it here is what makes that work visible to the next run. The vitest correction is mine; the rest is as found.

## Verification

Suite unchanged at **139 pass / 0 fail** (`./agent-test skills/fleet/scripts/*.test.mjs`, glob form). Markdown-only change, one file, two lines.

## Review pass

A review of this PR found the defect it fixes reproduced in its own headline (`0edcda0`, +1/-1, same file):

- the bullet header said **"Give specialists the docker-free config"** while the body it introduces establishes there is no such config here — the bolded lead is what an agent skims, so it sent readers hunting for the artifact the body calls absent. Now **"a stack-free test command"**: the property, not an artifact.
- **"the escape it names"** had no antecedent — the war story is the `./agent-test` collision, which names no escape. Now "the config-swap escape it calls for".

Re-verified after the fix: **139 pass / 0 fail** (`node --test skills/fleet/scripts/*.test.mjs`). Repo has no CI — `gh pr checks 57` reports none — so there is no run to bind.

Deferred, not dropped: the two dependency-guard claims noted on #58.


feigi:
**Review complete — applied 2, deferred 2.**

No specialists dispatched: docs-only, one file, ±2 lines. Findings verified directly against the tree and the object store rather than reasoned about.

**Applied** (`0edcda0`, +1/-1, same file — see the *Review pass* section added to the description):
1. The bullet header named the very artifact its body proves absent — *"Give specialists the docker-free config"* against *"there is no `vitest.ci.config.ts` to point anyone at"*. Header now names the property: *"a stack-free test command"*. This is the PR's own defect class reproduced in the line it fixes.
2. *"the escape it names"* had no antecedent after this diff — the war story names no escape, and the sentence that did is not in this PR's base. Reworded.

Also corrected in the description: hunk 2 was described as replacing an `npx vitest run --config vitest.ci.config.ts` instruction. `git show origin/main:skills/fleet/commands/review-and-fix.md` shows the base said *"Hand them the worktree's `./agent-test` runner"* — the vitest variant lives only in the maintainer's uncommitted working copy.

**Deferred** → #68 — `review-and-fix` steps 4/6 mandate a CI gate this repo cannot satisfy; `ci-state.mjs --pr 57` exits 2 on `ENOENT .github/workflows/ci.yml`. Filed `needs-triage`: three approaches, none decided.

**Verified true** — every factual claim the diff makes: no `.github/workflows/`, no compose file, no vitest, no `globalSetup`; `claim-ticket.sh:106` writes `TEST_COMPOSE_PROJECT`/`TEST_POSTGRES_PORT`/`TEST_OLLAMA_PORT` and nothing in the tree reads them; #42 open and documents the directory-argument failure with the glob form as the workaround.

**Suite:** 139 pass / 0 fail (`node --test skills/fleet/scripts/*.test.mjs`) on the pushed head. No CI in this repo — `gh pr checks 57` reports none — so there is no run to bind and no green to wait for.

Heads-up, not a finding: your main checkout carries an uncommitted edit to this same file holding the `npx vitest` variant. Merging this will not touch it — reconcile by hand.

Issue #58, opened 2026-07-30
candidates.mjs dependency scan misses to-tickets' blocking edges — every chain reaches the fleet with d: []


feigi:
`skills/fleet/scripts/candidates.mjs` reduces issue bodies to dependency references with the `d:[(.body//""|scan(…))]` fragment of its jq program:

```
(?i)(?:depends on|blocked by|requires|after)\s+#\d+
```

It requires the issue number to follow the phrase inline. `to-tickets` does not write them that way.

## Reproduction

Ran the live expression against the template shapes:

| body | `d` |
| --- | --- |
| `## Blocked by\n\n- #12\n- #13` | `[]` |
| `## Blocked by\n\n- None — can start immediately` | `[]` |
| `Blocked by #99 inline form` | `["Blocked by #99"]` |
| `- A reference to #77 blocking ticket` | `[]` |

The `-` list marker is not `\s`, so the heading-plus-bullet form never matches.

## Why it matters

`to-tickets` publishes chains in dependency order with a `## Blocked by` heading and list items (`skills/to-tickets/SKILL.md:99-101`), and prefers GitHub's native sub-issue links where available (`:63`) — which are not in the body under any regex.

So every chain `to-tickets` publishes reaches the fleet with no dependencies at all. Phase 0 step 2 has nothing to drop on, and admits blocked tickets. `skills/fleet/skills/run-team/SKILL.md:75-76` names this as the failure that "bites hardest at five wide."

Raised in priority by the FIFO ordering merged in #59: `to-tickets` publishes blockers first, so a chain's members now carry consecutive numbers and render adjacent in the shortlist. The sequencing annotation added at `run-team:93-96` is currently the only guard.

## Why this is `needs-triage`, not `ready-for-agent`

The defect is verified, but *what to build* is not decided — three viable approaches with different trade-offs:

1. **Section-aware regex** — parse the `## Blocked by` section, extract every `#\d+` within it. Fixes existing and future tickets, one file, no extra calls. Fiddliest to get right in jq.
2. **Fix the producer** — change `to-tickets`' template to the inline `Blocked by #12` form. One line, but only fixes tickets published afterwards, and `skills/to-tickets/` is an untracked vendored copy.
3. **Read native sub-issue links** — `gh api repos/:owner/:repo/issues/:n/sub_issues`. Complete, but one round trip per candidate, against the token-efficiency constraint the current design leans on.

Needs a decision before implementation.

## Context

[...]

PR #59, opened 2026-07-30
fleet: gate ticket admissibility on decided, not on size

feigi:
Phase 0 admitted a ticket only if `sizing-a-ticket` called it **light** — a rule that ANDed open-endedness together with file count. A large but fully-specified ticket was excluded forever, and the sizing subagent ran per candidate per wave with its verdict deliberately discarded, then re-derived a third time by the member.

The design of record already argued for the axis this PR adopts. `docs/specs/2026-07-22-run-team-agent-fleet-design.md` justified the filter as *"too open-ended to hand a background implementer"* — ambiguity — but implemented it by importing a row table that ANDs ambiguity with `>~3 files`. Intent was ambiguity; mechanism was size.

## What changes

Admission asks **is what to build decided?** — never how big it is. `ready-for-agent` is trusted as what it is: a maintainer-approved triage verdict, reached with the maintainer present and carrying an Agent Brief.

- **Phase 0** judges *decided?* during the full-issue read it already performs for `Out of scope` sequencing. No sizing subagent, no second fetch. Torn cases are surfaced in a third **unsure** group rather than guessed.
- **Phase 2** members read the issue before touching code and bail if it is still undecided, demoting by cause: `needs-triage` when under-specified so `/triage` can return it as `ready-for-agent`, `ready-for-human` when it needs human hands. Heavy is no longer a bail reason.
- **Ordering** is a script sort — `candidates.mjs` returns candidates oldest-first. Dependencies stay a constraint, never a ranking input.
- **The tick stages a wave**, it does not re-vet a ticket. The multi-select and its stop point remain; only the meaning changes.
- **to-spec specs are filtered by shape.** `/to-spec` publishes a whole spec as one issue stamped `ready-for-agent`; a spec is `to-tickets`' input, not a claimable ticket.

Design: `docs/specs/2026-07-30-fleet-trust-the-label-design.md`.

## Two opposite tie-breaks, deliberately

`sizing-a-ticket` answers *how much process* — torn → take the heavier row. Phase 0 answers *is it decided* — torn → surface. Both are now stated in both files, each naming which is its own. Nothing fails at runtime if an agent applies the wrong one; the old gate simply returns. This is the change's known failure mode and the reason for the redundancy.

## Review findings fixed

A whole-branch review found one Critical and two Important:

- **Critical:** "proceed on either row" sent unattended members into `superpowers:brainstorming`, whose `<HARD-GATE>` needs maintainer approval a member cannot obtain. Fleet members now enter the heavy path at `superpowers:writing-plans` — a decided ticket is already brainstormed, and the Agent Brief is that output. Solo `next-ticket` keeps the full path.
- **Important:** the 2026-07-22 design still stated the deleted gate imperatively, and `run-team:14` points agents at it. Retired with a supersession header.
- **Important:** `next-ticket:28` kept "Ticket unblocking others ranks higher" while `:48-50` deleted it — and step 2 performs the 3–5 cut, so the surviving rule had operational effect.

## Tests

`skills/fleet/scripts/candidates.test.mjs` is new — the script had no coverage. The `gh` stub runs real `jq` against a fixture with the expression under test; stubbing it to return pre-reduced JSON would leave the new predicate untested. Four tests: spec dropped, drop logged by number, `spec` key stripped from the payload, results ascending whatever order `gh` returned.

129 passing, 0 failing.

## Known limitations

- The spec filter is a plain regex with no markdown awareness, so a ticket quoting `## User Stories` inside a fenced block is dropped too. Accepted: every drop is logged by number, so it fails loudly. Named in a `ponytail:` comment.
- Heavy fleet members now run `writing-plans` → `subagent-driven-development`, spawning grandchildren the controller does not bound (`run-team:430-431` bounds members, not live agents). Correct per the design, untested under load — worth watching on the first heavy ticket in a wide fleet.
- Skill prose has no test coverage. The 129 tests are a regression gate on the script, not evidence about the prose.
- The final scoped re-review of the fix wave could not be obtained — two reviewers completed and went idle without delivering. The five findings were verified by hand instead: coherence across all three files the Critical touched, the old spec's renumbering, and the solo path left intact. That is weaker than independent review.

## Follow-up

#58 — `candidates.mjs`'s dependency scan misses `to-tickets`' blocking edges, so every chain reaches the fleet with `d: []`. Verified with a repro, no ordering dependency with this PR, and raised in priority by the FIFO ordering here.

https://claude.ai/code/session_01KwQx6q1m6VENzkTiZALtP1


feigi:
**Review complete — applied 0, deferred 1 (already filed).**

No specialists dispatched: verified directly, with mutations rather than reasoning.

**Mutation-verified, with negative controls** — both new behaviours discriminate, and each mutation fails only the tests it should:

| mutation | result |
|---|---|
| remove `rows = dropSpecs(rows)` | 3 spec tests ✖, ordering test ✔ |
| remove `rows.sort((a,b) => a.n - b.n)` | ordering test ✖, 3 spec tests ✔ |

The stub-real-`jq` choice is load-bearing and correct — the predicate under test lives in the `--jq` expression, so a stub returning pre-reduced JSON would have tested nothing. **129 pass / 0 fail** on a clean `pr59` worktree, matching the description.

**Verified true**, each against source rather than the description:
- `superpowers:brainstorming` really does carry `<HARD-GATE>` — *"until you have presented a design and the user has approved it"* — and `writing-plans` carries no equivalent. The Critical fix is grounded.
- `to-spec/SKILL.md:31` has `## User Stories` in its template and `:19` stamps `ready-for-agent` with *"no need for additional triage"*. No ticket template carries the heading.
- Phase 0's 4–7 → 4–6 renumbering leaves no dangling reference: the surviving `step 2` / `step 6` / `step 7` citations in `run-team` all resolve correctly.
- Every remaining "heavy row" mention across the tree is a deliberate process-depth use, not a missed admissibility leftover.
- The 2026-07-22 design carries the supersession header, and `run-team:14` — which points agents at it — now lands on that header.

**Deferred** → already filed, no duplicate created:
- **#60** — `dropSpecs` runs after the `--allow-fallback` emptiness test, so an all-specs labeled queue suppresses the fallback. Filed by an earlier review of this PR.
- **#58** — commented, not re-filed. `candidates.mjs:130` and `next-ticket:46` both justify deleting the "unblocks #X" ranking with *"the scan already drops anything with an open blocker"*, which #58 proves never fires (`d: []` for every `to-tickets` chain). The deletion stays right — the old ranking read the same empty `d` — only the stated reason is false, and #58 landing makes it true. Added to that issue's acceptance rather than shipping a qualifier this repo would delete a week later.

No CI in this repo — `gh pr checks 59` reports none — so there is no run to bind.

Issue #48, opened 2026-07-30
docs/specs/2026-07-23-fleet-plugin-design.md: three stale assertions (:208 section list, :225 caller count, :428 'all 33 patterns resolve')


feigi:
Deferred from PR #32 review. `docs/specs/2026-07-23-fleet-plugin-design.md` carries three assertions about the fleet skill tree that no longer hold. One is falsified by PR #32; two are pre-existing and were found while checking whether that PR falsified anything in files it does not touch (`review-and-fix.md:58` instructs reviewers to hunt exactly this).

### 1. `:208` — the `correction-tickets.md` row enumerates three sections; the file now has four

```
git show 70acb5c6:docs/specs/2026-07-23-fleet-plugin-design.md | sed -n '208p'
| `correction-tickets.md` | the four-for-four finding, why the mechanism is the ticket's framing, the implementer's clause-by-clause duty |

git show 70acb5c6:.../references/correction-tickets.md | grep -n '^##'
 5:## The four-for-four finding
 9:## The mechanism is the ticket's framing, not implementer sloppiness
13:## Second mechanism: minted, not inherited — scope creep in the prose      <-- added by PR #32
23:## The clause-by-clause duty
```
The row lists the sections in file order and PR #32 inserted a fourth between #2 and #3. The sibling rows in that table (`ci-and-staleness.md`, `member-lifecycle.md`) read as exhaustive enumerations, so this one now under-reports. **Caused by PR #32** — deferred rather than folded in because PR #31 is in flight and the spec is not on either PR's path.

### 2. `:225` — "both callers point at it" is one caller

```
git grep -n "correction-tickets" 70acb5c6
docs/specs/2026-07-23-fleet-plugin-design.md:141,208,225
skills/fleet/skills/run-team/SKILL.md:271: ... See references/correction-tickets.md.
```
Only `run-team/SKILL.md:271` points at the reference. `skills/fleet/commands/review-and-fix.md` carries the correction-ticket guidance **inline** at `:51` and never names the file — so it is not a caller in the sense the row claims. Pre-existing.

### 3. `:428` — "all 33 patterns resolve" — three do not

```
while IFS= read -r l; do [ -z "$l" ] && continue
  grep -qF "$l" <(git show 70acb5c6:skills/fleet/skills/run-team/SKILL.md) || echo "MISS: $l"
done < docs/fleet-rule-manifest.txt

MISS: newest run on a branch is frequently
MISS: Say the specialist count
[...]

PR #71, opened 2026-07-30
fix(fleet): 'both of these' undercounted three implementer prompt blocks

### LINKED ISSUE #13: run-team: 'both of these verbatim' undercounts three blocks — the dropped one is the commit-incrementally block :128 calls not optional

feigi:
`run-team/SKILL.md` told the controller each implementer prompt carries "**both of these** verbatim" and then listed **three** blockquotes. A controller following that literally copies two and drops the `Commit incrementally as you go` block — the one the very next paragraph calls not optional, and whose absence leaves a member's diff uncommitted and invisible to the `worktree-audit.sh` committed-vs-uncommitted split that decides whether a replacement redoes or destroys work.

## What changed

- **Stopped counting instead of correcting the number.** "both of these" → "each of these", the phrasing the issue itself offers. A fourth block can be added without re-staling the line.
- **Named the block semantically instead of positionally.** "That third block is not optional" → "The commit-incrementally block is not optional", so that sentence no longer depends on a count either. Both count claims in the file are now gone rather than merely made to agree.
- **Replaced the design spec's partial copy with a pointer.** `docs/specs/2026-07-22-run-team-agent-fleet-design.md` quoted two of the three blocks. It now points at the skill as the authoritative copy — the resolution the issue lists as acceptable. The spec is reachable (`run-team/SKILL.md:14` sends readers to it as "Rationale"). Its copy had drifted **by omission** — it reproduced two of the three blocks under a "carries, verbatim:" sentence implying exhaustiveness. It had *not* drifted textually: `cmp` of both deleted blocks against their SKILL.md counterparts exits 0.

All three prompt blocks now exist in exactly one place **as blockquotes**. That is narrower than repo-wide uniqueness — see the first bullet under "Known gaps" below.

## Verification

- **32** of the 33 patterns in `docs/fleet-rule-manifest.txt` resolve repo-wide — not 33. `Say the specialist count` resolves nowhere outside the manifest file itself; two others (`newest run on a branch is frequently`, `git cherry origin/main`) resolve outside `run-team/SKILL.md`, at `run-merge-bot.md:90` + `ci-state.mjs:103` and `references/reaping.md:24`. **This PR caused none of it**: the same 32 resolve against `origin/main`, per-pattern hit counts are identical between base and head, and no manifest pattern matches any line this PR changes. Tracked in #48, which measured against `run-team/SKILL.md` only — that scope difference is the whole of the apparent disagreement.
- Of those 33, the 30 that resolve inside `run-team/SKILL.md` are byte-identical to their pre-change baseline (the same 30 at base and at head).
- Every blockquote under `## Phase 2 — dispatch implementers` is a prompt block — three groups, lines 145-158, in a section spanning 140-181 — so the spec's new reference is exact **today**. The pointer is nonetheless prospectively over-broad, which is why the "cannot drift from them" clause was removed rather than kept.
- `./agent-test skills/fleet/scripts/*.test.mjs` → 145/145 pass. Docs-only change, so this is a regression guard, not coverage of the diff.
- Line numbers in the issue were stale against `origin/main` (`:111`→`:143`, `:128`→`:160`, spec `:240`→`:249`); all re-measured with `grep -n` before editing.

## Known gaps

- **The drift class is reduced, not removed.** `docs/agents/issue-tracker.md:34-37` carries the issue-read block's first two sentences word-for-word (a strict prefix of `run-team/SKILL.md:149-154`, differing only in line wrapping), with weaker paraphrases at `next-ticket/SKILL.md:41` and `sizing-a-ticket/SKILL.md:11`. So blocks 1 and 3 are genuinely unique repo-wide; block 2's text still lives in two places. Filed separately rather than widened into this PR.

## Out of scope, deliberately left

- The content of any individual block, per the issue.
- `docs/specs/2026-07-23-fleet-plugin-design.md:365` carries its own stale figures about this spec ("4 dead `~/.claude/commands/…` paths and 7 bare `/run-team` references") and cites `run-team` SKILL.md:13 where the "Rationale" line is at :14. Re-measured: dead `~/.claude/commands/…` paths are now **0**; bare `/run-team` *command* invocations are **0**, while the literal substring `/run-team` occurs **2** times — up from 1, because this PR's own pointer at spec:251 names the path `~/.claude/skills/fleet/skills/run-team/SKILL.md`. So that row is not quite "unchanged by this PR" under the literal reading. Pre-existing staleness; filed separately.
- That same file's `:367` row says every `run-team.md:NNN` citation in it is "+2 lines" off. `run-team/SKILL.md` gains a net +1 line here (628→629, the shift starting at old `:164`), so for cited lines at `:164` and beyond the offset is now +3. Those citations already resolved to nothing at `origin/main`, so this PR breaks nothing that worked. Filed separately.

Closes #13


feigi:
## Review deferrals

Two findings from this review were deferred rather than folded in, per the "do not widen the PR" call. Neither is filed against #13 — this PR's `Closes #13` would bury them on merge.

- **#77** (new, `needs-triage`) — `2026-07-23-fleet-plugin-design.md` migration-debt rows `:365` and `:367` carry stale figures. `:367`'s "+2 lines" becomes +3 for citations at `:164` and beyond once this PR lands. Those citations already resolved to nothing at `origin/main`, so this PR breaks nothing that worked.
- **#23** (comment, existing follow-up) — `docs/agents/issue-tracker.md:34-37` reproduces the issue-read block's first two sentences word-for-word. #23 already tracks the divergent-caveat problem across three skill files; this is a fourth site its table does not list, and the one case that is verbatim rather than paraphrased. Commented rather than filed new, since #23 owns the unify-or-not decision.

Not deferred here, already owned elsewhere: the manifest's one genuinely lost rule (`Say the specialist count`) is **#48**, which now carries the full 32-of-33 table.

## Applied in this PR

- `36e7f1c` — dropped `so it cannot drift from them` from the spec's new pointer. The clause was true of *content* but silently minted a **scope** claim: that every blockquote under `## Phase 2 — dispatch implementers` is a prompt block. That holds today (three groups, `:145-158`, section `140-181`) but is not guaranteed, and this same spec carries `**Superseded in part.**` banner blockquotes at `:6-20` — so a non-prompt blockquote under that heading is house style, not an edge case. The preceding clause already demonstrates the removal; asserting impossibility on top could only go false.
- PR body corrected: the manifest claim (33 → **32** of 33, with the scope difference against #48 spelled out), the "Both prompt blocks" miscount in a PR about miscounting, and the `:365` out-of-scope figures.

## Verification

- `./agent-test skills/fleet/scripts/*.test.mjs` → **145/145 pass** at `36e7f1c`, glob form. Docs-only diff, so this is a regression guard, not coverage.
- Pointer re-measured independently: `## Phase 2` at `:140`, next heading at `:182`; every `>`-prefixed line in range falls in exactly three prompt groups at `:145-147`, `:149-154`, `:156-158`. Heading byte-identical to the spec's quoted string, em-dash included.
- The removed spec copy had **not** textually drifted — `cmp` against its SKILL.md counterparts exits 0. Its drift was by *omission* (two of three blocks under a "carries, verbatim:" sentence implying exhaustiveness). The PR body now says so precisely.
- No collateral citation rot: the only numbered citation into the 07-22 spec is `2026-07-30-fleet-trust-the-label-design.md:19` → `:211-213`, above the edit site and identical at both revs.

This repo has no CI, so there is no run to bind; the label rests on the local suite plus the above.

feigi:
**Deferral list correction.** My earlier comment listed two deferrals; there are three, and one line in it was wrong.

- **#79** (new, `ready-for-agent`) — `docs/agents/issue-tracker.md:34-37` carries sentences 1-2 of the issue-read block verbatim and drops sentences 3-4 (the `## Agent Brief` authority rule and the `Respec` clause). Surfaced by the `comment-analyzer` specialist, which landed after that comment was posted. Pre-existing at `main`; not caused by this PR.
- **#23** — still cross-referenced, now pointing at #79 for the content gap. #23 keeps the caveat-wording question.
- **#77** — amended. My original framing there (":367's +2 becomes +3") was too generous to that row: its own worked example, `SKILL.md:423` for the `settings.json` prohibition, is actually at `:513`/`:514` — **~+90 off, not +2**. The stated offset was never a correct baseline. Conclusion unchanged: all nine affected citations were already broken at `main`, so this PR breaks nothing that worked.

**One specialist recommendation not applied.** `comment-analyzer` proposed the D3 sentence read "…this spec does not reproduce the block **text**." Its own finding #1 falsifies that: spec `:201-203` reproduces block 2's command string verbatim (line-wrapped), and `:326-329` carries a weaker paraphrase of the same rule. So "does not reproduce the block text" would be false as written. The shipped wording — "this spec does not reproduce the blocks", anaphoric to "the blockquoted blocks" in the preceding sentence — is true: the spec reproduces no blockquote. Keeping it. The finding itself is real and is the stronger argument for the D3 edit that shipped; only its replacement text is wrong.

No code change from any of this. Head remains `36e7f1c`.

PR #72, opened 2026-07-30
fix(fleet): a prunable worktree made release-ticket.sh permanently unanswerable

### LINKED ISSUE #37: release-ticket.sh: a prunable worktree makes the release permanently unanswerable
Deferred from PR #31 review.

`skills/fleet/scripts/release-ticket.sh:97`

When a claim's worktree **directory** has been deleted by hand but its admin files are still registered, `git worktree list --porcelain` keeps listing the entry (marked `prunable`). The dirty check then runs `git -C "$wt" status --porcelain` against a path that no longer exists, which fails, and the script dies:

```
release-ticket: cannot read the status of <path>: fatal: cannot change to '<path>': No such file or directory
```

Exit 2 — unanswerable — every time. Nothing is touched and nothing is lost, so this fails in the safe direction, but the claim can never be released by the script: the label stays, the branch stays, and the next run's in-flight probe keeps reading the ticket as taken. That is the state `release-ticket.sh` exists to clear.

Reproduced against a real throwaway repo:

```sh
git worktree add .worktrees/9-rel -b fix/9-rel origin/main
rm -rf .worktrees/9-rel          # directory gone, admin files remain
sh release-ticket.sh 9 rel fix --apply   # -> exit 2, no gh call, branch kept
```

`git worktree list --porcelain` already marks the entry `prunable`, so the fix is to treat a prunable worktree as one with nothing to check rather than as an unreadable one — the directory is already gone, so there is no uncommitted work in it to protect. `git worktree prune` (which the script already calls at the end of a successful apply) is the other half.

**Why deferred:** pre-existing in the shape of the check, not introduced by the two fixes applied in #31, and it fails closed and loudly rather than deleting anything. #31 was scoped to the false-accept direction.

feigi:
Closes #37

## The failure

A claim's worktree **directory** deleted by hand leaves its admin files registered, so `git worktree list --porcelain` keeps listing the entry (marked `prunable`). The dirty check — `release-ticket.sh:189`, not the `:97` the ticket cites — then ran `git -C "$wt" status --porcelain` against a path that is gone. It fails, and the `die` read that as an unknown answer: exit 2, every run.

Nothing is ever touched, so this failed in the safe direction — but permanently. The label stayed, the branch stayed, the stale entry stayed, and the next run's in-flight probe kept reading the ticket as taken. That is exactly the state `release-ticket.sh` exists to clear.

## The fix

A directory that does not exist holds no uncommitted work, so there is nothing there to protect — skip the check when it is gone.

That is the whole change. git already agrees at the delete: `git worktree remove` accepts the gone entry and clears the admin files (rc 0, verified against git 2.50.1), and the `git worktree prune` the script already runs at the end of a successful apply is the backstop. No other step needed touching.

## Why `-d`, not git's own `prunable` annotation

The ticket names the annotation, and it is the wrong test for this. Removing only a **live** worktree's `.git` file marks the entry `prunable` too — with the directory, and every uncommitted change in it, still sitting there:

```
$ rm .worktrees/8-x/.git      # directory and its work untouched
$ git worktree list --porcelain
worktree .../.worktrees/8-x
prunable gitdir file points to non-existent location
$ ls .worktrees/8-x
f  uncommitted.txt
```

Keying on `prunable` would skip the dirty check on that one and let `git worktree remove` take the member's work with it. `-d` makes exactly one previously-unanswerable case answerable — a gone directory — and leaves every other refusal alone: a worktree that exists but is dirty still blocks, and one that exists but cannot be read still dies as unknown.

## Tests

The existing case `an unreadable worktree is an unknown answer, never a clean one` set up this very scenario (`rmSync(c.wt)`) and pinned the bug as intended behaviour. It now uses a worktree whose `.git` file points nowhere — present, registered, still on the branch, still unreadable — so the `die` keeps a case named after it. Mutation-checked: swallow that status failure and the case fails.

Two new cases cover the gone directory, and both fail against the pre-fix script:

- `a worktree directory deleted by hand releases instead of blocking forever` — exit 0, all three artefacts gone including the stale entry, and the label actually comes off.
- `a worktree deleted by hand while its branch carries work still blocks` — the guards measured on the branch ref must not weaken because the directory went missing.

`./agent-test skills/fleet/scripts/*.test.mjs` — 147 pass, 0 fail.


feigi:
Review complete — six specialists, all delivered. Pushed `1e54bac`.

## Applied

**One regression fixed.** `[ -d "$wt" ]` is false for every `stat` failure, not only a deleted directory, so a worktree behind an unsearchable prefix (dropped mount, ejected volume, a path without Full Disk Access) read as gone and the claim was released outright:

| | base `73d7faf` | head `330047a` |
|---|---|---|
| unsearchable prefix, uncommitted work inside | exit 2, nothing touched | exit 0, `"released":true`, branch deleted, label dropped, work orphaned |

Against the base this is a regression, not a gap the fix left open. `git worktree remove` is no backstop: it gates its own clean check on the same `stat`, so guard and backstop fail on the same inputs.

The fix establishes absence instead of inferring it — walk up to the nearest ancestor that exists and require *that* to be searchable. Walking is load-bearing: four separately proposed one-level variants all read `rm -rf .worktrees` (parent gone with the child) as unknown and put that case back on the permanent exit 2 #37 exists to end.

**Three comment claims corrected**, all measured on git 2.50.1: the `prune` "backstop" (it finds nothing, and is only reachable once `remove` succeeded); "keying on `prunable` would let the remove take the work with it" (it refuses at rc 128 — the `-d` choice stands, on the true reason); and "every other refusal is measured on the branch ref" (two of six read the worktree list). The header at `:15-20` now states what the delete-time recomputation actually covers.

**Two regression tests**, plus `artefacts()` reused in place of a hand-rolled branch check and the label assertion tightened to the form used elsewhere in the file.

149 pass, 0 fail.

## Deferred — filed as their own issues

Not as comments here, since `Closes #37` would bury them on merge.

- #81 `stray` guard blocks forever when the stray directory is gone — the sibling of the bug this PR closes — and its remedy text names an action that cannot work (`ready-for-agent`)
- #82 `worktree-audit.sh:30` reports unreadable as `MISSING on disk` with `dirty=0`
- #83 `reap.sh` and this script take opposite defaults on an unreadable worktree
- #84 `:71` `wt_list` unguarded — missing guard, no demonstrated trigger
- #85 `:118-125` `PARTIALLY RELEASED` announced when nothing was touched
- #86 the dry run reports `released:true` where `--apply` exits 2 (`:188`/`:241` asymmetry)
- #87 `SKILL.md:387` / `reaping.md:34,43` "recomputes all four preconditions" is no longer accurate

## Confirmed, not changed

`impl-37`s rejection of the ticket's suggested `prunable` key was correct and is now confirmed four independent ways. The rewritten `an unreadable worktree is an unknown answer` case is legitimate — assertion set unchanged, only the fixture moved, and the old fixture was reassigned with the inverted expectation. Under mutation it is the only case that catches a swallowed `status` failure.

PR #75, opened 2026-07-30
fix(fleet): the emitted agent-test runner accepts a directory argument

### LINKED ISSUE #42: claim-ticket.sh: the emitted agent-test runner dies on a directory argument under Node 26

feigi:
Closes #42

`./agent-test <dir>` — the ergonomic way to say "run this suite" — died before
a single test ran. The emitted runner ends in `exec node --test "$@"`, and node
resolves a directory as a module specifier:

```
$ ./agent-test skills/fleet/scripts/
✖ skills/fleet/scripts (32.7ms)  Error: Cannot find module '.../skills/fleet/scripts'
  code: 'MODULE_NOT_FOUND'
$ echo $?
1
```

It failed in the safe direction, so nothing shipped green because of it. The
cost was misattribution: during the #32 review a member passed the directory
form, got a red, and spent a detour proving the red belonged to the invocation
rather than to the diff. A reviewer who does not spend that detour files it as
a finding against an innocent PR.

The runner now expands a directory to the test files under it, using the same
`\.(test|spec)\.[cm]?[jt]sx?$` shape the emit guard greps for.

## The zero-match case

The obvious one-liner is to rewrite the directory to a `**/*.test.*` pattern
and let node's own glob expand it. Measured on v26.5.0, that is wrong:

| invocation | result |
| --- | --- |
| `node --test sub` | `MODULE_NOT_FOUND`, exit 1 |
| `node --test sub/a.test.mjs sub/b.test.mjs` | exit 0, 2 pass |
| `node --test 'sub/**/*.test.mjs'` | exit 0, 2 pass |
| `node --test 'empty/**/*.test.mjs'` | **tests 0, exit 0 — silent green** |

That last row is the hazard `claim-ticket.sh:66` already refuses at emit time:
a runner that passes vacuously is worse than one that is dead, because the
review fan-out consumes it as a green suite. Handing a zero-match glob to node
would walk that failure right back in, past the guard, at run time. Counting
the matches in the shell is what makes a refusal possible, so the expansion
happens there and an empty directory exits 1 with `no test files under <dir>`.

The file-shape regex is now one variable feeding both the emit guard
(`claim-ticket.sh:68`) and the expansion it writes. Two copies would drift, and
drift means a directory the guard counted as a suite expands to nothing later.

Only the `node --test` entrypoint gets the shim. vitest and jest already accept
a directory, as a filter against naming conventions that need not match this
regex, so rewriting their arguments would refuse suites that are perfectly fine.

## Verification

`node --version` → **v26.5.0**. This repo has no CI; the local suite is the
evidence.

Tests assert against the **emitted** runner in a throwaway repo, not the
generator's source — the runner is what members invoke, and the generator's
source cannot show the bug. `--apply` labels the issue, so `gh` is stubbed;
the worktree, install, exclude and runner are all real.

- 150/150 pass, via both `./agent-test skills/fleet/scripts/` and
  `./agent-test skills/fleet/scripts/*.test.mjs` — identical counts.
- The two new directory tests were confirmed red before the fix, and red for
  the right reasons (`MODULE_NOT_FOUND`, and a missing refusal message).
- Also exercised by hand against a freshly emitted runner: recursion into
  subdirectories, a trailing slash, `.`, paths containing spaces, mixed
  directory-plus-file argv, and a zero-match directory (exit 1).

One trap worth recording: `node --test` marks the processes it spawns with
`NODE_TEST_CONTEXT`, and a nested runner that inherits it reports to a parent
that is not listening — status 0 and no stdout at all. The helper strips it.
That is an artifact of testing a test runner from inside one, not of this fix.

## Left out

`workflows/review-pr.js:241` tells specialists to use `${testCmd}` and "never a
bare test command" without a worked example, and #42 flags that as worth
checking. I left it alone: the guesswork it invites was only harmful because
the directory guess was rejected, and this PR makes that guess work. Changing
it now would be documenting a hazard that no longer exists.

`skills/fleet/skills/run-team/SKILL.md` is owned by PR #71 this wave and was
not touched.


feigi:
Review complete. Six specialists — code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, code-simplifier (type-design-analyzer skipped: shell, no types) — plus independent verification of every load-bearing claim against a runner emitted by this branch, never against the generator source.

**The load-bearing measurement holds.** `node --test 'empty/**/*.test.mjs'` → `tests 0`, exit 0, reproduced. The one-liner really would smuggle a vacuous pass past the `:66` guard, so counting matches in the shell is justified and the loop stays. `node --test <dir>` → `MODULE_NOT_FOUND`, exit 1: the bug was real. All four rows of the PR body table reproduce exactly. Zero-match refusal verified against the emitted runner for an empty directory, a directory of only non-test files, and a nonexistent path. npm entrypoint emits `exec npm test -- "$@"` with no shim traces. `testfile_re` is byte-identical between the emit guard and the emitted expansion. Baseline `main` 145/145 → this branch 150/150, nothing removed. The five new tests survive nothing they should — deleting the refusal, flipping its exit code, dropping the shim, and forcing it onto npm each kill exactly the right test, with a comment-reword control staying green.

**Two silent greens applied in `08fc89c`**, both reaching exit 0 having run no tests — the one failure this script exists to refuse:

- `set -f` settles only the *shell*'s handling. Node globs its own argv, where a literal `[` is a bracket expression that cannot match itself. `./agent-test 'my [odd] dir'` → **exit 0, tests 0**; `./agent-test app` silently dropped `app/[slug]/page.test.mjs` and reported `pass 1` of 2, with zero diagnostics. Escaped to the `[[]` idiom. The comment claiming the split was "safe for … glob characters" now says what is true.
- `find`'s exit status was discarded by the pipe. An unreadable subtree still yields the part find reached, grep still matches, the count guard still passes: `exit 0, tests 1, pass 1` where the readable tree gives `exit 1, fail 1`. Now read before grep overwrites it.

Also: the fixture gained a root-level test file, because bare `node --test` discovery and the two-file cases both landed on `pass 2` — so "the expanded glob form still works" could not distinguish argv honoured from argv discarded, and no mutation killed it. Nested, spaced and bracketed directories are pinned too; each was claimed by a comment and tested by nothing. And `review-and-fix.md` cited `claim-ticket.sh:106` for the port exports, which this branch moved to `:112`.

153/153 on v26.5.0. Both new guards mutation-verified with a passing control.

**Deferred:** #97 (bare `./agent-test` runs node's discovery, which disagrees with the emit guard — a `.spec.`-only repo is certified as having a suite and then reports a green having run none of it), #98 (no `node_modules` prune; nested vendored tests execute), #99 (a symlinked directory is falsely refused), #100 (node silently drops argv it cannot resolve, so a typo'd path reads as a pass), #101 (`review-pr.js`'s default `testCmd` is `./agent-test`, which is untracked and so never exists in the `git archive` snapshot line 241 tells specialists to run it from), #102 (docs still say `./agent-test <file>`).

**On the two items this PR deliberately left out**, ruled independently rather than inherited:

- `run-team/SKILL.md` — agreed, not falsified. A file argument still works, so the sentence stays true; it is merely no longer exhaustive. Filed as #102 for discoverability only, since PR #71 owns the file.
- `workflows/review-pr.js:241` — **disagreed**, on grounds #42 did not name. The reasoning was that the guesswork was only harmful because the directory guess was rejected. The guess space narrowed; it did not close. A quoted glob is not a directory, passes straight through the shim, and node's zero-match glob is still `tests 0` / exit 0 — the same row this PR's own table flags as the hazard. Separately, the default `testCmd` is a bare `./agent-test`, which is untracked and cannot exist in the snapshot. Filed as #101.

feigi:
Follow-up push `c79f383` — comment accuracy, no behaviour change. 153/153.

Three assertions in the shim claimed more than they could:

- The zero-match refusal justified itself with "`node --test` with nothing to run exits 0". True of node, but wrong about the alternative it argues against: appending nothing leaves argv **empty**, and bare `node --test` then discovers the whole worktree — a green for a suite nobody asked for, not a green over nothing. Conclusion was right, stated reason was not.
- The absence of `set -e` in the emitted runner is load-bearing and read as an oversight. grep exits 1 on no match, so under `-e` the shell aborts before the refusal can print (measured: `exit=1`, output empty). The generator opens with `set -eu`, which makes adding it to the runner look like obvious hardening. Now named.
- `$testfile_re` at `:117` is inside a `#` comment and does not expand; the identical token at `:146` is inside an unquoted heredoc and does. Two meanings a screen apart — the test file's parallel comment already said "this regex", so the generator now does too.

Also: the glob-form test records which glob it covers — a *matching* one; a zero-match glob is not a directory and misses the shim entirely (#100) — and the npm test is named for what it asserts, which is emitted text rather than a directory being run.

**One thing to watch at merge:** the `claim-ticket.sh:106` → `:112` citation fix in `skills/fleet/commands/review-and-fix.md` lands in a file owned by PR #80 this wave. It is a one-token change to a citation this branch itself staled, so I shipped it rather than filing — but it may want a trivial conflict resolution.

PR #73, opened 2026-07-30
docs(fleet): give members a report contract, and gate the finisher's label

feigi:
Two run-team rules that a fleet run kept losing work to. Both were applied to the maintainer's working tree during that run and never committed. Ticket-less: they were authored as mid-run tooling fixes under the skill's own "fix the tooling mid-run" rule, which is why no issue backs them.

**This branch no longer lands them unchanged** — see *Review corrections* below.

## 1. A report is a `SendMessage`, not the end of a turn

Six of seven members in one run went idle without reporting — three finishers, a reviewer, and two merge bots. All but one answered a controller ping with a full, high-quality report, so **the work was done and only the delivery was missing**. That is what makes it worth a rule rather than better prompting: the members were not failing at their jobs, they were failing at an unstated step.

Placed in the top `rules that fail silently` section rather than patched into each of the three separate "report X, exit" sites, so implementers, reviewers, finishers and merge bots all route through one statement of it.

The header drops its count in the same hunk. It read "Two rules that fail silently" while already listing three, and this change would make four. A count in a header rots every time the section grows, so it stops counting rather than being corrected to a number that will rot again.

## 2. The finisher's duties are an ordered list, and the worktree audit is first

- **"Filed" means *as an issue*, not parked on the PR's own source issue.** The PR's `Closes #N` buries every such comment on merge. Caught once at seven findings.
- **The dirty-worktree halt is duty 1, ahead of the label**, and names its check and its failure branch. A finisher read the rule's *purpose* (protect unverified work), verified the dirt was a strict subset of the pushed head, and labelled anyway. Its reasoning was sound and its compliance was not — an exit condition that trails the action it guards reads as advisory.

## Review corrections (applied on this branch)

Reviewed by `review-pr-73` with `code-reviewer` + `comment-analyzer`. Four convergent findings, all applied:

1. **The rename orphaned a citation.** `references/member-lifecycle.md:3` quoted `SKILL.md's "Two rules that fail silently"` verbatim — the only file in the repo that did. Updated. The PR argued header counts rot and rotted a citation in the same commit.
[...]

Issue #45, opened 2026-07-30
fleet specs: both design docs describe inflight.sh with a payload and an exit contract it does not have


feigi:
## Problem

Two design docs describe `skills/fleet/scripts/inflight.sh` incorrectly. Neither was touched by PR #40, which is what made the divergence visible: #40 promotes `0 free, 1 taken, 2 unanswerable` to a documented contract in `next-ticket/SKILL.md:33`.

### 1. `docs/specs/2026-07-23-fleet-plugin-design.md:165`

The script table row reads:

| `inflight.sh` | `<N>` | `{taken, probe, evidence}` | never — reports |

Both the Out column and the Non-zero column are wrong.

Actual payload:

```
$ ~/.claude/skills/fleet/scripts/inflight.sh 41
{"issue":41,"taken":false,"hits":[],"evidence":{"pr":"","remote":"","localBranch":"","worktree":""}}
```

- `probe` does not exist. The key is `hits`, and it is an array.
- `issue` is undocumented.
- "Non-zero when: **never** — reports" is false. Measured: free → 0, taken → 1, nonexistent issue / non-numeric / no args / not a repo → 2. The script's own header says so: `# Exit 0 free, 1 taken, 2 the question could not be answered.`

`reap.sh` and `pr-overlap.mjs` genuinely are "never" rows, so the column is meaningful and this row is not a stylistic shortcut.

### 2. `docs/specs/2026-07-22-run-team-agent-fleet-design.md:187-189`

Still specifies the three inline commands as the in-flight check:

```
3. In-flight check — `next-ticket` step 3, all three probes per candidate
   (`gh pr list --state all --search`, `git ls-remote --heads origin`,
   `git worktree list` + `git branch -vv`).
```

Those are exactly the hand-rolled probes issue #33 removed, and `inflight.sh:26-29` documents the first as unusable on its own. Lower priority than the row above — this is a historical design record, so the fix may be a dated note rather than a rewrite.

## Suggested

Correct the table row to `{issue, taken, hits, evidence}` and `1 taken, 2 unanswerable`. For the older spec, mark the superseded step rather than silently rewriting history.

Deferred from PR #40 review

PR #76, opened 2026-07-30
fix(fleet): inflight.sh probe 2 reports "no remote branch" when it could not look

### LINKED ISSUE #56: inflight.sh probe 2: `git ls-remote ... || true` renders an unanswerable question as `no remote branch for #N`
## Problem

`skills/fleet/scripts/inflight.sh:103-104`:

```sh
remote=$(git ls-remote --heads origin 2>/dev/null | … | paste -sd, - || true)
```

`2>/dev/null` discards the cause and `|| true` converts **every** failure into "no match". The probe then prints `no remote branch for #N` — a positive assertion of fact for what is actually *"I could not look"*.

```
$ git ls-remote --heads origin   # origin -> /nonexistent/unreachable.git
exit 128
swallowed: fatal: '/nonexistent/unreachable.git' does not appear to be a git repository
swallowed: fatal: Could not read from remote repository.
```

## Why now — PR #53 makes it bite

Pre-existing, but **not** at its old severity. Probe 1 used to hit on *any* linked PR, which accidentally covered probe 2's blindness. PR #53 narrows probe 1 to OPEN-only and strips that cover.

Measured — ticket #8, linked PR #20 MERGED, `ls-remote` unable to answer:

```
BEFORE  {"issue":8,"taken":true, "hits":["pr"],…}  EXIT=1
AFTER   {"issue":8,"taken":false,"hits":[],…}      EXIT=0
```

PR #53 is correct about #20. It also turns a latent, masked silent failure into a reachable wrong **"free"** — the direction that puts two agents on one ticket, which the 0/1/2 contract exists to prevent.

Triggers that do not take GitHub down (so probe 1 will not `die 2` first): no `origin` configured, SSH key or agent unavailable, host-key prompt suppressed, `origin` pointing at a stale URL after a rename.

## Suggested

Capture the status separately and `die` on a non-zero `git ls-remote`, keeping `|| true` only for `grep`'s no-match exit 1. `:113-114` and `:119-122` have the same shape with lower exposure — worth settling in the same pass.

Deferred from PR #53 review (silent-failure-hunter, F2). Verified by the controller: the finding is about probe 2, and PR #53's own fix does not change it either way.

feigi:
## The bug

`skills/fleet/scripts/inflight.sh` probe 2 ran the remote lookup inside its own filtering pipeline:

```sh
remote=$(git ls-remote --heads origin 2>/dev/null | awk … | grep -E … | paste -sd, - || true)
```

A pipeline reports its **last** command's status — `paste`, which always succeeds — so `ls-remote`'s exit 128 never reached the script, and `2>/dev/null` discarded the reason with it. What arrived was zero matching lines: exactly what a clean ticket produces. The probe then printed `no remote branch for #N`, a positive assertion of fact for what was really *"I could not look"*.

**The issue's diagnosis is off by one step, and it changes the fix.** It blames the trailing `|| true`. Measured in a throwaway repo with `origin` pointed at a nonexistent path:

```
$ git ls-remote --heads origin            → exit 128
$ remote=$(git ls-remote … | … | paste)   → remote=[] status=0   # no `|| true` at all
```

Removing `|| true` alone changes nothing — the pipeline had already thrown `git`'s status away. The lookup has to be hoisted **out** of the pipeline. `|| true` stays exactly where it was, now covering only what it was meant to: `grep`'s no-match exit 1.

## Why it is urgent

It yields a wrong **"free"**, which puts two agents on one ticket — the one direction the 0/1/2 exit contract exists to prevent. 2 means "the question could not be answered", and this path could never reach it.

None of its triggers need GitHub to be down, so probe 1 does not `die 2` first: no `origin` configured, an unavailable SSH key or agent, a suppressed host-key prompt, or `origin` pointing at a stale URL after a rename. PR #53 narrowing probe 1 to OPEN-only removed the cover that had been masking this.

## The fix

`inflight.sh:156` — the guard `release-ticket.sh:168` already uses for the same call, down to leaving stderr on stderr rather than folding it into the value (the emptiness of `$remote` *is* the answer, so a host-key notice on a fine query would otherwise read as a branch that exists).

The two conditions now stay apart: `grep` exiting 1 is "looked, found nothing" and leaves the ticket free; a failed `ls-remote` is "could not look" and reaches exit 2.

`inflight.sh:175` (`git for-each-ref`) and `:184` (`git worktree list`) had the same shape at lower exposure — local reads that fail only on a corrupt ref or a broken worktree admin file. Rarely is not never and the failure points the same wrong way, so they are guarded alike rather than left as the last paths that can answer "no" without having looked.

## Tests

Four cases in `inflight.test.mjs`, committed failing first (2ed3991) and passing in 9de019b. Both directions are pinned, because a fix that over-reaches would make every clean ticket unanswerable:

| Case | Before | After |
|---|---|---|
| unreachable `origin` | exit 0, "no remote branch" | **exit 2**, "could not look" |
| no `origin` configured | exit 0, "no remote branch" | **exit 2** |
| reachable `origin`, no match | exit 0, free | exit 0, free (unchanged) |
| reachable `origin`, match | exit 1, taken | exit 1, taken (unchanged) |

The first two failed on `0 !== 2` before the fix; the last two passed before and after by design — they are the regression pins for the opposite direction.

The fixture gains a real bare `origin` (probe 2 now requires one) with branches built via `commit-tree` directly in the bare repo — never pushed from the clone, since a local branch of the same name would let probe 3 answer for probe 2. `GIT_TERMINAL_PROMPT=0` is set so no case can block on a prompt.

`./agent-test skills/fleet/scripts/*.test.mjs` — **149 pass, 0 fail** (145 before). No CI in this repo, so that local run is the evidence.

## Left out

- **`GIT_TERMINAL_PROMPT=0` in the script itself.** A suppressed host-key prompt is a named trigger, so the environment already sets it where it matters; an unattended probe that *hangs* is a real and distinct bug, but it is not this one and forcing it would change behaviour for interactive users. Worth its own ticket.
- **`run-team/SKILL.md:54`** describes step 3 as "any hit = taken" without naming exit 2, while `next-ticket/SKILL.md:36` documents the contract correctly. Pre-existing, and this change makes exit 2 materially more reachable — but that file has uncommitted changes in the working tree from another agent, so I did not touch it.

Closes #56


feigi:
## Verification: the greens are not vacuous

Prompted by the `NODE_TEST_CONTEXT` trap another member hit. Nothing in this suite spawns a nested runner (`inflight.sh` shells out to `git`/`gh`/`python3`, never node), so the trap does not bite here — but I checked rather than assumed, and asserted counts rather than exit status:

```
NODE_TEST_CONTEXT=child ./agent-test …inflight.test.mjs   → EXIT=0, zero tests, no output
env -u NODE_TEST_CONTEXT …                                → tests 18, pass 18, fail 0
```

Worth knowing fleet-wide: `./agent-test` inherits a poisoned env and reports **exit 0 having run nothing**. Every run cited here is a clean-env run quoting its count.

## Mutation evidence, per test

Two of the four cases pass both before and after the fix by design, so a count alone does not prove they are live. Each was given a failing mutant:

| Mutant | free-pin | taken-pin | exit-2 pins |
|---|---|---|---|
| pre-fix (lookup inside pipeline) | pass | pass | **RED**, `0 !== 2` |
| A: `grep` no-match → `die` (the over-reach regression) | **RED** | pass | pass |
| B: probe 2 discards its own lookup | pass | **RED** | pass |

Mutation A is the one that matters for review: it is exactly the "fix that over-reaches and makes every clean ticket unanswerable" direction, and it also reddened 10 unrelated cases — every ticket without a matching remote branch. That is the blast radius this guard avoids.

Script restored between mutants; tree clean at 9de019b, full suite **149 pass, 0 fail**.

feigi:
## Review: the load-bearing claim is independently confirmed, nothing blocking

The PR's central claim — that `|| true` was never the mechanism, and the pipeline discarded `ls-remote`'s 128 on its own — was re-measured from scratch rather than taken from the description. `pipefail` is unset in this script (grepped), so:

```
git ls-remote --heads origin                                  -> exit 128
remote=$(git ls-remote … | awk | sed | grep | paste -sd, -)   # NO `|| true` anywhere
   -> assignment-status=0   remote=[]   script-exit=0         # under set -eu
```

Hoisting the lookup out of the pipeline is aimed at the right mechanism, and **the `|| true` deletion the issue text prescribes is correctly absent** — a reviewer reading #56 will expect one and should not.

End-to-end from the object store, same throwaway repo, `origin` → nonexistent path:

| | stderr | exit |
|---|---|---|
| `main` | `no remote branch for #8`, `taken=false` | **0** — the wrong free |
| `9de019bc` | git's own `fatal: … does not appear to be a git repository`, then `git ls-remote failed, so whether #8 has a remote branch is unknown` | **2** |

**Mutation-verified independently**, not read off the table in #issuecomment-5137397106:

- Revert the fix (lookup back inside the pipeline) → exactly the two exit-2 pins redden, 16/18.
- Over-reach (`grep` no-match → `die`) → **12 of 149 redden**: the named free-pin, 10 further cases in `inflight.test.mjs`, plus `release-ticket.test.mjs`'s cross-file probe. That matches the "10 unrelated" figure in the comment above exactly; it simply did not count the cross-file one. First attempt at this mutant was **inert** — wrapping a pipeline that still ended in `paste` reproduces the bug rather than the over-reach, which is itself a neat demonstration of the thing being fixed.

`release-ticket.sh:168` was checked at the specific line rather than trusted by reputation: `if ! remote=$(git ls-remote --heads origin "refs/heads/$branch"); then die …` — genuinely the same shape.

Suite: **149 pass, 0 fail**, run as `env -u NODE_TEST_CONTEXT -u NODE_TEST_WORKER_ID`, count asserted rather than exit status.

## One inaccuracy, non-blocking

`inflight.sh:148-150` lists four triggers and says probe 1 "sails through all four". Measured: real `gh issue view` in a repo with **no remotes** exits 1 with `no git remotes found`, so probe 1 dies 2 *first* and probe 2 is never reached — unless `GH_REPO` is set, in which case the trigger is real. The test passes only because `gh` is stubbed. Outcome is exit 2 either way, so no behavioural defect; the other three triggers hold unconditionally.

## Judged sufficient, not filed

The `for-each-ref` and `worktree list` guards ship untested, and that is fine. Corrupting `packed-refs` (`printf 'GARBAGE\n' > .git/packed-refs`, one line) makes `for-each-ref` exit 128 — but `ls-remote` reads the same local ref store and **dies at probe 2 first**, so probe 3 is never reached. `git worktree list` resisted four separate forcing attempts. A test would need a corruption that breaks `for-each-ref` while leaving `ls-remote` healthy, and none could be constructed. Defense-in-depth, not a live path.

## Deferred — all pre-existing, none introduced here

| # | finding | label |
|---|---|---|
| #88 | probe 3 truncates worktree paths containing a space → **wrong free** (verified with a control differing only in the space; needs a detached worktree, since `add -b` masks it via the branch half) | `ready-for-agent` |
| #89 | three filtering pipelines still discard mid-pipeline status — same class, one level down; `pipefail` is not portable to `dash` so the approach is open | `needs-triage` |
| #90 | a `"` in a ref or worktree name emits unparseable JSON; `release-ticket.sh:110`'s `jstr()` already solves it | `ready-for-agent` |
| #91 | `/tmp/.inflight.$$` is symlink-truncatable (measured: victim file 37 → 14 bytes) | `ready-for-agent` |
| #92 | no `GIT_TERMINAL_PROMPT=0` — an unattended probe can hang instead of answering | `needs-triage` |
| #93 | `run-team/SKILL.md:54` says "any hit = taken", never names exit 2, which this PR makes far more reachable | `ready-for-agent` |

#89 was argued for landing here, on the grounds that this PR's own comment at `:121-127` already makes the fork-failure case. Deferred anyway: the honest fixes are guard-every-stage (verbose) or collapse `awk|sed|grep|paste` into one `awk` (cleaner, but rewrites the matching logic the fleet depends on and breaks the shape-match with `release-ticket.sh:168`). Different change, different risk, wants its own tests.

Nothing applied to the branch — the PR is correct as it stands.

feigi:
## Fan-out complete — six specialists, five applied fixes, `543faa3`

Applied here, each mutation-verified rather than taken on report:

| fix | evidence |
|---|---|
| `awk '/^worktree /{print substr($0,10)}'` at `:189` | four specialists found it; `release-ticket.sh:70` — the file this PR cites as its reference — already reads the field that way and documents why. Reverting reddens the spaced case **alone, 1 of 151** |
| probe-3 comment now names its ceiling | it claimed the guards cover "a corrupt or unreadable ref, a broken worktree admin file". Measured: `chmod 000 .git/refs/heads` → `for-each-ref` **rc 0, empty**, and `fix/55-real` reads as **free, exit 0** |
| "probe 1 sails through all four" cut to what holds | real `gh` resolves owner/repo from git remotes, so with none configured it exits 1 and probe 1 dies first. Trigger 4 (rename redirect) was never measured by anyone |
| host-key rationale corrected | 4 realistic notices × 4 ticket numbers through the actual filter: **zero survived**. `awk '{print $2}'` reduces them to `Permanently`/`authenticity`. The reason was transplanted from `release-ticket.sh:168`, which tests `$remote` raw and where it genuinely applies |
| `\|\| die` on the final `printf` | `sh inflight.sh 9 >&-` gave **rc 1**, which the contract reads as *taken* — the hazard `:121-127` already names, live on the last line. Now 2 |
| two `GIT_CONFIG_*` shutouts in the fixture | this PR made probe 2 a hard dependency of every case in the file. Under `protocol.file.allow=never`: **14 of 20 red** without them, 20/20 with |

**151 pass, 0 fail** (149 before).

## Deferred, with the reasoning

- **#89** residual pipeline-status sites — `pipefail` is absent from `dash`, so the honest fix is guard-every-stage or collapse into one `awk`; the latter rewrites the matching logic and breaks the shape-match this PR leans on. Different change, different risk.
- **#95** git's *degraded* reads (rc 0, output missing) — the ceiling `:168-174` now documents. Detecting it means cross-checking counts, roughly doubling each probe's cost.
- **#96** `die` aborting the disjunction — with a worktree present and origin unreachable, `rc=2` while the true answer is 1 and `hits:["local"]` is discarded. The escalation is right, the placement is wrong; ~15 lines and a payload contract change.
- **#90** `jstr()` for JSON escaping · **#91** `/tmp/.inflight.$$` symlink truncation · **#92** prompt suppression *and* a timeout — measured that `GIT_TERMINAL_PROMPT=0` does **not** bound a stalled transport (8s block either way, `rc=142`).
- **#93** `run-team/SKILL.md`'s missing exit-2 arm. Two specialists independently found that one environment fault empties a whole wave's shortlist, and I nearly landed it here — but this branch is **42 insertions behind `origin/main`** on that file after #71 and #73. Editing the stale copy would conflict or revert their work. Needs a small PR off current `main`.
- The stale spec row at `2026-07-23-fleet-plugin-design.md:165` is **already #45** (`ready-for-agent`) — not re-filed.

## Corrections to earlier claims, mine included

- My own first over-reach mutant was **inert** — it wrapped a pipeline still ending in `paste`, reproducing the bug rather than the over-reach. Corrected, it reddens 12 of 149.
- The `10 further cases` figure above is right **in-file**; suite-wide it is 11, the extra being `release-ticket.test.mjs:154`.
- The no-push fixture rationale credits the wrong mechanism: `assert.deepEqual(r.json.hits, ["remote-branch"])` would have caught a pushed fixture anyway. The no-push design is belt, that assertion is braces.
- The worktree-space defect as first reported needed a precondition it omitted: with `worktree add -b` the branch half still catches it. The escape needs a detached or renamed-branch worktree, which is why the new test uses `--detach`.

PR #107, opened 2026-07-31
fix(fleet): run-team called the six-specialist fan-out "five"

### LINKED ISSUE #26: run-team: SKILL.md:238 says "the full five" for a six-specialist catalog that review-and-fix.md:49 enumerates
Deferred from PR #20 review.

## Problem

`skills/fleet/skills/run-team/SKILL.md:238` describes the fan-out as dropping dead dimensions — *"a docs-only change runs correctness+comments, **not the full five**"*.

The catalog has **six** entries, and `skills/fleet/commands/review-and-fix.md:49` enumerates them: `code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, type-design-analyzer, code-simplifier`, and says *"Profile unknown → **the full six**, the safe direction."*

So the same fan-out is "five" in one file and "six" in the other, and only the six-item version matches the enumerated catalog.

## Evidence

```
$ grep -n "full five\|full six" skills/fleet/skills/run-team/SKILL.md skills/fleet/commands/review-and-fix.md
skills/fleet/skills/run-team/SKILL.md:238:correctness+comments, not the full five — and skips the adversarial pass on
skills/fleet/commands/review-and-fix.md:49:... Profile unknown → the full six, the safe direction. ...
```

## Why deferred

Pre-existing, and outside PR #20's subject. That PR corrects the *delivery* documentation (issue #8); `:238` is a fan-out **sizing** claim in a paragraph the PR does not touch. Folding an unrelated count fix into a correction ticket is how correction tickets acquire new wrong claims — the failure mode PR #20 exists to document.

## Fix

Confirm the catalog size against `selectDimensions` / `DEFAULT_DIMENSIONS` in `workflows/review-pr.js`, then make `SKILL.md:238` agree with it. Check whether any other doc restates the count while you are there.

Found by the `comment-analyzer` specialist during the PR #20 review.

feigi:
`run-team/SKILL.md` described the review fan-out as dropping dead dimensions —
"a docs-only change runs correctness+comments, **not the full five**" — while
`skills/fleet/commands/review-and-fix.md:49` calls the same fan-out a
"six-specialist catalog" and says "the full six". One of the two was wrong.

## Settled against the code, not the issue

The issue's every line number was stale (`:238`, then `:242`, then `:245`); the
string is at `skills/fleet/skills/run-team/SKILL.md:308` at this branch's merge
base. Nothing below is carried over from the issue body — each claim has a
command behind it.

`DEFAULT_DIMENSIONS` at `workflows/review-pr.js:73` is the source of truth, and
it has **six** entries:

```
$ node -e 'const s=require("fs").readFileSync("workflows/review-pr.js","utf8");
  const b=s.slice(s.indexOf("const DEFAULT_DIMENSIONS = ["));
  console.log([...b.slice(0,b.indexOf("\n];")).matchAll(/key: "([^"]+)"/g)].map(m=>m[1]))'
[ 'correctness', 'silent-failure', 'tests', 'comments', 'types', 'simplify' ]
```

`selectDimensions` (`workflows/review-pr.js:139`) only ever filters that array,
so six is the ceiling the docs should name. `review-and-fix.md` was right and
`SKILL.md` was wrong — hence the one-word change, in the file the issue names.

The surrounding paragraph needed nothing else: its docs-only claim
("correctness+comments") matches `selectDimensions`' `docsOnly === true` branch
exactly, and its manual fallback already says "the full set" rather than a
number.

## Scope searched for other restatements

The issue asks whether any other doc restates the count. Searched the whole
worktree excluding `.git`, case-insensitively, on four patterns: `full
(five|six|four|seven|N)`; `(two|three|four|five|six|N)[- ]?(specialist|
dimension|reviewer|agent)s?`; `specialist (catalog|set|list)` / `full set`; and
the six agent type names. `grep -rn "full five"` now returns nothing repo-wide.

Sweep caveat worth carrying forward: **`git grep -E` silently ignores `\b`**, so
a word-boundary pattern reports a clean sweep it never actually ran. A
review-side sweep here hit exactly that false negative on its first pass. Use
`-P`, or anchor on surrounding literals.

**Corrected in review: `SKILL.md:308` was NOT the only wrong restatement.**
The original version of this section claimed it was, and ruled `SKILL.md:437`
("fans out 4-5 specialists") correct-as-is on the grounds that it is a
throughput estimate rather than a catalog count. That ruling was wrong, and the
second commit on this branch fixes :437 too. See "The second stale count" below.

One other numeric mention was checked and is correct, so it is untouched:

- `workflows/review-pr.js:20` — "five specialists on one PR and four on another"
  is a war story about one observed run, not the catalog size.

**Two more this section originally cleared as correct are in fact wrong.** Both
were passed over without being read in their actual context; both are
pre-existing defects in files this PR does not touch, and both are deferred to
their own issue rather than folded in. Re-derived by executing the real
`computeStats` and `selectDimensions` out of the head blob:

```
computeStats([])  = {... docsOnly:false, hasSrc:false, hasTests:false, profile:"empty"}
WITH the guard    -> 6 specialists
WITHOUT the guard -> 2 specialists [correctness,comments]   (4 dropped, not 3)
```

- `workflows/review-pr.js:143` — "drop three dimensions" does **not** describe
  the `hasSrc === false` branch. The comment sits on the *empty-profile guard*
  one line below it. Strip that guard and an empty profile falls through to
  `hasTests === false` **and** `hasSrc === false`, dropping tests + types +
  silent-failure + simplify — **four**, not three.
- `skills/fleet/scripts/diff-stats.mjs:105` — "down to two specialists" is not
  about the docs-only pair at all. It is in the `run()` catch, describing what a
  failed `gh` query would cost if it read as an empty diff. It would not cost
  two: `computeStats([])` yields `profile: "empty"`, which `review-pr.js:145`
  widens to **all six**. The comment names a harm that guard already prevents.

## The second stale count (`SKILL.md:437`) — found in review, fixed here

`635cffb` ("run the full 6-agent pr-review-toolkit suite") raised the catalog
5→6 and updated `review-and-fix.md` + `workflows/review-pr.js`, but touched
`run-team/SKILL.md` **not at all** — leaving two 5-era counts in it. The first
commit fixed :308; :437 was the other one.

"4-5 specialists" is wrong under both readings, not merely stale as a catalog
count. Executing the real `selectDimensions` over every input combination:

```
docsOnly=true                  -> 2 [correctness,comments]
docsOnly=false hasSrc  hasTests-> 6 [correctness,silent-failure,tests,comments,types,simplify]
docsOnly=false hasSrc  !tests  -> 5
docsOnly=false !hasSrc  tests  -> 3
docsOnly=false !hasSrc !tests  -> 2
profile unknown                -> 6
REACHABLE SIZES: [2,3,5,6]    is 4 reachable? false
```

4 is unreachable, and 6 — the ordinary source+tests PR — sits above the stated
ceiling. Read as the throughput estimate it is, it still understates, because
the paragraph spends that number on reviewer capacity ("default 2 implementers
/ 5 reviewers"). Changed to "up to six specialists": an upper bound that is
actually reached, naming no impossible value.

Manifest-safe: pattern "Reviews are the bottleneck" resolves against :436, not
the edited line. All 33 patterns resolve identically before and after. (:437 is
itself pinned by manifest line 23, so the line was load-bearing, not incidental
prose.)

## One more word in the same sentence (`SKILL.md:309`)

`review-pr.js:122` says a `suggestion` is "never **auto**-applied". SKILL.md:309
dropped the "auto" and asserted they are never applied at all — an overstatement,
since nothing forbids a reviewer applying one and `review-and-fix.md` never
mentions `suggestion` (`grep -ci` returns 0). Restored, in the sentence this
branch already edits.

The remaining half — that the defer policy the sentence cites is written down
nowhere — is deferred to #112.

`docs/specs/2026-07-22-run-team-agent-fleet-design.md:115` enumerates five
specialist names (no `code-simplifier`), but it is a dated design record
describing what was true when written, not a live count — left alone
deliberately.

## Manifest check

Every pattern in `docs/fleet-rule-manifest.txt` was resolved against
`skills/fleet` + `workflows` before and after the edit. **No pattern stops
resolving because of this change** — the edited words appear in no manifest
pattern, and the diff is one word inside a paragraph no pattern anchors to.

## Pre-existing break found by that check — NOT fixed here

Manifest line 15, `Say the specialist count`, resolves to **zero** files, and it
was already orphaned before this branch:

```
$ git grep -F "Say the specialist count" a56c01a -- skills/ workflows/
(no output)
```

Commit `8b46fdb` replaced the rule `**Say the specialist count in the prompt.**`
with the "**The fan-out scales itself to the diff.**" paragraph this PR edits,
and never updated the manifest — the same defect `975c3d0` fixed for manifest
line 6. So the no-rule-lost gate has one dead entry and the "all 33 patterns
resolve" claim at `docs/specs/2026-07-23-fleet-plugin-design.md:428` is stale.

Left out on purpose. It is a different defect class (a dead gate pointer, not a
wrong number), fixing it means *choosing* a replacement phrase rather than
correcting one, and this issue's own body argues the case: "Folding an unrelated
count fix into a correction ticket is how correction tickets acquire new wrong
claims." It wants its own ticket.

**Review update:** that ticket already exists — #48 section 3 tracks both the
dead manifest entry and the stale `:428` "all 33 patterns resolve". Settled
there in a review comment rather than duplicated: resolving the manifest
tree-wide (not just against `run-team/SKILL.md`, which was #48's method) shows
only **one** of its three claimed misses is genuinely orphaned — `newest run on
a branch is frequently` resolves in `run-merge-bot.md:90` / `ci-state.mjs:103`,
and `git cherry origin/main` in `references/reaping.md:24`. So `:428` is stale
by one, not three.

## Testing

`./agent-test` — 170 pass, 0 fail, before and after rebase onto `origin/main`.
No test asserts this prose, so the standing check is the `DEFAULT_DIMENSIONS`
extraction above.

Closes #26


feigi:
Review complete (fleet reviewer). Fan-out for this PR's profile (`diff-stats.mjs` → `profile=docs`, `docsOnly=true`): code-reviewer + comment-analyzer, per `review-and-fix.md:49`.

**Applied here** — one finding, from the correctness specialist, re-verified by execution before ruling:

- `run-team/SKILL.md:437` carried a second 5-era fan-out count ("4-5 specialists"). Reachable sizes are `{2,3,5,6}` — 4 is unreachable, 6 is the ordinary source+tests PR. Fixed to "up to six specialists" in `fdac65d`. The PR body has been corrected too: it previously ruled :437 correct-as-is, which this commit falsifies.

**Deferred, each filed as its own issue** (not parked here — `Closes #26` would bury a comment):

- #112 — `SKILL.md:309` cites a `suggestion`→defer policy that `review-and-fix.md` never states (`grep -ci suggestion` over it returns 0). `needs-triage`: the skip is verified, but whether to write the policy down or reword the claim is open.
- #113 — `review-and-fix.md:49` promises `args.dimensions` accepts keys; `review-pr.js:222` has no normalization, so keys silently interpolate `undefined` into the specialist prompt. `needs-triage`: honour the doc or correct the doc, not settled.
- #48 — the pre-existing dead manifest entry the PR body deferred already had a ticket. Commented there rather than duplicating, with a correction: only one of its three claimed misses is genuinely orphaned.

No CI in this repo (no `.github/` tracked), so there is no green to gate on. Not adding `ready-to-merge`.

feigi:
Second review pass — the comment-analyzer dimension landed late and is now covered; the earlier partial-coverage caveat is withdrawn.

**Applied here** (`89c8310`): `SKILL.md:309` said `suggestion`s are "deferred and never applied"; `review-pr.js:122` says "never **auto**-applied". Restored the dropped word — same sentence this branch already edits.

**PR body corrected again.** It cleared two more claims as verified that are not: `review-pr.js:143` ("drop three dimensions" — the comment sits on the empty-profile guard, and stripping it drops **four**) and `diff-stats.mjs:105` ("down to two specialists" — `computeStats([])` gives `profile: "empty"`, which `review-pr.js:145` widens to **all six**). Both re-derived by executing the real functions, both pre-existing, both now **#118** along with their root cause: `selectDimensions` cannot be imported, so every count claim about it is hand-derived — which is how both comments went wrong, and why both specialists had to build a harness to check a number.

**Not changed:** the body’s "the issue’s every line number was stale (`:238`, `:242`, `:245`)". A finding claimed `:242`/`:245` appear nowhere in #26; that grep required a colon prefix, and #26 writes them bare and bolded (`already at **242** on origin/main`). The sentence is accurate as written.

Deferrals now: #112, #113, #118. Still no `ready-to-merge`, still no CI in this repo to gate on.

PR #108, opened 2026-07-31
docs(fleet): both specs described inflight.sh with a payload and an exit contract it does not have

### LINKED ISSUE #45: fleet specs: both design docs describe inflight.sh with a payload and an exit contract it does not have

feigi:
Closes #45

Two design docs described `skills/fleet/scripts/inflight.sh` wrongly. Every claim below was settled by running the script at this HEAD, not read off the issue — the issue's line numbers were stale (the second doc's step is at 198, not 187).

## 1. `docs/specs/2026-07-23-fleet-plugin-design.md` — script table row

Measured payload:

```
$ ./skills/fleet/scripts/inflight.sh 41
{"issue":41,"taken":false,"hits":[],"evidence":{"pr":"","remote":"","localBranch":"","worktree":""}}
```

- `probe` does not exist; the key is `hits`, an array. `issue` was undocumented. Out column → `{issue, taken, hits[], evidence}` (`[]` marks arrays, matching the neighbouring rows).
- "never — reports" was false. Drove the script into each state: free → 0, taken → 1 (issue 45, hit `local`), and exit 2 from no args, a non-numeric arg, a nonexistent issue, and running outside a repository. `inflight.test.mjs` additionally pins exit 2 for an unreachable GitHub and a repository-level failure, so the Non-zero column names "or a probe failed".

The script's header already stated the contract — `# Exit 0 free, 1 taken, 2 the question could not be answered.` — so header and behaviour agree; only the spec diverged.

Worth flagging: a "never" row also contradicted this document's own compensating rule 1, *"fail closed — unknown or unreachable state exits non-zero with a reason"*. A caller trusting the row would read exit 2 as free and hand out a ticket already taken.

`reap.sh` and `pr-overlap.mjs` are left as "never" rows — that column is meaningful and only this row was wrong.

## 2. `docs/specs/2026-07-22-run-team-agent-fleet-design.md` — Phase 0 step 3

Still specifies the three hand-rolled inline probes. Marked rather than rewritten: the step text stands and gets a dated `> **Superseded in part.**` blockquote, the shape this document already uses twice. Rewriting a historical design record would erase what was actually designed on 2026-07-22.

The note's date is PR #40's merge date (2026-07-30), matching the existing notes' convention of dating by the superseding artifact.

## Verification

`./agent-test skills/fleet/scripts/inflight.test.mjs` → 20/20 pass, before and after rebase onto `origin/main`. Docs-only diff (8 insertions, 1 deletion); no test asserts spec-doc contents.


feigi:
Review deferrals, both filed as their own issues, both `needs-triage` (defect verified, fix approach open):

- **#114** — the three remaining `never` rows in this file's Script surface table (`pr-overlap.mjs`, `reap.sh`, `worktree-audit.sh`) all exit 2 when measured. Issue #45's premise that they "genuinely ARE never rows" is false, and this PR's body repeats it. **The diff itself is correct** — not widening into those rows was the right call; only the PR-body claim is wrong. Whether the column means "non-zero as a verdict" or "non-zero when, literally" is the open question.
- **#117** — `docs/specs/2026-07-30-fleet-trust-the-label-design.md:18-26` carries two line-number citations into `2026-07-22-run-team-agent-fleet-design.md`. Both are already stale on `main` (one target step was deleted outright); this PR's 7-line insert shifts both a further 7. Not caused here.

Neither blocks this PR.

feigi:
Review follow-up — one fix applied at `1ca997d`, deferral scope widened.

**Applied** (both defects were in prose *this PR* adds, so no history was rewritten — Phase 0's step 3 remains byte-identical to `main` and the diff is still a single pure-insertion hunk):

- `Phase 0's step 3` → `Phase 0's *In-flight check* step`. Positional references rot, and that rot is already live two paragraphs up: the note above points at "Phase 0's step 5 ... removed below", and a later renumbering means there is now an unrelated step 5 to land on. The step's own title supplies a semantic anchor verbatim.
- `In-flight probing (2026-07-30) moved into a script` → `... is now a script call`. `inflight.sh` was created by `aad0796` on **2026-07-23**; 2026-07-30 is when PR #40 made `next-ticket` call it. The parenthetical stays (siblings date the superseding change); only the predicate, which dated the script's creation a week late, changed.
- `issue #33 replaced them` → `PR #40 replaced them`. The issue asked; the PR replaced.

**#114 widened** from "the three `never` rows" to a full audit of the table's eleven `Non-zero when` rows. `candidates.mjs:164` turned out to state the *opposite* of its code — the row promises "empty result is exit 0 with `[]`", while `candidates.mjs:157` is `process.exit(rows.length === 0 ? 1 : 0)` (measured: `rc=1`). That is worse than the `never` rows and is wrong under either reading of the column.

#117 unchanged. Neither deferral blocks this PR.

feigi:
Second review pass — three more fixes at `3c7f05b`, after the comment-analyzer's report landed. Both specialists independently reached the first two.

**Date named the wrong event.** `(2026-07-30)` → `(2026-07-23)`, `See` → `docs/specs/2026-07-23-fleet-plugin-design.md`. In-flight probing moved into a script on 2026-07-23: `aad0796` created `inflight.sh` and `9f328de` wired it into `run-team/SKILL.md` — **this document's own artifact** (`:29`). 2026-07-30 is when PR #40 made `next-ticket` call it, a follow-up to a different artifact. Dating it 07-23 also restores the convention both sibling notes follow (`<Topic> (<date>)` … `See docs/specs/<same-date>-*.md`, self-verifying); the previous `See` target was undated, so nothing verified the date. Side benefit: the two halves of this PR now reference each other.

**"Runs all three itself" implied parity, and that was the riskier half.** The step is deliberately left intact, so it still shows a reader three inline commands with no warning attached — and probe 1 is exactly the one the script refuses to use bare (`inflight.sh:26-29`: `gh pr list --search 41` measured six PRs, five merely containing the digits). A maintainer treating them as a fallback hits the precise failure #33 was filed about. The note now says so.

**Fixed the citations this PR was breaking** (`2026-07-30-fleet-trust-the-label-design.md:19-22`) rather than only filing them. Both were already stale before this PR — `:211-213` off by 7, and `:193` pointing at a step that the change *that document argues for* had deleted — and this PR pushed them to ~14. Replaced with semantic anchors that cannot drift.

**Scope note:** this takes the PR to 3 files / 16 insertions / 6 deletions, past the ticket's 8/1. The third file is outside issue #45. I judged leaving a citation more broken than I found it worse than the overrun, but that is a judgment call worth a second opinion — say the word and I will revert that file and hand it back to #117.

Unchanged: Phase 0 step 3 remains byte-identical to `main` (verified again at this SHA), suite 170/170.

Issue #99, opened 2026-07-31
claim-ticket.sh: the emitted runner refuses a symlinked directory, because [ -d ] follows symlinks and find does not


feigi:
Deferred from PR #75 review.

`claim-ticket.sh:143` tests `[ -d "$arg" ]`, which follows symlinks, and `:145` then runs `find "$arg" -type f`, which does not descend a symlinked argument. A symlink to a directory full of tests therefore passes the branch test, yields zero matches, and hits the refusal:

```
$ ls -l tlink
tlink -> t          # t/ holds two passing tests

$ ./agent-test tlink
agent-test: no test files under tlink
EXIT=1

$ ./agent-test tlink/     # trailing slash makes find descend
ℹ tests 2  ℹ pass 2  EXIT=0
```

A false refusal, so it fails in the safe direction — loud and wrong rather than silent and green — but the two forms disagreeing on the same target is a trap, and the inconsistency is invisible until someone hits it.

Inert for `feigi/claude-config` today: `git ls-files -s | awk '$1=="120000"'` returns zero tracked symlinks.

## Fix

One character — `find "$arg/" -type f`. Prefer it over `find -L`, which would also follow symlinks *inside* the tree and can sweep into symlinked `node_modules` (see the prune issue) or loop.

No test covers symlinks; the case is worth pinning alongside the fix.

Verified by the reviewer on Node v26.5.0.

Issue #97, opened 2026-07-31
claim-ticket.sh: a bare `./agent-test` runs node's own discovery, which disagrees with the emit guard's test-file shape


feigi:
Deferred from PR #75 review.

`claim-ticket.sh:60` defines `testfile_re='\.(test|spec)\.[cm]?[jt]sx?$'` and it feeds two consumers — the emit guard (`:69`) and the directory expansion in the runner (`:145`). There is a third consumer nobody reconciled: **node's own default discovery**, which runs whenever the runner is invoked with no arguments. `for arg do` never enters its body on empty argv, so the expansion and its zero-match refusal are both bypassed, and `exec node --test "$@"` falls through to node's defaults — a different pattern set that matches neither `.spec.` nor the `x` in `[jt]sx?`.

Measured on v26.5.0, a repo whose only test file is `t/a.spec.mjs`:

```
$ sh claim-ticket.sh 42 slug fix --apply
    test entrypoint → node --test        <- the emit guard accepted the repo

$ ./agent-test          # bare
EXIT=0 tests=0 pass=0                    <- vacuous green, one real test in the repo

$ ./agent-test .        # through the shim
EXIT=0 tests=1 pass=1
```

So the guard at `:67-73` certifies "this repo has a suite", and the runner it emits then reports a green having run none of it. That is the exact failure the guard's own comment says it exists to refuse.

**Pre-existing** — the pre-#75 runner did the same. Not a regression, but #75 edits precisely this guard and its comment at `:56-59` claims one shape shared by "the emit guard and the directory expansion", which is true as written and still leaves this third disagreeing consumer unnamed.

## Why the approach is open

The obvious fix is one line above the loop:

```sh
[ $# -gt 0 ] || set -- .
```

which routes a bare invocation through the guard that already exists (verified: the `.spec.`-only repo becomes `tests 1`, and a repo with no test files refuses with `no test files under .`). But it interacts with the missing `node_modules` prune — `find` has no filter where node's discovery does, so defaulting to `.` would start sweeping vendored test files in any repo that has them. Settle that first, or settle both together.

Verified by the reviewer on Node v26.5.0; harness in the PR #75 review scratch.

feigi:
Blocker discharged by #109 (merged pending), which added `-not -path '*/node_modules/*'` to the runner's directory expansion.

This issue's body says:

[...]

PR #109, opened 2026-07-31
fix(fleet): the runner's directory expansion ran vendored tests

### LINKED ISSUE #98: claim-ticket.sh: the runner's directory expansion has no node_modules prune, so vendored tests execute
Deferred from PR #75 review.

`claim-ticket.sh:145` expands a directory with `find "$arg" -type f | grep -E "$testfile_re"` and no prune. Node's own discovery excludes `node_modules`; this does not.

Top-level `node_modules/` happens to be inert — node refuses those paths outright:

```
$ ./agent-test node_modules
agent-test: ... Could not find 'node_modules/dep/a.test.mjs, node_modules/dep/b.test.mjs'
EXIT=1
```

**Nested** `node_modules` — the pnpm / npm-workspaces shape — is not filtered, and runs:

```
find sees:  mono/packages/api/src/s.test.mjs
            mono/packages/api/node_modules/vendor/v.test.mjs
$ ./agent-test mono
✔ VENDOR test — not ours
✔ our real test
ℹ tests 2   EXIT=0
```

Third-party test code executes and the suite result depends on it passing.

Related, same root cause: when a mixture is handed to node, the entries it refuses are dropped **silently** — `node --test ours.test.mjs node_modules/dep/x.test.mjs` reports `tests 1, pass 1, exit 0` with no diagnostic. Only when *every* argument is refused does it exit 1.

## Scope

Inert for `feigi/claude-config` itself: `origin/main` has no `package.json` and none of the three lockfiles, so `install="true"` and no `node_modules` is ever created. Live for any other repo the fleet claims into that has a lockfile and no `scripts.test` — which is exactly the shape that gets a `node --test` runner rather than an `npm test` one.

## Fix

```sh
find "$arg" -type f -not -path '*/node_modules/*'
```

Note #97 depends on this: defaulting a bare invocation to `.` is unsafe until the prune exists.

feigi:
`claim-ticket.sh:147` expanded a directory with `find "$arg" -type f` and no prune. Node's own test discovery excludes `node_modules`; this did not, so third-party test code executed and the suite's result depended on vendor tests passing.

The fix is the one flag the issue names: `-not -path '*/node_modules/*'`.

## Reproduced first

Nested `node_modules` — the pnpm / npm-workspaces shape — is the live case. Fixture of `pkg/src/x.test.mjs` plus `pkg/node_modules/vendor/v.test.mjs`, Node v26.5.0:

```
before:  ✔ VENDOR test — not ours
         ✔ our real test
         ℹ tests 2   EXIT=0
after:   ✔ our real test
         ℹ tests 1
```

Confirmed on the real emitted runner too, not only on a bare `find`: with a vendored test present, `./agent-test .` now reports `tests 1, pass 1`.

The top-level `node_modules/` case was already inert and stays that way — the pattern needs a leading component, so a bare `node_modules` argument is still handed to node, still refused, still `EXIT=1`. A refusal, never a false green.

## Scope

Inert for this repo: `origin/main` has no `package.json` and none of the three lockfiles, so no `node_modules` is ever created here. Live for any other repo the fleet claims into that has a lockfile and no `scripts.test` — exactly the shape that gets a `node --test` runner rather than an `npm test` one. The test therefore builds the shape itself.

Held to the one flag. The adjacent `[ -d "$arg" ]` symlink defect (#99) and the bare-invocation default (#97) are untouched — `git diff` shows zero changed lines matching either.

## The test

It writes `node_modules` into the worktree rather than through the `repo()` helper: that is where `node_modules` actually comes from — the install step, not a commit — and committing it would rest the test on whatever `core.excludesFile` the machine happens to have.

Its vendor test fails on purpose. `pass 3` alone cannot separate "vendor was pruned" from "vendor ran and failed", and the exit status alone cannot separate "pruned" from "the expansion dropped everything"; both assertions or neither pins it. Seen red for exactly that reason before the fix, green after.

171/171 green across `skills/fleet/scripts`.

Closes #98


feigi:
Review complete — six specialists, all clearing the fix. **Approved.** Head is now `308b776`.

The flag is correct and unchanged. Scope held: re-derived from the object store, stripping `-not -path '*/node_modules/*'` and its comment from the head file yields a file byte-identical to base. `[ -d "$arg" ]` at `:131` byte-identical (#99 untouched), nothing added before `for arg do` at `:129` (#97 untouched).

Verified rather than read — freshly generated runners at base and head, never an existing worktree:

| | vendor passes | vendor fails |
|---|---|---|
| base | `tests 2` (vendor ran) | `fail 1`, **exit 1** |
| head | `tests 1` | `pass 1`, exit 0 |

The added test genuinely gates it: with the flag reverted, exactly that one test goes red.

`308b776` corrects three measurably wrong claims in the new comments, no behaviour change:

- "the top-level one is inert **only** because node refuses those paths outright" — node refuses only when the relative path's first segment is `node_modules`; a deeper segment or an **absolute** path runs. Measured at base, `./agent-test /abs/dir` ran the vendored test. The fix closes a hole its own comment said could not exist, and as written the comment justifies deleting the prune.
- "the pnpm / npm-workspaces layout" — neither. pnpm's per-package `node_modules` is a symlink farm that `find` without `-L` never descends; its real files sit under the root `node_modules/.pnpm/`, and npm workspaces hoist to root.
- "the exit status alone cannot tell pruned from expansion dropped **everything**" — dropping everything refuses (`no test files`, exit 1). The status misses a **partial** drop: exit 0 with `pass 2`. Same conclusion, right reason.

Also recorded why the spelling is not a `-prune` (it would return 0 on an unreadable dir under `node_modules` and defeat the readability guard), and hoisted a 4-segment fixture path that was spelled twice — a typo in either wrote the fixture elsewhere and the test still reported `status 0` / `pass 3`.

171/171 green.

Deferred, filed as their own issues: #123 (`-prune` vs `-not -path` — two reviewers reached opposite verdicts, genuine design call), #124 (the runner is pinned per worktree with no version stamp; **all eight** live worktrees are stale, including this one, so merging this fixes zero existing runners), #125 (a mixed argv greens having run less than asked). Commented on #97 — its stated blocker is discharged by this PR and its body still says otherwise.

This repo has no CI, so nothing gates on checks; the above is the verification.

feigi:
Second review pass — a seventh specialist (test coverage) landed after `308b776`. Head is now `2af7d65`.

**Ruling on the `-prune` contradiction: ship `-not -path` as-is.** Confirmed by measurement that both spellings (`-path '*/node_modules/*' -prune` and `-name node_modules -prune`) exit 0 on an unreadable dir inside `node_modules` while still exiting 1 on our own code — so the two reviewers measured the same mechanism and split on whether the change is desirable. Decisive point: **the current behaviour is identical at base**, so this PR neither caused nor worsened it. Changing it here would be a behaviour change outside #98's scope. Filed as #123 with both verdicts, the measurement table, and the `-prune` cost figure (68ms → 20ms over 8000 vendored files; 3 files printed, 303 visited).

`2af7d65` fixes two real gaps in the test this PR added:

**It pinned one direction only.** Over-pruning is the worse bug — it deletes real tests and still exits 0, a green over a shrunken suite — and mutating the pattern to `*node_modules*` passed **30/30**. `git grep node_modules` over `skills/fleet/scripts/*.test.mjs` returns only this test, so it was the fleet suite's sole coverage of the flag and it guarded one side. `node_modules_old/` must now survive.

**The count was an unanchored substring.** `/pass 3/` matches `pass 31`, so the pin dissolves once a fixture grows past 30 — demonstrated fully vacuous against the pre-fix script with a grown fixture. Now `/^ℹ pass 4$/m`. The file's five other `/pass N/` assertions have the same hole → #127 (`ready-for-agent`).

Three-way mutation check: flag reverted → red, over-broad pattern → red, unmutated → green. 171/171.

Also corrected: the two-assertion justification in the comment. Ablation shows **status alone catches the regression and the count alone misses it entirely** — and the previously stated reason for keeping the count ("status cannot tell pruned from expansion-dropped-everything") is false, since that case hits `[ -n "$files" ]` and exits 1. The count is still load-bearing, for a partial over-prune that only triggers in the presence of `node_modules` — which test #1 cannot catch, its fixture having none.

Deferred: #123, #124, #125, #127.

Issue #106, opened 2026-07-31
no-undo-audit.sh: conflicting paths with spaces or quotes silently break atRisk and the JSON


feigi:
Found while reviewing #105. All three are pre-existing, unrelated to that PR's gate change, but `no-undo-audit.test.mjs` now presents itself as this script's regression gate and covers only the refusal condition — no fixture ever produces a conflict, so `conflicts[]`/`atRisk[]` have zero coverage and hold three live defects.

## 1. A conflicting path with a space reports `atRisk: []` — the exact failure the script exists to prevent

`skills/fleet/scripts/no-undo-audit.sh:97` — `git log --oneline "$fork".."$base" -- $conflicts`, unquoted (SC2086 disabled on purpose), errors swallowed by `2>/dev/null || true`.

Reproduced: single conflicting file `has space.txt`, main holding a later commit that touches it.

```
ground truth: git log ... -- 'has space.txt'  →  f96d07d MAIN COMMIT AT RISK
script:       {"conflicts":["has space.txt"],"atRisk":[]}   exit=0
```

The word-split makes the pathspec `has` + `space.txt`, neither of which exists, so git returns nothing and the swallow hides it. The script's header says its job is stopping "a question that was never actually answered" being reported safe. This is that.

## 2. A conflicting path containing `"` emits malformed JSON on a *passing* audit

`skills/fleet/scripts/no-undo-audit.sh:87` wraps each path in quotes without escaping; `:101` (`at_risk_json`) does `gsub(/"/,"\\\"")`. Only the conflicts line was missed.

Reproduced with `has"quote.txt`:

```
{"...","conflicts":[""has\"quote.txt""],"atRisk":[]}   exit=0
JSON.parse → Expected ',' or ']' after array element at position 170
```

Exit 0, so a caller parsing the payload crashes on an audit that passed. git C-quotes such paths, which is where the inner quotes come from.

## 3. The exit-2 merge-tree guard cannot fire for the realistic bad-ref case

[...]

PR #110, opened 2026-07-31
fix(fleet): a quote in a ref or worktree name broke inflight.sh's JSON

### LINKED ISSUE #90: inflight.sh: a quote in a ref or worktree name emits a JSON payload no parser accepts — release-ticket.sh already has the fix
## Problem

`skills/fleet/scripts/inflight.sh:208-209` interpolates `$pr`, `$remote`, `$local_b` and `$wt` raw into JSON string positions:

```sh
printf '{"issue":%s,"taken":%s,"hits":[%s],"evidence":{"pr":"%s","remote":"%s","localBranch":"%s","worktree":"%s"}}\n' \
  "$n" "$taken" "${hits%,}" "$pr" "$remote" "$local_b" "$wt"
```

Git ref names permit `"`; worktree paths permit `"` and `\`. Either produces stdout no JSON parser accepts.

## Measured at `9de019bc`

Local branch `fix-42-say"hi`, `inflight.sh 42`:

```
{"issue":42,"taken":true,"hits":["local"],"evidence":{"pr":"","remote":"","localBranch":"fix-42-say"hi","worktree":""}}

$ python3 -c 'import json,sys; json.load(open(sys.argv[1]))' out.json
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 1 column 101 (char 100)
```

## Severity: evidence corruption, not decision corruption

The **exit code stays correct** (1, taken). Both `next-ticket/SKILL.md:36` and `run-team/SKILL.md:54` document the exit code as the decision, so nothing currently mis-decides on this. What breaks is any consumer that parses the payload for diagnostics.

## Suggested fix

The sibling script already solved this. `release-ticket.sh:110` defines:

```sh
jstr() { printf '%s' "$1" | tr '\001-\037\177' ' ' | sed 's/\\/\\\\/g; s/"/\\"/g'; }
```

used at five sites there and pinned by two passing tests. `inflight.sh` has no equivalent — `jstr` appears in exactly one file tree-wide. Lift it and wrap the four fields.

Deferred from PR #76 review. Pre-existing.

feigi:
`inflight.sh`'s final `printf` interpolated `$pr`, `$remote`, `$local_b` and `$wt` raw into JSON string positions. git accepts `"` in a ref name, and a worktree path is a filename, so it accepts `\` as well — either emitted a payload no JSON parser accepts.

## Reproduced before the fix

At `a56c01a`, throwaway fixture, local branch `fix-42-say"hi`, `gh` stubbed so probes 2 and 3 run against real git:

```
{"issue":42,"taken":true,"hits":["local"],"evidence":{... "localBranch":"fix-42-say"hi", ...}}
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 1 column 101 (char 100)
```

Same fixture after the fix parses, and both vectors round-trip exactly:

```
localBranch = 'fix-42-say"hi'
worktree    = '.../fix-42-pa"th\\x'
```

## The change

Lift `jstr()` from `release-ticket.sh:115` — the same `tr` + `sed` pipeline, already pinned by two tests there — and wrap the four evidence fields. The other three interpolations stay raw and need nothing: `$n` is refused unless it is all digits, `$taken` is the script's own `true`/`false`, and `$hits` is built only from the three fixed literals `add_hit` is called with.

Copied rather than shared: these are standalone `sh` scripts run by path, so a sourced helper would buy one line at the cost of a resolution mechanism. `jstr` now appears in two files; before this it appeared in exactly one, confirmed by grepping the whole worktree tree excluding `.git`.

## Severity boundary

Evidence corruption, not decision corruption — the exit code was always right. No caller in this repo parses this payload: `next-ticket/SKILL.md:36` documents the exit code as the contract, `run-team/SKILL.md:65` states "any hit = taken", and `release-ticket.test.mjs:161` spawns the script and reads only `.status`. The reproduction above exits 1 before and after.

## Tests

Two, shaped after the two `jstr` cases in the release-ticket suite — one per reachable vector, since a ref and a path admit different characters. Both fail on the parse at the pre-fix script.

`./agent-test skills/fleet/scripts/inflight.test.mjs` 22/22, `release-ticket.test.mjs` 32/32 (the cross-suite consumer), both re-run after rebase onto `origin/main`.

Not covered: a control character or tab in a worktree path. `jstr`'s `tr` would neutralise it, but probe 3's `awk -F'\t'` pipeline splits on a tab and `read -r` on a newline before the payload is ever built — a separate pre-existing limit, out of scope here.

Closes #90


feigi:
Review deferrals, all filed as their own issues:

- **#119** — five more fleet scripts emit unescaped JSON (`verify-sha`, `claim-ticket`, `worktree-audit`, `prove-merge`, `reap`); `prove-merge.test.mjs` `JSON.parse`s its stdout so it has a real consumer. Also carries the lift-vs-share decision: keeping the copy was right at two sites, and the extraction point is that sweep. `ready-for-agent`
- **#120** — a broken `sed`/`tr` makes probe 2 report a genuinely-taken ticket as **FREE** (exit 0). Pre-existing, reproduced identically at `a56c01a` and at this PR's head, and a *decision* bug rather than an evidence one. Also covers the evidence-blanking this PR introduces at `inflight.sh:240-242`, where a command substitution's status is discarded so `|| die` structurally cannot fire. `needs-triage`
- **#121** — `inflight.sh 007` clears the all-digits guard and emits `{"issue":007,...}`, which no parser accepts. `ready-for-agent`
- **#122** — a tab in a worktree path truncates the reported path to a value that exists nowhere on disk (payload still parses, verdict still correct). `needs-triage`

`no-undo-audit.sh` has the same escaping defect but is already **#106** item 2 — referenced there rather than duplicated.

Applied here instead of deferred: a mutation matrix showed `$remote` and the `tr` stage were both free to delete with the suite staying green, so both are now pinned (4 of 24 tests fail against the pre-fix script, was 2).

feigi:
Second review pass — the comment-analyzer and pr-test-analyzer reports landed after the first push.

**pr-test-analyzer independently reproduced both gaps already fixed in `a707462`** (unwrapping `$remote` and dropping the `tr` stage each survived its 12-mutant matrix). No new code findings.

**comment-analyzer found three false assertions in the new comments**, now fixed in `4a669d0`:

- `inflight.sh` — "Byte-safe for UTF-8, whose bytes are all `>= \200`". Half of UTF-8 is ASCII (`printf 'aé'` → `97 195 169`). `release-ticket.sh:113` scoped this to *"the UTF-8 in these messages"*; the lift dropped the scope and made it universally false.
- `inflight.sh` — "The four fields are names chosen elsewhere". Three are; `$pr` is composed by this script.
- `inflight.test.mjs` — claimed next-ticket **and** run-team both read the exit code as the decision. Only next-ticket does; run-team says "any hit = taken".

Remaining cross-file line citations in those blocks are **dropped rather than re-pointed** — **#129** records four already-stale ones in these same two files, one naming a construct absent from the cited file entirely.

Also fixed: the `git()` helper this PR added passed `process.env` rather than the fixture's hardened `env`, so an inherited `GIT_DIR` left a `fix-42-say"hi` branch in whatever repo it named. Victim now stays clean.

Deferrals total: **#119, #120, #121, #122, #129**.

Head `4a669d04fae006a0be71cad03ced73d4821eae9c` — `./agent-test skills/fleet/scripts` 174/174.

Issue #64, opened 2026-07-30
candidates.mjs: "every candidate was filtered out" is byte-identical to "the queue is empty"


feigi:
## What to build

The file states its own contract: the caller must distinguish outcomes without reading stderr (1 = successful query, no survivors; 2 = query broke). After the to-spec filter landed there is a third outcome with no signal — the query returned rows and every one was dropped. On stdout and exit code it is identical to a genuinely empty queue.

A fleet controller should be able to say "the only `ready-for-agent` items are specs — run `to-tickets`" instead of "no work".

**Undecided, needs a call:** distinct exit code for all-filtered, versus a machine-readable `dropped: N` line on stdout alongside exit 1. Both satisfy the contract; they differ in what breaks for existing callers.

## Acceptance criteria

- [ ] A caller can tell all-filtered from genuinely-empty without parsing stderr
- [ ] The chosen mechanism is documented in the file header next to the existing exit-code contract
- [ ] Tests cover all three outcomes

## Blocked by

- Blocked by #60 — that ticket moves where filtering happens relative to the emptiness test, which is the same code this one changes

Found by the review of #59; deferred from that PR.

feigi:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** The candidate scan reports "queue is empty" when the query actually returned rows and every one was filtered out — the caller cannot tell the two apart.

**Decision made during triage:** the undecided call in the issue body is settled — **use a distinct exit code (3)**, not a `dropped: N` line on stdout. Rationale below under *Why exit code, not stdout*.

**Verified during triage.** Reproduced against merged `main` (PR #59 landed 2026-07-30T22:11:05Z) by stubbing `gh` to return two rows that both carry the to-spec signature heading:

| case | stdout | exit |
| --- | --- | --- |
| 2 rows returned, both filtered as specs | `[]` | `1` |
| query genuinely returned 0 rows | `[]` | `1` |

Byte-identical on both channels. The only difference is stderr.

**Current behavior:**
[...]

Issue #114, opened 2026-07-31
fleet-plugin-design: the Script surface table's Non-zero column was written from intent — audit all eleven rows (candidates.mjs states the opposite of its code)


feigi:
Deferred from PR #108 review. **Widened after review:** this started as "the three `never` rows are wrong" and turned out to be a whole-column defect, including one row that states the *opposite* of its code. One audit of all eleven rows closes the class instead of one row per wave.

The column is `Non-zero when` in the Script surface table of `docs/specs/2026-07-23-fleet-plugin-design.md:160-174`. PR #108 corrected exactly one row (`inflight.sh`) and correctly stayed in scope. Issue #45's premise that the other rows were sound is false.

### 1. `candidates.mjs:164` — the row states the opposite of the code

Row: *"`gh` failure only; empty result is exit 0 with `[]`"*

```
$ sed -n '155,157p' skills/fleet/scripts/candidates.mjs
# Exit 1 for a successful query with no survivors. The caller must be able to
# tell "the queue is empty" from "the query broke" (2) without reading stderr.
process.exit(rows.length === 0 ? 1 : 0);

$ ./skills/fleet/scripts/candidates.mjs --require-label no-such-label-xyz ; echo rc=$?
[]
rc=1
```

Empty result is exit **1**, not 0. This is the worst of the set: a caller trusting the row reads an empty queue as a broken query, or gates on exit 0 and never sees the queue at all.

### 2. Three `never` rows that all exit 2

```
$ ./skills/fleet/scripts/reap.sh a b                -> rc=2   (row :170 says "never — refusals are findings")
$ (cd /tmp && .../reap.sh)                          -> rc=2
$ ./skills/fleet/scripts/pr-overlap.mjs             -> rc=2   (row :163 says "never — it reports, the model rules")
$ ./skills/fleet/scripts/pr-overlap.mjs --a 108     -> rc=2
$ (cd /tmp && .../worktree-audit.sh)                -> rc=2   (row :171 says "never")
```

`pr-overlap.mjs` is self-refuting: its own header at `:9-10` reads *"Exit 2 only when a query failed and the answer is therefore unknown."* `reap.sh:18` dies 2 on bad usage `:22`, not-a-repo `:25`, fetch failed `:28`, unresolvable base `:29` — its *refusals* are findings, but its *preconditions* exit 2. `worktree-audit.sh` dies 2 at `:17,20,21`.

### 3. Two incomplete rather than false

[...]

Issue #63, opened 2026-07-30
candidates.test.mjs overstates its jq fidelity — gh applies --jq with embedded gojq, not the system jq the stub execs


feigi:
## What to build

The header comment claims the stub runs "the real jq … the same expression gh would have received". The expression is real, but the *engine* is not: `gh` applies `--jq` with its own embedded gojq, and the stub execs system jq. Flavour-sensitive predicates are therefore not covered by the suite.

Comment-only change — the current predicate is confirmed portable across both engines by a live run against `feigi/claude-config`.

## Acceptance criteria

- [ ] The comment says the stub exercises the expression under system jq
- [ ] It names embedded gojq as a different engine, so regex-flavour differences are out of the suite reach
- [ ] No test behaviour changes

## Blocked by

- None — can start immediately

Found by the review of #59; deferred from that PR.

Issue #61, opened 2026-07-30
candidates.mjs fails open on malformed input — a broken query exits 1, the code reserved for "queue empty"


feigi:
## What to build

Two paths let a broken invocation read as a successful-but-empty one, defeating the fail-closed design the file states for itself:

1. **No shape check after the parse.** `gh` output is consumed as `{n,t,l,d,spec}[]` with no validation. Any throw after the parse exits 1 — the code reserved for "successful query, no survivors" — so a `--jq` reduction that silently did not apply is indistinguishable from an empty queue.
2. **`--limit` as the last argv element.** `arg()` returns `undefined`, and `??` treats that as "flag absent", so the value falls back to 500 instead of hitting the existing positive-integer guard.

Both should `die()` (exit 2).

## Acceptance criteria

- [ ] Output that is not an array of the expected row shape exits 2 with a message naming the reduction
- [ ] `--limit` with no value exits 2 rather than defaulting to 500
- [ ] Tests assert the exit codes, not just stderr

## Blocked by

- None — can start immediately

Found by the review of #59; deferred from that PR.

feigi:
Found a third instance of the same `arg()` root cause during the PR #116 review — same fix location, so recording it here rather than as a duplicate issue.

**`--require-label` as the final argv element.** `arg("require-label")` returns `undefined`, and `query()` at `:50` treats a falsy label as "no label", so the script runs the **unfiltered** query and exits 0. Verified at head:

```
$ gh issue list --state open --limit 500 --search -label:in-progress -label:onhold … --json … --jq …
    3 candidate(s)
exit=0
```

Note the `--search` string carries no positive `label:` term at all. This is worse than the `--limit` case already in the body: it does not merely pick a wrong default, it reaches **the exact effect `--allow-fallback` gates** — widening the net to `ready-for-human` — without the flag being passed. The file's header states the fleet must never widen on an empty `ready-for-agent` queue, and `run-team/SKILL.md:61` deliberately withholds `--allow-fallback` to enforce that. A trailing `--require-label` silently defeats it.

Not reachable from either real caller as currently written, but both callers are markdown read by a model, so a malformed invocation is more plausible here than in a hand-typed CLI.

Suggest extending the acceptance criteria:

- [ ] `--require-label` with no value exits 2 rather than running the unfiltered query

The general form — `arg()` returning `undefined` for any trailing flag — may be worth fixing once in `arg()` itself rather than at each call site.

PR #116, opened 2026-07-31
fix(fleet): a queue of only specs suppressed --allow-fallback and reported "no work"

### LINKED ISSUE #60: candidates.mjs: a labeled queue of only specs suppresses --allow-fallback and reports "no work"
## What to build

`dropSpecs` runs after the `--allow-fallback` block, so the emptiness test that gates the fallback sees the *unfiltered* row count. A `ready-for-agent` queue containing only to-spec specs therefore never triggers the unfiltered retry: the script reports an empty queue while real work exists.

Filtering must happen before the emptiness test that decides the fallback, while the cap check keeps seeing the raw pre-filter length (the ordering the file already calls load-bearing).

## Acceptance criteria

- [ ] A labeled query returning only specs triggers the unfiltered fallback
- [ ] `refuseIfCapped` still runs against the pre-filter length for both the labeled and the fallback query
- [ ] The existing cap-before-drop test still passes
- [ ] A test covers the all-specs-then-fallback path

## Blocked by

feigi:
## What broke

`dropSpecs` ran *after* the `--allow-fallback` block, so the emptiness test gating the fallback read the **unfiltered** row count. A `ready-for-agent` queue holding nothing but to-spec specs never triggered the unfiltered retry — every row was filtered out one line later, the payload came back `[]` and the script exited 1 ("the queue is empty") while claimable work sat in `ready-for-human`.

Reproduced before the change, with a label-aware `gh` stub — two specs labeled `ready-for-agent`, one real ticket only reachable unfiltered:

```
    2 candidate(s) with label:ready-for-agent
    dropped #10 — to-spec spec, not a ticket (## User Stories)
    dropped #11 — to-spec spec, not a ticket (## User Stories)
[]
exit=1
```

One `gh issue list` invocation — the fallback was never attempted. After:

```
    2 candidate(s) with label:ready-for-agent
    dropped #10 …
    dropped #11 …
candidates: empty with label:ready-for-agent; --allow-fallback given, retrying unfiltered
    1 candidate(s) unfiltered
    #12 [ready-for-human] real ticket
[{"n":12,…}]
exit=0
```

## The fix

Drop specs **between** the cap check and the emptiness test, in both the labeled query and the fallback. The two orderings are in tension and both are load-bearing, so neither moved off its anchor:

- `refuseIfCapped` still runs **first** and still sees the raw pre-filter length — filtering first shrinks the array below `--limit` and the cap check stops seeing a truncated list.
- The emptiness test now sees the **post-filter** length — a queue whose every row was filtered out is an empty queue.

`--allow-fallback` has exactly one caller, `next-ticket`. `run-team` deliberately does not pass it, so the fleet's "empty means no work" rule is untouched.

## Tests

Two added, each mutation-checked against the ordering it pins:

| Mutation | Fails |
|---|---|
| `dropSpecs` back to after the fallback block (the bug) | both new tests, nothing else |
| drop before `refuseIfCapped` **inside** the fallback | only the new fallback-cap test |
| drop before `refuseIfCapped` in the **labeled** query | only the pre-existing cap-before-drop test |

The pre-existing "cap is checked before specs are dropped" test still passes and still discriminates. `172/172` in `skills/fleet/scripts`.

The `gh` stub now serves a fixture per query, so "the fallback ran" is provable from the payload rather than from a log line the script prints before it has an answer. The discriminator matches `` label:`` **with the leading space**: every exclusion is spelled `-label:`, so a bare `label:` match also catches the unfiltered query and silently serves it the labeled fixture — that bug in the harness made both new tests fail against an already-correct fix.

## Not in scope

"Every candidate was filtered out" and "the queue is empty" remain byte-identical to a caller reading only the exit code — that is #64, left alone deliberately.

Closes #60


feigi:
## Review deferrals

Applied in `8e40b97`; everything below is deferred, none dropped.

| # | Finding | Label |
|---|---|---|
| #132 | `process.exit()` truncates the payload at 64 KiB on a pipe — 499 rows deliver 124797 bytes to a file, **65536 to a pipe**, unparseable, exit 0. `refuseIfCapped` never fires (count guard vs. byte truncation). Pre-existing line; this PR newly routes the all-specs case into the largest payload. | `ready-for-agent` |
| #133 | The fallback path logs every spec drop twice (production unfiltered ⊇ labeled) — 4 `dropped #` lines for 2 specs. Approach genuinely open, including "accept it". | `needs-triage` |
| #61 (comment) | `--require-label` as the final argv element runs the **unfiltered** query at exit 0 — the exact effect `--allow-fallback` gates, without the flag. Same `arg()` root cause already filed there, so recorded as a comment rather than a duplicate. | existing |

Not re-filed, already tracked: the non-array-payload → exit 1 contract violation is **#61**; `docs/specs/2026-07-23-fleet-plugin-design.md:164` claiming this script exits 0 on empty is **#114** (four independent rediscoveries this run).

Confirmed out of scope and untouched, by direct comparison against the base: **#64** — the `empty with label:X` message is byte-identical with only its line number moved, and head vs. base stdout **and** stderr are byte-identical on the no-`--allow-fallback` path. **#63** — the `JQ` constant is byte-identical, so no new jq-flavour-sensitive assertion was added; the two new tests inherit that gap without widening it.

feigi:
## Review deferrals — updated after the full six-specialist fan-out

Second round of fixes in `a0722ab` (all six specialists now in; three of them independently found the same unpinned line, and the last two landed better fixes than my first pass).

**Additional deferral:**

| # | Finding | Label |
|---|---|---|
| #134 | Nothing in the suite asserts exit **1**. `process.exit(0)` hardcoded leaves 8/8 green, so the new test's `assert.equal(status, 0)` never discriminates on its own and its comment asserts a contract only half present. Pre-existing gap, surfaced by the new comment. | `ready-for-agent` |

Previously filed, unchanged: **#132** (64 KiB pipe truncation), **#133** (double-logged spec drops), **#61** (comment: trailing `--require-label` runs the unfiltered query).

**What changed in `a0722ab`, and why it supersedes my first pass:**

- The fallback fixture now carries spec **#13**, so the assertion already written discriminates — `[12, 13]` vs `[12]` on the mutant. Deleting the fallback's own `dropSpecs` previously left **all 8 green** while a spec shipped as a claimable ticket. Cheaper than the extra assertion I first added, and it pins the real harm rather than the key set.
- The stub's subject is **padded** (`" $search "`). `gh` ignores qualifier order, so moving the positive term to the front of `search` is behaviour-preserving for real `gh` but inverted the discriminator, serving the **labeled** query the unfiltered fixture — byte-identical payload, identical exit code, test green, fallback never taken. Padding fixes that at the source instead of at each assertion. Verified: un-pad and the reorder is caught rather than silently served.
- Added `assert.equal(stderr.match(/^\$ gh /gm).length, 2)` — the `retrying unfiltered` line prints *before* the query, so it pins the branch being entered, never that its answer shipped.

**Final mutation matrix** (172/172 throughout): the original bug fails both new tests only; each cap swap fails its own test only; deleting the fallback strip fails the fallback test; and a variable rename, a log reword, and the search reorder all stay green.

Issue #111, opened 2026-07-31
fleet: the whole phase-3 merge path gates on CI, and this repo has none — every PR strands unlabelled


feigi:
Hit live during a `/fleet:run-team` run with four open PRs (#107, #108, #109, #110). The controller could not reach the finisher gate for any of them, because every gate in the skill is expressed in terms of a CI signal this repository does not produce.

## Measured

```
$ git ls-tree -r origin/main --name-only | grep -i "\.github"
(no output)

$ gh run list --limit 5
(empty)

$ gh pr checks 107
no checks reported on the 'fix/26-specialist-count' branch

$ skills/fleet/scripts/ci-state.mjs --pr 107 --quiet
ci-state: cannot read .github/workflows/ci.yml: ENOENT: no such file or directory
exit 2
```

No `.github/` is tracked at `origin/main`, there has never been a workflow run, and no PR carries a check. `ci-state.mjs:54` defaults `workflowFile` to `.github/workflows/ci.yml` and dies at exit 2 — correctly, by its own fail-closed contract. It is not wrong; it is being asked an unanswerable question.

## Why that stops the loop rather than degrading it

`skills/fleet/skills/run-team/SKILL.md` routes the entire post-review path through CI:

- the event-loop edge is *"**Monitor: CI run completes** → bind it (`ci-state.mjs --pr <N>`); the diff-validating `check` job green with no heavy job in `failure` → dispatch a finisher to label"*. With no runs, that edge never fires.
- the finisher gate is *"When the diff-validating `check` job is green **and no heavy job is in `failure`** … dispatch a finisher"*, and explicitly *"Gate on the `check` job, **not** on `ci-state --quiet` exit 0"*. Both predicates are unreachable — there is no `check` job to be green and none to be in `failure`.
- `skills/fleet/commands/review-and-fix.md` tells reviewers to watch checks until green.

So `ready-to-merge` can never be added by the documented path, and since a reviewer must not add it and the controller must not either, **every PR strands unlabelled and the merge bot never has a wave**. Nothing errors; the queue just stops. Same silent-stall shape as #3, from a different direction.

Note the failure is *safe* — nothing merges unverified — but it is total, and it is invisible: "no checks reported" reads identically to "checks have not started yet".

## The decision this needs (why `needs-triage`)

What should the fleet do when a repo has no CI? None of these is obviously right:

[...]

PR #130, opened 2026-07-31
fix(fleet): the default testCmd named a file the snapshot cannot contain

### LINKED ISSUE #101: review-pr.js: the default testCmd is `./agent-test`, which never exists in the snapshot the same prompt tells specialists to run it from
Deferred from PR #75 review. This is the reviewer's ruling on the `workflows/review-pr.js:241` item flagged in #42; PR #75 deliberately left it alone and the ruling here disagrees, on different grounds than #42 named.

Two separate defects in the same instruction.

## 1. The default command cannot exist in the snapshot

`workflows/review-pr.js:113`

```js
const testCmd = (args && args.testCmd) || "./agent-test";
```

`workflows/review-pr.js:241` then tells every specialist:

> Tests: use '${testCmd}' from the snapshot; never a bare test command, …

The snapshot is cut with `git archive HEAD | tar -x` (`review-pr.js:174`), which carries **tracked files only**. `agent-test` is written into the worktree by `claim-ticket.sh` and added to `.git/info/exclude` — untracked by construction. So whenever the caller does not pass `args.testCmd`, the instruction resolves to `./agent-test` and every specialist that follows it literally gets `No such file or directory` in the snapshot. `review-and-fix.md:41` already names this exact failure ("`git archive` carries tracked files only … specialists silently have no runnable suite and reason from source instead of measuring") without connecting it to the default.

Callers that pass `testCmd` explicitly are fine; the default is the trap.

## 2. "never a bare test command" with no worked example still invites a silently-unsafe guess

PR #75 argued this hazard was cured, since the guess it invited — a directory — now works. The guess space narrowed; it did not close. A quoted glob is not a directory, so it passes straight through the new shim, and node's zero-match glob is a silent green:

```
./agent-test 'empty/**/*.test.mjs'      ℹ tests 0  EXIT=0
./agent-test 't/*.nomatch.mjs'          ℹ tests 0  EXIT=0
./agent-test 'nosuchdir/**/*.test.mjs'  ℹ tests 0  EXIT=0
```

That is the same row PR #75's own table flags as "the hazard". A specialist that guesses a glob still reports a green having run nothing.

## Fix

Give `:241` a worked example rather than a prohibition, and make the default something that exists in a snapshot — for this repo, `node --test skills/fleet/scripts/*.test.mjs`, which is what `review-and-fix.md:43` already tells reviewers to hand out by hand.

feigi:
`review-pr.js` tells every specialist to run `testCmd` **from the snapshot**, and defaulted it to `./agent-test` — a file the snapshot cannot contain.

Closes #101

## The defect, measured

The snapshot is cut with `git archive HEAD | tar -x`, which carries tracked files only. `agent-test` is written into the worktree by `claim-ticket.sh:27` and added to `.git/info/exclude` at `:166` — untracked by construction. Cutting a snapshot from this branch's base and running the old default as a specialist would:

```
$ ls snapshot/agent-test         ->  No such file or directory
$ git ls-files agent-test        ->  (empty)
$ cd snapshot && ./agent-test    ->  bash: ./agent-test: No such file or directory   EXIT=127
```

Nothing in this repo passes `args.testCmd` (`grep -rn testCmd` finds only the definition, the prompt, and `review-and-fix.md:33` naming it as an arg), so the default is what actually fired. Every specialist following the instruction literally got `No such file or directory` and reasoned from source instead of measuring — the failure `review-and-fix.md:41` already names without connecting it to the default.

## The second defect: "never a bare test command" invited an unsafe guess

The prohibition had no worked example, and the guess space was never closed. A glob that matches nothing is a **silent green** — measured on Node v26.5.0:

```
bash:  node --test nosuchdir/*.test.mjs   ->  tests 0   EXIT=0     <- green, ran nothing
zsh:   node --test nosuchdir/*.test.mjs   ->  no matches found     EXIT=1
```

`node --help` lists no flag that fails a zero-test run, so the shell cannot be the guard and the reading rule has to be. `node --test <dir>` is not an escape either: on v26.5.0 it loads the directory as a module and dies `MODULE_NOT_FOUND`, so the glob form is the only working shape.

## The fix

- Default is now `node --test skills/fleet/scripts/*.test.mjs` — tracked, so it survives the archive.
- The prompt hands that command over **verbatim and interpolated**, and carries the reading rule: `'tests 0' is a FAILED run`, report the dimension unrun.
- `review-and-fix.md:43` hands out the same glob by hand and is the most copy-pasted command in the fleet. It now carries the same rule, and both sites name one command. (That file was owned by PR #80, which has merged — verified its merge commit is an ancestor of this branch.)

## Verification

Snapshot cut from this branch's HEAD, default read out of the snapshot's own `review-pr.js` and run inside it:

```
agent-test:  absent
node --test skills/fleet/scripts/*.test.mjs   ->  tests 173  pass 173  fail 0  skipped 0  EXIT=0
```

Local suite in the worktree: **173/173 pass, 0 skipped**. There is no CI in this repo (#111); the local suite is the gate.

## On the new test

> **Superseded by review commit `1e158b7`.** The design below — three tests parsing the command string and checking its arguments against `git ls-files` — was replaced during review. It reported **three green** on the default `./agent-test skills/fleet/scripts/*.test.mjs`, i.e. this ticket's own defect, because `args.slice(1)` drops the executable and only the arguments were ever checked. It also passed `node --test skills/*.mjs`, which runs zero tests at exit 0, because `git ls-files` and `fs.globSync` use different glob languages. The file now cuts a `git archive` snapshot and **runs** the default in it, asserting exit 0 and `tests N > 0`. The probe table below describes the superseded design; see the review comment for the current one.

`review-pr-testcmd.test.mjs` guards the invariant, not the string: whatever the default becomes, its file arguments must be tracked. Mutation-probed — each of these fails it and the tree restores clean:

| probe | result |
|---|---|
| default back to `./agent-test` | 2 of 3 fail |
| default to a zero-match glob | 2 of 3 fail |
| reading rule deleted from the prompt | 1 fails |
| worked example hardcoded instead of interpolated | 1 fails |

Two of those probes were worth the trouble. The reading-rule probe **initially passed** — the assertion matched `/tests 0/`, which the surrounding explanation also contains, so it could not fail. Same silent-green class as the ticket. It now matches the ruling itself.

And the end-to-end snapshot check caught the guard breaking the command it guards: the test shells to `git ls-files`, it is one of the files the new default matches, and an archive copy is not a git repo — `fatal: not a git repository`, EXIT=1, which would have painted the default red for every specialist. `tracked()` now asks git only where there is git; inside a snapshot every file present is tracked by construction of the archive, so `fs.globSync` answers the same question there. The worktree branch stays `ls-files`, the only check that can see the original defect, since `agent-test` exists on disk and is untracked.

Scope is the two files the ticket names plus the test; `claim-ticket.sh`'s runner is untouched (#97/#98/#99 territory).


feigi:
Review complete — six specialists (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, type-design-analyzer, code-simplifier), fan-out dispatched from one immutable snapshot.

**The fix itself holds.** Every empirical claim in the PR body reproduces: `agent-test` is absent from `git archive HEAD` by construction (79 files extracted = 79 tracked), the new default runs 173/173 in the snapshot, bash zero-match glob → `tests 0` exit 0 while zsh exits 1, `node --test <dir>` → `MODULE_NOT_FOUND`, no `node --test-*` flag fails a zero-test run, and the two command strings in `review-pr.js:120` and `review-and-fix.md:43` are byte-identical.

**Applied here (1e158b7)** — the guard, not the fix. The test file passed the very default this ticket is about:

| probe | old guard | new guard |
|---|---|---|
| `./agent-test skills/fleet/scripts/*.test.mjs` | **pass 3** | fail |
| `./agent-test skills/fleet/scripts/` | **pass 3** | fail |
| `node --test skills/*.mjs` (runs 0 tests, exit 0) | **pass 3** | fail |
| ruling moved out of the specialist prompt | **pass 3** | fail |
| snapshot inside an unrelated repo | **2 false reds** | skip, with reason |
| different valid tracked default | pass | pass |

Root causes: `args.slice(1)` dropped the executable, which is where `./agent-test` lives; `git ls-files` and `fs.globSync` are different pattern languages (git's `*` crosses `/`, node's does not); the prompt slice ran to EOF; and `rev-parse --git-dir` walks up. The two parsing tests are replaced by one that cuts the archive and **runs** the default in it, asserting exit 0 **and** `tests N > 0`. Also corrected `claim-ticket.sh:27` → `:110` (`:27` is the `runner=` assignment, not the write site).

Suite: **172/172, 0 skipped** in the worktree at `1e158b7`.

**Deferred, each filed as its own issue:**

| # | finding | label |
|---|---|---|
| #137 | `FINDINGS_SCHEMA` has no field for the test run — a dimension that never ran is indistinguishable from one that ran green; `dimensionsRun` names what was dispatched | ready-for-agent |
| #138 | `review-pr.js:267` — a crashed reviewer yields `[]`, read as "ran, found nothing"; the verifier path at `:290-297` handles the same case correctly | ready-for-agent |
| #139 | `review-pr.js:31` — `scope_searched` documented "Required", absent from `required` | ready-for-agent |
| #140 | `snap.path` interpolated into every specialist prompt without an existence check | ready-for-agent |
| #141 | `run-team/SKILL.md:515-519`, `isolation.md:11`, `inflight.test.mjs:3` still pair the snapshot with `./agent-test` | ready-for-agent |
| #142 | the new default is claude-config-specific in a **global** workflow — off-repo it turns a loud exit 127 into a silent `tests 0` exit 0 | needs-triage |
| #143 | the `tests 0` rule is too narrow — a partial tree and an all-skipped run both pass it | ready-for-agent |

Branch is behind `main` but `MERGEABLE`/`CLEAN`; not rebased (staleness is not a failure, and the merge bot re-establishes currency). This repo has no CI (#111), so the gate is the local suite plus the verification above.

Issue #119, opened 2026-07-31
fleet scripts: five more emit unescaped JSON, and jstr is now duplicated


feigi:
Deferred from PR #110 review (which fixed `inflight.sh`).

Eight shell scripts build JSON with `printf '{'`; only two escape their string fields. Re-derived at `d663193`:

```
$ git grep -ln "printf '{" -- 'skills/fleet/scripts/*.sh'   -> 8 files
$ git grep -ln jstr          -- 'skills/fleet/scripts/*.sh'   -> inflight.sh, release-ticket.sh
```

Unescaped, with the raw string fields:

| script | line | fields |
|---|---|---|
| `verify-sha.sh` | 41-42 | `$branch`, `$tip` |
| `claim-ticket.sh` | 170-171 | `$branch`, `$wt`, `$install`, `$runner` |
| `worktree-audit.sh` | 53-54 | `$wt`, `$short` |
| `prove-merge.sh` | 117-118 | `$second`, `$first`, `$proof_path` |
| `reap.sh` | 104-105 | `${reaped}`, `${kept}` |

(`no-undo-audit.sh` has the same defect but is already filed as #106 item 2 — fix it there, not here.)

Same vectors as #90: branch names accept `"`, worktree paths accept `"` and `\`.

**Priority: `prove-merge.sh`.** `prove-merge.test.mjs` `JSON.parse`s its stdout, so it has a real consumer that breaks exactly as `inflight.sh` did.

## On centralising `jstr`

`jstr` now lives in two files, byte-identical. PR #110 kept the copy deliberately and that was right at two sites — no script in `skills/fleet/scripts/` sources another (`git grep -nE '^[[:space:]]*(\.|source)[[:space:]]+'` over `*.sh` → zero hits), so sharing meant inventing a resolution mechanism to save one line.

At seven sites it stops being defensible. If this ticket is taken as a sweep, extract `skills/fleet/scripts/json.sh` sourced via `. "$(dirname "$0")/json.sh"` — every one of these is already run by path, so `$0` is the mechanism and nothing new is needed.

**One hazard to handle explicitly:** under `set -e` a failed `.` exits non-zero, and `inflight.sh`'s contract reads exit 1 as *taken*. A missing lib would fabricate a claimed ticket. Guard the source with an explicit `|| die` rather than letting `set -e` decide.

feigi:
Deferred from PR #227 review (issue #146) — commenting rather than filing, since this issue already owns the duplication.

[...]

PR #131, opened 2026-07-31
fix(fleet): a conflicting path with a space made no-undo-audit answer "nothing at risk"

### LINKED ISSUE #106: no-undo-audit.sh: conflicting paths with spaces or quotes silently break atRisk and the JSON

feigi:
`no-undo-audit.test.mjs` called itself this script's regression gate while covering only the refusal. No fixture ever produced a conflict, so `conflicts[]` and `atRisk[]` — the half of the audit that runs on a **passing** run — had no coverage at all and held three live defects. All three are exit 0 with a payload emitted and the answer wrong, which is the one outcome the script's own header says it exists to prevent.

## The three named defects

**1. A conflicting path with a space reported `atRisk: []`.** The pathspec was unquoted, so `has"quote and space.txt` became several pathspecs matching nothing. Measured on that fixture with main holding the later commit:

```
ground truth:  git log -- 'has"quote and space.txt'  ->  MAIN COMMIT AT RISK
script:        {"conflicts":[...],"atRisk":[]}            exit 0
```

**One correction to the issue's account.** It attributed the silence to `2>/dev/null || true`. Measured here, `git log` spends **exit 0** on a pathspec that matches nothing — so there was never an error to swallow, and nothing was hidden. There was simply no output. The swallow goes anyway, because a `git log` that genuinely fails leaves the identical empty answer, and that one is unanswerable rather than safe. `git merge-base` loses its `|| true` for the same reason: an empty `$fork` skipped step 3 outright.

**2. A path containing `"` emitted malformed JSON on a passing audit.** `jstr()` is lifted from `release-ticket.sh:115` — the same helper and the same pipeline PR #110 lifted into `inflight.sh`, backslashes before quotes, the order a reviewer there measured as load-bearing. `jarr()` is its array form, sparing `\012` because that is the record separator here rather than part of an element.

`-z` on `merge-tree` is the other half, and the part the issue did not name: without it git C-quotes such a path, so what reached the caller — and the `git log` above — was the *rendering* `"has\"quote and space.txt"`, which is neither a real pathspec nor a JSON string. The test asserts the real path, so it fails on `-z` alone. NUL not surviving a command substitution is why the output now lands in a file.

**3. The exit-2 guard could not fire for the case that reaches it.** The issue flagged this one as not re-run; reproduced here on git 2.50.1, `BASE_REF` naming a tag that dereferences to a blob (which satisfies the `rev-parse --verify` guard and gets as far as merge-tree):

```
merge-tree: blobtag - not something we can merge     exit 1
script: {"conflicts":[],"atRisk":[]}                 exit 0   <- safe
```

Exit 1 is not only "ran fine, found conflicts" — git spends it on that refusal too, so the exit code cannot carry the distinction. Every run that ran at all prints the tree OID first, so the guard is on empty output: the outcome the exit code will not name. Now exit 2, no payload.

## Gating

Tests were written first and committed red in e025e88 (15 tests, 11 pass, 4 fail), then greened. Each fix was **mutation-checked individually** — reverting it on its own turns the test that claims to gate it red:

| Mutation | Goes red |
|---|---|
| drop `-z` | the quoted-path test |
| `jarr` stops escaping quotes | that test + the commit-subject test |
| `jarr` escapes quotes *before* backslashes | both, confirming PR #110's order finding |
| drop the empty-output guard | the blob-tag test |
| restore the word-split pathspec | the quoted-path test |
| final `printf` stops escaping | the branch/worktree test |

`node --test skills/fleet/scripts/*.test.mjs`: **178 tests, 178 pass** on the whole fleet suite after rebase. `shellcheck` clean. (No CI in this repo — issue #111.)

## Weigh these separately — beyond the three named defects

The issue invited these as "adjacent gaps worth folding in". Each is separable:

- **`atRisk`, `worktree` and `branch` gained escaping the issue did not ask for.** `atRisk` escaped quotes but not backslashes; `worktree` and `branch` were raw. A worktree path is a filename so it admits both `"` and `\`; git accepts `"` in a ref name (`check-ref-format`, measured — `\` it rejects). Same defect class, same `printf`, and `review-pr-110`'s follow-up ticket deliberately excludes this file because #106 covers it, so no one else will take it.
- **Exit-2 coverage** — five paths had no test; one compact table-driven test asserts exit 2 *and* that no payload is emitted. I changed exit-2 behaviour, so this guards my own change.
- **Two dirty shapes** — every existing dirty fixture is untracked-only and so never touches the index. Adds a modified tracked file and a staged change.
- **`GIT_CONFIG_COUNT` / `GIT_CONFIG_PARAMETERS`** joined the unset list; they outrank the config files and bypass `GIT_CONFIG_GLOBAL=/dev/null`. Two lines against a false red nobody could reproduce by hand.
- **One spec doc line.** The contract table listed only the dirty-worktree refusal. Exit 2 always existed but is newly reachable for a realistic bad ref, so the row would now mislead a caller.

## Known ceiling

Marked `ponytail:` in place: NUL is translated back to newline for the section split, so a filename containing a literal newline would still split. Closing it needs a NUL-capable reader (`perl -0`, GNU `sed -z`) — macOS awk 20200816 reads `RS="\0"` as `RS=""` and silently switches to paragraph mode (measured), and POSIX sh has no `read -d ''`. Spaces and quotes, the shapes actually at issue, are covered.

Closes #106


feigi:
Review complete. Six specialists dispatched over an immutable snapshot; the three named defects are genuinely fixed and genuinely gated — each of the six claimed mutations reddens only its own test, re-derived independently, and the tests really were red first (15 tests, 11 pass, 4 fail against the pre-fix script; the test file is byte-identical at `e025e88` and the fix commit, so the pairing is valid).

Two of the PR's own claims were checked hardest and both hold. `git log` exits **0** with empty stderr on a pathspec matching nothing, so issue #106's attribution to `2>/dev/null || true` is the misleading text and this PR's correction is right. And quoting alone is insufficient: without `-z` git renders `has"quote and space.txt` as `"has\"quote and space.txt"`, which matches no file — though `-z` earns that on the `"`, not on the space, which git does not quote at all.

## Fixed here (pushed as `c17d4b6`)

**One blocker.** `-z` moved the section split to NUL, and the split translates NUL back to newline so macOS awk can read it — so a filename holding a literal newline manufactures the empty record that ends the filename list. Measured with main really holding the commit: base emitted unparseable JSON (the caller crashed, loudly), head emitted `{"conflicts":[],"atRisk":[]}` at exit 0. A strict downgrade to a silent false safe, on the one input class this script exists to catch. Newline is now parked on `\001` before NUL becomes the separator, and a path that still holds one is exit 2.

Also fixed: `:(literal)` on each pathspec — `--` ends the options, not the magic, so a real file named `:colon.txt` matched nothing and reported `atRisk: []` at exit 0 (the test fixture needed the same prefix to `git add` it). A `|| die` on the payload `printf`, which exited **1** on a failed write — and 1 means "REFUSED, worktree dirty". Two decoy commits in `conflictRepo`, because without them dropping the pathspec filter *or* the range from `git log` left the suite green; both mutations now redden three tests. A control byte in the commit-subject fixture, pinning a C0 scrub that could previously be deleted whole. Four comment claims that measurement contradicted, including a `jstr` citation to `inflight.sh:227` — a line that did not exist when written (the helper was unmerged on #110) and is `:233` now that it has landed.

19 tests in this file, 190 across the fleet suite, `shellcheck` clean.

## Deferred

- **#146** (`needs-triage`) — `jstr`/`jarr` *replace* control bytes rather than escaping them, so a conflicting path holding a TAB is reported as one that does not exist, and the runbook feeds those paths straight into `git diff`. Convention decision spanning three scripts, and #119 wants the helper in five more.
- **#147** (`needs-triage`) — the stash count reads 0 for an unknown count. Sharper than it looks: `git stash list` **exits 0** when it cannot read the reflog, so the `2>/dev/null` is not the mechanism — the same trap #106 fell into about `git log`.
- **#148** (`ready-for-agent`) — above `ARG_MAX` xargs splits the pathspec list and `atRisk` repeats each commit once per batch. Measured with `xargs -s 300`; safe direction, one-line dedupe.

Not deferred, recorded here instead: the `mt_rc` half of the merge-tree guard is shadowed by the emptiness check for every input I could construct, and the test comment now says so rather than implying coverage it does not have.

The single-row edit to the spec's contract table stands, and now lists all six payload fields and ends its exit-2 causes with a general clause instead of reading closed — **#114 still owns the full eleven-row audit** and this does not settle the column question for it.

No CI in this repo (#111), so the gate is the suite above plus the measurements in the commit body.

PR #150, opened 2026-07-31
ledger.mjs check: query the tracker and report near-misses instead of a bare boolean

### LINKED ISSUE #145: ledger.mjs check is a token-subset match against one ledger, so it misses the duplicates that actually happen
Measured during a `/fleet:run-team` session that filed 33 issues across 8 PRs with ~40 specialist reports.

## What the guard does

`run-team/SKILL.md` requires `ledger.mjs check "<subject>"` before every `gh issue create`, and members complied — reviewers consistently reported "ledger-checked, not previously filed". The path resolution is already careful: `defaultLedgerPath()` resolves via `git rev-parse --git-common-dir` precisely so a member running from a worktree reads the run's ledger and not a cwd-local file that does not exist, with a warning when it cannot. That fail-open is closed.

## What it actually matches

Token-subset against the ledger's `filed` list. Measured against a live ledger holding 180 entries:

```
$ ledger.mjs check "fleet-plugin-design Non-zero column written from intent — audit 11 rows (review-pr-108)"
{"found":true,"match":"#114 fleet-plugin-design Non-zero column written from intent — audit 11 rows (review-pr-108)"}

$ ledger.mjs check "Non-zero column audit 11 rows"            # strict subset
{"found":true,"match":"#114 ..."}

$ ledger.mjs check "candidates.mjs row states the opposite of its code"   # same defect, different words
{"found":false,"match":null}
```

The third string describes the **same finding as #114** — the spec table row that contradicts `candidates.mjs:157`. It reads as safe to file.

**And the check never queries GitHub.** There is no `gh` call in the check path. An issue that exists on the tracker but was never recorded in this ledger's `filed` list is invisible.

So the guard reliably catches one thing: re-filing, in near-verbatim wording, something already recorded this run. That is the case least likely to occur, because the filer usually remembers.

## The case that does occur, five times in one run

Issue **#114** was independently rediscovered by five different specialists on four different PRs — #108 (code-reviewer), #116 (code-reviewer and silent-failure-hunter), #130, #131 — each describing it in its own words:

- *"the Script surface table's Non-zero column was written from intent"*
- *"`candidates.mjs` row `:164` states the opposite of its code"*
- *"the `2026-07-23` spec's exit-0 claim"*
- *"`docs/specs/…:164` claims empty result is exit 0 with `[]`"*

**None was caught by `ledger check`.** Every one was caught by the controller recognising the finding while relaying reports and telling the reviewer "already #114, do not re-file". Absent that human-in-the-loop step, four duplicates ship.

The same pattern, smaller, for the `run-merge-bot.md` exit-2 documentation gap (three specialists on PR #131) and the `-e`/`.git`-linkage class (four specialists on PR #115).

## Why the subset direction is also backwards for the common case

`check` returns found when the **checked** subject is a subset of a **filed** row. So a *narrower* new finding matches a broader filed one — which is right. But a *broader* or differently-worded finding does not match a narrower filed row, and the second discoverer usually words it differently rather than more narrowly. The asymmetry favours exactly the wrong direction.

## Options

1. **Query the tracker.** `gh issue list --search "<subject>" --state open` before filing. Catches pre-existing issues the ledger never recorded — the class the ledger structurally cannot see. Costs one network call per filing; filings are rare.
2. **Match on file paths and symbols rather than prose.** Most duplicates in this run named the same `file:line` or the same identifier (`candidates.mjs:164`, `Non-zero`, `run-merge-bot.md:121`). A path/symbol index over filed rows would have caught all five.
3. **Report near-misses rather than a boolean.** Return the top-N filed rows by token overlap with a score, so the filer decides. Cheapest, no network, and it puts a human-or-model judgement where the ambiguity actually is.
4. **Accept it and make the controller's relay the designated dedupe point** — which is what this run did in practice, undocumented. If this is the answer, `SKILL.md` should say so, because it currently implies the script is the guard.

Option 3 plus option 1 would have caught every duplicate observed here. Option 4 is the honest description of the status quo.

## Note

Filed from the controller seat, not a reviewer's deferral. `ledger.mjs` is touched by no open PR, so this is safe to pick up now.

feigi:
Closes #145

Builds options **1 + 3** from the ratified approach comment: a tracker query *and*
near-miss reporting. Options 2 (path/symbol index) and 4 (document the status quo)
are not taken.

## What changed

**1. Tracker query.** When the ledger misses, `check` runs
`gh issue list --search <terms> --state all --limit 5` and reports each hit with its
number, state, url and an overlap score.

- **`--state all`, not `--state open`.** A closed issue is exactly as invisible to
  this run's `filed` list as an open one, and re-filing it is still a duplicate. The
  state is reported so the caller can tell "closed as fixed → this may be a
  regression worth filing" from "open → do not file".
- **Three search terms, longest-first.** `gh` ANDs search terms, so every extra term
  can only subtract. Measured against the live tracker: `candidates opposite states`
  returns #114; adding a fourth, less distinctive term (`open`) drops it. Longest-first
  is a crude distinctiveness proxy — there is no corpus to weigh terms against — and
  the comment says so.
- `norm()` already strips punctuation, which is what stops a subject containing
  `is:open` or `file.mjs:164` from smuggling a qualifier into the search. A subject
  with no distinctive terms skips the query entirely: an empty `--search` matches
  every issue in the repo.

**2. Near-misses.** The top 3 `filed` rows by token overlap, each with a score,
printed under the verdict and returned as `near[]`. Scored on its own token set
(stopwords dropped, naive singular fold) so ranking sharpens without widening what
counts as ALREADY FILED. Overlap coefficient rather than Jaccard: filed rows carry
`#NNN` prefixes and `(review-pr-108)` tags the subject can never match, and Jaccard's
union flattens every row to a similar small number.

The floor is "shares one content word", not a score threshold — the measured #114
rewording shares exactly one, and a tidier threshold would drop the case this exists
for. The top-3 cap keeps the output short instead.

## Exit-code contract — changed deliberately, not silently

| Exit | Meaning |
|---|---|
| 0 | ledger clean **and** tracker clean |
| 1 | already in this run's `filed` list (unchanged) |
| 2 | usage error (unchanged) |
| **3** | **new** — ledger clean, but tracker issues match |

0 means safe to file, which is the claim this change exists to stop making; 1 means
ALREADY FILED in this run, which a fuzzy tracker hit does not establish; 2 is `die()`.
Near-misses stay **exit 0** — a ranked suggestion is not a finding of duplication.
No script caller of `check` exists (`board.mjs` uses `read`), so the blast radius is
doc-level; `run-team/SKILL.md` now states all of it, including that exit 0 is not
automatically "safe to file".

## The offline path

Any `gh` failure — no network, no auth, rate limit, `gh` absent (ENOENT), timeout,
or exit-0-with-unparseable-stdout — degrades to the ledger-only answer, prints
`WARNING — TRACKER NOT CHECKED (<reason>)`, and its verdict line reads
`not previously filed in this run's ledger — ledger-only answer, tracker unchecked`.
Never a bare "safe to file". Two tests cover it (`gh` exiting 1, and `gh` absent from
PATH), and swallowing the error into a clean tracker reds exactly those two.

**`defaultLedgerPath()` is untouched.** `git diff origin/main...HEAD` shows the only
occurrence of `--git-common-dir` in the diff is a `+` comment line referring to it;
the resolution and its warning are unmodified.

## Verification

Every claim below was run against this tree.

- `node --test skills/fleet/scripts/*.test.mjs` → **209 pass, 0 fail** (post-rebase onto
  `origin/main`; the glob form, since `node --test <dir>` is MODULE_NOT_FOUND on v26.5.0).
- New tests run against `origin/main`'s `ledger.mjs`: **10 of 15 fail**. The 5 that pass
  are the 4 characterisation tests pinning the pre-existing exact/subset/usage
  contract — that is their job — plus *"a ledger hit short-circuits: gh is never
  invoked"*, which passes **vacuously** on main because there is no query to skip. It
  is pinned by mutation instead, not by failing first.
- Mutation-checked, each reds exactly the test naming it: dropping the near-miss sort;
  raising the floor to 0.5; swallowing the `gh` error; removing the short-circuit;
  exiting 0 on a tracker hit; widening to four terms; dropping `--state all`; dropping
  the empty-terms guard.
- End-to-end against the real tracker, the exact subject from the issue body that
  `check` reported safe to file:

```
$ ledger.mjs check "candidates.mjs row states the opposite of its code"
ledger: TRACKER HIT — #114 (OPEN) fleet-plugin-design: the Script surface table's Non-zero column … — https://github.com/feigi/claude-config/issues/114
ledger: TRACKER HIT — #145 (OPEN) ledger.mjs check is a token-subset match against one ledger …
ledger: not in this run's filed list, but 2 tracker issue(s) match 'candidates opposite states' — review before filing
exit=3
```

  #114 scores 1.0, #145 0.17. An unrelated subject still exits 0 with
  `tracker.hits: []`.

## Noted, not fixed

`docs/specs/2026-07-23-fleet-plugin-design.md:172` describes `ledger.mjs` as
`set/filed/ruled/read` with "ledger unreadable" as its only non-zero — already wrong
before this PR (`set` is `row`, `check` is missing, exit 1 on ALREADY FILED
contradicts it), and now also missing exit 3. That row is inside the eleven-row table
**open issue #114** exists to audit, so fixing it here would collide with that ticket.

## Out of scope, per the issue

Ledger location and format; the `filed` / `ruled` / `row` subcommands; auto-filing or
auto-closing anything — this reports, the caller decides.


feigi:
## Review complete — six specialists dispatched, five delivered, one died

Applied in `cf47718`; suite **209 → 212, all passing**.

**Verdict: approve.** Every acceptance criterion on #145 is met, verified independently rather than read off the PR body. The live case the ticket exists for — `check "candidates.mjs row states the opposite of its code"` — surfaces **#114 at exit 3** where it previously read safe-to-file.

### Applied

One real fix: `gh` printing JSON that parses but is not an issue list produced `ok:true`, three `TRACKER HIT — #undefined` lines and **exit 3** — a confident duplicate ruling, blocking a legitimate filing, from an answer nobody can read. Now validated inside the existing `try`, so it degrades exactly like unparseable output.

Three behaviours had no regression gate, each confirmed by mutation before and after:

| Decision | Before | Now |
|---|---|---|
| near-miss `.slice(0, 3)` cap | `→ .slice(0, 10)` red **nothing** | reds its own test |
| overlap coefficient | `Math.min → Math.max` red **nothing** | reds its own test |
| unparseable `gh` output | swallowing it red **nothing** | reds its own test |

The cap test was vacuous for a specific reason: only 3 of its 5 fixture rows scored above the floor, so the assertion rode the filter and never reached the cap.

Four comment claims corrected — the `#NNN` prefix never reaches the scorer (`subjectOf()` strips it); there is no asymmetry `isMatch()` "gets backwards" (it already picks the smaller set); a fourth search term does not simply drop #114 (`code` keeps it, `open` drops it, and `gh` matches bodies as well as titles); and exit 0 is still returned when the tracker was never read. `SKILL.md` said the ledger means *"no finding is filed twice"* — the exact over-claim #145 was filed against — four lines above the paragraph this PR adds.

### Verified, not taken on trust

- **No script caller of `check` exists**, so exit 3 breaks nothing. Scope: all tracked files, six pattern families; `board.mjs:227` uses `read`. The only consumer is an agent following `SKILL.md:604`.
- **`defaultLedgerPath()` does not regress** — byte-identical base↔head; the sole `--git-common-dir` occurrence in the diff is a comment.
- **The honesty claims hold.** 10 of 15 new tests fail against `origin/main`; the 5 passing are the 4 characterisation tests plus the one the PR itself flagged as vacuous — and that one is genuinely pinned by mutation here. All 8 claimed mutations red the test naming them; a control mutation stays green, so the suite discriminates.
- **Degradation is honest across every path probed**: auth failure, `gh` absent, non-JSON stdout, JSON `null`, wrong-shape JSON, cwd not a repo, no distinctive terms.

### Deferred — filed, not dropped

| # | Finding | Label |
|---|---|---|
| #151 | `review-and-fix.md` files issues without running `check` — the path the measured duplicates came through | `ready-for-agent` |
| #152 | exit 0 cannot distinguish a clean tracker from an unread one | `needs-triage` |
| #153 | longest-first selection drops the most distinctive query term | `needs-triage` |
| #154 | tracker hits and near-misses truncate silently | `ready-for-agent` |
| #155 | queries the cwd repo's tracker, not the ledger's | `needs-triage` |
| #156 | two remaining unpinned behaviours, one duplicated token filter | `ready-for-agent` |

**#151 is the one to read first.** The guard is now strong on the controller path, while the reviewer filing step — which produced the measured 5× rediscovery of #114 — still hand-rolls its own `gh issue list --search` and contains zero references to the ledger.

**#152 is a genuine split**: one specialist argued for minting exit 4 on the `candidates.mjs:64-66` fail-closed precedent, another argued any non-zero blocks `check "$s" && gh issue create` and contradicts the tested `offline must not block filing`. Left as-is because the maintainer's ratifying comment already ruled this path — *"degrade to the ledger-only answer"* — and the PR implements exactly that.

`docs/specs/2026-07-23-fleet-plugin-design.md:172` confirmed as pre-existing #114 territory and correctly left alone.

### Coverage gap

The **test-analysis specialist died mid-run** (`Connection closed mid-response`) and never reported. I substituted for it directly: the full 8-mutation matrix, the 10/5 split against `origin/main`, the vacuous-test claim, and a discrimination control are all measured above. Its broader test-quality sweep is the dimension that stayed thin — though the two findings it was most likely to surface, the vacuous cap test and the unpinned `Math.min`, were each independently found twice by other specialists.

PR #157, opened 2026-07-31
ci: rebase gate and a check job, carried over from agent-brain

feigi:
Two workflows, both carried over from `agent-brain`, plus the checks this repo could already run locally but never ran anywhere else.

## `ci.yml`

- **`rebase-check`** (PR-only, verbatim from agent-brain) — fails a PR whose merge-base is not the tip of its base, or that carries merge commits above it. This is what keeps main's merges empty and strictly sequential going forward, rather than depending on whoever merges remembering to rebase.
- **`check`** — `node --check` over tracked `.mjs` (there is no build; scripts ship straight to the runtime, so this is it), `jq -e` over tracked JSON, `shellcheck -x -S warning`, and `node --test skills/fleet/scripts/*.test.mjs`.

Two deliberate choices, both explained in-file: `-S warning` because the only info-level shellcheck hits in the tree are false positives (SC2016 on inflight.sh's python3 heredoc, SC1003 on statusline-command.sh's OSC-8 escape), and an explicit test glob rather than the directory, because `node --test <dir>` also picks up vendored trees — the defect already fixed once in the runner.

`defaults.run.shell: bash` for `pipefail`: every step is `git ls-files | xargs <checker>`, and without it a broken left-hand side is a green run that checked nothing.

## `rebase-check-refresh.yml`

`rebase-check` only runs on `pull_request`, so a PR that passed keeps a stale green check after main advances. This re-runs the real job on open PRs with `behind_by > 0` when main moves. Only PRs actually behind are touched — the rerun endpoint re-runs dependents too, so firing unconditionally would flip an already-green required check back to in-progress.

The job name `rebase-check` is load-bearing: this workflow finds it by `select(.name == "rebase-check")`.

## Verification

- `actionlint` clean on both files
- 212/212 tests pass with `gh` removed from PATH (the suite stubs `gh` and sets its own git identity — no auth, no network)
- `node --check`, `jq -e`, `shellcheck -x -S warning` all exit 0 on the current tree

## Follow-up

`rebase-check` should be added as a required status check on main for it to actually block anything.


feigi:
## Review complete — 3 specialists, applied + deferred

Fan-out: `code-reviewer`, `silent-failure-hunter`, `comment-analyzer`. All three reported.

### Applied here (`eb9195f`, `87fcfd6`)

**Behaviour — every one of these was a green that verified nothing:**
- `Tests` ran a bare glob; unmatched it exits 0 with `tests 0`, silently dropping all 212 assertions. `shopt -s failglob`. This repo already documents the hazard in `review-and-fix.md` and then hand-wrote the unguarded form into CI.
- `jq -e .` exits 1 on a file holding `null` or `false` — both valid JSON — under a step named "Validate JSON". → `jq empty`.
- Added a python syntax check; 7 tracked `.py` files run as modules and nothing else would see a syntax error in them.
- Refresh workflow: hardcoded `--base main` → `$GITHUB_REF_NAME`; numeric guard on `behind_by`; an in-progress/queued 403 is now a per-PR skip instead of a fatal "token lacks actions:write"; `NO_RUN`/`NO_JOB`/`REJECTED` now warn instead of being laundered into green by one accepted rerun; `TOTAL` via `wc -l` instead of `grep -c . || true`.

**Comments — seven claims measured false against the tree:**
- "`node --test <dir>` also picks up vendored trees" is backwards; node's own discovery *excludes* `node_modules` (measured), and the cited defect was a `find` in the emitted runner — `claim-ticket.sh` says so itself.
- Renaming the `rebase-check` job was said to make the refresh workflow "go green while refreshing no PRs"; it exits 1 with a per-PR reason.
- "the branch protection still blocks the merge" — `main` is not protected (404).
- "settings.json and the caches are read by hooks that die on a parse error" — every cache is gitignored so CI cannot see one, and the only in-repo reader swallows parse errors in a `try/catch`.
- "Every check below is `git ls-files | xargs`" — two of five steps are not.
- The `first(...)` note justified itself against SIGPIPE-ing `gh`; `gh` is not in that pipeline.
- `inflight.sh` has no heredoc; line 80 is a single-quoted `python3 -c`.

Also fixed `review-and-fix.md`, which still told review specialists this repo has **no CI** — falsified by this very PR.

### Deferred

| Issue | Label | Why deferred |
|---|---|---|
| #158 | `needs-triage` | `node --check` skips 2 tracked `.js`; widening the glob turns the step **red** (`review-pr.js` is ESM in `.js`, no package.json). Fix needs a maintainer call. |
| #159 | `ready-for-agent` | Rejected rerun never falls back to older candidate runs — restructures the loop, wanted reviewable alone. |
| #160 | `ready-for-agent` | Jobs-API failure misreported as "no rebase-check job". |
| #161 | `ready-for-agent` | The four `xargs` checks look identical in the log whether they checked 23 files or 0. |
| #162 | `ready-for-agent` | Three accounting holes that exit green having evaluated nothing. |
| #163 | `needs-triage` | `node-version: 22` is a version the suite had never run against; which node this repo targets is a decision. |
| #164 | `needs-triage` | Make `rebase-check` actually required — needs admin scope + ruleset choices. |
| #165 | `ready-for-agent` | Four remaining comments describing behaviour the code does not have. |

### CI

Green on `87fcfd6`, run-bound: run `30630559731`, `runHeadSha == prHead`, `rebase-check` and `check` both `success`, `behind_by=0`.

### One correction worth recording

A specialist reported that a non-numeric `behind_by` would make `set -e` abort the step. It does not — `set -e` is suspended inside an `if` condition, so bash prints `integer expression expected`, the test reads FALSE, and execution continues. The real defect was different and smaller: the PR falls through to the rerun branch and gets re-run without ever being shown to be behind. Fixed on the corrected mechanism, not the reported one.

Issue #132, opened 2026-07-31
candidates.mjs: process.exit() truncates the payload at 64 KiB on a pipe — a silently short list, exit 0


feigi:
Deferred from PR #116 review.

## What to build

`candidates.mjs:157` ends with `console.log(JSON.stringify(rows))` immediately followed by `process.exit(...)`. Node stdout to a **pipe** is asynchronous, and `process.exit()` does not drain it. Both real callers capture stdout through a pipe.

Measured at head (499 rows of realistic shape, so `refuseIfCapped` does **not** fire — 499 ≠ the 500 default):

| sink | bytes | parses |
|---|---|---|
| file redirect | 124797 | yes |
| **pipe** | **65536** | **no — `Unterminated string in JSON at position 65536`** |

The script exits 0 believing it handed back a complete list. This is precisely the failure the file's own header exists to prevent — *"a truncated list is indistinguishable from a complete one"* — reintroduced on the way out the door. `refuseIfCapped` and this defect do not compose: the cap guard only fires on a row **count**, and this truncates on **bytes**.

Verified fix, one line, same 499-row fixture then delivers all 124797 bytes intact with exit codes preserved:

```js
process.exitCode = rows.length === 0 ? 1 : 0;   // was: process.exit(...)
```

## Why deferred

The defective line is untouched by PR #116 and the defect is pre-existing. Stated carefully, though: #116 newly routes the "labeled queue is all specs" case into the **unfiltered** query — the largest payload the script can emit. Before it, that case exited 1 with `[]`, two bytes, which cannot truncate. So the PR moves a case from never-truncates to most-likely-to-truncate without introducing the bug.

Not currently biting: this repo's ~106 surviving issues serialize to ~14 KB. Bites at roughly 480+ rows, or any raised `--limit`.

## Acceptance criteria

- [ ] A payload larger than the pipe buffer arrives complete through a pipe
- [ ] Exit codes unchanged: 0 with survivors, 1 empty, 2 broken
- [ ] A test asserts the full payload parses when read through a pipe, not a file

## Blocked by

- None — can start immediately

feigi:
Sharper measurement from the PR #167 review (`silent-failure-hunter` specialist), confirming this ticket and pinning the crossover.

Through `spawnSync` with piped stdio — the shape both a fleet caller and `candidates.test.mjs:72` use — with 400 rows:

```
run 0: status=0  stdout.length=65536  parses=NO (Unterminated string in JSON at position 65536)
[...]

PR #167, opened 2026-07-31
fix(fleet): candidates.mjs exited 0 having widened the net on a malformed flag

### LINKED ISSUE #61: candidates.mjs fails open on malformed input — a broken query exits 1, the code reserved for "queue empty"

feigi:
Closes #61

Exit 2 is "the query broke", exit 1 is "there is no work". Every caller reads
only the status, so a path that swaps them is silent by construction. Three did.

## Root cause

All three trace to one line. `arg()` returned `process.argv[i + 1]`, which is
`undefined` for a flag given no value, and every caller read that falsy result
as "the flag was absent":

- **`--limit` trailing** — fell back to 500, past the positive-integer guard
  that exists to catch exactly this.
- **`--require-label` trailing** (the maintainer comment's third defect) — the
  worst of them. `query()` reads a falsy label as "no label", so the
  **unfiltered** query ran and the script exited **0**. Not a wrong default: it
  is the exact widening onto `ready-for-human` that this file's header forbids
  and that `run-team/SKILL.md:61` withholds `--allow-fallback` to prevent,
  reached without the flag being passed.
- **`gh` output never validated** — the reduction runs server-side inside `gh`,
  so nothing local guarantees it applied. The wrong shape flowed on until the
  first use of it threw, and an uncaught throw exits **1**.

Fixed in `arg()` itself rather than at each call site — the one place all the
flag routes converge, and it covers any future flag for free.

## Beyond the stated criteria

Two deliberate extensions, both the same defect by a different spelling:

- **`--require-label --allow-fallback`** (value is the next flag) reaches the
  same widening at exit 0. Guarded with `startsWith("--")`; neither option can
  take a value that starts with `--` — a count and a GitHub label. Both real
  callers pass values and are unaffected.
- **Empty `gh` output no longer returns `[]`.** (covered by `54ce6f7`) The reduction is `[…]`-wrapped
  and emits an array for every input including none, so nothing at all means it
  never ran — the same fact as the shape check, not an empty queue.

Rows are named by index in the error, never dumped: an unreduced payload
carries every issue body, ~97% of what this file exists to avoid fetching.

## Verification

TDD throughout. The 5 tests for the stated criteria were committed red first
(`0c30849`), then the fix (`b476678`). A 6th (`54ce6f7`) covers the empty-output
extension below — it was the one refusal nothing pinned, since it never reaches
the parse the shape checks guard; verified red by restoring `return []`, which
fails it at exit 1.

Full suite `./agent-test skills`: **218 tests, 218 pass, 0 fail**, re-run after
rebase on `origin/main`.

Every new test asserts the **exit code**. The `--limit` and `--require-label`
tests additionally assert that **no query ran at all** — status 2 alone is
satisfied by a run that widened and then died, and for `--require-label` the
query that runs *is* the harm.

`JQ_OVERRIDE` is the stub's one new seam: the real expression always produces
the reduced shape whatever the fixture holds, so no fixture can reach a
reduction that did not apply.

## Not fixed here

`pr-overlap.mjs`, `diff-stats.mjs` and `ci-state.mjs` carry their own `arg()`
copies with the same defect. Separate files, not callers of this one, and
outside #61 — worth a follow-up ticket.

Per the controller, `candidates.test.mjs`'s header claim about jq fidelity is
left alone for #63.


feigi:
Six-specialist review applied at `2055fa6`. Suite **221/221** (was 218/218); three tests added, all red-verified against the previous head.

## Applied

**A fourth and fifth spelling of this PR own defect still widened the net.** Found independently by three specialists and reproduced against `54ce6f7`:

```
--require-label ""                 exit=0, search carried NO label: term  -> unfiltered
--require-label=ready-for-agent    exit=0, search carried NO label: term  -> unfiltered
```

`""` is the likeliest of all four spellings — `--require-label "$LABEL"` with the variable unset leaves an empty argv slot, not a missing one — and its stderr reads `N candidate(s)` with no label suffix, indistinguishable from a clean run. The `=` form is invisible to `indexOf` entirely. Both now refuse at the same choke point this PR already chose. `--limit 0` still reaches its own positive-integer guard.

**Two properties this PR asserts and nothing tested:**
- **Exit 1.** Every refusal pinned 2, nothing pinned 1. Flipping `rows.length === 0 ? 1 : 0` to `? 2 : 0` — the swap the rewritten header forbids — passed the whole suite.
- **The row-shape refusal names the row, never prints it.** A `die()` interpolating the row instead of its index kept the suite fully green while emitting every issue body, the ~97% payload this file exists to avoid fetching.

**Three comment claims corrected**, all added by this PR:
- `--jq` does not run "server-side" — it runs in the local `gh` process (`gh issue list --jq _error("X")_` raises locally). Only the two sites this diff added are touched; the five inherited ones remain #63.
- "No option here takes a value that could start with `--`" is false — GitHub permits such a label and `denoland/deno` has `--watch` with 54 open issues. The guard stays, restated as the trade it is rather than an impossibility.
- The row-shape `die()` claimed "the --jq reduction did not apply", but also fires when the reduction ran perfectly and an upstream field type was wrong. It now states what is wrong, not a cause it cannot know.

## Verified, no change needed

- The `queriesRun` assertions are **not** vacuous — neutering the helper lets a validate-too-late fix through undetected. Worth noting the base defect exits **0**, so status alone would also have caught it; the comments are phrased as conditionals and are literally true, so no edit.
- The empty-output refusal is correct, not an over-correction: real `gh` 2.96.0 with a no-match search returns `[]` (2 bytes, exit 0), so empty output cannot mean "no issues".
- **#63 is unaffected** — its jq-fidelity paragraph is byte-identical to `main` (SHA-1 over the five lines matches). The one new line that would have left that file self-contradicting after #63 lands is fixed here.

## Deferred

- #175 — `label:${label}` interpolated unquoted; `--require-label "good first issue"` exits 1 against a non-empty queue (0 vs 5 results measured). Pre-existing.
- #176 — `die()` duplicates gh stderr unbounded; 1.06 MB measured on one failed query. Pre-existing.
- #132 — commented with a sharper measurement: the 64 KiB pipe truncation crossover is ~341 rows for realistic rows, i.e. *inside* the 500 default this file permits.
- #169 — commented: `ci-state.mjs:52-54` is the one site of the four not saved by its own guard (`--base` with no value silently compares against `main`), and any shared fix should cover all four spellings, not two.

Branch had been rebased onto `d856185` by the time I pushed; both reviewed files are byte-identical across that rebase, so every finding above stands against the current head.

Issue #129, opened 2026-07-31
fleet scripts: four cross-file line citations in comments point at the wrong line


feigi:
Deferred from PR #110 review. All pre-existing — none introduced by that PR, which corrected its own new ones.

Comments in this repo cite sibling files by `file:line`, and four have drifted. Verified at `d663193`:

| citation | claims | actually |
|---|---|---|
| `inflight.sh:188` | `release-ticket.sh:70` does `substr($0,10)` | **not in that file at all** — `git grep -n 'substr($0,10)' release-ticket.sh` returns nothing |
| `inflight.test.mjs:330` | same `release-ticket.sh:70` | same |
| `inflight.sh:151` | `release-ticket.sh:168` for the `ls-remote` | `:173` |
| `inflight.test.mjs:129` | `release-ticket.test.mjs:30` for the git-config shutout | `:40` |

The first is the one worth caring about: it does not merely point at the wrong line, it points at a construct that is no longer in that file, so a reader following it learns nothing and may conclude the rationale is stale.

Also seen in the same sweep, already fixed in #110: `run-team/SKILL.md:65` quoted for "any hit = taken" when the string is on `:66`, and issue #90 cited `:54` for it, which was wrong at head.

## Suggested fix

Re-point the four. Consider dropping the line number and citing the file plus the construct (`release-ticket.sh`'s `substr($0,10)`) — this class of rot is structural, since nothing recomputes a line number when a sibling file grows, and the repo now has five instances of it in two files.

feigi:
A fifth instance, found in the PR #168 review. Same class, same file-plus-construct remedy this issue already recommends.

| citation | claims | actually |
|---|---|---|
| `skills/fleet/scripts/candidates.mjs:105` | `run-team:180` for "heavy is explicitly never a bail reason" | **`run-team:185`** — `:180` is the `worktree-audit.sh` sentence |

The comment reads:

```js
// This is the ONLY line of defence — nothing downstream catches a spec. A
// leaked one is decided and needs no human hands, so it passes both of phase
// 2's bail tests, and heavy is explicitly never a bail reason (run-team:180).
```

Verified off by 5 **on both sides of PR #168** — the cited claim sits at `run-team:185` in that PR's base and at `:185` in its head, so the diff did not move the target and did not cause or worsen this. It was already stale on `main`; it was exact when written and a later +5 shift on main broke it.

[...]

PR #168, opened 2026-07-31
docs(fleet): phase 2 stated the bail rule and the heavy-row rationale twice

### LINKED ISSUE #62: run-team phase 2 states the bail rule and the heavy-row rationale twice each
## What to build

Phase 2 repeats two things the member already has:

- Its opening sentence restates a bail instruction the member receives verbatim ~16 lines above, inside the brief block the controller must paste into the prompt. Hang the rule off that block instead of reordering it.
- The heavy-row entry point and its full rationale are written out in both `run-team` and `sizing-a-ticket`, for a rule that only ever fires inside `sizing-a-ticket` — which the member runs. `run-team` should keep its own policy ("heavy is never a bail reason", "brief will not support a plan → that is the undecided case, bail and demote") and delegate the rest.

## Acceptance criteria

- [ ] The bail rule appears once in phase 2, attached to the existing brief block
- [ ] The heavy-row rationale lives only in `sizing-a-ticket`
- [ ] `run-team` still states its own policy: heavy never bails; an unsupportable brief is the undecided case
- [ ] No rule in `docs/fleet-rule-manifest.txt` stops resolving

## Blocked by

feigi:
Closes #62

Two duplications in `run-team` phase 2, both prose the member already receives by another route.

## The bail rule

It opened a paragraph seventeen lines below the `gh issue view` block the controller pastes verbatim into the member's prompt. The member was told to read the issue in the block, then told again — out of band — to read it before touching code and bail if it is undecided. Only the block reaches the member as written; the paragraph reaches it only if the controller paraphrases.

Hung the rule off that block instead, in the block's second person, so read-then-bail arrives as one instruction.

## The heavy-row rationale

The entry point and its `<HARD-GATE>` reasoning were written out in full in **both** `run-team` and `sizing-a-ticket`, for a rule that only ever fires inside `sizing-a-ticket` — which the member runs, and which states it unconditionally at `sizing-a-ticket:18-22`. Deleted the copy and delegated.

`run-team` keeps its own two policies, which are dispatch decisions rather than sizing ones:

- heavy is never a bail reason
- a brief that will not support a plan is the undecided case, not a heavy row

`demote` drops from the member-facing text because demotion is the controller's action — phase 3's implementer-bails table already owns it.

## Verification

- **`docs/fleet-rule-manifest.txt`** has no checker, so every pattern was resolved by hand against `skills/ workflows/` before and after and the two results diffed: **identical**. Resolved against the whole tree, not `run-team/SKILL.md` alone, since patterns legitimately live in sibling files.
- One manifest entry, `Say the specialist count`, was **already** orphaned on main by 8b46fdb and is tracked in #48. Orphan count is 1 on both sides — unchanged by this PR, not introduced by it.
- `node --test skills/fleet/scripts/*.test.mjs` — **212 pass, 0 fail**, same as baseline, before and after the rebase. No test covers prose; this only proves nothing else moved.
- Change is confined to phase 2 (`:167-191`). Nothing past `:197` is touched.

Net −3 lines (8 insertions, 11 deletions), as a de-duplication should be.


feigi:
## Review complete — approved, two prose fixes applied in `db909f0`

Fan-out: **code-reviewer + comment-analyzer** (docs-only profile per `review-and-fix.md`'s heuristic). Both reported; no dimension unrun. Both recommended merge independently.

### Acceptance criteria — all four met

| # | Criterion | Verdict |
|---|---|---|
| 1 | Bail rule once in phase 2, attached to the brief block | met (`:170-172`) |
| 2 | Heavy-row rationale only in `sizing-a-ticket` | met — `writing-plans`, `brainstorming`, `HARD-GATE` all absent from `run-team` at head, present at base; target `sizing-a-ticket:18-22` carries entry point + brainstorming skip + `<HARD-GATE>` reason |
| 3 | `run-team` keeps its own two policies | met (`:185`, `:186-187`) |
| 4 | No manifest rule stops resolving | met — see below |

**Criterion 4 was re-resolved by hand, three times independently.** It has no checker. The reviewer used `git grep -F -c -e <pat> <rev> -- skills workflows`; the comment-analyzer used `grep -rF` over `git archive` extracts; both against the implementer's own artifacts. All three agree: **33 entries, 32 resolving, 1 orphan, before/after byte-identical.**

The single orphan `Say the specialist count` resolves 0 at head, at base, **and on `origin/main`** — dead since `8b46fdb`, tracked in #48. Not introduced here and not this PR's to fix.

### On the ticket's premise

The implementer flagged that #62's premise is partly wrong. **Confirmed, and raising it was correct.** Base `:165-170` is a *read-the-issue* block containing no bail rule; the bail rule appeared exactly **once** in the pre-image, at `:182-184`. What was duplicated is the read instruction, not the bail rule — so #62's body and title are both imprecise on that half. The "~16 lines" distance is right (17).

The prescribed fix is correct regardless, because the real defect is a **delivery-route** defect: the read instruction reached the member verbatim, the bail rule depending on it only via controller paraphrase. Criterion 1 is met on its literal wording. The heavy-row half of the title was accurate.

Dropping `demote` from the member-facing blockquote is likewise verified correct — phase 3 `:212-224` owns demote-by-cause, and its two rows map 1:1 onto the blockquote's two bail conditions. The surviving "bail and demote" at `:186-187` is controller-facing prose and is the policy #62 explicitly requires be kept.

### Applied — `db909f0`

Two defects in the sentence this PR added, both inside the block pasted into the member's prompt **verbatim**:

1. `Read it` → `Read the issue` — the pronoun's nearest antecedents were "its `Respec` block" / "the `## Agent Brief` comment", not the issue the deleted prose named.
2. The qualifier `with the repo in front of you` was fronted so it governs **both** disjuncts again. The deleted prose trailed it after both; the rewrite had attached it to the first only, letting a member claim it needs human hands without looking at the repo.

Wording only. Manifest re-verified after the fix: identical to base **and** to `32cfd91`.

### Deferred

- **#172** (`needs-triage`) — the member's route to `sizing-a-ticket` is paraphrase-dependent (`:184` sits outside the verbatim blocks), and nothing tells the member it is a fleet member, so `sizing-a-ticket:18`'s `**Fleet member on heavy:**` condition may not fire — routing it to `brainstorming` → `<HARD-GATE>` → stall. Includes the dangling `"the member starts there"` referent at `:188`. **Pre-existing; verified net-zero change** — the pre-image kept the heavy-row copy in the same non-pasted prose. `needs-triage` because the approach is genuinely open.
- **#129** (comment, not a new issue) — `candidates.mjs:105` cites `run-team:180` for a claim at `run-team:185`. Off by 5 **on both sides**; the target line did not move. Same defect class #129 already owns.

### Verification

- `node --test skills/fleet/scripts/*.test.mjs` → **tests 212 / pass 212 / fail 0**, before and after the fix. Not a `tests 0` phantom. **No test in this repo asserts on prose** — green proves only that nothing else broke, not that this diff is right. Every claim above is settled against the object store.
- Cross-reference sweep over all seven top-level tree entries: **nothing outside `docs/specs/` is falsified**, and those are dated design docs already stale against the tree.
- PR #166 **merged mid-review** (it read `OPEN` on first query). Its only `run-team/SKILL.md` hunk was `:350-361`; this PR's highest changed line is 191. Separation held — `git merge-tree --write-tree origin/main db909f0` is **clean, no conflicts**.
- CI: `check` **success**; `rebase-check` red solely from `behind=5`. Standing staleness — not rebased.

Not adding `ready-to-merge`; a finisher does that on green.

PR #171, opened 2026-07-31
fix(fleet): the stray guard demanded a hand-release of a directory that is gone

### LINKED ISSUE #81: release-ticket.sh: the `stray` guard blocks forever when the stray worktree directory is gone — and tells the operator to tidy a directory that is not there
Deferred from PR #72 review.

## Problem

`skills/fleet/scripts/release-ticket.sh:134-136` blocks when a worktree is this claims but is not on its branch:

```sh
if [ -z "$wt" ] && [ -n "$stray" ]; then
  block "worktree $stray is this claims but is not on $branch — release it by hand"
fi
```

`$stray` comes from the `worktree list --porcelain` **registration**, matched on the directory-name suffix. It never asks whether that directory still exists. So a worktree that went detached and then had its directory deleted by hand blocks permanently — **the exact sibling of the bug #37 closed on the dirty-check path, and it survives PR #72 untouched.**

## Measured

```
$ git -C .worktrees/9-x checkout --detach && rm -rf .worktrees/9-x
$ sh release-ticket.sh 9 x fix --apply
    BLOCKED: worktree .../.worktrees/9-x is this claims but is not on fix/9-x — release it by hand
exit 1
$ sh release-ticket.sh 9 x fix --apply      # identical, every run
    BLOCKED: worktree .../.worktrees/9-x is this claims but is not on fix/9-x — release it by hand
```

The registration keeps `inflight.sh` probe 3 reading the ticket as taken every run, which is the silent queue shrink `release-ticket.sh` exists to clear.

## The remedy text is also wrong in this state

"release it by hand" instructs the operator to tidy a directory that does not exist. The action that actually clears it is `git worktree prune`.

## Desired

The stray branch should establish whether the directory is there before demanding hand-release, the same way `:188` now does, and name `git worktree prune` as the remedy when it is gone. PR #72 added an ancestor-walk predicate at `:188` for exactly this question — reuse it rather than inventing a second one.

Verified on git 2.50.1 against the post-#72 script (`1e54bac`).

feigi:
`worktree list --porcelain` keeps listing a worktree whose directory was removed by hand — annotated `prunable` — so a claim whose worktree went detached and was then deleted still matched `stray`, and the guard asked only whether it was REGISTERED, never whether the directory was there. It named "release it by hand" for a directory that is not there: an instruction with no action behind it, so nothing cleared the entry and every later run blocked identically, with the label and the branch standing and the in-flight probe still reading the ticket as taken.

## What changed

The ancestor-walk predicate the dirty check already carried is now a function, `gone`, called from both sites — one question, answered once, rather than a second predicate free to drift from the first. When the stray's directory is established gone the blocker names `git worktree prune`; otherwise the hand-release wording is unchanged.

Still a blocker, never a release. The worktree is not on the claim's branch, so releasing would delete a different ref and then reach the `git worktree prune` every `--apply` ends with — unanchoring a detached HEAD's commits, which no ref points at, as a side effect of releasing something else. Naming the prune hands the operator the command that clears the entry and leaves the discard their decision.

Absence stays ESTABLISHED, not inferred from a failed `-e`: a stray inside a prefix we may not search keeps the hand-release wording, since pruning there would unregister a worktree still sitting on disk with uncommitted work in it.

`skills/fleet/skills/run-team/references/reaping.md` documented the old blanket remedy, so it gets the same distinction.

## Measured

Before, on the reported state (detached, then `rm -rf`), every run:

```
BLOCKED: worktree .../.worktrees/9-x is this claim's but is not on fix/9-x — release it by hand
exit 1
```

After, the loop terminates (git 2.50.1):

```
$ sh release-ticket.sh 9 x fix --apply
    BLOCKED: worktree .../.worktrees/9-x is this claim's and its directory is gone — git worktree prune to clear the registration
exit 1
$ git worktree prune
$ sh release-ticket.sh 9 x fix --apply
{"issue":9,...,"released":true,"applied":true,"blockers":[]}
exit 0
```

No stale entry and no claim branch left afterwards.

## Tests

`node --test skills/fleet/scripts/*.test.mjs` — CI's own command: **214 passing, 0 failing, 0 skipped**, up from 212 on `main`.

Two new cases, both written first and watched fail:

- *a stray worktree whose directory is gone names the prune that clears it* — red against the current script on the exact assertion (`actual: "... — release it by hand"`, `expected: /git worktree prune/`).
- *a stray worktree the script may not stat keeps the hand-release remedy* — the reason the walk is reused instead of a bare `-e`. Verified by mutation: swapping `gone "$stray"` for `[ ! -e "$stray" ]` turns this case red and leaves every other test green.

`shellcheck -x -S warning` and `node --check` clean over all tracked files.

Closes #81


feigi:
Six-specialist review complete (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, type-design-analyzer, code-simplifier), each on its own `git archive` tree so no mutation probe was visible to a reader.

**Verdict: the PR is correct and delivers #81.** No specialist raised a blocking finding. The three claims I was asked to check rather than assume all held up under execution:

- **Reuse, one predicate.** `gone` at `:156`, exactly two callers. The hoist is behaviour-preserving on the delete gate: under the preceding `[ ! -e "$wt" ]`, `! gone "$wt"` reduces to base's `[ ! -x "$look" ]`. Verified over a 15-case matrix under both `sh` and `dash`, identical on every case.
- **Blocking rather than releasing is right.** `git worktree prune` is the last statement of the `--apply` branch, and on this path `$wt` is empty so only the *claim's* branch would be deleted. Shown physically: while the entry stays registered `git gc --prune=now` keeps a detached-HEAD commit; after the prune it destroys it.
- **The mutation claim is exact.** `gone "$stray"` → `[ ! -e "$stray" ]` reddens the may-not-stat case and nothing else; reverting the guard to the pre-fix unconditional block reddens the other case and nothing else. Two tests, two different mutations — neither is decoration. Reproduced independently three times.
- **The corrected assertion is right.** `artefacts().worktree` matches `branch refs/heads/<branch>`, which a detached worktree never prints, so it was blind to the stale entry; the path-based replacement observes it, proven by inserting a `prune` before it and watching it fail.

## Applied here

- **A coverage gap on this PR's own new branch.** Both added cases remove the worktree directory and leave `.worktrees` standing, so the stray guard never reached the ancestor *walk*. A stray-only precondition (`[ -d "${stray%/*}" ] &&`) therefore ran **37/37 green** while reinstating exactly the #81 refusal. Added the `rm -rf .worktrees` case: that mutant now dies to it alone, and it stays green under the bare-`-e` mutation, so it discriminates. **215 tests, all passing.**
- Corrected three added claims that outran the code: the header promised "the nearest ancestor that does exist" when the walk never reaches `/`; the return contract (0 = established absent, 1 = present *and* cannot-stat) was stated only above one call site; `prune` ends an apply that *completes*, since `halt` exits ahead of it. Also documented that a bare `gone` call exits at rc 1 under `set -e` — this script's own blocked-run code, with no receipt — a shape only the hoist made possible.
- `reaping.md` gained the new distinction as a fresh paragraph but left the one above it asserting the superseded remedy unconditionally and in bold. Amended so it cannot be read standalone. The author's negative claim that this was the only doc reader is **correct** — confirmed by my own sweep of six phrasings over tracked files at the head SHA, and independently by comment-analyzer's eight greps.
- CI's zero-test guard comment counted 212; the glob it guards now matches 215.

## Deferred

- **#178** (`ready-for-agent`) — `gone`'s walk never tests `/`, so an absent path with no surviving ancestor below root reads "cannot tell". Pre-existing and byte-identical at base; the one-token fix changes the irreversible-delete gate, which wants its own regression test rather than a drive-by edit on a PR whose point was that the hoist changed nothing.
- **#179** (`needs-triage`) — the stray guard gives an authoritative hand-release remedy for a worktree it never managed to stat. Behaviour is deliberate and safe; only the message asserts an unmeasured state. Three defensible approaches including doing nothing, so it wants a maintainer's call.
- **#184** (`ready-for-agent`) — the three `0o000` permission fixtures leak on failure and go vacuous under euid 0. Pre-existing pattern.

Not actioned: code-simplifier's suggestion to cut the four-line "one predicate" rationale — comment-analyzer wanted that same area expanded, and it records the ticket's own requirement, so I left the author's text.

Every empirical claim in the added comments was checked against git 2.50.1, the version they cite. All held: both absent and unsearchable entries are marked `prunable`, `worktree remove` accepts a prunable-because-absent entry at rc 0 while refusing a live worktree whose `.git` was deleted, and the blocked → `prune` → release loop terminates with `{"released":true,"blockers":[]}`.

Issue #96, opened 2026-07-31
inflight.sh: a failed probe dies immediately, discarding a sufficient hit another probe already found


feigi:
## Problem

`inflight.sh` runs three probes whose results form a **monotone disjunction**: any single hit is *sufficient* to answer "taken". But the fail-closed guards added by PR #76 `die` immediately, so a lookup failure in a later probe discards a positive answer already in hand.

## Measured

Ticket 56 with a local worktree **and** a local branch present, `origin` unreachable:

```
rc=2          <- the true answer is 1; a worktree demonstrably exists
stdout: ''    <- hits:["local"] and the worktree path both discarded
```

Control, same repo with `origin` reachable: `rc=1`, `hits:["local"]`.

Same shape from the other direction: an open linked PR **plus** an unreachable origin prints `HIT: pr` to stderr and then exits **2**.

The operator is sent to debug the remote when the answer was sitting on local disk, and the ticket is reported unanswerable when it is provably taken.

## Why this is not simply "the guards are wrong"

The escalation is right; the **placement** is wrong. Failing closed does not require abandoning a sufficient positive. `die` is correct when nothing has been established, and premature when something has.

Note the safety direction is unaffected — 2 and 1 are both non-free, so no ticket gets double-claimed either way. What is lost is diagnosis quality and the payload.

## Suggested approach (~15 lines)

Accumulate rather than abort: replace each `die` with an append to an `unknown` list, run all three probes, then decide

- any hit → 1
- else any unknown → 2
- else → 0

and emit the payload on **all** paths with an added `"unknown":[…]` field, so exit 2 stops meaning "no stdout at all". That last part is a payload contract change and wants its own look — today every consumer can assume stdout is either a full verdict or absent.

Deferred from PR #76 review: a design change to the probe loop, not a fix to the mechanism that PR addresses.

feigi:
> *This was generated by AI during triage.*

**Verified against `main` (`796f330`), git 2.50.1.** Open linked PR plus an unreachable origin:

```
    HIT: pr
inflight: git ls-remote failed, so whether #42 has a remote branch is unknown
exit=2      stdout: ''        <- hits:["pr"] discarded
```

**The payload-contract worry in the body is smaller than it looks.** Every reference to this script in the repo was checked: the test suite, and prose in `next-ticket` and `run-team`. Nothing machine-parses its stdout, so adding a field and emitting on the exit-2 path breaks no consumer.

## Agent Brief

**Category:** bug
[...]

Issue #95, opened 2026-07-31
inflight.sh probe 3: git's degraded reads exit 0 with output missing, so a real branch still reads as "free"


feigi:
## Problem

PR #76 guards probe 3's two local lookups with `if ! x=$(...); then die; fi`. Those guards catch a lookup that **could not run** (rc 128 — an unreadable `packed-refs`, git missing, a fork failure). They do **not** catch git's *degraded* reads, which exit **0** with output silently missing.

A real branch then reads as **free** — the same wrong answer #76 exists to eliminate, one probe down.

## Measured at `543faa3`

Repo with branch `fix/55-real`, origin reachable:

```
BASELINE:                      local branches: fix/55-real   taken=true   EXIT=1

$ chmod 000 .git/refs/heads
$ git for-each-ref --format='%(refname:short)' refs/heads ; echo $?
0                                                    <- rc 0, output EMPTY
$ inflight.sh 55
    no local branch or worktree for #55   taken=false  EXIT=0             <- wrong free
```

A corrupt loose ref behaves the same way: `git for-each-ref` prints `warning: ignoring broken ref refs/heads/fix/55-real` on stderr, **omits the ref**, and exits 0.

`git worktree list --porcelain` is the same story across four corruption variants (gitdir emptied, removed, garbage, baseline): rc 0 every time, entries silently skipped.

`543faa3` corrected the comment to name this ceiling honestly rather than claim coverage it lacks, but the underlying gap is real and unfixed.

## Why the approach is open

The exit-status contract cannot express this — git reports success. Options:

1. **Cross-check the counts.** `git for-each-ref --count` vs the parsed line count, `git worktree list --porcelain` entry count vs `.git/worktrees/` directory count. Detects omission without trusting rc.
2. **Capture stderr and treat `warning: ignoring broken ref` as unanswerable.** Narrow, string-matched, brittle across git versions.
3. **Pre-flight the readability of `.git/refs` and `.git/worktrees`.** Cheap, but a proxy for the thing rather than the thing.
4. **Accept the ceiling.** The trigger is a permission or corruption fault on the local repo, which usually breaks much else first — note it and move on.

Option 1 is the only one that detects the actual condition, and it roughly doubles each probe's cost. Genuinely a judgement call.

Deferred from PR #76 review. The comment at `inflight.sh:168-174` already documents the ceiling, so this issue is the record of the gap, not of a lie in the tree.

feigi:
> *This was generated by AI during triage.*

[...]

Issue #92, opened 2026-07-30
inflight.sh: an unattended probe can hang on a credential or host-key prompt — GIT_TERMINAL_PROMPT is never set


feigi:
## Problem

`skills/fleet/scripts/inflight.sh` never sets `GIT_TERMINAL_PROMPT=0`. Its `git ls-remote --heads origin` at `:156` reaches the network, so on a host without cached credentials or a known host key it can **block on a prompt** rather than returning any of 0, 1 or 2.

For an unattended fleet a hang is arguably worse than the wrong answer #56 fixed: exit 2 stops one member with a diagnosable message, whereas a blocked probe holds a slot indefinitely and reports nothing at all.

The script's own comment at `:148-150` names "a suppressed host-key prompt" as one of the four triggers for the fail-closed path it just added — which is the *suppressed* case. The unsuppressed case has no handling.

`inflight.test.mjs` sets `GIT_TERMINAL_PROMPT: "0"` in its fixture env (added by PR #76) with a comment explaining exactly this hazard, so the test suite is protected. Production callers are not.

## Why this is not obviously a one-line fix

Setting it unconditionally at the top of the script would change behaviour for a human running `inflight.sh` by hand: today a missing credential prompts and can be answered, after the change it fails. That is right for the fleet and arguably wrong for interactive use.

Options, none clearly best:

1. `export GIT_TERMINAL_PROMPT=0` unconditionally — simplest, breaks interactive credential entry.
2. Set it only when stdin is not a tty (`[ -t 0 ] || export GIT_TERMINAL_PROMPT=0`) — preserves both cases, one extra line, but "is stdin a tty" is a proxy for "is this unattended" rather than the thing itself.
3. Leave the script alone and set it in the fleet's callers (`run-team`, `next-ticket`) — no behaviour change for humans, but every future caller has to remember.
4. Do nothing and document it.

Option 2 looks right to me, but it is a judgement about who the script's users are, so triage rather than assume.

Deliberately left out of PR #76 by its implementer and independently ruled the same way in review: real and distinct from the bug that PR fixes.

Related: #56, PR #76.

feigi:
Refinement from PR #76 review, measured rather than assumed: **`GIT_TERMINAL_PROMPT=0` does not fix the hang.** With `origin` over ssh and a stalled transport, `git ls-remote` blocks identically with and without it — killed at 8s, `rc=142`. It suppresses *credential prompts*, not a hung connection.

[...]

PR #174, opened 2026-07-31
fix(fleet): inflight.sh reported a failed filter as a clean ticket

### LINKED ISSUE #89: inflight.sh: three filtering pipelines still discard mid-pipeline exit status, the same class PR #76 fixed one level up
## Problem

PR #76 fixed one instance of "a pipeline's status is its last command's, so a failed lookup reads as no-match". Three filtering pipelines in `inflight.sh` still have the residual form of it, where a **mid-pipeline** failure is invisible and the trailing `|| true` discards the last command's status too:

- `:159-160` — probe 2's remote filter, four forks deep, downstream of the network call
- `:178-179` — probe 3's branch filter
- `:187-190` — probe 3's worktree filter

Measured at `9de019bc` with `fix/42-thing` present on the remote (ground truth: taken, exit 1):

```
sed shimmed to exit 1   (mid-pipeline)  ->  "no remote branch for #42", taken=false, EXIT 0
paste shimmed to exit 1 (last command)  ->  "no remote branch for #42", taken=false, EXIT 0
```

A shimmed `sed` is not the realistic trigger — **fork failure under process-table pressure** is. The PR's own new comment at `:121-127` already makes exactly that argument to justify the `|| die` on `:129`:

> a fork failure under process-table pressure is exactly what a parallel fleet approaches by construction… a crash rendered as a decision.

That reasoning applies verbatim to `:159`, `:178` and `:187`, where it is not applied.

Same class, one line down: `$(basename "$p")` at `:188` discards its status, so a failure yields an empty first field and a silent non-match.

## Why the approach is open

Three candidates, none obviously right:

1. **`set -o pipefail`** — one line, but **not portable**. The script is `#!/bin/sh`; `pipefail` is bash/ksh/zsh and is absent from `dash`, which is `/bin/sh` on Debian-family systems. Rejected unless the shebang changes.
2. **Guard each stage** — correct but verbose, turning each 2-line pipeline into ~8 lines across three sites.
3. **Collapse `awk | sed | grep | paste` into a single `awk`** — removes three forks *and* makes the status directly observable, so it is both the smaller runtime and the smaller failure surface. But it rewrites the matching logic in the code path the fleet depends on for claim arbitration, and it breaks the "identical shape to `release-ticket.sh:168`" property PR #76 deliberately leans on.

Option 3 looks best to me but is a real change with real regression risk, so it wants its own tests rather than riding along on a fix for a different mechanism.

Deferred from PR #76 review. Pre-existing. Not applied there because the fix is not a finishing touch and the likelihood class differs: probe 2's `ls-remote` fails on mundane, reproducible conditions (no origin, unavailable key, stale URL), whereas this residual needs a fork failure.

feigi:
Closes #89

Each of `inflight.sh`'s three filters was a pipeline ending in `paste -sd, - || true`. A pipeline reports only its last command's status and the `|| true` discarded even that, so a stage that could not run produced an empty result — exactly what a genuinely free ticket produces. Three outcomes were two.

Measured on this script before the change, each against the same fixture with the stage intact (which answers taken, exit 1):

| filter | stage broken | result |
|---|---|---|
| probe 2, remote branches | `sed` | `no remote branch for #42`, taken=false, **exit 0** |
| probe 3, local branches | `grep` | `no local branch or worktree`, taken=false, **exit 0** |
| probe 3, worktrees | `basename` | `no local branch or worktree`, taken=false, **exit 0** |

Three wrong "free"s — the one answer this script must never invent, and the one that puts two agents on the same ticket.

## The change

Each `awk | sed | grep | paste` chain collapses into a single `awk` whose status is read with `|| die`. That removes the forks that are the actual failure mode — a fork failure under process-table pressure is what a parallel fleet approaches by construction, the argument the script already makes for the `|| die` on its PR-counting call — and it dissolves the worktree filter's per-line `basename` subshell, which had the same defect one level down.

`grep` is what made this delicate: it exits **1** for "looked, found nothing", which is legitimate and must never become exit 2. A lone awk needs no such case separated out — these programs contain no `exit`, so they return 0 whether or not anything matched, and every non-zero status is a real failure. `set -o pipefail` was not an option: the script is `#!/bin/sh` and `pipefail` is absent from `dash`.

Exit contract (0 free, 1 taken, 2 unanswerable) and the JSON evidence shape are unchanged.

## Tests

Four new cases, 212 → 216. The three forced-failure cases were verified red first, the worktree one failing with the exit **0** above.

The shim selects a filter by a substring of its awk program, never by counting invocations. A count cannot address one filter here: the worktree filter *was* two awks in one pipeline, and a pipeline starts its stages concurrently, so both read the same counter before either wrote it — measured, both were "the second awk" and no invocation was ever the third, so a count-built case broke nothing and passed the script a clean bill of health.

The fourth case pins the substring guard the rewrite had to preserve: 41 is not claimed by a remote branch, a local branch or a worktree named for 341. The existing space-in-the-path case pins the raw `substr($0,10)` read.

Dropping the `-F'\t'` join fixed corrupt evidence on the way past: measured, a worktree named `fix-88-a<TAB>b` arrived as the bare string `b`, its path cut at the tab it had just been joined on. It now arrives whole, the tab neutralised by `jstr` like any other C0 byte. The verdict was right either way.

Four stale citations corrected against the tree: `release-ticket.sh:70` → `:78`, `release-ticket.sh:168` → `:173`, `release-ticket.test.mjs:642` → `:743`, and the claim that `\t` is eaten upstream in probe 3 (the code that ate it is gone).

## Out of scope

Left alone deliberately, all separate open tickets: **#95** (git's degraded reads, which exit 0 with output missing), **#96** (a failed probe discarding a sufficient hit another probe already found), **#92** (prompt suppression and transport timeouts).


feigi:
Review complete — six specialists (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, type-design-analyzer, code-simplifier), all reports retrieved and adjudicated. Verdict: **approve, with four fixes applied**. Head `e10fb91`, suite **217/217, 0 skipped**.

**Applied**

- `a7f441f` — four comment corrections. The PR's own citation pass fixed two of four: the `release-ticket.sh:70` twin at `inflight.test.mjs:359` and the 101-line drift at `:394` (`:628` → `:729`) were left. `awkFailAt` at `:62` is an identifier that exists nowhere and reads as the *rejected* counting design. `:533`'s "there was no single awk to break" is false for the remote filter, which had exactly one — `{print $2}` occurs twice on the merge base, so that lone awk was byte-identical to the worktree filter's second and no substring could address it alone.
- `9feb8cb` — a regression this branch introduced. One non-UTF-8 byte in any ref takes **every** ticket to exit 2 permanently: base matched in `grep` (byte-oriented), a lone awk in a UTF-8 locale aborts rc 2 on a record it only scans past. `LC_ALL=C` on all three, with a test that reddens without it.
- `e10fb91` — the segment guard pinned one anchor of two. Dropping the trailing `([-/]|$)` left the whole suite green while 41 read as taken off `410` on all three probes. 410 folded into the existing fixture.

**Deferred**

- **#185** (new, `ready-for-agent`) — a newline in a worktree path yields a wrong "free": a checked-out worktree for #33 reads `taken=false` exit 0. Pre-existing and byte-identical on base, so not this branch's regression — but it is #89's own **unchecked** acceptance criterion, and this PR closes #89, so it would otherwise close untracked. `(#122)` was repointed at it.
- **#93** (already open) — `run-team/SKILL.md` states no exit-2 doctrine; this branch adds three `die` sites, making the state it cannot name more reachable. Not re-filed.

**Confirmed, not assumed** — every pinned claim measured: the awk programs cannot exit non-zero on a legitimate no-match; all nine outcome-states distinct with exit 2 naming its own probe and empty stdout; the shim keys are a clean bijection onto the three filters at `169/195/216`; the space and 341 criteria byte-identical to base; and the JSON evidence *shape* unchanged, with only the corrupt tab value repaired.

`rebase-check` red is behind-count staleness, not this diff — deliberately not rebased.

Issue #127, opened 2026-07-31
claim-ticket.test.mjs: five pass-count assertions are unanchored substrings, so they stop discriminating once a fixture grows


feigi:
Deferred from PR #109 review.

`assert.match(r.stdout, /pass 3/)` is an **unanchored substring** against node's reporter output. It matches `pass 3`, and it also matches `pass 30`, `pass 31`, … So the moment a fixture grows past the assertion's decade, the pin silently stops discriminating — the test keeps passing while no longer checking anything.

Five remaining instances, all in `claim-ticket.test.mjs`:

```
:101  assert.match(r.stdout, /pass 3/);
:112  assert.match(r.stdout, /pass 1/);
:121  assert.match(r.stdout, /pass 1/);
:136  assert.match(r.stdout, /pass 1/);
:147  assert.match(r.stdout, /pass 2/);
```

`/pass 1/` is the worst of them — it matches every count from 1 and 10–19 through 100+.

## Demonstrated

A reviewer forced it against the **pre-fix** script with the fixture grown into the 30s and a passing vendored test: `status === 0` **and** `/pass 3/` both held while the vendored test ran. Fully vacuous — the exact failure the file exists to refuse. The window is narrow today (4–29 fails loudly, 30–39 silent), which is what makes it a latent trap rather than a live bug: it waits for someone to add fixtures.

## Fix

Anchor to a whole line, multiline flag:

```js
assert.match(r.stdout, /^ℹ pass 3$/m);
```

PR #109 already did this for the assertion it added (`:191`, `/^ℹ pass 4$/m`) and left these five alone as out of scope. Mechanical change, one line each; the suite is 171/171 and any breakage is a real count mismatch worth looking at.

`ready-for-agent` — the defect and the fix are both settled, nothing to decide.

Surfaced by the test-coverage reviewer on PR #109.

PR #177, opened 2026-07-31
fix(fleet): the emitted runner refused a symlink to a directory of tests

### LINKED ISSUE #99: claim-ticket.sh: the emitted runner refuses a symlinked directory, because [ -d ] follows symlinks and find does not

feigi:
Closes #99.

## The defect

In the `agent-test` runner `claim-ticket.sh` emits, the directory shim's two
halves disagreed about one target. `[ -d "$arg" ]` follows symlinks, so a
symlink to a directory of tests entered the branch. The `find` under it does
not descend a symlinked *argument*, so it matched nothing, and the count guard
then refused a suite that was right there:

```
agent-test: no test files under tlink     EXIT=1
```

A false refusal — loud rather than silently green — but the inconsistency is
invisible until someone hits it. Inert in this repo today: `git ls-files -s |
awk '$1=="120000"'` returns zero tracked symlinks.

## The fix

`find "$arg/"`. One character; it settles the argument alone.

Deliberately **not** `find -L`. That agrees with `[ -d ]` too, but by also
following symlinks *inside* the tree — which walks into a symlinked
`node_modules` that the neighbouring `-not -path '*/node_modules/*'` filter
only recognises by path, and loops on a cycle.

## Test

Written first and watched fail against the current emitter, with exactly the
refusal above at exit 1. It builds the symlink in the emitted worktree rather
than through the `repo()` fixture — the runner only stats a path in its cwd,
so where the symlink came from is invisible to it, and the shared `SUITE`
fixture stays untouched. `pass 3` is `t/`'s two files plus the one under
`t/nested/`, so a symlink followed only one level deep reads as a red too; the
assertion is anchored (`/^(?:ℹ|#) pass 3$/m`) after the vendored-test count's
model.

Because `SUITE` is unchanged, the unanchored pass-count substrings that #127
reports are unaffected — none of their fixtures grew, so none of them was
silently defeated.

## Verification

`node --test skills/fleet/scripts/*.test.mjs` — CI's own command:

- origin/main `fe70bf5`: **tests 212, pass 212, fail 0** (measured on `d8561853`;
  `fe70bf5` is that plus two docs-only commits, merged, touching no `.mjs`)
- this branch: **tests 213, pass 213, fail 0, skipped 0**

`node --check`, `jq empty` and `shellcheck -x -S warning` over the tracked
globs all pass locally.

The third commit moved the `212` the failglob rationale in `ci.yml` cites to
`213`. **Review superseded that** (`002c6e7`): the count is now removed rather
than maintained, because it was already wrong on `main` — 212 stated against a
221-test suite — and PR #171 had independently edited the same line to 215.
Nothing in CI validates it. Its neighbouring "29 flat files" claim is untouched
and still holds; unlike the test count it does not move when a case joins an
existing file.

Review also corrected two false claims in the `-L` rationale ("loops on a
cycle" — neither find loops; GNU exits 1, BSD skips silently) and added a
fixture pinning the choice of the trailing slash over `-L`, which previously
passed the whole suite. See the review comment below.


feigi:
## Review complete — six specialists, applied + deferred

**Verdict: the fix ships.** `find "$arg"` → `find "$arg/"` is correct, minimal, and at the right level. Base-vs-head over one fixture changes exactly one behaviour and it is the reported bug. Declining `find -L` was right on both stated grounds, and the root-cause sweep found no sibling caller with the same `[ -d ]`/`find` mismatch inside the shim.

### Applied here (`e4c0000`, `002c6e7`)

**1. The `-L` rationale stated two behaviours neither `find` has.** Both were the load-bearing prose justifying the chosen form, and a maintainer who checked either would have discounted the paragraph.

- *"a symlinked `node_modules`"* names the shape the filter **does** catch — if the symlink is itself named `node_modules`, the printed path contains it and `-not -path` excludes it. The escape needs a symlink named something *else*.
- *"loops on a cycle"* is false on both implementations; neither hangs. GNU find detects the loop and **exits 1**, which the `||` on that same line turns into a refusal wearing the unreadable-directory message — so on CI, `-L` would hard-refuse a good suite. BSD find skips it silently at exit 0. The true behaviour is the stronger argument.

The test file restated the same mechanism in seven lines and inherited both errors; replaced with a pointer so the reasoning lives in one place.

**2. The rationale was prose only.** `find -L "$arg"` passed all 213 tests, so the likeliest regression was never deleting the character — it was "clarifying" it to `-L`, which reads more intentional. The symlink test now builds `t/vendor -> ../node_modules`, vendored code behind a symlink named something else. The slash form does not descend it and still reads `pass 3`; `-L` sweeps the vendored test in and reads 4.

Mutation-checked in both directions: reverting the slash **and** switching to `-L` each fail this test alone, while breaking the vendor filter fails only the vendored test. No new test case, so the count is unchanged.

**3. Two smaller corrections.** The anchoring comment cited "the vendored-test count above" — it is at `:226`, below. And "a symlink followed only one level deep" is not a `find` mode; `-maxdepth 1` is the real hazard `pass 3` guards, which is how the neighbouring directory test already words it.

**4. The `ci.yml` commit now removes the count instead of bumping it.** Verified comment-only first — filtering the diff to lines that are neither hunk headers nor `#` comments returns empty, and both job keys are intact.

The bump was a treadmill. The number is **already wrong on `main`**, which says 212 while its suite measures **221** — #167 landed nine tests, left the comment alone, and CI went green, because nothing validates it. Rebased onto current `main` this branch would assert 213 against 222. It was also contended: #171 independently edited the same line to 215 in the same hour. The rationale never needed the cardinality — "not one test" is the part that makes a zero-test run concrete.

`29 flat files` left alone and re-measured at 29; unlike the test count it does not move when a case joins an existing file.

### Confirmed, no action

- **#127 interaction, both halves.** The base→head diff of the test file is *only* the `symlinkSync` import plus the appended test — `SUITE` byte-identical, no fixture object touched. #127's five assertions are unchanged in content. The new assertion `/^(?:ℹ|#) pass 3$/m` is genuinely anchored, and better than #127's own proposal since it handles the piped `#` reporter. `pass 3` arithmetic checks out.
- **Cross-platform** settled by execution, not reasoning: `check` was already green on ubuntu at `43e7441` with the suite under `failglob`.
- **Counts**: PR base `fe70bf5` = 212, head = 213, current `main` = 221 — all re-measured independently.

### Deferred

| # | Finding | Label |
|---|---|---|
| #186 | The emitted runner runs vendored tests when the **argument** is a symlink into `node_modules`. Narrow regression from this PR: `-not -path` matches the printed path, which no longer contains `node_modules`. Bounded — the bare-relative route is pre-existing (both forms admit it), the ordinary shape is still excluded, and this repo has zero tracked symlinks. Two reviewers split on fix-vs-document, so the disposition is open. | `needs-triage` |
| #187 | `recall-lib.sh:146` — the stale-recall prune no-ops when `TMPDIR` is unset, because `/tmp` is a symlink. **Same defect class, same one-character fix** as this PR. Fails safe. | `ready-for-agent` |
| #188 | `claim-ticket.sh:30` — `[ -e ]` on a dangling symlink at the worktree path gives the wrong refusal. Fails closed; only the message is wrong. Filed so it is not dropped, not because it earns a claim. | `ready-for-agent` |

Negative-direction coverage (symlink to an empty dir / to non-test files) was measured rather than assumed: every case already refuses at exit 1, so it is YAGNI for a one-character fix and was not filed.

<sub>Note for readers of the history: `43e7441` bumps the `ci.yml` count and `002c6e7` then removes it. Kept as two commits rather than rewriting, to avoid a force-push mid-review.</sub>

Issue #147, opened 2026-07-31
no-undo-audit.sh:72 the stash count reads 0 for an unknown count — and git stash list exits 0 while doing it


feigi:
Deferred from PR #131 review. Pre-existing — that PR removed two swallows of exactly this shape elsewhere in the same script but did not touch this one.

`skills/fleet/scripts/no-undo-audit.sh:72`:

```sh
stash=$(git -C "$wt" stash list 2>/dev/null | wc -l | tr -d ' ')
```

Measured at `c17d4b6`, two real stash entries, then the stash reflog made unreadable:

```
$ git stash list | wc -l          -> 2
$ audit                            -> {"stash":2,...}
$ chmod 000 .git/logs/refs/stash
$ git stash list ; echo rc=$?      -> (no output) rc=0
$ audit                            -> {"stash":0,...}
```

## The mechanism is not the swallow

This is the same trap issue #106 fell into about `git log`, and it is worth stating plainly because the obvious fix does not work. **`git stash list` exits 0** with empty output when it cannot read the reflog. So neither `2>/dev/null` nor the pipe is what hides this — there is no error to catch, and adding `|| die` changes nothing. (`wc` is last in the pipeline anyway, so the exit code never belonged to git.)

## Why it matters despite being non-gating

The count is deliberately reported and never gated (`:56-71` explains why at length, and that reasoning is sound). But `skills/fleet/commands/run-merge-bot.md` instructs the operator to read `git stash list` **when the count is nonzero**, warning that an entry naming this branch may be a dead member's only copy. A silent `0` is precisely the value that stops the operator from ever looking — so the failure mode is "unrecoverable work is not read about", which is the hazard the surrounding comment block exists to flag.

## Why `needs-triage`

The defect is verified; what to do is open, and cheap options exist that are not obviously right:

1. **Cross-check the ref** — `git rev-parse --verify --quiet refs/stash` succeeding while the list is empty means the count is wrong, not zero. Two lines, catches this exact case, and says nothing about a reflog that is merely truncated.
2. **Report unknown instead of a number** — `"stash": null` when the cross-check disagrees. Honest, but changes the payload's type for a field callers may already read as a number.
3. **Leave it** — decide the field is advisory, and move the "read the list yourself" instruction in the runbook so it does not depend on the count being nonzero. Zero code.

[...]

PR #180, opened 2026-07-31
fix(fleet): the audit's worktree gate walked up, so it answered for the parent

### LINKED ISSUE #126: no-undo-audit.sh:21 — the `rev-parse --git-dir` worktree gate walks UP, so a deleted .git file passes it and the audit reads the PARENT as clean
`no-undo-audit.sh:21` opens with what looks like exactly the right guard — an explicit "is this a git worktree at all" gate, the very thing PR #115 added to `release-ticket.sh`:

```sh
git -C "$wt" rev-parse --git-dir >/dev/null 2>&1 || die "$wt is not a git worktree"
```

**That gate does not establish the target.** `rev-parse --git-dir` walks UP exactly like `status` does. Delete the worktree's `.git` file — directory and every uncommitted file still on disk — and the gate passes at rc 0, having resolved the **parent's** git dir.

Measured, git 2.50.1, clean parent with `.worktrees/` gitignored (the fleet's own layout):

```
$ rm .worktrees/9-x/.git          # directory and precious.txt still on disk
$ git -C .worktrees/9-x rev-parse --git-dir; echo "rc=$?"
/…/parent/.git                    # the PARENT's git dir, not the worktree's
rc=0                              # gate PASSES

$ git -C .worktrees/9-x status --porcelain; echo "rc=$?"
rc=0                              # empty — line 35 reads CLEAN
```

So the gate at `:21` waves the run through, and `porcelain=$(git -C "$wt" status --porcelain)` at `:35` then answers about the parent repository. With a clean parent the answer is **empty at rc 0**: a positive assertion that the worktree holds no uncommitted work, produced without ever having looked at it.

This is the same defect class as #74 (fixed for `release-ticket.sh` by PR #115), but it lands differently here and worse in one respect: `release-ticket.sh` had no identity gate at all, whereas this script **has one and it does not work**. A reader auditing `no-undo-audit.sh` sees line 21 and concludes the target is established.

It also matters more per-occurrence: this script's entire job is gating an irreversible action. A manufactured "clean" is the one answer it must never produce.

Note this is *not* the hole `53d423c` ("the audit's sole gate could still answer 'clean' without looking") closed, and that commit does not incidentally cover it either. `53d423c` replaced `porcelain=$(git -C "$wt" status --porcelain || true)` with a `die` on failure, so an *rc-non-zero* status can no longer be read as clean:

```sh
porcelain=$(git -C "$wt" status --porcelain) \
  || die "git status failed in $wt — cannot tell a clean worktree from a dirty one"
```

The deleted-`.git` input exits **rc 0**, so that new `die` never fires. The `||` branch is the wrong side of the problem: this failure mode is a *successful* command answering about the wrong repository. The linkage gate at `:21` is the only thing that could have caught it, and it does not.

## Suggested fix

Establish that the git dir belongs to **this** worktree rather than an ancestor, and refuse as unknown when it cannot be established. PR #115's precedent for the sibling script is `[ -e "$wt/.git" ] || die "…is unknown"`, chosen over comparing `rev-parse --show-toplevel` against `$wt` because the path comparison adds a false-refusal class (macOS resolves the tmpdir through `/private`). The same reasoning applies here.

**Do not** key on git's `prunable` annotation: it means "the gitdir link is broken", which includes a fully-populated worktree holding work that exists nowhere else — i.e. it would skip the check on exactly the worktree that needs it.

`no-undo-audit.test.mjs` should gain a case that fails against the current script. Assert the refusal happens **before** the audit reports anything, not merely that the exit code is non-zero.

Deferred from PR #115 review (which closes #74 and is correctly scoped to `release-ticket.sh` only).

feigi:
Closes #126

## What was wrong

`no-undo-audit.sh` opened with what reads as the right guard — `git -C "$wt" rev-parse --git-dir || die "$wt is not a git worktree"` — but `rev-parse --git-dir` **walks UP**, exactly as `status` does. Delete a nested worktree's `.git` (directory and every uncommitted file still on disk) and the gate passes at rc 0, having resolved the **enclosing** repo's git dir. `status --porcelain` then answers for that repo, and with a clean parent — `.worktrees/` gitignored, the fleet's own layout — the answer is **empty at rc 0**.

Measured on this branch before the fix, with `precious.txt` untracked in the worktree the whole time:

```
{"worktree":".../w/.worktrees/9-x","branch":"fix/9-nested","clean":true,"stash":0,"conflicts":[],"atRisk":[]}
```

`"clean":true`, exit 0, for a tree the script never looked at — out of the one script whose whole job is gating an irreversible action.

Not what `53d423c` closed. That commit made an rc-**non-zero** `status` a `die`; this input exits **rc 0**, so that `die` never fires. The failure here is a *successful* command answering about another repository, which only a linkage gate can catch.

## The fix

```sh
[ -f "$wt/.git" ] || [ -e "$wt/.git/HEAD" ] \
  || die "$wt has no .git of its own — cannot tell a clean worktree from a dirty one"
```

The `.git` file a linked worktree has, or the HEAD every real `.git` directory has. Exit 2, unknown — never "clean".

Two clauses because neither half-measure holds here (measured, git 2.50.1):

- **`-f` alone**, which `release-ticket.sh` spends on the same hole, refuses every main checkout — `$wt` here is any worktree the caller names, and a main checkout's `.git` is a **directory**. That is what every fixture in this suite is.
- **`-e "$wt/.git"` alone**, the spelling the ticket suggests, calls an **empty `.git` directory** present, and git walks up past that one too — the same manufactured clean, one byte over. This is precisely why the sibling guard is `-f`.

Rejected alternatives, both named in the ticket and both still rejected: comparing `rev-parse --show-toplevel` against `$wt` buys a false-refusal class (macOS resolves tmpdirs through `/private`), and git's `prunable` annotation means "the gitdir link is broken" — which *is* a populated worktree holding work that exists nowhere else, so keying on it would skip the check on exactly the worktree that needs it.

The pre-existing `rev-parse` gate stays: it answers "can git operate here at all" and keeps its own message for a directory with no repo above it.

## Tests

Three cases, in a new section. The fixture nests a linked worktree inside the clone with `.worktrees/` gitignored, because that is the only layout where the lie is available — a worktree with no repo above it has nothing to walk up to, and the script already refused there.

Each asserts the refusal lands **before the audit reports anything** (`doesNotMatch(stderr, /status --porcelain/)`), not merely a non-zero exit: a script that audits, prints "clean", then refuses has already put the wrong answer on the caller's screen.

Every clause is pinned — measured by mutating the guard and re-running:

| guard | what fails |
|---|---|
| `true` (guard deleted) | both refusal cases |
| `[ -e "$wt/.git" ]` (ticket's literal) | the empty-`.git`-directory case |
| `[ -e "$wt/.git/HEAD" ]` (`-f` dropped) | the intact-linked-worktree case |

That last row is why the passing case is here: dropping `-f` refuses **every linked worktree on disk** — the fleet's own shape, since `claim-ticket.sh` builds them with `git worktree add` — and without that test the suite stays green while every real caller is refused.

## Verification

- `./agent-test skills` on this branch: **tests 224, pass 224, fail 0, skipped 0**.
- `origin/main` at `14b765c`, measured from a `git archive` extract: **tests 221** (1 skip, the archive-extract skip). Exactly +3.
- Both new refusal cases were red against the unmodified script first, with the manufactured `"clean":true` payload quoted above as the failure.

Spec table updated: the exit-2 enumeration for `no-undo-audit.sh` now names the linkage refusal.


feigi:
Six-specialist review complete. Head is now `6e36f5f`.

## Applied

**The linkage guard was one byte short of the bug it closes.** `-e "$wt/.git/HEAD"` tests that HEAD *exists*; git requires HEAD **and** `objects/` **and** `refs/` before it will call a directory a git dir, and walks up past anything less. So a `.git` directory holding a lone HEAD passed the guard, `status --porcelain` answered for the enclosing repo, and the audit printed `"clean":true` at **exit 0** with the uncommitted work still on disk — byte for byte the failure this branch exists to fix. Four of six specialists found it independently. Not a hand-built shape only: a **real** `.git` whose HEAD an interrupted write truncated, and one whose `refs/` went missing, land in the same gap.

Swapped the predicate to ask git which tree it resolved rather than infer it from `.git`:

```sh
prefix=$(git -C "$wt" rev-parse --show-prefix) || die "$wt is not a git worktree"
[ -z "$prefix" ] || die "git answers for the repo above $wt, not $wt — ..."
```

`--show-prefix` is the path asked about **relative** to the worktree root git picked — empty iff `$wt` **is** that root. No string compare, so the three false-refusal classes that correctly ruled out `--show-toplevel == "$wt"` (relative `$wt` from `claim-ticket.sh:26`, symlinked path, macOS tmpdir reported back through `/private`) cannot arise. Measured: **8 leak shapes refused** (including the `core.worktree` redirect `release-ticket.sh` defers to its own ticket), **5 legitimate shapes admitted** — relative arg, symlink, trailing slash, main checkout, intact linked worktree.

Also applied: 3 tests (each red against the previous predicate, green after, other 22 unmoved); the comment claiming *"every fixture in this file"* is a main checkout, which `nestedWorktree` in the same commit contradicts; and `run-merge-bot.md:124`'s exit-2 list — the copy the merge bot actually reads — left behind when the spec table was updated.

**`./agent-test skills`: tests 227, pass 227, fail 0, skipped 0.** (`origin/main` 221, this PR was 224.)

## Confirmed, not changed

- **The deviation from #126's literal `-e "$wt/.git"` was correct.** Both halves verified by construction: `-e` returns true for an empty `.git` directory and git walks up past it at rc 0 (`status` empty → manufactured clean); a main checkout's `.git` is a directory, so `-f` alone false-refuses it — and 17 of this file's fixtures are main checkouts.
- **All three claimed mutations reproduce exactly**, plus a negative control (clauses swapped → stays green, so the tests pin behaviour not spelling).
- **The passing case genuinely fences dropping `-f`** — it is the sole failure under that mutation, and it is a passing-path case, not another refusal.
- **`doesNotMatch(stderr, /status --porcelain/)` discriminates.** Moving the guard below the status echo fails on exactly that assertion's message. `assert.equal(r.stdout, "")` is independently load-bearing.
- **Both rejected alternatives stayed rejected** — `--show-toplevel` and `prunable` appear only in comments, never in code. The `prunable` rationale was *inverted*, though (it marks a worktree whose **directory** is gone — nothing left to lose — and stays silent for one still holding work whose linkage broke); corrected in the rewritten comment.
- **The pre-existing `rev-parse --git-dir` gate is not dead** — it catches an unsearchable `$wt` with the accurate message, which is why this script needs no analogue of `release-ticket.sh:248`'s `! -x` clause.
- **Scope of the spec edit: keep.** The PR added an exit-2 condition, so enumerating it is obligatory, not scope creep. #114 covers rows that are *wrong*, and none of its defects touch this row — commented there.

## Deferred

- **#189** `needs-triage` — a `.git` file naming another worktree's admin dir still passes (gitdir identity ≠ worktree-root identity); `release-ticket.sh:229-233` defers the same residual, should close for both at once
- **#190** `ready-for-agent` — `jstr` failure inside the `printf` argument list is invisible to `set -e`
- **#191** `ready-for-agent` — test hygiene: dead fixture push, unused parameter default, overclaiming assertion message
- **#147** (existing) — commented: rc capture fixes the corrupted-ref variant but a `chmod 000` reflog returns empty at rc 0, so no shell-level fix reaches it

## Not fixed — history

Commit `36db033`'s body says *"dropping the HEAD clause fails the empty-directory case"*. It does not: that mutation leaves `-f` alone, which still refuses an empty `.git` **directory**, and instead false-refuses 17 main-checkout tests. The outer two attributions in that sentence are correct. Left alone rather than force-pushed over three commits mid-run.

feigi:
Follow-up to the review comment above. Head is now `a3bbbff`.

**One dead line the predicate swap created, now removed.** `--show-prefix` returns the **same rc as `--git-dir` on every shape reachable here** — plain directory, garbage `.git` file, `.git` file whose gitdir target is gone, dangling symlink and unsearchable worktree all `128` for both; main checkout, linked worktree, bare repo, and `.git` deleted / empty / holding a lone HEAD all `0` — and it already dies with that gate's own message. Removing the gate leaves the suite at **227/227**.

Removed rather than kept as cheap redundancy, because it read as an ordering dependency that no longer exists. The old `-f`/`-e` guard genuinely needed something ahead of it: `chmod 644` on the worktree makes `-f "$wt/.git"` false **with the `.git` sitting right there**, so it would have asserted an absence nothing established — the hazard `release-ticket.sh:248` spends `[ ! -x "$wt" ]` on. Asking git removes the stat, so it removes the hazard. A comment documenting the ordering as load-bearing — which was the reasonable next step under the old shape — would have been false the moment it was written.

**Adjudication of the third fix proposal, for the record.** Extending the predicate to git's own `is_git_directory()` triple:

```sh
[ -f "$wt/.git" ] || { [ -e "$wt/.git/HEAD" ] && [ -d "$wt/.git/objects" ] && [ -d "$wt/.git/refs" ]; }
```

is the smallest delta and it **does not work**. Measured against the same shape set: it closes `.git` dir with a lone HEAD and `.git` dir missing `objects/`, but still **admits** a real `.git` whose HEAD is garbage and one whose HEAD is a directory — both leak `"clean":true` at exit 0. `-e` tests HEAD's *filename*; git tests whether HEAD *validates as a ref*, and no `[ ]` spelling reaches that. Re-deriving git's rules in shell is the same class of mistake as the original guard, one rung further along. That is why the swap asks git instead.

**Also on the record:** the mutation *table* in the PR body is accurate — its row 2 is `[ -e "$wt/.git" ]`, a rewrite of both clauses, which does fail exactly the empty-directory case. Only commit `36db033`'s prose describes that row as *"dropping the HEAD clause"*, which is a different mutation. The measurement was done correctly and then described incorrectly. Still not force-pushing over three commits to amend it: the predicate it describes no longer exists in the tree, so the inaccuracy is archaeological.

PR #193, opened 2026-07-31
docs(fleet): the source-of-truth rule set no size bound, so 88 whole-file reads landed in agent prefixes

### LINKED ISSUE #181: review-and-fix.md:47 makes whole-file `git show` the default with no size bound — 88 unpiped calls, 434 KB into agent prefixes in one run
Found by a token-spend audit of the fleet plugin, not by a PR review. Measured with the plugin's own `computeSpend` over one real run's subagent transcripts (session `e8fa298a`, 92 subagents).

## The finding

`skills/fleet/commands/review-and-fix.md:47` makes whole-file `git show` the default read, with no bound on size:

> Tell readers **`git show <sha>:<path>` is the source of truth** — the object store cannot be contaminated by any agent, needs no copy, and touches nothing. Make it the default for settling *what the PR contains*, not a fallback.

The rule is right and should stay. What is missing is any instruction to bound the output. Measured across that one run's 92 subagent transcripts, classifying each `git show <sha>:<path>` call by whether it pipes into `sed`/`head`/`grep`/`awk`/`wc`/`rg`:

| form | calls | tool_result bytes |
| --- | --- | --- |
| piped / ranged | 319 | — |
| **unpiped, whole file** | **88** | **433,784** |

434 KB is roughly 110k tokens, and it is the single largest attributed line item in the run — larger than every `Read` call combined (856 KB across 106 calls, but spread over far more agents). Worst individual case: `git show <sha>:skills/fleet/scripts/release-ticket.test.mjs` at ~37 KB per call, 10 calls.

It lands in the worst place. A specialist or reviewer that dumps a 37 KB file at turn 5 carries it in its prefix for every remaining turn, so it re-bills as cache-read on each one. Reviewers in this run peaked at 241k context and account for 53% of run spend; specialists another 30%.

Per-agent totals for the unpiped form:

```
52839  15x  Comment analysis PR 110
49816   9x  Simplify review PR 150
44862  13x  Review PR 115
42934  14x  Code review PR 110
40690  15x  Comment/claim audit PR 107
```

That the same run also made 319 *piped* calls shows agents already know the bounded form. Nothing tells them when it is required, so the tail goes unbounded.

## The fix

Extend `:47` with a size rule. The object-store guarantee is unaffected — `git show <sha>:<path> | sed -n 'A,Bp'` reads the same uncontaminatable blob:

- default to a bounded read: `| sed -n 'A,Bp'` around the hunk, or `git show <sha> -- <path>` when the diff is what is being settled
- whole file only when it is genuinely small (a line count, not a feeling — `git show <sha>:<path> | wc -l` costs nothing)
- name the reason: an unbounded read is permanent prefix, not a one-off cost

## Acceptance criteria

- [ ] `review-and-fix.md:47` states when the whole-file form is warranted and what to use otherwise
- [ ] The object-store-is-authoritative rule is preserved verbatim in substance — this ticket does not weaken it
- [ ] No new required file read is introduced to carry the rule

## Out of scope

- Reviewer/controller doc compression generally — that belongs to #66
- Moving text into `references/` — blocked on the #78 call

feigi:
`review-and-fix.md:47` makes `git show <sha>:<path>` the default read for settling what a PR contains, and says nothing about its size. Measured over one run's 92 subagents (session `e8fa298a`, via the plugin's own `computeSpend`): 88 unpiped whole-file calls carried 434 KB (~110k tokens) — the single largest attributed line item in that run — against 319 piped ones. An unbounded read is prefix, not a one-off cost; it re-bills as cache-read on every remaining turn.

## What changed

One line, `:47`, extended in place. No new file, no new required read.

- **Bounded default** — `| sed -n 'A,Bp'`; whole file only after `| wc -l` says it is small.
- **Diff forms for settling a change** rather than a file.
- **The reason named** — prefix cost, not one-off.

## One correction to the ticket's own suggestion

The ticket proposed `git show <sha> -- <path>` "when the diff is what is being settled". Shipped with a qualifier, because measured it has a silent-zero mode:

```
$ git show $(git rev-parse origin/main) -- skills/fleet/commands/review-and-fix.md | wc -l
0
```

It is **one commit's** diff and prints nothing at exit 0 when that commit did not touch the path — which a reviewer reads as "the PR changed nothing here". Same silent green as the `tests 0` trap this file already documents. So `git diff <base>...<head> -- <path>` is named for a PR's whole change and `git show <sha> -- <path>` scoped to one named commit.

## Acceptance criteria

- [x] `:47` states when the whole-file form is warranted and what to use otherwise
- [x] The object-store-is-authoritative rule preserved verbatim — a pipe reads the same blob, and the original three sentences are untouched
- [x] No new required file read introduced

## Verification

`node --test skills/fleet/scripts/*.test.mjs` → 236 pass, 0 fail, before and after rebase on `origin/main`. Docs-only diff; no test asserts on this file.

Closes #181


feigi:
Review: \`review-pr\` workflow, 9 agents, dimensions \`correctness\` + \`comments\` (docs-only diff), 2 findings, **0 survived adversarial verification**.

**Refuted — the bounded form is not a silent-zero.** The claim was that `git show <sha>:<path> | sed -n` masks git's exit 128, so the prescribed `| wc -l` gate answers `0` ("small") on a mistyped path or unfetched ref. The mechanics reproduce, but both verifiers refuted the equivalence: the bounded read prints `fatal: path ... does not exist` on stderr, which a Bash tool surfaces inline. The two cases `:47` actually warns about (`tests 0`, and `git show <sha> -- <path>` on an untouched path) emit **zero bytes on both streams** — that silence is the whole point of the warning. Loud and self-diagnosing is the opposite class. Not applied.

**Refuted as stated, half deferred → #194.** Second finding: the size rule reaches only this file, while `workflows/review-pr.js` — the path `:33` tells controllers to prefer — hands its specialists neither the size rule nor the source-of-truth rule. Verifiers split 1-1. The `:187` half is genuinely refuted: that line is a byte-identity check whose natural `| cmp -` execution is O(1) in file size (187 bytes on an 18.6 KB file), and a `sed` range would break it. The prompt-block half (`:239-259` carries neither rule) neither verifier disputed, and it reproduces directly — filed as #194, `ready-for-agent`.

**Merge-commit nit, twice refuted.** `git show <sha> -- <path>` is also silent on a merge commit under default `--cc`. True, and an additional trigger — but the sentence states a sufficient condition, not an exclusive one, and its operational advice is unchanged. Incompleteness, not a wrong claim. Not applied.

Nothing to apply; no new commit. CI green and run-bound: run `30640146793`, attempt 1, head `27c004b`, `rebase-check` + `check` both success, `behind_by=0`.

Issue #78, opened 2026-07-30
run-team: references/ do not hold the war stories the preamble promises


feigi:
`skills/fleet/skills/run-team/SKILL.md:14-16` promises:

> The war story behind each rule lives in `references/`; load one only when a member needs the *why*.

Several inline anecdotes have no reference backing them, so a member that follows the pointer finds nothing.

Concrete case found reviewing PR #73:

- `SKILL.md` finisher paragraph ends `See references/ci-and-staleness.md.` and now carries the deferral gate and the dirty-worktree halt. That reference has **zero** hits for `dirty`, `worktree`, `deferral` or `finisher`:
  `git grep -inE "dirty|worktree|deferral|finisher" <sha>:skills/fleet/skills/run-team/references/ci-and-staleness.md` → no output. It covers CI and staleness only, which is what it is named for.
- `Caught once at seven findings.` is uncitable anywhere in the tree: `git grep -in "seven findings" <sha>` returns only that line.

So the finisher's two newest rules point a reader at a reference that explains neither.

**Why deferred rather than fixed in PR #73:** the approach is open, and all three options are defensible.

1. Add the war stories to `references/ci-and-staleness.md` — but they are not CI or staleness material, so this arguably wants a new `references/finisher.md`.
2. Add a new reference file and repoint the paragraph.
3. Relax the `:14-16` claim to "most rules", and accept inline anecdotes as self-contained — which is the file's *de facto* house style already: `SKILL.md` and `references/member-lifecycle.md` both carry uncited inline tallies.

Option 3 is the smallest and may be the honest one; it is also the one that quietly weakens a promise other rules lean on. Needs a call before anyone writes text.

Scope checked: `skills/fleet/skills/run-team/` at `cce753e`, all five files in `references/`.

Deferred from PR #73 review.

feigi:
## On hold — this ticket's call now gates other work

Parked with `onhold` (excluded from the candidate queue at `candidates.mjs:43`), not because it is stale but because a token-spend audit of the plugin found a second problem that **option 3 would decide against, silently**.

## What the audit found

Measured over the corpus, counting shared 8-word spans between `run-team/SKILL.md` and the files it defers to:

| reference | shared 8-grams with SKILL.md |
| --- | --- |
| `isolation.md` | 38 |
| `member-lifecycle.md` | 34 |
| `reaping.md` | 21 |
| `ci-and-staleness.md` | 11 |

[...]

PR #195, opened 2026-07-31
docs(fleet): review-and-fix steps 4 and 6 retold the CI rules their own blocks below already state

### LINKED ISSUE #182: review-and-fix.md and run-merge-bot.md each restate the whole ci-state ruleset, and the reference that owns it is cited by neither
Found by a token-spend audit of the fleet plugin, not by a PR review.

## The finding

`references/ci-and-staleness.md` exists to hold the CI reading and staleness material. **Neither command file cites it, and both restate it in full:**

```
$ grep -c 'ci-and-staleness' skills/fleet/commands/review-and-fix.md skills/fleet/commands/run-merge-bot.md
skills/fleet/commands/review-and-fix.md:0
skills/fleet/commands/run-merge-bot.md:0
```

The same four facts are told independently in both files, and again in `run-team/SKILL.md`:

| fact | review-and-fix.md | run-merge-bot.md |
| --- | --- | --- |
| `gh pr checks` aggregates across runs, inherits a `pass` from a cancelled run | `:13` | `:80` |
| what `ci-state.mjs` binds; absent job reads `pending`, inherited reads `pass` | `:19` | `:86` |
| a conclusion is not stable under a fixed run id | `:23` | `:88` |
| `--limit 1` can hide the CI result | `:23` | `:90` |

Measured over the corpus, counting shared 8-word spans: **47** between `review-and-fix.md` and `run-merge-bot.md`, **59** between `review-and-fix.md` and `run-team/SKILL.md`.

Within `review-and-fix.md` alone the run-binding / behind-count / `skipped`≠`passed` story is told five times — steps 4 (`:9`) and 6 (`:11`) are paragraph-length restatements of the standalone bold blocks that follow them at `:13`, `:21`, `:23`, `:25`, `:27`.

## Why it costs

Both files are read *in full* by the members that use them, once per member. In one measured run: `review-and-fix.md` 11 reads / 179 KB, `run-merge-bot.md` 6 reads / 92 KB. That text then sits in the prefix of the longest-lived agents in the fleet — reviewers peaked at 241k context and are 53% of run spend — so it re-bills as cache-read on every subsequent turn. Roughly 4.7k tokens x ~55 turns x 10 reviewers ≈ 1.2M cache-reads for `review-and-fix.md` alone.

It is also the failure mode `run-team/SKILL.md`'s own "Fix the tooling mid-run" section warns about: *"Prefer moving text to adding it, and delete the copy you superseded; a duplicated rule becomes a contradiction."* That has already happened once — see the `attempt`-key contradiction filed separately.

## The fix — compress in place, do not add a required read

Fold each file's five-times-told version down to one statement per fact, keeping it **inline**. Specifically in `review-and-fix.md`, steps 4 and 6 stop restating what `:13`-`:27` already say and reference their own bold blocks instead.

**Deliberately not "move it into `references/ci-and-staleness.md`".** A member reading `run-merge-bot.md` needs these facts on the path it is already on; replacing an inline paragraph with a second mandatory file read costs more tokens, not fewer. The move-to-`references/` direction is also exactly the call parked on #78 and must not be pre-empted here.

## Acceptance criteria

- [ ] Each of the four facts above appears once per file, not two-to-five times
- [ ] No instruction is lost — every rule still stated, just once
- [ ] No new mandatory file read added to either command file
- [ ] Both files still agree with each other and with `run-team/SKILL.md` on every fact they share

## Out of scope

- `references/ci-and-staleness.md` itself, and whether text should move there — #78
- General prose-compression criteria — #66

feigi:
Steps 4 and 6 of `review-and-fix.md` carried paragraph-length restatements of the run-binding, behind-count and `` `skipped` ``-is-not-`` `passed` `` rules that the standalone bold blocks at `:13`-`:27` already state in full. A member read the same rule up to five times in one file, and that file sits in the prefix of the longest-lived agents in the fleet.

## What changed

- **`review-and-fix.md` steps 4 and 6** state their own operative rule and point at the block that owns the rationale — `**`skipped` is not `passed`**` and `**Do NOT rebase to label**`. Every gate rule survives; only the re-derivation went.
- **`ci-state.mjs`'s per-job paragraph in both files** stopped re-deriving the inherited-pass hazard that its own bold block states one paragraph earlier.
- **`run-merge-bot.md` step 2** stopped restating step 3's re-query rule, keeping the pointer to it.

## What deliberately did not change

Nothing moved to `references/ci-and-staleness.md` and no new mandatory read was added — that direction is #78's call, and replacing an inline paragraph with a second file read costs a member tokens rather than saving them. Cross-file duplication between the two command files is left in place for the same reason: each is read by a member already on that path.

## Verification

`node --test skills/fleet/scripts/*.test.mjs` → 236/236 pass.

Acceptance criteria were encoded as a check and watched fail before any edit (6 real duplications), then pass:

| criterion | result |
| --- | --- |
| each of the four CI facts told once per file | `aggregat` / `inherited` / `rewrit` / `newest run` / ``as `pending` `` = 1 each, both files; steps 4 and 6 restate none of them |
| no instruction lost | all 122 rule anchors present at `origin/main` still present |
| no new mandatory file read | set of file paths each command references is unchanged; neither cites `ci-and-staleness.md` |
| both files still agree with each other and `run-team/SKILL.md` | every shared fact still present in all three |

Shared 8-word spans between the two command files (whitespace-split tokenization): **39 → 38**. An earlier revision pasted one identical replacement sentence into both files and pushed this to 63; each file now carries its own phrasing.

One clause was caught and corrected during the pass: a rewrite claimed a cancelled run's finished jobs "keep being reported as `pass`", which is false for a job that concluded `failure`. Corrected against `ci-state.mjs:156-157`, which says they keep *their* conclusions.

Closes #182


feigi:
Review pass: 3 specialists (`correctness`, `rule-loss`, `comments`) over an immutable snapshot of `283d985`, 15 findings, each critical/important one faced 2 adversarial refuters. 1 survived, 6 refuted, 8 unverified suggestions.

**Applied here** (`c940227`)

- `review-and-fix.md:9` — restored `solely from a non-zero behind-count` to step 4. `.github/workflows/ci.yml` fails `rebase-check` on five conditions (fetch failure, unresolvable base, no common ancestor, not rebased, merge commits above base); only one is the behind-count. Without the qualifier a standalone reviewer reads any `rebase-check` red as staleness, stops watching and goes to step 6 — where the label gate deliberately ignores `rebase-check`. Restored the six words only; step 4 already says "fix a **genuine** failure and repeat from 3" three clauses earlier.

**Deferred**

- #196 `ready-for-agent` — step 6 dropped `behind-count` (diverging from `run-team/SKILL.md:327` and `board.mjs:54`) and the mechanism behind "which a behind PR never reaches"
- #197 `ready-for-agent` — step 4's pointers do not match their bold targets verbatim (trailing punctuation sits inside the emphasis at `:25`/`:27`)
- #198 `needs-triage` — the ci-state paragraph lost `force-push` in one file and the aggregation scope on "reads as `pending`" in the other; open whether to tighten the docs or `ci-state.mjs:155`, which carries the same looseness

**Rejected**

- Three findings (2 critical, 1 important) claimed the rewritten sentences misattribute the inherited-pass hazard to the job-presence check rather than the head-SHA filter. Refuted 2/2 each, with probes against a stubbed `gh`: `review-and-fix.md:13` and `run-merge-bot.md:80` both state head-SHA binding **misses** that case ("the head is genuinely right"), so crediting the presence check is correct.
- "236/236 pass overstates — 235 pass, 1 skipped": disproved by running the suite. `node --test skills/fleet/scripts/*.test.mjs` → 236 tests, 236 pass, 0 skipped.

**Corrected in the PR body** — "40 → 38" did not reproduce; whitespace-split gives **39 → 38**. Body updated and the tokenizer named. The fix commit leaves it at 38.

PR #200, opened 2026-07-31
docs(fleet): candidates.test.mjs claimed jq fidelity the stub cannot have — gh applies --jq with embedded gojq

### LINKED ISSUE #63: candidates.test.mjs overstates its jq fidelity — gh applies --jq with embedded gojq, not the system jq the stub execs

feigi:
The header said the stub "runs the real jq ... with the same expression gh would have received". The expression is real; the engine is not.

Verified against the tree, not taken from the ticket:

- `go version -m $(which gh)` — gh 2.96.0 embeds `github.com/itchyny/gojq v0.12.17`, and applies `--jq` in its own process. The stub `exec jq`s the system jq (`jq-1.7.1-apple`).
- Flavours differ: `gh api ... --jq '"xy"|test("x(?=y)")'` fails with `error parsing regexp: invalid or unsupported Perl syntax: `(?=``; the same expression under system jq returns `true`. So a flavour-sensitive predicate was never inside this suite's reach, however real the expression was.
- The current predicate is unaffected: reducing the 93 live open issues of `feigi/claude-config` through both engines yields identical `spec` and `d` for every row. (The serialisations differ — gojq sorts keys and gh escapes `<`/`>` — but `JSON.parse` erases both, and nothing in the suite reads bytes.)

The header also called `--jq` server-side, which the stub's own comment in the same file already contradicts ("gh applies `--jq` itself, inside its own process"). Dropped rather than restated.

Comment only. 17/17 pass, unchanged.

Closes #63


feigi:
Review of `243cca0`. Applied one fix in `a4606bc`; three findings deferred and filed.

**Applied — the correction shipped a new false claim.** The header asserted "a predicate that turns on flavour is out of this suite's reach. The current one does not: it reduces live issues to identical rows under both." The live reduction does agree under both engines (I reproduced it: 103 open issues, 0 mismatched parsed rows). The conclusion does not follow. The current predicate *does* turn on flavour — through `\s`, not through lookahead:

```
$ jq -n '"## User Stories<U+00A0>"|test("(?m)^##\\s+User Stories\\s*$")'
true      # system jq (Oniguruma) — what this suite runs
$ gh api repos/feigi/claude-config --jq '<same expression>'
false     # gojq (Go RE2) — what production runs
```

`\s`/`\d` are Unicode-aware in Oniguruma and ASCII-only in RE2. Production is the stricter matcher, so a `## User Stories` heading padded with U+00A0 is `spec:false` under `gh`, `dropSpecs` never fires, and the spec ships as a claimable ticket — with this suite green throughout. `candidates.mjs:155`: "This is the ONLY line of defence — nothing downstream catches a spec."

The lookahead illustration was true but taught the wrong lesson: gojq *rejects* that pattern and exits non-zero, so it makes divergence look loud. The divergence that actually applies is silent. Swapped the illustration to `\s` so the header names the live exposure.

Ruled out by test, not by argument: `(?m)` behaves identically in both engines (both anchor per-line with it, at string bounds without it, neither reads `m` as dot-matches-newline), and gojq's sorted-key output is harmless because `candidates.mjs:111-118` type-checks properties rather than key order.

Also narrowed "The expression is under test" → "The spec predicate is under test": neutering `d:` to `[]` leaves the suite at 17/17.

**Deferred and filed**

- **#204** (`needs-triage`) — the `\s`/`\d` divergence as a production defect in `candidates.mjs:70-73`. Triage rather than ready-for-agent because two materially different fixes are open: make the expression engine-portable (explicit ASCII classes), or make the suite run the engine production actually uses.
- **#205** (`ready-for-agent`) — `candidates.test.mjs` pins none of the `d:` dependency scan; neutering it to `[]` leaves 17/17 green. Distinct from #58, which is the scan under-firing rather than being unguarded.
- **#206** (`ready-for-agent`) — `skills/fleet/skills/next-ticket/SKILL.md:95` reads "server-side `--search` + `--jq`". Outside #201's stated scope (`grep -rn 'server-side' skills/fleet/scripts/`), so owned by neither #201 nor this PR. Repo-wide sweep in the issue body; the two remaining `server-side` hits are accurate.

`candidates.mjs:6` is #201's and was left alone.

**Acceptance criteria** — all three met, verified independently. Comment-only holds after my change: zero non-comment lines in the diff against `main`, and the executable remainder is byte-identical (`shasum 682c4dc4a4b47ed4190335aadf348b49a08095e5` on both sides). 17/17 pass, `tests 17` — not a vacuous zero-test green.

PR #203, opened 2026-07-31
fix(fleet): halt announced a partial release over a detail line reading all-false

### LINKED ISSUE #85: release-ticket.sh:118-125 `halt` announces PARTIALLY RELEASED in cases where nothing at all was touched
Deferred from PR #72 review. Pre-existing shape.

## Problem

`skills/fleet/scripts/release-ticket.sh:118-125`:

```sh
halt() {
  echo "$NAME: #$issue PARTIALLY RELEASED — $1" >&2
  echo "    worktree removed: $done_wt, branch deleted: $done_branch, in-progress: still on the issue" >&2
```

`halt` is reached from the very first mutation (`:247`, `git worktree remove`), so the common case is `done_wt=false, done_branch=false` — **zero of three artefacts touched**. The headline says PARTIALLY RELEASED; the line immediately below correctly says nothing landed.

## Measured

```
release-ticket: #9 PARTIALLY RELEASED — git worktree remove refused .../9-x: fatal: validation failed...
    worktree removed: false, branch deleted: false, in-progress: still on the issue
    the ticket still reads as taken — finish or restore it by hand
exit 2
```

Nothing was released. The claim is intact.

## Why it is minor

The detail line is accurate and sits one line below, and the JSON is accurate (`"released":false`). This is operator-facing prose only. But this script is unusually careful about not misreporting state — that care is its entire design — so a headline that overstates is off-key for the file it lives in.

## Reachability

Ordinary, and independent of #72: any `git worktree remove` refusal reaches it, including the #74 path.

What it should say when zero artefacts landed. `NOT released` collides with the `:207` blocked message which means something different (preconditions refused, nothing attempted); this case attempted and was refused mid-flight. Needs a third phrasing, and picking it is a judgement call.

feigi:
`release-ticket.sh`'s `halt` is reached from the first mutation onward, and the first mutation attempted is the worktree removal — so the ordinary refusal is one where zero of the three artefacts were touched, printed under a `PARTIALLY RELEASED` headline with the detail line right under it correctly reading all-false.

The headline is now decided by the two booleans that track what landed, never by the call site:

```
release-ticket: #9 HALTED mid-release — nothing landed: git worktree remove refused …
    worktree removed: false, branch deleted: false, in-progress: still on the issue
```

`PARTIALLY RELEASED` is kept for a refusal that followed a successful mutation. Keying on the call site instead would still be wrong: the same `git branch -d` refusal is reached with nothing removed on a claim that has no worktree registered (measured).

Unchanged: the detail lines, the compensating-action line, exit 2, and the JSON receipt — the new test asserts both receipts whole, byte for byte. The pre-mutation blocked message (`NOT released — nothing was touched`) and its exit 1 are untouched, and now pinned, since the distinction it carries — refused before anything was attempted, versus attempted and refused — is what the new wording has to stay clear of.

## Tests

`skills/fleet/scripts/release-ticket.test.mjs`: one new case pins both headline forms against a `git` shim that refuses on command, so both receipts are deterministic to the byte. Two mutants killed — the pre-fix unconditional headline (2 failures) and `&&` for `||`, which survives both-false and both-true and dies on the one-landed boundary (1 failure).

Full suite: `./agent-test skills/fleet/scripts` → tests 237, pass 237, fail 0.

Closes #85


feigi:
Review pass (5 specialists: code-reviewer, silent-failure-hunter, comment-analyzer, code-simplifier, pr-test-analyzer; type-design-analyzer not applicable to a `/bin/sh` script plus a `node:test` file).

**Behaviour ships as-is.** Verified rather than inherited: the receipt is byte-identical by construction (one `printf` outside the branch, mutation-proven — a JSON-equivalent key reorder kills the new test and nothing else); both mutants re-run independently (unconditional headline → 2 failures, `&&` for `||` → 1 on the one-landed boundary); the second reachability path reproduces with an intact receipt (`"worktree":""`, exit 2); the retired `/PARTIALLY RELEASED|contains modified or untracked/` alternation was an OR whose second arm passed even with a wrong headline, so its removal is a strengthening, and the string has no other referent tree-wide.

**Applied in 3705027 — comments only, no behaviour change.** The rewording left five comments quoting the string it replaced, two inside this ticket's own added prose:

- `release-ticket.sh:117` — opener true of only one of the two branches
- `release-ticket.sh:123` + `test.mjs:606` — "the FIRST mutation" unqualified; it is guarded by `[ -n "$wt" ]`, so `git branch -d` is first on a claim whose registration is already cleared
- `release-ticket.sh:265` — "a partial release that never happened", now inaccurate in the other direction
- `test.mjs:569` — "the other two halt() sites are reached with nothing deleted"; the `branch -d` site is ordinarily reached with the worktree already removed, as this PR's own new test asserts fifty lines below
- `test.mjs:616` — "the only thing a partial release is"; the label-edit refusal after both deletes is also partial
- `test.mjs:716`, `:778` — counterfactual halts described as announcing a partial release

`./agent-test skills/fleet/scripts` → 237 tests, 237 pass, 0 fail.

**Deferred → #208** (`needs-triage`): `done_wt` records the exit code, not what landed. `git worktree remove` clears `$GIT_DIR/worktrees/<id>` even when the directory removal fails — measured against bare git 2.50.1, registry 2→1 at rc 255 with the directory still on disk. So `HALTED mid-release — nothing landed` is falsifiable on that path, where the old wording was accurate by luck. Not the wording question #85 settled; the boolean behind it is not a measurement. Filed rather than fixed because there are three states and one boolean, and the compensating-action line derives from it. `release-ticket.sh:129-132` now records the limit so the file no longer claims the booleans measure the filesystem.

Not filed, judged not worth it: consolidating the `git` shim into a shared helper is +6 lines and edits a passing neighbour test outside #85.

One correction for the record: the PR body's `set -e` rationale is wrong, though harmless. The `&&`-chain hazard needs the chain to be the function's **last** command; here `printf` and `exit 2` follow it, and sh/dash/bash all exempt a non-final AND-OR member from `set -e`. Three specialists ran the rejected `&&` form and it behaves identically. `if`/`else` is still the right choice on readability — the reasoning given for it just is not load-bearing, and it appears nowhere in the tree, so nothing needed changing.

feigi:
Addendum — `e2c5181`, one test added after the pr-test-analyzer report landed.

Two mutants survived the suite as pushed, both in the same cell of the truth table:

- keying the headline on the **call site** instead of the booleans — 39/39 green while printing `PARTIALLY RELEASED` where nothing landed, i.e. #85 reintroduced undetected
- dropping the `done_branch` disjunct — 39/39 green while printing `nothing landed` over a deleted branch

Both close with one fixture: a claim whose worktree was removed by hand, which is the state the new case above already cites as its justification without exercising it. Two runs on it — `git branch -d` refused (false, false, same call site as the partial case) and the tracker refused after the branch went (false, true).

Verified: each mutant now fails on this test alone, the suite is green on revert, and the test stays green under a reworded command trace it does not claim to pin. `./agent-test skills/fleet/scripts` → 238 tests, 238 pass, 0 fail.

Written rather than deferred because the claim it pins is this PR's own, and it was the only unpinned one.

Issue #194, opened 2026-07-31
the bounded-read rule reaches only hand-dispatched reviewers — review-pr.js's specialist prompt carries neither it nor the source-of-truth rule


feigi:
Deferred from PR #193 review (#181). Found by the correctness specialist, verified 1-1 by the adversarial pass; the half filed here is the half neither verifier disputed.

## The gap

PR #193 extended `skills/fleet/commands/review-and-fix.md:47` with the size rule for `git show`. That file is the **hand-dispatch** path. The workflow path — `workflows/review-pr.js`, which `review-and-fix.md:33` tells controllers to *prefer* — hands its six specialists neither the size rule nor the source-of-truth rule it extends.

Verified directly on the tree:

```
$ sed -n '239,259p' workflows/review-pr.js | grep -E "source of truth|object store|uncontaminat|sed -n|wc -l|git show"
(no match)
```

The prompt block at `:239-259` carries snapshot-isolation, the `testCmd` + `tests 0` reading rule, the scratch-dir rule and report-what-you-ran — and nothing about how to read a file at a revision, bounded or otherwise. So the specialists dispatched by the preferred path are exactly the population #181 measured (88 unpiped whole-file reads, 434 KB, the largest line item in that run) and they receive none of the guidance that run produced.

This is the drift shape already on record: a rule lands in `review-and-fix.md` and not in its sibling path. `run-team/SKILL.md:517` avoids it deliberately ("Specialist tree isolation is `review-and-fix.md`'s job — do not restate it. It owns the object-store rule"), but `review-pr.js` has no such delegation — its specialists never read `review-and-fix.md` at all.

## Do NOT bound line 187

`review-pr.js:187` also names `git show HEAD:<path>` unpiped, and the original finding proposed bounding it. **That was verified and rejected** — recorded here so it is not re-litigated:

- The line reads "confirm a couple of the diff's files are byte-identical between the snapshot and `git show HEAD:<path>`". It is a byte-identity **check**, whose natural execution is a comparator: `git show HEAD:<path> | cmp - <snap>/<path>` emitted 187 bytes on an 18,651-byte file. Output is O(1) in file size, so the prefix-cost hazard cannot arise there.
- A `| sed -n 'A,Bp'` on that line would **break** the instruction — a partial read cannot establish byte-identity.
[...]

PR #207, opened 2026-07-31
docs(fleet): two files cited a suggestion-defer policy the reviewer doc never stated

### LINKED ISSUE #112: run-team/SKILL.md:309 cites a suggestion→defer policy that review-and-fix.md never states
Deferred from PR #107 review.

`skills/fleet/skills/run-team/SKILL.md:308-309` justifies the workflow skipping adversarial verification on low-severity findings by citing a policy:

> ... and skips the adversarial pass on `suggestion`s, **which are deferred and never applied**.

The *skip* is real and verified — `workflows/review-pr.js:128` sets `suggestion: 0` verifiers, and the finding comes back `verdict: "unverified"` with no refuters rather than being dropped.

The *policy it cites* is written down nowhere:

```
$ git show ee81d14:skills/fleet/commands/review-and-fix.md | grep -ci "suggestion"
0
```

`review-and-fix.md` step 2 says only "Split explicitly into **apply now** / **defer**. A finding is deferred, never dropped." Nothing keys the defer bucket to `severity: "suggestion"`. So the design intent is asserted in two places (SKILL.md:309 and the `review-pr.js:122` comment, "A `suggestion` is deferred by review-and-fix, never auto-applied") and enforced in neither.

Why this matters: it is the load-bearing justification for spending zero verification budget on that severity. A reviewer that applies a `suggestion` is doing nothing the docs forbid, and that finding reached them unverified.

**Approach still open** — hence `needs-triage`, not `ready-for-agent`. Two defensible fixes and the choice is not settled:

1. Make review-and-fix.md step 2 state the rule (`severity: "suggestion"` → defer bucket), so the cited policy exists.
2. Or reword SKILL.md:309 + the review-pr.js:122 comment to describe the budget rationale without asserting a policy that is not written.

Found by the correctness specialist during the PR #107 fan-out; sub-threshold for that PR (pre-existing, outside its diff).

feigi:
`review-pr.js` and `run-team/SKILL.md` both justified spending zero verification budget on the lowest finding severity by citing a policy — that `suggestion`s are deferred and never auto-applied — that the reviewer-facing document did not contain.

Settled against the tree, not the ticket (its line numbers had already moved):

```
$ grep -ci "suggestion" skills/fleet/commands/review-and-fix.md
0
```

Step 2 said only that findings split into **apply now** / **defer** and that a deferred finding is never dropped. Nothing keyed the defer bucket to a severity, so a reviewer applying a `suggestion` broke no documented rule — and that finding had reached it unverified by design.

## The change

**`skills/fleet/commands/review-and-fix.md` step 2** now states the rule and its reason: the severity goes to defer, and each such finding arrives unverified, so applying one applies a claim nothing checked. A reader of that step alone can now act correctly without opening the workflow source.

**`skills/fleet/skills/run-team/SKILL.md`** cited the policy unsourced; it now names `review-and-fix.md` step 2 as its home — the same attribution `review-pr.js` already made. That follows the file's own norm of pointing at `review-and-fix.md` rather than restating it (SKILL.md:517, and the same `step N` form at :339).

`review-pr.js` needed no edit: its comment already attributed the policy to review-and-fix, and that citation now resolves.

## Verification

Each factual clause was settled by command before it was written:

- `review-pr.js:135` — `suggestion: 0`, the budget the rule justifies
- `review-pr.js:127-131` — the budget follows apply-probability *because* the severity is deferred, so the rule is the cause and the budget the effect, not the reverse
- `review-pr.js:269-272` — 0 verifiers yields `verdict: "unverified"` with `votes: []`, returned rather than dropped

`./agent-test skills/fleet/scripts/*.test.mjs` → 236 pass, 0 fail (non-zero count, so not a vacuous green).

## Scope

The verifier budget and the severity vocabulary are unchanged — `git diff` touches no `.js` file.

Closes #112


feigi:
Review complete (`394481a`). Two specialists ran — correctness and comment-accuracy — and converged independently on one Important finding, which is fixed here.

**Applied:** step 2 justified the defer rule with `review-pr.js` machinery, but a dispatched reviewer never reaches it (`:33` routes it away) and step 1 runs the `pr-review-toolkit` plugin, which has no adversarial pass and no `severity` field. The rule's only stated reason evaporated on the path most readers are on. Naming `unverified` was independently unsafe: that bucket is not suggestion-only (`review-pr.js:317-318` — "or one whose refuters all crashed"), so "every suggestion reaches you unverified" plus an unconditional defer rule inverts to "defer the unverified array", silently deferring a `critical` whose verification crashed. Now keyed to the band on every path, mechanism dropped: 38 words → 34, and `unverified` appears once in the file instead of twice with two meanings four lines apart.

**Deferred:** one finding, recorded on #194 (same axis, no duplicate ticket) — the file still does not state that under hand-dispatch *no* severity is adversarially verified.

**Verified clean, not re-derived:** `workflows/review-pr.js` untouched (`git diff --stat origin/main...HEAD -- workflows/review-pr.js` empty), so the verifier budget is unchanged by construction; `SKILL.md` is two token changes plus reflow; "step 2" is a durable citation (6 steps, step 2 = "Plan the actions", across every commit that ever touched the file; `SKILL.md:339` already cites "step 5"); all three policy sites agree. `./agent-test skills/fleet/scripts/*.test.mjs` → 236 pass, 0 fail.

CI `rebase-check` red is behind-count staleness (`check` green, behind 3 after #200 merged) — not rebasing; the merge bot owns that.

PR #224, opened 2026-08-10
fix(fleet): prune node_modules in the vendored-file walk instead of filtering it

### LINKED ISSUE #123: claim-ticket.sh: the node_modules filter does not prune traversal, so an unreadable dir inside node_modules refuses a suite it cannot affect
Deferred from PR #109 review.

`claim-ticket.sh:152` (post-#109) filters with `find "$arg" -type f -not -path '*/node_modules/*'`. `-not -path` is a **filter, not a `-prune`** — find still descends `node_modules`, so an unreadable directory in there still makes find exit non-zero and the readability guard above still refuses:

```
$ chmod 000 t/node_modules/vendor/deep
$ ./agent-test t
agent-test: cannot read every path under t
EXIT=1
```

Identical at base, so the *mechanism* is not a regression. What #109 changed is whether it is **defensible**: before the prune those files were part of the run, so refusing was correct. Now they are excluded, so the suite refuses over content that cannot affect its result. A false refusal — loud and safe, never a false green, but a stall.

## Two reviewers reached opposite verdicts

- Switching to `-prune` **defeats the readability guard** for vendored paths. Measured, both spellings:

  | form | unreadable inside `node_modules` | unreadable in our own code |
  |---|---|---|
  | `-not -path` (shipped) | exit **1** | exit 1 |
  | `-path '*/node_modules/*' -prune -o -type f -print` | exit **0** | exit 1 |
  | `-name node_modules -prune -o -type f -print` | exit **0** | exit 1 |

  So a `-prune` rewrite silently stops seeing unreadable directories under `node_modules`.

- The other reviewer calls exactly that **the fix**: refusing a whole suite over permissions on third-party code that was excluded anyway is a stall, not safety. It verified `-name node_modules -prune -o -type f -print` as a drop-in — genuine refusals on our own code still fire (`claim-ticket.test.mjs:154` stays green), 30/30 of the suite passes unmodified, and it is alias-independent where the path glob needs a leading component. It also measured `-prune` as faster: 68ms → 20ms over 8000 vendored files.

## The call

**Should an unreadable directory inside `node_modules` refuse, or proceed?** Both are defensible; only one ships.

The reviewer kept `-not -path` for #109 on scope grounds — #98 specified that exact flag, and the shipped form's worst case is a loud exit 1 while a mis-spelled `-prune` fails to a silent exit 0. That is a holding position, not a ruling on the merits. `claim-ticket.sh:153-154` now carries a comment explaining why it is not a `-prune`; if this ticket goes the other way, that comment must go with it.

Measured on Node v26.5.0, BSD find. Note `find` in an interactive zsh here is a function wrapping `bfs`; the runner's `#!/bin/sh` resolves `/usr/bin/find`. Pin the binary when re-measuring.

feigi:
## Summary

The emitted runner's `find` for a directory argument filtered vendored paths out of its *output* with `-not -path '*/node_modules/*'` while still descending into `node_modules`. An unreadable directory anywhere under there flipped find's exit status, and the readability guard refused the whole suite over content already excluded from the run — a loud but pointless stall. That is #123.

Fixed by pruning instead: `find "$arg/" -name node_modules -prune -o -type f -print`. An unreadable directory under `node_modules` is never descended, so it can no longer affect the run; a genuine unreadable directory in our own code still refuses as before.

**This overturns a decision that was documented in the code.** `claim-ticket.sh` carried a comment reading *"It is deliberately not a `-prune`: pruning skips the walk, and the readability check above would stop seeing an unreadable directory under `node_modules`"* — kept that way for #98/#109 on scope grounds. #123 settled the two-reviewer split in favour of pruning, on the merits: the unreadable content is excluded from the run either way, so its permissions cannot change the verdict. The comment goes with the decision.

## The prune alone was not enough

`-prune` fires only where the walk descends *through* a dirent named `node_modules`. Where the argument is itself at or under one, find starts inside it, prunes nothing, and prints every file. Node does not backstop it — it refuses an argv path only when the relative form begins with `node_modules/`, which these do not. Measured against the generated runner with the prune alone:

```
t/node_modules/pkg          -> status 0, "pass 1"   <- vendored test RAN, suite GREEN
./t/node_modules/pkg        -> status 0, "pass 1"
<abs>/node_modules/pkg      -> status 0, "pass 1"
<abs>/t/node_modules/pkg    -> status 0, "pass 1"
```

The `-not -path` filter refused all four. So the prune needed a second mechanism, not a replacement: a `case` on the argument, before the find, covering the disjoint set prune cannot reach.

The `case` also names the real reason. Falling through to the emptiness guard printed `no test files under <arg>` while the files sat there, deliberately excluded — a reader sent after a discovery bug that does not exist.

Its one side effect — refusing an absolute path whose ancestor is a `node_modules` — is what the previously-shipped filter already did. Measured identical to base on all four spellings, so no regression; filed as #230 since it is a real (pre-existing) defect.

## Test plan

- [x] An unreadable directory under `node_modules` no longer refuses the suite — mutation-checked: reverting the walk to the `-not -path` filter fails this test and nothing else
- [x] A vendored directory argument refuses in all six spellings, and the message names exclusion rather than absence — mutation-checked: removing the `case` guard fails this test and nothing else
- [x] Existing runner tests pass unmodified (symlink handling, own-code unreadable directory still refuses, nested `node_modules` vendored test excluded, `node_modules_old` lookalike not pruned)
- [x] `node --test 'skills/fleet/scripts/*.test.mjs'` — **294/294**

## Review deferrals

- #229 — `-type d` on the prune clause (refuted as behaviourally inert; the grep downstream kills the only delta)
- #230 — absolute path with a `node_modules` ancestor is refused while the relative spelling runs (pre-existing, needs a project-root notion the runner lacks)

Closes #123


feigi:
Review deferrals filed: #229 (`types` — `-type d` on the prune clause, refuted as behaviourally inert) and #230 (`silent-failure` — absolute path under a `node_modules` ancestor refused while the relative spelling runs, pre-existing). Applied findings are in 91b8e15.

PR #225, opened 2026-08-10
fix(fleet): ledger check payload names its verdict, hits absent on failure

### LINKED ISSUE #152: ledger.mjs check: exit 0 cannot distinguish a clean tracker from one that was never read
Deferred from PR #150 review. **Two specialists reached opposite conclusions; this needs a maintainer decision, which is why it is `needs-triage`.**

## The finding

`skills/fleet/scripts/ledger.mjs:343` exits 0 for two epistemically different states:

- the tracker was searched and came back clean (`tracker.ok: true`, `hits: []`)
- the tracker was never read at all (`tracker.ok: false`) — `gh` auth failure, `gh` absent, timeout, unparseable output, or a subject with no distinctive terms

Both print to stderr and the second prints a loud `WARNING — TRACKER NOT CHECKED`, so it is **not** a bare "safe to file". But `$?` alone cannot tell them apart.

## The two positions

**Mint exit 4** (silent-failure-hunter). The repo's fail-closed convention is a stated rule, not a habit — and `skills/fleet/scripts/candidates.mjs:64-66` is the identical construct (same `execFileSync` + `JSON.parse`, same `String(e.stderr || e.message)` idiom) and dies:

> Fail closed. A failed query and an empty queue are different facts, and only one of them means "there is no work".

Not exit 2, because `die()` owns 2 and prints no JSON, while this path has a real ledger answer on stdout.

**Do not split** (type-design-analyzer). Any non-zero blocks the natural `ledger.mjs check "$s" && gh issue create`, turning every offline run into a hard stop on filing — which contradicts the behaviour PR #150 deliberately tested at `ledger.test.mjs:209` (`offline must not block filing`). Its alternative puts the distinction in the payload instead of the exit code:

```js
verdict: "already-filed" (1) | "tracker-hit" (3) | "clean" (0) | "unverified" (0)
process.exit(verdict === "tracker-hit" ? 3 : 0)
```

## Related, same decision

`tracker.hits` is `[]` on **both** failure arms, so `tracker.hits.length === 0` reads as "clean" when it means "unknown" — the #145 fail-open class one layer up. No test asserts `hits` on a failure arm. Proposed fix: `hits: null`, so `.length` throws loudly rather than lying quietly. This only makes sense settled together with the above.

## Why deferred

The maintainer's ratifying comment on #145 already ruled on this path — *"Degrade to the ledger-only answer and say so on stderr"* — and PR #150 implements exactly that, so current behaviour is spec-compliant, not a defect. Changing it is a second exit-contract change on the final PR of a run, after the contract was already changed once with explicit permission. That is a maintainer's call.

Note there are currently **zero non-test consumers of the stdout JSON** and no script caller of `check` at all (`board.mjs:227` uses `read`), so any payload change here is breaking in principle and free in practice.

feigi:
## Summary

Closes #152.

`ledger.mjs check` had two failure arms — an unreachable tracker (gh
absent, auth failure, timeout, unparseable/wrong-shape output) and a
subject with no distinctive terms to search — that both constructed
`tracker.hits: []`. A consumer testing `.length` reads that as "clean"
when the truth is "never searched," the #145 fail-open class one layer
up.

- `hits` is now **absent** (not `[]`) on both failure arms, so `.length`
  throws loudly instead of lying quietly. A searched-and-clean tracker
  still reports `hits: []`, now distinguishable from an unread one.
- The top-level payload now carries an explicit `verdict`:
  `already-filed | tracker-hit | clean | unverified`, so a consumer no
  longer has to reconstruct the outcome from `tracker.ok` plus
  `tracker.hits`.
- **Exit codes are unchanged**, per triage's ruling: the payload
  carries the distinction, the exit code does not. Minting a non-zero
  code for "unverified" would break the documented
  `check "$s" && gh issue create` flow and contradict the suite's own
  offline-must-not-block-filing test. An unreachable tracker still
  exits 0.

There are currently zero non-test consumers of this stdout JSON and no
script caller of `check` at all (`board.mjs` uses `read`), so this
payload shape change is breaking in principle and free in practice.

## Test plan

- [x] `env -u NODE_TEST_CONTEXT -u NODE_TEST_WORKER_ID node --test 'skills/fleet/scripts/*.test.mjs'` — 292/292 pass
- [x] New/updated assertions cover the hit list's value on **both**
      failure arms (the coverage gap #152 names) plus the new
      `verdict` field across already-filed, tracker-hit, clean, and
      unverified outcomes


feigi:
## Review follow-up: applied / deferred split

Zero findings survived adversarial verification in the `critical`/`important` bands. The 18 `suggestion`-band findings arrive unverified by policy, so each was scope-split and the in-scope ones each got one refuter dispatched biased to refuse.

**Applied** (`e5ec1ac`, tests 292 pass / 0 fail):

- `ledger.mjs:362` — exit code now derives from `verdict` instead of re-deriving from `tracker.hits?.length`. Refuter reproduced real drift under a plausible edit to the verdict ternary: stdout `"verdict":"clean"` with the process exiting 3.
- `ledger.mjs:336` — TRACKER HIT loop moved inside the branch that already knows `hits` is present; `?? []` guard and its comment deleted. stderr/stdout verified byte-identical. Ordering is now pinned by a test (mutation-checked three ways).
- `run-team/SKILL.md` — documents the `verdict` field. With zero script consumers of `check`'s stdout, documentation is the delivery path for this ticket.
- `ledger.test.mjs` — the stderr-capture comment trimmed 13 lines → 4.

**Deferred and filed:**

| issue | finding | label |
|---|---|---|
| #231 | `verdict: "clean"` emitted when the ledger file was never read | `needs-triage` |
| #232 | shape guard validates only `h.number`; a partial gh row prints `TRACKER HIT — #114 (undefined) — undefined` | `needs-triage` |
| #233 | REFUTED — jq cannot distinguish absent from empty `tracker.hits` | `wontfix`, closed |
| #234 | REFUTED — the five `hits, undefined` assertions are not near-duplicates | `wontfix`, closed |
| #235 | the discriminant is still written twice (stderr chain + verdict ternary); blocked by #231 | `needs-triage` |

Nine of the 18 were confirmations that existing code is correct — no fix, no issue.

`ledger.mjs:351` is byte-identical to what the maintainer ruled on.

PR #226, opened 2026-08-10
fix(ci): syntax-check hooks/statusline.js and workflows/review-pr.js

### LINKED ISSUE #158: ci.yml: node --check is scoped to .mjs, so review-pr.js and statusline.js are never parsed
Deferred from PR #157 review.

`.github/workflows/ci.yml` — the `Syntax-check scripts` step globs `*.mjs` only. Two tracked `.js` files get no syntax check at all:

- `hooks/statusline.js`
- `workflows/review-pr.js`

`workflows/review-pr.js` is exactly the case the step's own comment describes — "a script no test imports". `review-pr-testcmd.test.mjs:20` only ever `readFileSync`s it and regex-matches the text, so a syntax error in it surfaces at workflow-execution time and nowhere earlier.

## Why this is not just "widen the glob"

Measured on v26.5.0:

```
$ node --check hooks/statusline.js        # OK
$ node --check workflows/review-pr.js
SyntaxError: Unexpected token 'export'
```

`review-pr.js` is ESM in a `.js` file and the repo has **no package.json**, so `node --check` parses it under CJS rules and dies on the first `export`. Adding `'*.js'` to the glob turns the step red.

(Note: one reviewer in the #157 fan-out asserted "`node --check` accepts both files fine". That is wrong — the command above is the refutation.)

## Why deferred

The defect is confirmed; the approach is not. Three options with different blast radii:

1. rename `workflows/review-pr.js` → `.mjs` — does the Workflow loader resolve it by exact name?
2. add a minimal `package.json` with `"type": "module"` — changes how every `.js` in the repo is interpreted, and the repo deliberately ships without one (`ci.yml`: "There is no build")
3. keep the glob and check `.js` separately with an ESM-aware invocation

Picking one is a maintainer call. #157 records the constraint in-file so the gap is not silently re-closed wrong.

feigi:
## Summary

`Syntax-check scripts` in `.github/workflows/ci.yml` only globbed `*.mjs`. Two tracked `.js` files got no syntax check at all — including `workflows/review-pr.js`, which is exactly the case the step's own comment describes: "a script no test imports." Its only prior coverage read it as text and regex-matched, so a syntax error surfaced only at workflow-execution time.

Widening the glob to `*.js` is not a fix: `review-pr.js` is a Workflow-harness script — the harness reads `export const meta = …` off it without running it, then wraps the *rest* of the body in an async function it executes, so the file also has top-level `await`/`return` (legal only inside that wrapper). That combination parses as neither a plain CJS script (the `export`) nor an ES module (top-level `return` is always illegal there) — `node --check` rejects it under both interpretations.

Per triage's ruling on #158 (option 3): added a second CI step that checks each tracked `.js` file with an invocation matching what it actually is:
- `hooks/statusline.js` — ordinary CommonJS script → `node --check`
- `workflows/*.js` — Workflow-harness script → `new AsyncFunction(...)` over the source with the top-level `export` stripped, the same technique `skills/fleet/scripts/review-pr-reads.test.mjs:439` already uses for exactly this, reused rather than reinvented.

No file renamed, no `package.json` added — both ruled out on blast radius by triage. The workflow comment explains why the two invocations differ, so the next person to widen a glob doesn't re-close this wrong.

## Verification

Demonstrated both directions locally (not assumed):
- Appended `const x = ;` to `workflows/review-pr.js` → the `AsyncFunction` check fails with `SyntaxError: Unexpected token ';'` (exit 1)
- Appended `const x = ;` to `hooks/statusline.js` → `node --check` fails the same way (exit 1)
- Reverted both, re-ran clean → both invocations pass

## Test plan

- [x] `node --check` on `hooks/statusline.js` passes on the clean tree
- [x] `AsyncFunction` parse on `workflows/review-pr.js` passes on the clean tree
- [x] Deliberate syntax error in `workflows/review-pr.js` fails the check (demonstrated, reverted)
- [x] Deliberate syntax error in `hooks/statusline.js` fails the check (demonstrated, reverted)
- [x] Full suite: `env -u NODE_TEST_CONTEXT -u NODE_TEST_WORKER_ID node --test 'skills/fleet/scripts/*.test.mjs'` → 292 passing, 0 failing (matches baseline)

Closes #158


feigi:
## Review applied — `b7c828c`

Four `suggestion` findings, all arriving unchecked (the workflow budgets that band 0 refuters). Each was scope-split, then the in-scope ones went to refuters instructed to refute. **Three of four refuted; the outcome is comments only, no behaviour change.**

| # | Finding | Outcome |
|---|---|---|
| 1 | `workflows/*` branch duplicates `review-pr-reads.test.mjs:439` | **Refuted** — applied as a comment |
| 2 | `/^export /m` lacks `g` | **Refuted** — `/m` is correct; documented |
| 3 | `for f in $(git ls-files …)` word-splits | **Deferred** → #237 |
| 4 | ESM helper under `workflows/` fails the step | **Refuted** — nothing to do |

**Finding 1.** I initially ruled for deleting the branch and was wrong: I checked that the two AsyncFunction invocations were identical and stopped before checking they were pointed at the same *set* of files. The step is glob-driven; the test pins `review-pr.js` by name. Measured — a second `workflows/*.js` carrying `const = ;` gives the step exit 1 while the suite stays 292/292 green — and `docs/specs/2026-07-23-fleet-plugin-design.md:151` plans `merge-wave.js`. The real defect was the comment never stating that justification, so the next reader reconstructs the redundancy and deletes the branch. That is what this commit fixes.

**Finding 2.** Reframed and refuted on the harness contract: exactly one top-level `export` (`meta`, first in the file), and nothing consumes a second. So `/m` is enforcement, not an oversight — across 12 contract-violating arrangements `/m` fails all 12 while `/gm` passes 7, i.e. `/gm` would green a file the harness rejects at load time. `review-pr-reads.test.mjs:452` keeps `/m` too; the two now agree deliberately rather than by accident.

**Finding 4.** Static `import` does throw, but the sandbox has no filesystem or Node API access and nothing in `workflows/` imports anything, so the red is correct — it refuses a file the harness cannot load. The proposed comment would document a false affordance. `workflows/*.mjs` is already covered by the `.mjs` step at `ci.yml:86`.

### Acceptance criterion re-proved

Against the step extracted from the YAML itself, not a hand-retyped copy:

```
clean                                   exit 0
const = ;  in workflows/review-pr.js    exit 1
const = ;  in hooks/statusline.js       exit 1
second top-level export                 exit 1
```

Gate: `tests 292 / pass 292 / fail 0`.

### Correction to #158's premise

#158 says a syntax error in `review-pr.js` "surfaces at workflow-execution time and nowhere earlier". That is now stale — the AsyncFunction parse test landed later, in #223 (`52fbf29`). It is why the redundancy exists at all, and it is not this PR's doing.

### Coverage caveat

**This review ran one dimension, `correctness`.** The single-`.yml` diff trimmed the fan-out, so `silent-failure`, `tests`, `comments`, `types` and `simplify` never ran. Read their silence as **unmeasured, not clean** — thin for a change to CI's own gating logic. Note the sharpest finding here was itself a *comment* defect, surfaced by `correctness` because the `comments` dimension did not run.

PR #227, opened 2026-08-10
fix(fleet): escape tab/CR/LF short forms in jstr/jarr, preserve DEL, flag rewrites

### LINKED ISSUE #146: no-undo-audit.sh: jstr/jarr replace control bytes with a space, so conflicts[] can name a file that does not exist
Deferred from PR #131 review. Pre-existing in `release-ticket.sh`, inherited by `no-undo-audit.sh` in that PR, and #119 proposes spreading the same helper to five more scripts — so the decision here sets the convention for all of them.

`jstr` and `jarr` do not escape control bytes, they **replace** them:

```
skills/fleet/scripts/no-undo-audit.sh:21  tr '\001-\037\177' ' '
skills/fleet/scripts/no-undo-audit.sh:24  tr '\001-\011\013-\037\177' ' '
```

Measured at `c17d4b6`, a real conflicting file whose name holds a TAB (shown here through `git ls-files`, which C-quotes it):

```
on disk:  "has\tab.txt"
payload:  {"conflicts":["has ab.txt"],"atRisk":["a3adba2 MAIN COMMIT AT RISK"]}   exit 0
```

`atRisk` is **correct** — the pathspec is built from the raw shell variable, not the escaped one — so this is not a repeat of the false safe #106 fixed. What breaks is fidelity: `conflicts[]` names `has ab.txt`, which exists nowhere on disk.

The consumer is what makes it worth a ticket. `skills/fleet/commands/run-merge-bot.md` step 5 tells the operator to run `git diff origin/main...HEAD -- <conflicting files>` with exactly those paths, and a rewritten path matches nothing — so the step that exists to prove nothing was undone would silently prove it about no files at all.

`\177` (DEL) is scrubbed gratuitously: it is not a C0 byte and JSON does not forbid it unescaped (`JSON.parse` accepts a raw DEL inside a string). CR and TAB are likewise legal once escaped as `\r` and `\t`.

## Why `needs-triage`

The defect is verified; the approach is a convention decision rather than a lookup:

1. **Escape properly** — `\t`, `\r`, `\n`, and `\uXXXX` for the rest. Faithful, but per-byte hex in POSIX `sh` is a real step up in machinery from a one-line `tr`, and it has to land in every copy of the helper at once.
2. **Keep replacing, and say so** — the payload stops claiming to carry a usable path, and any consumer that acts on one (the runbook above) needs a different source for it.
3. **Split the difference** — escape the three bytes with short forms, keep replacing the rest.

Whoever takes this: decide across `release-ticket.sh`, `inflight.sh` and `no-undo-audit.sh` together, since the helper is copied byte-identically between them and #119 wants it in five more. Drive each shape into existence rather than reasoning about it — a TAB, a CR and a DEL behave differently in the three fields.

Same fidelity class as #122 (`inflight.sh` truncates a worktree path at a TAB), different mechanism — that one loses the bytes two stages upstream of the payload, where `jstr` cannot reach. Related: #119.

feigi:
## Summary

`jstr`/`jarr` replaced every C0 byte (and DEL) with a space instead of escaping it where JSON allows escaping, so a worktree path, conflicting-file path, or commit subject containing a tab, CR or LF came back as a different string than the real one — `no-undo-audit.sh`'s `conflicts[]` could name a file that exists nowhere on disk, and the runbook step that pastes those paths into `git diff` would silently prove nothing.

Per triage's ruling on #146 (option 3, plus a disclosure signal):

- Escape tab, CR and LF with their JSON short forms — backslash rule first, same ordering constraint the existing quote rule already has.
- Stop scrubbing `\177` (DEL) — it is not a C0 byte and JSON permits it unescaped.
- Keep replacing every other C0 byte (no short form for it).
- Disclose via a sibling `*Rewritten` field whenever a value was actually replaced, so a consumer can refuse to treat it as the real path/ref.

Applied identically to the three byte-identical copies of `jstr` (`no-undo-audit.sh`, `release-ticket.sh`, `inflight.sh`) and to `jarr`, which exists only in `no-undo-audit.sh` — per #119's drift concern, all four copies remain byte-identical after the change (verified by diff).

`skills/fleet/commands/run-merge-bot.md` step 5 now tells the operator to skip the `git diff -- <path>` step for any path the audit flags `*Rewritten: true`, since a rewritten path won't match a real pathspec.

## Test plan

- [x] `node --test 'skills/fleet/scripts/*.test.mjs'` — 303/303 green
- [x] New tests drive a real tab, CR and DEL byte into worktree paths, conflicting paths, and commit subjects across all three scripts; every assertion parses the script's actual stdout via `JSON.parse`
- [x] A companion no-short-form byte (`\x02`) pins the opposite case: replaced and `*Rewritten: true`
- [x] Mutation-checked: reverting either `jstr` or `jarr` to the old scrub-everything behavior turns the corresponding new tests red (not vacuous)
- [x] Full suite parses every emitted payload as JSON

Closes #146


feigi:
Review outcome applied at `70fde20`. Suite: **`tests 305 / pass 305 / fail 0`** (baseline on this branch was 303), `node --test 'skills/fleet/scripts/*.test.mjs'`, no `--no-verify` (repo has no pre-commit hook).

**Applied** — the 5 double-refuted survivors, each mutation-checked both ways:
- `\b` and `\f` **do** have RFC 8259 short forms. The comment in all three copies said otherwise and the code implemented the comment, scrubbing `\010`/`\014` to a space and flagging the value rewritten. Now escaped; `\013` (VT) stays scrubbed, since JSON has no `\v`. The rules match a literal byte via `printf` — `\b` in a BRE is a zero-width word boundary to GNU sed and a literal `b` to BSD sed (measured, GNU sed 4.9 / macOS sed), so that spelling corrupts rather than no-ops.
- The multi-line slurp and the LF rule are now pinned, in the one script where a newline is reachable.
- `jrewritten`'s `X` sentinel was inert and its comment named a trap a trailing `\r` cannot spring — both removed.
- Every `$(jstr …)` / `$(jrewritten …)` moved out of printf's argument list, where the `|| die` structurally could not reach it.

**Deferred** — #242 (nine refuted findings, recorded with their refutations, incl. one whose refutation measurement contradicts), #243 (`release-ticket.sh` exits 2 through awk's own error on a newline slug), and a comment on #119 (the three-copy duplication and the "worth 160 lines?" question).

Two measured corrections to the review itself:
1. The proposed pin for the newline half — `sh release-ticket.sh 9 "$(printf 'a\nb')" fix` — cannot work: the branch reaches `awk -v b="refs/heads/$branch"` first and awk refuses a newline in a `-v` value (rc 2, before any receipt). Pinned through `no-undo-audit.sh`'s `wt` instead, which is argv `$1` and therefore a filesystem path.
2. The slurp in `release-ticket.sh` and `inflight.sh` is **unreachable**, not merely unpinned — deleting it from either, or deleting `s/\n/\\n/g` from `release-ticket.sh`, leaves the full suite green. Recorded on #119.

Verification: 20 mutations on fresh snapshots — every new pin fails only its own script's tests, and 5 controls (the two unreachable slurps, an unreachable LF rule, a semantics-preserving rule reorder, a variable rename) stay green. Also run under Linux/GNU sed 4.9 in Docker: all three helpers produce byte-identical output, and the container's 2 failures are pre-existing (root bypasses the directory permissions those tests rely on — the unmodified base fails the same two).

feigi:
Follow-up at `2ea3bfc`. Suite: **`tests 306 / pass 306 / fail 0 / skipped 0`**. All 13 `unverified` findings are now settled; nothing is left unexamined.

## A hazard this branch cannot pin, and CI cannot see

**The `$!` guard on `sed`'s `N` is undiscriminable on GNU sed.** Removing it — `-e '$!N'` → `-e 'N'` — fails 3 tests on BSD sed, because BSD `N` at EOF discards the pattern space and `jstr` returns empty for any single-line value. GNU sed prints the pattern space there instead, so the two spellings are **identical** on Linux. CI runs Linux.

The consequence is worth stating plainly: **a future edit dropping that guard ships green through CI and breaks only on macOS.** No test can close this — the behaviour being guarded does not exist on the platform CI runs. The comment above each `jstr` says why the guard is there; that comment is the only defence, which is precisely why it must not be trimmed as redundant.

## Also unpinnable, for the opposite reason

The multi-line slurp in `release-ticket.sh` and `inflight.sh` is **unreachable**, not unpinned — `release-ticket.sh` hits `awk -v b="refs/heads/$branch"` before any receipt and awk refuses a newline in a `-v` value (#243); `inflight.sh` builds every evidence string from line-oriented git output. Deleting the slurp from either, or `s/\n/\\n/g` from `release-ticket.sh`, leaves the suite green. Pinned in `no-undo-audit.sh` instead, whose `wt` is argv `$1`. Do not bolt a test onto the other two copies to "close the gap" — there is no gap, and any such test would assert on an input the scripts cannot receive.

## Settled since the last comment

- **Design spec** — `no-undo-audit.sh`'s `Out` cell listed the pre-#146 payload. Fixed and pinned: the test derives the key set from a real run, word-boundary matched so `worktree` cannot satisfy `worktreeRewritten`. Mutation-checked three ways. Heads-up left on #114, which owns a different column of the same table.
- **`jarr_rewritten` cardinality** — already pinned as code, measured, no change. Three discriminating mutations.
- **Twelve repeated subshells** — already collapsed in `70fde20`. The suggested lazily-evaluated function is unnecessary: `wt` is single-assignment at `:77`, not mid-script.
- **The `X` sentinel and the `orig=$(printf '%sX' "$1")` fork** — one defect, gone together in `70fde20`. Note for anyone revisiting: with the sentinel removed, both operands must still pass through the same `$()`, because `$()` strips a trailing `\012` from `raw` and would not from a bare `"$1"` — a value ending in a newline would report `true`.
- **DEL** — independently confirmed against the real consumer: Node's `JSON.parse` accepts a literal `0x7F` inside a string, per RFC 8259's U+0000–U+001F escape range.
- **Four suggestions refuted** by dispatched refuters (dead `*Rewritten` fields, `printf false || printf true`, `raw`/`orig` globals, unwired flags) — all recorded on #242 with their evidence, including a place where two refuters disagree and the demonstrated result wins over the one derived from a rule.

Issue #169, opened 2026-07-31
fleet scripts: arg() returns undefined for a trailing flag in four more scripts — the defect PR #167 fixed in candidates.mjs only


feigi:
Found by `impl-61` while fixing #61; controller verified all four sites independently before filing.

PR #167 fixes `arg()` in `candidates.mjs` — `process.argv[i + 1]` is `undefined` for a flag given no value, and every caller reads that falsy result as "the flag was absent". The fix landed **inside `arg()`**, which covers that one file.

**Four other scripts carry their own copy with the identical defect.** `impl-61` reported three; the fourth is `board.mjs`, which spells it as an arrow function and so does not match a `function arg` grep.

```
$ grep -rn "const arg\|function arg" skills/fleet/scripts/*.mjs
skills/fleet/scripts/candidates.mjs:23:function arg(name) {      <- fixed by PR #167
skills/fleet/scripts/diff-stats.mjs:94:function arg(name) {
skills/fleet/scripts/ci-state.mjs:24:function arg(name) {
skills/fleet/scripts/pr-overlap.mjs:22:function arg(name) {
skills/fleet/scripts/board.mjs:17:const arg = (n) => { const i = process.argv.indexOf(`--${n}`); return i === -1 ? null : process.argv[i + 1]; };
```

The three `function arg` bodies are byte-identical to each other and to the pre-#167 `candidates.mjs`:

```js
function arg(name) {
  const i = process.argv.indexOf(`--${name}`);
  return i === -1 ? null : process.argv[i + 1];
}
```

`board.mjs:17` is the same expression inlined.

## Why it is not the same severity everywhere

In `candidates.mjs` the defect was reachable as a **silent widening** — a trailing `--require-label` ran the unfiltered query at exit 0, reaching the exact behaviour `--allow-fallback` gates. That is what made #61 a bug rather than a nit.

The consequence in each of these four has **not** been established, and that is the work here — not a mechanical copy of the #167 patch. Each needs its own answer to "what does a caller do with `undefined`, and is the wrong answer silent?". `ci-state.mjs` is the one to look at first: the fleet reads its verdict to decide whether a PR is genuinely green.

## Not simply "extract a shared helper"

[...]

PR #228, opened 2026-08-10
fix(fleet): phase 0's rule named --label, and candidates.mjs ignored it

### LINKED ISSUE #173: run-team/SKILL.md:60 prose names --label where the flag is --require-label — and candidates.mjs silently ignores unknown flags, so copying it widens at exit 0
Found by the `code-simplifier` specialist reviewing PR #167; controller reproduced both halves before filing.

## The typo

`skills/fleet/skills/run-team/SKILL.md:59-61`, phase 0 step 1:

```
1. **Candidate scan** — `~/.claude/skills/fleet/scripts/candidates.mjs
   --require-label ready-for-agent`. **`--label ready-for-agent` mandatory, no
   fallback** — do NOT pass `--allow-fallback`.
```

The command is right. **The emphasised rule one line below it names a flag that does not exist.** `candidates.mjs` accepts exactly:

```
$ git show origin/main:skills/fleet/scripts/candidates.mjs | grep -n "arg(\"\|has(\""
29:const requireLabel = arg("require-label");
30:const allowFallback = has("allow-fallback");
37:const limit = Number(arg("limit") ?? 500);
```

No `--label`.

## Why it is not merely cosmetic

`candidates.mjs` **silently ignores unknown flags**. So a controller that follows the bolded rule rather than the code block above it gets no error — it gets the unfiltered query. Measured by `code-simplifier` against PR #167's head with a label-honouring `gh` stub:

> `--label ready-for-agent` … ships a `ready-for-human` ticket at **exit 0**.

That is the exact widening this very sentence exists to prevent, and the sentence is the thing that causes it. The bolded, emphasised text is also the more memorable of the two forms — an agent skimming for the rule finds `--label` before it finds the command.

It compounds a defect already tracked: `#169` covers `arg()` returning `undefined` for a trailing flag across four scripts. **Unknown-flag rejection is a different mechanism** (nothing validates the flag *name*), so it is not a duplicate of #169, though `parseArgs({strict: true})` would close both.

## Two independent fixes, neither depending on the other

1. **The prose** — one word, zero code: `--label` → `--require-label` at `:60`. Note open PR #168 also edits this file, at `:167-191`; `:60` is clear of it.
2. **The script** — reject unknown flags instead of ignoring them. Larger, and it is a design choice (hand-rolled name check vs adopting `parseArgs`), which is why this is `needs-triage` rather than `ready-for-agent`.

Fix 1 alone removes the reachable harm from this repo's own instructions. Fix 2 is what stops the next such typo being silent.

## Scope searched

`git grep -n "\-\-label " skills/ docs/` over tracked files at `origin/main`, plus every `arg(`/`has(` site in `candidates.mjs`. **Not searched:** whether other fleet scripts have the same prose-vs-flag drift in their own callers — worth a sweep when fixing.

feigi:
Closes #173

## The two halves

**Prose.** `skills/fleet/skills/run-team/SKILL.md` phase 0 step 1 stated the mandatory rule as `` `--label ready-for-agent` mandatory, no fallback ``, one line under a command spelling it `--require-label`. The bolded form is the more memorable of the two. It now names `--require-label`. `docs/fleet-rule-manifest.txt` carried the same string and moves with it, so the manifest still resolves.

**Script.** `candidates.mjs` ignored unrecognised flags, so following the bolded rule ran the UNFILTERED query at exit 0 — the widening onto `ready-for-human` that `--allow-fallback` gates. `parseArgs({ strict: true })` now refuses an unknown flag by name, through the existing `die()` (exit 2, before any `gh` call):

    candidates: Unknown option '--label' — accepted: --require-label, --allow-fallback, --limit

The accepted set is one object literal; the refusal message derives its list from it.

## What is deliberately untouched

Only `ERR_PARSE_ARGS_UNKNOWN_OPTION` is refused here. `parseArgs` also rejects a missing value, and those spellings fall through to `arg()`, whose four refusals are #169's rule — a trailing flag, a blank value, a `--flag` as the value, and the `--flag=value` form. All four still refuse at exit 2 with their own wording. #169 stays a fix-in-place, sequenced after this.

Accepted flags are unchanged, so both live call sites — run-team's `--require-label ready-for-agent` and next-ticket's `--require-label ready-for-agent --allow-fallback` — are unaffected, and both are exercised by the suite.

## Tests

292 → 294. Both new pins were mutation-tested in both directions:

- the unknown-flag refusal: neutering the guard reds it (`fail 1`)
- the corrected instruction: pinned against `run-team/SKILL.md` itself, sliced to the step-1 bullet, positive match paired with a `doesNotMatch` for `--label`. Restoring `--label` reds it; rewording the sentence around it ("is strictly mandatory here, and there is no fallback", "An empty result means there is no work") stays green.

## Sweep — other fleet scripts' callers

Searched: all 16 tracked scripts under `skills/fleet/scripts/` (excluding `*.test.mjs`), against all 46 tracked `.md`/`.txt`/`.mjs`/`.js`/`.sh`/`.yml` files outside that directory. Accepted sets derived from each script's `arg()`/`has()`/`parseArgs` options/`case` labels; two passes, same-line and a 6-line window (the same-line pass alone would have missed #173, whose wrong flag sits one line below the script name); every hit attributed by hand.

Two drifts found, **both in dated design specs, none in a live caller**:

- `docs/specs/2026-07-23-fleet-plugin-design.md:162` gives `ci-state.mjs` as `--pr --branch`. It has no `--branch` — that flag is what it passes to `gh` internally (`ci-state.mjs:134`). `run-merge-bot.md:86` already states "No `--branch` flag".
- `docs/specs/2026-07-24-fleet-cockpit-design.md:42,74,175` write `build --once`. `board.mjs` reads no `--once`; its usage line lists `[--ledger] [--port] [--interval] [--open]`, and `build` is already the one-shot.

Neither reaches #173's harm: `ci-state.mjs` refuses without `--pr`, `board.mjs build` already does what `--once` implies, and neither can widen a query. Left as-is — they are historical design records, and the live skills, commands and workflows are clean.


feigi:
## Review applied — deferrals filed

Pushed `56db4a1`. Suite: **301 tests, 301 pass, 0 fail, 0 skipped** (294/294/0 on this branch before).

### Applied

**The three criticals were one root cause, fixed once.** The catch this branch added refused only `ERR_PARSE_ARGS_UNKNOWN_OPTION` and swallowed every other `parseArgs` error. Since `parseArgs` throws on the *first* offending argument, any argument raising a dropped code masked every unknown flag behind it. Reproduced at `e840bc5` against a stub `gh` returning one `ready-for-human` row:

| argv | before | after |
|---|---|---|
| `ready-for-agent` | UNFILTERED query, exit 0 | exit 2 |
| `junk --label ready-for-agent` | UNFILTERED query, exit 0 | exit 2 |
| `--allow-fallback=true --label ready-for-agent` | UNFILTERED query, exit 0 | exit 2 |
| `--label ready-for-agent` | exit 2 | exit 2 |
| `--require-label ready-for-agent --bogus` | exit 2 | exit 2 |

The fix deletes the code test and moves the `try`/`catch` **below** the `arg()` value guards, so #169's wording still wins where both would refuse. `--allow-fallback=true` is closed too: it previously exited **1** ("queue empty, query fine") for a run that never applied the flag, because `has()` compares exactly and never saw the `=` form.

The comment that shipped with the guard is replaced. It claimed `parseArgs` "also rejects a missing value, and those spellings are left to `arg()` below" — false, since positionals and `--boolflag=value` were dropped entirely rather than deferred. Every clause of the replacement was checked by running it. One correction to the review's own framing: **position was not what decided the old guard** — `--require-label x --bogus` refused correctly with the bad flag last. What mattered was only whether something earlier threw a dropped code.

`strict: true` dropped — it is `parseArgs`' default, measured identical across 48 comparisons including with `allowPositionals`/`tokens`/`allowNegative` added.

**Seven pins added, each mutation-tested both ways.** Restoring the code test fails exactly the five new `parseArgs` pins and leaves the pre-existing bad-flag-first test green — the measure of how vacuous that test was against this defect. Hoisting the block back above the value guards fails only the wording pin. A cosmetic reword of the accepted-flag list fails none. The `--allow-fallback=true` pin carries its own control so it cannot be satisfied by a blanket refusal of `--allow-fallback`, which `next-ticket/SKILL.md:15` passes legitimately.

### Deferred

- **#240** (`ready-for-agent`) — `docs/specs/2026-07-22-run-team-agent-fleet-design.md:201` still names `--label ready-for-agent`. Out of `docs/specs/`-scope for this PR. The issue records that the string is hand-duplicated across three files, and flags that `docs/specs/2026-07-24-fleet-cockpit-design.md:72` must **not** be swept — there `--label` is a correct `gh` flag.
- **#241** (`needs-triage`) — `docs/fleet-rule-manifest.txt` is a hand-synced grep checklist nothing reads; it has now drifted and been hand-patched twice. Approach is genuinely open (pin the 33 patterns vs delete a spent restructure gate), so triage rather than agent-ready.

feigi:
## Addendum — final head `8110005`, and two measurement notes worth carrying

Suite unchanged at **301 tests / 301 pass / 0 fail / 0 skipped**.

Two more commits since the last comment, both small: `56db4a1` drops `strict: true` (parseArgs' own default — 48 comparisons, 0 differences, and it does not flip when `allowPositionals`/`tokens`/`allowNegative` are added), and `8110005` inlines `phase0Step1`, a wrapper defined once and called once that re-implemented an anchor-slice four sibling test files write inline. Both mutation legs re-run after inlining: breaking the pinned rule still fails the test, and moving the anchor still reports `run-team phase 0 step 1 moved — update this test` rather than a slice error.

### The mutation result that matters most

Restoring the `e.code === "ERR_PARSE_ARGS_UNKNOWN_OPTION"` test fails **exactly the five new parseArgs pins** and leaves the branch's pre-existing bad-flag-first test **green**:

```
✖ a bare positional refuses before any query …
✖ a boolean flag handed a =value refuses before any query …
✖ an unknown flag behind a positional refuses before any query …
✖ an unknown flag behind a =value boolean refuses before any query …
✖ --allow-fallback=true never reaches the fallback …
   tests 31 / pass 26 / fail 5
```

That green is the measurement of how vacuous the original test was against this defect: it asserted the guard worked, and the guard did not work, and it passed anyway. It only ever placed the bad flag first.

### Two measurement hazards this review surfaced

**1. A `git archive` snapshot reports a skip that does not exist in the real tree.** The baseline was circulated as `294 / 293 pass / 1 skipped`. In the worktree it is **294 / 294 pass / 0 skipped**, measured twice, gated and ungated. The cause:

```
skills/fleet/scripts/review-pr-testcmd.test.mjs:51:  return t.skip("not a git checkout — cannot cut an archive here");
skills/fleet/scripts/review-pr-testcmd.test.mjs:54:  return t.skip("not this repo's toplevel — cannot cut an archive here");
```

Review specialists work from detached copies, so that test skips itself there. A fix-applier taking the number on trust would go hunting for a test that never went missing. Same family as the `derive_workspace_id` hazard already documented in `review-and-fix.md`.

**2. zsh does not word-split unquoted expansions.** A repro loop of the form `for a in "--flag x"; do node script.mjs $a; done` passes `--flag x` as **one** argv entry under zsh, which silently converts every multi-arg case into a single-token case — and for this PR that made a correct invocation look like an unknown-flag refusal. Everything here was re-measured through a node driver spawning explicit argv arrays.

### One correction to the review's own framing

The finding was written as "the guard works only when the bad flag happens to be first". That is false: `--require-label x --bogus` refused correctly at `e840bc5` with the bad flag **last**. Position never decided it — only whether something *earlier* in argv threw a code the catch dropped, since parseArgs stops at the first offender. The comment in the code is written to the measured rule, not the stated one.

### Deferral update

`docs/specs/2026-07-23-fleet-plugin-design.md:164` (the table row claiming an empty result exits 0, where the script exits 1) was **not** filed as a new issue — it is already **#114**, open and `ready-for-agent`, whose section 1 is this finding verbatim. Commented there instead, with two citation refreshes the audit will need. Deferrals for this PR remain **#240** and **#241**.

Issue #117, opened 2026-07-31
fleet-trust-the-label spec: the remaining bare :NNN citations are unverified, and no cross-docs/specs sweep has been run


feigi:
Deferred from PR #108 review. Found by checking whether that PR falsifies claims in files it does not touch (`review-and-fix.md:58`).

`docs/specs/2026-07-30-fleet-trust-the-label-design.md:18-26` ("Wrong axis") carries two line-number citations into `docs/specs/2026-07-22-run-team-agent-fleet-design.md`. **Both are already stale on `origin/main`**, and PR #108 inserts 7 lines at the top of the cited file, pushing both a further 7 lines.

### 1. `:211-213` — off by 7 already, 14 after PR #108

The citation attributes this quote: *"A ticket can be correctly labeled `ready-for-agent` and still be too open-ended to hand a background implementer."*

```
# where the sentence actually is
$ git show origin/main:docs/specs/2026-07-22-run-team-agent-fleet-design.md | grep -n "too open-ended to hand"
219:correctly labeled `ready-for-agent` and still be too open-ended to hand a
# (sentence spans 218-220)

# what :211-213 actually contains on main
$ git show origin/main:docs/specs/2026-07-22-run-team-agent-fleet-design.md | sed -n '211,213p'

`ready-for-agent` means triage has already specified the ticket for an unattended
agent. `ready-for-human` means the opposite: it needs a human to brainstorm the
```

At PR #108's head the sentence moves to 225-227, so the citation is then ~14 lines off.

### 2. `:193` — the cited step no longer exists

> But `:193` implements it by importing `next-ticket`'s row table wholesale

That step was **deleted** — by the very change this document argues for. The 07-22 doc's own superseded note says so at `:13`: *"Phase 0's \"admit light-row only\" step is deleted"*. Nothing in the file now imports the row table:

```
$ git grep -n "sizing-a-ticket\|row table\|light-row\|one-two files" origin/main -- docs/specs/2026-07-22-run-team-agent-fleet-design.md
docs/specs/2026-07-22-run-team-agent-fleet-design.md:13:> Phase 0's *"admit light-row only"* step is deleted; the gate asks **is what to
```

`main:193` is Phase 0 step 1's label-filter prose; at PR #108's head `:193` lands on the `### Phase 0 — shortlist` heading.

### Why the approach is open

[...]

PR #244, opened 2026-08-10
fleet: the fix-applier cannot fetch a workflow specialist's report, and verbatim means all of them

feigi:
Ad-hoc tooling fix from the 2026-08-10 `/fleet:run-team` run, under the skill's own **Fix the tooling mid-run** rule. Trigger met: *identical failure twice*. Prose only, no code, no ticket — the defect was produced by this run and is scoped to a file this run invoked.

## What went wrong, twice

Both fix-appliers asked me for specialist **output-file paths** so they could run the retrieval the prompt prescribes:

```
tail -1 <output-file> | jq -r '.message.content[]?|select(.type=="text").text'
```

**Those files do not exist on the default review path.** The `review-pr` workflow has `agent()` return findings *into the script* — that is the delivery guarantee that makes it the default over hand-dispatch — so there is no per-specialist transcript, and the controller's message is the only copy of a review's findings.

The retrieval block was written for the **hand-dispatch fallback**, where a specialist genuinely does leave a transcript. It reads as covering every specialist a fix-applier might hear about.

## The other half is the controller's, and it is the one that actually lost findings

`the returned survived / unverified findings verbatim` was already the rule. I read it as licence to relay a *ranked selection*:

- `fix-pr-228` received **5 of 7** unverified findings
- `fix-pr-227` received **7 of 13**

Both then reported findings as *"looked at by nobody"* that had simply never been sent. One of them had independently rediscovered a finding I had not relayed, without recognising it as one.

Ranking was the second-order harm. Twice I attached an adjective to an *unverified* `suggestion` — the band the workflow budgets **0 refuters** by policy — and passed that emphasis on as weight:

- I called one finding *"the sharpest of the four"*. Its refuter killed **both halves**: 6 of 8 fields could never be `true` or merely echoed the caller's argv, and the proposed fix would have appended a warning to stderr — the one channel that already prints the raw byte-exact value the JSON scrubbed.
- I relayed a refuter's string-match inventory as a finding, which would have rewritten `gh issue list --label ready-for-agent` — gh's own correct flag — into a bug. The fix-applier caught it by running `gh issue list --help`.

## The change

Two edits in `skills/fleet/skills/run-team/SKILL.md`:

1. The retrieval block now states what it covers — **the refuters the fix-applier itself spawns** — and says the review's specialists have no file to fetch, so the thing to ask for is the text, not a path.
2. `verbatim` now says **every one, with no adjective**, and records the two measured costs of doing otherwise.

## Verification

- Suite on this branch: **303 tests / 303 pass / 0 fail / 0 skipped** — unchanged from `main`. Several sibling tests pin `run-team/SKILL.md` prose by anchor slice; none broke.
- Prose only: `git diff --stat` is one file, +19/-5.

## Not in scope

Two hazards this run surfaced that need a design call rather than a text edit, deliberately **not** bundled here:

- The `$!N` EOF guard in `jstr` is **undetectable on GNU sed**, so a future edit deleting it ships green through CI (Linux) and breaks only on macOS.
- `review-pr-testcmd.test.mjs:51` self-skips in any `git archive` snapshot, so every review specialist working from one reports a phantom `skipped 1`.


feigi:
Fix-applier pass complete. Pushed `560de8b`.

**Applied (1)** — the PR shipped a false claim, in a PR about not shipping false claims. It said the review specialists "left no transcript you can read" / "there is no output file to fetch" / "nothing is on disk for a member to fetch". They do leave transcripts (101 `subagents/workflows/wf_*/agent-*.jsonl` in one session, 33 of them `pr-review-toolkit:*`); the `tail -1 … select(.type=="text").text` recipe returns empty only because a schema-bound specialist ends on a `tool_result`, with the payload present as a `StructuredOutput` input. The operational rule — *ask for the text, not the paths* — is unchanged and now rests on what is true: the transcripts are not **addressable** by a member. `review-pr.js:623` returns no transcript path, and the `.meta.json` sidecars carry only `agentType`/`model`/`spawnDepth`.

**Applied (1, survived)** — both of the PR's new rules were unpinned; either hunk was deletable with the suite fully green. Two pins added to `review-path-default.test.mjs`, in the `fixApplierLeadIn()` / `fixApplierPrompt()` slices that exist for exactly this block. Each is a positive assert plus a `doesNotMatch`, because a re-inserted falsehood leaves every pinned phrase intact and only the exclusion catches it. Mutation-tested both ways, six mutants, each redding only its intended pin (M1 delete paragraph, M2 revert hunk 2, M3/M4 re-insert each false claim → RED; B1/B2 benign rewords → GREEN).

**Deferred (1)** — #245: `SKILL.md:609` "no relay ever occurs". **Refuted**, one refuter biased to refuse, three measurements. The premise misattributes the clause to this PR (it is from `bb79c7a`, present at `ff55adb~1:474`); the suggested Phase 3 event-loop bullet would delete an existing pin (`review-path-default.test.mjs:155` bans the word `relay` in that slice outright); and `:369-370` fixes the document's referent for bare "relay" as specialist-report delivery, so `:609` is consistent, not self-contradictory. Filed with the full refutation.

**No action (1)** — `SKILL.md:418`, the "5 of 7 and 7 of 13" anecdotes. The specialist itself said nothing contradicts the tree; agreed, nothing filed. One clause of that anecdote *was* false and is fixed above: "each asked for paths that cannot exist" → "each asked for paths nothing had given it".

**Out of scope (1)** — the `docs/specs/…-specialist-read-rules-design.md:47` → `run-team/SKILL.md:684` citation. Verified **pre-existing**: at the spec's own commit `c3c49a46` the quoted text was already at `:758`, not `:684`. This PR shifts it 14 further, to `:772`. Commented on #117 rather than filing — same class, same structure, and #117 already owns the `docs/specs/` sweep question.

Gate, run in the worktree (not a snapshot — a `git archive` copy self-skips `review-pr-testcmd.test.mjs` and reports a phantom `skipped 1`):

```
env -u NODE_TEST_CONTEXT -u NODE_TEST_WORKER_ID node --test 'skills/fleet/scripts/*.test.mjs'
ℹ tests 305  ℹ pass 305  ℹ fail 0  ℹ skipped 0
```

Baseline before this commit was `303 / 303 / 0 / 0`. No `--no-verify`. Not rebased.

PR #247, opened 2026-08-10
fix(fleet): stamp emitted runners with a template content hash

### LINKED ISSUE #124: claim-ticket.sh: the emitted agent-test is pinned per worktree with no version stamp and no refresh path, so runner fixes reach zero existing worktrees
Deferred from PR #109 review.

`claim-ticket.sh` writes `agent-test` into a worktree **once, at claim time**, from the template in force then. Nothing ever rewrites it. So a fix to the emitted runner — like #109's `node_modules` prune — lands in the template and reaches **no worktree that already exists**.

Measured across every live worktree, classified by grepping for #109's own added text:

```
STALE  .worktrees/101-snapshot-testcmd     STALE  .worktrees/26-specialist-count
STALE  .worktrees/106-audit-path-quoting   STALE  .worktrees/45-inflight-spec-contract
STALE  .worktrees/60-candidates-fallback   STALE  .worktrees/74-dirty-check-target
STALE  .worktrees/90-inflight-jstr         STALE  .worktrees/98-runner-prune
```

All eight — **including `98-runner-prune`, the worktree #109 was written in**. Merging #109 fixes zero runners that currently exist.

## Three mechanisms combine

1. `claim-ticket.sh` emits the runner only on the claim path.
2. `[ -e "$wt" ] && die "$wt already exists — ticket may already be claimed"` (`:29`) rules out re-claiming as a refresh route.
3. `.git/info/exclude` gets `agent-test` appended (`:163-164`), so git never touches it — not on checkout, not on rebase, not on merge.

## Why it is worse than a stale file

Nothing in the emitted runner identifies its template version. The only way to classify the eight above was to grep for the *content* of the fix being reviewed — which works exactly once, for one known fix, and is unavailable to anyone who does not already know what changed. A reviewer measuring the fix in their own worktree sees the old runner and can conclude the fix is inert; this reportedly bites people repeatedly.

## Proportionate suggestion, approach open

One line into the heredoc, making staleness *detectable* rather than solved:

```sh
# agent-test template: <sha>
```

Explicitly **not** a version-negotiation scheme. Open questions this ticket needs settled: stamp only, or also a refresh route (a `--refresh-runner` flag, or relaxing `:29` for a runner-only rewrite)? Should `worktree-audit.sh` or `reap.sh` warn on a stale stamp? Filed as `needs-triage` because the stamp is decided but the refresh half is not.

Surfaced by the interface-design reviewer on PR #109; staleness table re-derived independently.

feigi:
## Summary

`claim-ticket.sh` writes `agent-test` into a worktree once, at claim time, from the template in force then, and nothing ever rewrites it. Nothing in the emitted file identified which template produced it, so classifying a runner as current or stale required grepping for the content of a specific known fix — a technique that works once, for one change, and only for someone who already knows what to look for.

This stamps every emitted runner with a `cksum` of the script that produced it, as a comment near the top, written before the branch that tells the two entrypoint forms (`node --test` with the directory shim, and everything else) apart — so both carry it. A content hash of the script *is* the template, so the stamp changes exactly when the template changes and stays put otherwise.

Per triage's ruling on #124: the stamp only, nothing else. No refresh route, no relaxed re-claim guard, no rewrite of existing runners, no `worktree-audit.sh`/`reap.sh` warnings, no programmatic version check. Nothing reads the stamp; its only job is to make staleness legible to a human comparing it against a fresh `cksum` of the current script.

Closes #124

## Test plan

- [x] `runner: carries a template stamp on the npm entrypoint`
- [x] `runner: carries a template stamp on the node --test entrypoint`
- [x] `runner: the stamp is stable across claims of the same template`
- [x] `runner: the stamp changes when the template's content changes`
- [x] Full suite: `./agent-test skills/fleet/scripts/claim-ticket.test.mjs` — 37/37 pass


feigi:
Review findings adjudicated. Applied at `45df187`; suite green (`node --test "skills/fleet/scripts/*.test.mjs"` → tests 323, pass 323, fail 0).

**Applied (4)**
- `survived-1` — the stamp comment's "changes exactly when the template does and stays put otherwise" was false; it now says plainly that `cksum "$0"` is a whole-script checksum that over-reports, one-way.
- `survived-2` — `workflows/review-pr.js:262`: `claim-ticket.sh:110` / `:166` → the `cat > "$runner"` heredoc and the `grep -qx agent-test` line. Both refs were stale (119 and 215); semantic anchors instead of renumbering, per the `#117` class.
- `survived-3` — `skills/fleet/commands/review-and-fix.md:43`: `claim-ticket.sh:112` → the `export TEST_COMPOSE_PROJECT=…` line.
- `unverified-7` — `apply(files, script = SCRIPT)` in the test helper; the content-changes test stops re-implementing it inline (net −16). Refuter confirmed it still discriminates and the dropped `chmodSync` was dead.

**Deferred and filed (6 issues, 8 findings)**
| issue | finding | label | why |
|---|---|---|---|
| #252 | `unverified-1` | `needs-triage` | doc gap real, but SKILL.md:171 covers an *absent* runner, not a stale-template one — landing spot open |
| #253 | `unverified-9` | `wontfix` | inlining forecloses any exit-status guard; finding concedes both forms correct |
| #254 | `unverified-6` | `wontfix` | 20000 revisions → 0 cksum collisions; `sha256sum` not portable on macOS; docstring half already fixed |
| #255 | `unverified-4` | `wontfix` | premise false — `:310`/`:315` already name the missing stamp; helper makes the report 2.2x noisier |
| #259 | `unverified-8` | `wontfix` | subsumption real, but the "~2.7 s" saving measures 0.5 s and is invisible under scheduling noise |
| #263 | `unverified-2` + `unverified-3` | `needs-triage` | one defect, filed once; unreachable from every real invocation form, and the guard as specified fires *after* the worktree exists |

`unverified-5` needed no action — its own `suggested_fix` is "None needed" (a positive data point on documented-vs-enforced).

The two `refuted` findings were not applied as such. Both restate defects also live in the `unverified` band, and both were judged there on their own refuters: `refuted-1` = #263, `refuted-2` = #252.

Each of the 9 `unverified` findings was scope-checked, and the 8 with an action got one refuter biased to refuse. 7 refuted, 1 survived.

PR #249, opened 2026-08-10
docs(fleet): make run-team's references promise accurate, and drop two pointers that misdirect

### LINKED ISSUE #78: run-team: references/ do not hold the war stories the preamble promises

feigi:
Closes #78

Triage ruled option 3 plus repointing the pointer that misdirects, and ruled out writing the missing war stories: the evidence does not exist in retrievable form, so composing a reference means inventing detail nobody can verify.

## Evidence, re-run at this branch's HEAD

- `git grep -inE "dirty|worktree|deferral|finisher" HEAD:skills/fleet/skills/run-team/references/ci-and-staleness.md` → no output (exit 1). Nor any hit for `quiet`.
- `git grep -in "seven findings" HEAD` → one line, `SKILL.md:537`, the line that cites it. The anecdote has no source in the tree.
- `grep -inE "install|lockfile|npm|frozen" references/isolation.md` → no output (exit 1). That rule's evidence lives in `skills/fleet/scripts/claim-ticket.test.mjs:303-326`, not in a reference file.
- 74 bold rule lead-ins in `SKILL.md`, 23 `See references/` pointers — so most rules never had a reference, which is what made the preamble's claim false.

## Changes

**The preamble** now describes the directory as it stands: a rule ending `See references/<file>` has its war story there, and a rule without that line carries its reasoning inline — nothing is missing. A member no longer has to guess whether an uncited anecdote means a file was lost.

**Two pointers dropped**, both sending a reader to a file that carries nothing on the rule it is attached to:

- the finisher's CI gate (`check` job, not `ci-state --quiet` exit 0) → `ci-and-staleness.md`
- `Infer <install> — never default to npm install` → `isolation.md`

Neither rule loses anything: both already state their own reason inline, and only the trailing pointer is removed. No new reference file, no invented anecdote — a fabricated war story reads as evidence and is not.

## Pointer sweep (acceptance criterion 1)

All 5 files in `skills/fleet/skills/run-team/references/` read in full, and every one of the 26 `references/` mentions in `SKILL.md` checked against its target's content rather than its filename. The two above were the only ones whose target does not address its rule; the remaining 23 `See references/` pointers plus the one inline citation (`references/correction-tickets.md records four tickets in one run` — matched by that file's "Four for four in one run") resolve to a document that addresses the attached rule.

## Test

`node --test "skills/fleet/scripts/*.test.mjs"` → tests 319, pass 319, fail 0. No pin covers the edited prose (checked: no test string matches the preamble, the `--quiet` gate, or the `<install>` rule).


feigi:
**Review deferrals filed.**

| finding | dimension | outcome | landed |
|---|---|---|---|
| `unverified-1` — `SKILL.md:15` "lives there" binds to the spec path | correctness | refuter returned `refuted=false` — **survived** | applied in 6b07afa |
| `unverified-2` — selector misses the mid-prose `references/` citation at `:102` | correctness | refuter returned `refuted=true` — **refuted** | #261 (`needs-triage`, filed for the record) |
| `survived-1` — manifest pin `Say the specialist count` resolves nowhere, so the design doc's "all 33 patterns resolve" is false | comments | survived both refuters | commented on **#48**, not duplicated |
| `unverified-3` — both deleted pointers were genuinely dangling | comments | confirmatory — supports this diff | no action |

`survived-1` is deferred rather than fixed here for two reasons: it is pre-existing and outside ticket #78's scope, and **PR #251 is open and edits `docs/specs/2026-07-23-fleet-plugin-design.md`**, so touching it here would collide at merge. It already had an open `ready-for-agent` issue (#48) covering the same assertion, so it went there as a comment rather than a duplicate ticket.

Suite: `node --test "skills/fleet/scripts/*.test.mjs"` → **319 pass, 0 fail** at 6b07afa.

Issue #250, opened 2026-08-10
reap.sh guards arity only, so a typo in --apply is silently treated as "not --apply" and reports a clean no-op


feigi:
Found by the #114 exit-code audit while measuring `reap.sh`. Filed, not fixed: #114 is a documentation audit and changing a script's exit codes is out of its scope.

`skills/fleet/scripts/reap.sh:20-22`:

```sh
apply=false
[ "${1:-}" = "--apply" ] && apply=true
[ $# -gt 1 ] && die "usage: reap.sh [--apply]"
```

The usage guard is **arity only**. One argument passes whatever it says, and the line above it treats anything that is not exactly `--apply` as "not `--apply`".

Measured in worktree `114-script-surface-exit-audit`, each as `<cmd> >/dev/null 2>&1; echo rc=$?`, from inside a git repository:

```
$ reap.sh --nonsense    -> rc=0    runs the full dry run, reports, deletes nothing
$ reap.sh a b           -> rc=2    the arity check, not a flag check
```

An operator who types `--aply` gets a report indistinguishable from a successful reap: exit 0, the same `KEEP`/`reaped` output, and the branches still there. The dry-run banner does print (`DRY RUN — nothing will be deleted`), but on stderr among the per-branch lines, and it is the same banner an intentional dry run prints — nothing in the run says the flag was not understood.

Failing closed here is one line: reject any single argument that is not `--apply`, the same way the arity check already rejects two.

The direction of the silence is the mild one — a typo'd `--apply` under-deletes rather than over-deletes, which is why this is filed rather than treated as urgent. It is still a refusal the caller never sees.

Noted originally by a reviewer in a #114 comment ("probably worth folding into this ticket or splitting out — flagging rather than deciding"). #114's table row now states the guard as measured ("more than one argument") rather than as intended, so the documentation is correct either way this is resolved.

Related: #83 (reap.sh and release-ticket.sh disagree on an unreadable worktree).

feigi:
## The dry-run path has no test coverage, which is what a fix here needs

Deferred from the review of PR #426 (fix for #264), which added `skills/fleet/scripts/reap.test.mjs` — the first test file this script has ever had.

[...]

Issue #232, opened 2026-08-10
ledger.mjs check: shape guard validates only h.number, so a partial gh row prints "TRACKER HIT — #114 (undefined) — undefined" at exit 3


feigi:
Deferred from PR #225 review (ticket #152).

**Dimension:** `silent-failure-hunter` (attribution per controller relay).

**A refuter was dispatched against this finding, biased to refuse. It came back `refuted: false` — the finding stands.** It is filed rather than applied because the defective line is **outside the PR's diff**, and the maintainer's ruling on the sibling question (#231) set the precedent that #225 ships to its own ticket's scope.

`needs-triage` rather than `ready-for-agent`: the defect is verified but the right width of the guard is an open call — see *Open question* below.

## The finding

`skills/fleet/scripts/ledger.mjs:305` validates one of the four fields it goes on to interpolate:

```js
if (!Array.isArray(parsed) || parsed.some((h) => !h || typeof h.number !== "number")) {
  throw new Error("gh returned JSON that is not an issue list");
}
```

A payload of objects carrying only `number` passes it. Reproduced literally, with fixture `[{"number":114}]` through the suite's own `gh` stub:

```
STATUS: 3
STDERR: ledger: TRACKER HIT — #114 (undefined)  — undefined
STDOUT: {... "tracker":{"ok":true,"hits":[{"number":114,"title":"","score":0}]}, "verdict":"tracker-hit"}
```

A confident tracker hit, at the blocking exit code, built from a payload the code cannot describe.

## Scope — why this is deferred, not applied

`git diff 52fbf29...cc4fa87 -- skills/fleet/scripts/ledger.mjs` touches only lines 198, 278-281, 324-337 and 344-362. The guard at 305 is untouched, and the `undefined` stderr plus exit 3 are **pre-existing**. What PR #225 newly adds on this path is `verdict:"tracker-hit"` — a faithful derivation of `tracker.hits.length` — so it *propagates* the defect into one more consumer-facing field rather than creating it.

## Severity, honestly stated

Fail-**closed**, not fail-open. Exit 3 blocks the filing and `undefined` is visibly garbage, not a silent clean bill. Narrower than the #152 class.

**Why it still counts:** the repo has already *declared* this invariant, in `ledger.test.mjs`:

```js
assert.doesNotMatch(r.stderr, /undefined/, "never print a hit the payload cannot describe");
```

The partial-object case walks straight through it.

## Realism

[...]

PR #251, opened 2026-08-10
docs(fleet): the Non-zero column was written from intent — audit all eleven rows against the scripts

### LINKED ISSUE #114: fleet-plugin-design: the Script surface table's Non-zero column was written from intent — audit all eleven rows (candidates.mjs states the opposite of its code)

feigi:
Every cell of the Script surface table's `Non-zero when` column now states the exits its script can actually produce — verdicts and preconditions alike. That is triage's ruling on the column's one open question (reading 2, "*non-zero when*, literally"): a caller cannot act on a verdict-only column while operational failures carry the same codes.

All eleven rows, in one pass. Fixing one row per wave is what produced this ticket.

### What changed

- **The column's meaning is stated once above the table**, so the next row added has a rule instead of an intent to follow: 0 the answer asked for, 1 a verdict, 2 the question could not be answered. `ledger.mjs check` mints the one exception (3) and its own row says so.
- **`candidates.mjs`** said "empty result is exit 0 with `[]`". It is exit **1** — deliberately, so an empty queue is distinguishable from a broken query — and since #167 a malformed invocation is **2**, so the three-way split the column wants now exists in the code.
- **`pr-overlap.mjs`, `reap.sh`, `worktree-audit.sh`** each said "never". All three exit 2 on their preconditions. Their rows now name those and say what stays a finding at exit 0.
- **`verify-sha.sh`, `claim-ticket.sh`, `ledger.mjs`, `prove-merge.sh`, `ci-state.mjs`, `inflight.sh`** under-reported: exits their scripts produce that no caller could have known about, including `ledger.mjs check`'s exit 3 and its exit 1 on `ALREADY FILED` — a normal result a caller must not read as an error.
- **`no-undo-audit.sh`'s row is untouched.** PR #180 already rewrote it and it measures correct. Its `Out` pin (`no-undo-audit.test.mjs`) stays green.

### The column's second copy

`run-merge-bot.md` carries its own prose enumeration of `no-undo-audit.sh`'s exit-2 causes — the one the merge bot actually reads — and it still omitted "bad argument, no such worktree" after #180 resynced the linkage clause. It now carries the same list as the table.

`pr-overlap.mjs`'s own header claimed a failed query was the ONLY exit-2 cause. Bad usage is one too.

### Evidence

Every claimed exit was driven into existence, never reasoned about. The full command list and its measured `rc` for all eleven scripts is in the commit message, so the next reader can re-run rather than re-trust. Sandbox: a bare origin plus a work clone under a scratch dir, and a stub `gh` that exits 1 for the query-failure paths. `--apply` was never invoked on `reap.sh` or `claim-ticket.sh`.

### Filed, not fixed

**#250** — `reap.sh:20-22` guards arity only, so `reap.sh --aply` is silently "not `--apply`" and reports a clean no-op at exit 0. Changing a script's exit codes is out of this ticket's scope; the row states the guard as *measured* ("more than one argument") rather than as intended, so the table is correct either way #250 resolves.

Noticed and left alone as #48's territory (a different column, different assertions): the `ledger.mjs` row's `In` cell reads `set/filed/ruled/read`, but the subcommands are `row`, `filed`, `ruled`, `check`, `read` — there is no `set`, and `check` is the one that reaches a verdict.

Suite: 319 passing, unchanged from baseline.

Closes #114


feigi:
## Review applied — `848e727`

Five dimensions, 34 agents, then 5 more refuters on the in-scope suggestion band. **9 survived, 5 refuted, 12 unverified.** Split below.

### The headline: one row, seven findings

`survived-1`, `survived-4`, `survived-5`, `survived-6`, `refuted-5`, `unverified-8` and `unverified-12` are all **one defect** in the `ci-state.mjs` row. Fixed once; all seven resolve together.

The row documented exit 2 on "bad or missing `--pr`/`--branch`". **`ci-state.mjs` has no `--branch` flag.** `:86` is `const branch = prInfo.headRefName;` — derived from `gh pr view` — and the usage string at `:76` names only `--pr/--base/--workflow/--workflow-file/--quiet`. The row's own `In` cell carried the same stale flag, which is what seeded the exit claim, so both are gone.

This is worth stating plainly rather than filing as a nit. **#114's first acceptance criterion was that every claimed exit be "derived from its script and *demonstrated by invoking it* — each claimed exit driven into existence, not reasoned about"**, and this PR's body claims that was done. A row asserting an exit for a flag that does not exist is proof it was not done for that row: there was nothing to invoke.

The rest of the same cell was written from intent too:

| Was | Is |
|---|---|
| "a `gh` or `git` query failed" | The sole `git` call (`:223 tryRun("git", ["remote","get-url","origin"])`) is deliberately fail-soft — `:196-201` says it must not use `run()` precisely so it cannot die. A totally broken `git` still yields `behind: null`, `verdict=green`, **exit 0**. |
| "output that is not the JSON it asked for" | True only of the parse failure `runJson` guards. A reply that parses but is the **wrong shape** is unguarded: `runs.filter` throws and node exits **1** — the code this script uses for "not bound-green". Now documented on exit 1, where it happens. |
| (absent) | Three `die()` sites the row never listed: workflow file unreadable, a job-level `name:`, zero jobs derived. |

### Applied (5 rows, 1 preamble line)

| Finding(s) | Row | Change |
|---|---|---|
| `survived-1/4/5/6`, `refuted-5`, `unverified-8/12` | `ci-state.mjs` | `In` cell → `--pr [--base --workflow --workflow-file --quiet]`; exit-2 list corrected as above |
| `survived-2`, `survived-3` | `reap.sh` | "exit 2 only" was false — documents the exit-1 prune path, and that exit 0 is not a closed account of what was safe to delete |
| `survived-8`, `refuted-1` | `verify-sha.sh` | exit 1 also carries a `merge-base` that failed rather than answered |
| `survived-9` | `candidates.mjs` | adds `refuseIfCapped` — a result landing exactly at `--limit` |
| `unverified-1` (part), `unverified-7` | `prove-merge.sh` | adds `is_ancestor`'s own probe failure, and says that is where a base resolving to a non-commit lands |
| — | preamble | "A script with no verdict to report **never exits 1**" was a false universal the moment `reap.sh`'s exit-1 path was written down. Now normative ("must never"), with the row naming itself as the violation. |

### Filed — code defects, out of scope per #114 ("if a code looks wrong, file it, do not fix it here")

- **#264** `needs-triage` · silent-failure · **data loss.** `reap.sh:44`'s `git cherry ... 2>/dev/null | grep -q '^+'` takes **grep's** status. A `git cherry` that dies (exit 128) emits nothing and is indistinguishable from "no unmerged commits" — `--apply` then deletes a `[gone]` branch whose commits exist nowhere else, at exit 0, error suppressed. Reproduced with **real corruption**, then again with a PATH wrapper. Bounding fact in the issue: **this run's own `reap.sh --apply` is verified unaffected** — it KEPT `feat/212-implementer-model-by-class` for "unmerged commits" in the same invocation, a verdict reachable only when `git cherry` actually emits `+`.
- **#265** `needs-triage` · correctness. `reap.sh:102` — failing `git worktree prune` exits 1 with no payload, after the branches are already deleted. Two viable fixes, so triage picks.
- **#266** `ready-for-agent` · silent-failure. `verify-sha.sh:31` — probe exit 128 reported as `reachable:false`. **This exact collapse was already found and fixed in the sibling: #17 (closed) added the `[ $? -eq 1 ] || die` guard to `prove-merge.sh:39-44`.** The sweep stopped at the site that motivated it. A false `reachable:false` stalls the controller for every SHA it enqueues.
- **#267** `ready-for-agent` · silent-failure. `prove-merge.sh:63` — `die` inside `$(...)` kills only the subshell, so the caller overwrites a correct "cannot prove anything" with "that merge did not land". Reported as an unchecked suggestion; **two refuters on unrelated findings both drove it into existence**, so it is filed verified.

### Commented, not filed

- **#147** — `unverified-4` (no-undo-audit stash count reads 0 on an unreadable reflog) is **already this issue**. `ledger.mjs check` scored it 0.67 and caught the duplicate. Commented with the corrected line number (this PR moves it from `:72` to `:188`) and the constraint that git exits 0, so only a positive `refs/stash` check works.
- **#48** — `survived-7`, `unverified-3`, `unverified-6`, `unverified-9` are all the **`In`/`Out` columns**, which #114 does not scope. Four findings, one comment: the `ledger.mjs` `In` cell advertises `set` (not a subcommand, always exits 2), omits `row`, and omits `check` — the entire subject of the cell beside it; and the `Out` cells are stale for `ci-state.mjs`, `verify-sha.sh` and `worktree-audit.sh` (verified from each script's payload construction, including one **type** error: `worktree-audit`'s `dirty` is a count, `dirtyFiles` is the array). The `In` cell keeps `--branch` removed here only because that token was the source of the false exit claim.

### Refuted — not applied

Four of the five in-scope suggestions were refuted by running something, not by reasoning:

- **`unverified-2`** (pr-overlap zero-files as a third exit-2 cause) — the die is reachable, but `pr-overlap.mjs:58-59` explicitly classifies zero files **as** a failed query: *"a PR with no changed files is not a legitimate 'no overlap' answer, it is a query that did not work."* The suggested fix would have split the two apart and made the doc contradict the comment it describes.
- **`unverified-7`** (prove-merge base-ref wording) — loose, not wrong; every case the phrase covers does land on exit 2. Its substance is folded into the `prove-merge.sh` row edit above rather than dropped.
- **`unverified-10`** (ten rows re-gloss the preamble in four wordings) — the counts were real but the grouping was not. **"exit 2 only" is a different assertion**, not a redundant gloss: it claims the exit surface is closed at 2, and the refuter verified it holds for all four scripts that carry it (`claim-ticket.sh`'s `exit 1`s are inside a heredoc — they belong to the generated `agent-test` runner). The suggested fix would have deleted that claim from four rows. Real drift is 3 wordings over 6 rows; residual noted below.
- **`unverified-11`** (247-char clause duplicated with `run-merge-bot.md:124`) — the LCS number is correct, but the load-bearing claim is **false**: the PR has **two** hunks in that file, the second an unrelated `conflictsRewritten[]` paragraph. The file was already being edited, so removing the duplication saves no file-touches.

`unverified-1` split: line 42 **survived** and is applied; line 80 (`cannot read the parents`) was refuted as unreachable behind the line-62 probe, so it is not enumerated.

The five originally-`refuted` findings were reviewed by the controller and none reversed — in each case the substance is live elsewhere (`refuted-5` → the cluster above, `refuted-1` → #266, `refuted-2/3/4` → the `ci-state.mjs` and `prove-merge.sh` row edits).

### One residual, not applied — flagging rather than deciding

`no-undo-audit.sh`'s row says "exit 2 the question **is unanswerable**" where four sibling rows say "the question **could not be answered**". A bare synonym, in the column #114 owns. Left alone deliberately: #114 is an accuracy audit and both phrasings are accurate, so this is polish, not a defect. One-word change if wanted.

### Gate

`node --test "skills/fleet/scripts/*.test.mjs"` → **319 pass, 0 fail** (baseline 319). No `--no-verify`.

This PR is **behind=3** — a `rebase-check` red or heavy jobs `skipped` off that behind-count is staleness, not failure.

feigi:
## Deferral record filed — #268

Correcting the previous comment on one point: the four refuted suggestions were described here but not **filed**. A finding is deferred, never dropped, and "refuted" is a deferral state like any other. They are now in **#268** (`wontfix`), one record rather than four stubs, each with its measurements:

- **`unverified-2`** — remedy would have split zero-changed-files away from "a query failed", contradicting `pr-overlap.mjs:58-59`, which classifies it as exactly that.
- **`unverified-7`** — the base really is not commit-checked, but the column documents causes-to-codes, not messages. Substance folded into #251's `prove-merge.sh` row edit instead.
- **`unverified-10`** — remedy would have stripped `exit 2 only` from four rows. That phrase is a **verified-true closed-exit-surface claim**, not a redundant gloss — precisely the class #114 exists to audit.
- **`unverified-11`** — the 247-char LCS is real; the causal claim is not. `git diff main...HEAD --numstat` on `run-merge-bot.md` returns `3 1` — **two** hunks, the second unrelated `conflictsRewritten[]` guidance — so removing the duplication saves zero file-touches.
- **`unverified-1`'s line 80** — recorded as a **negative result**: it could not be driven, because the `:62`/`:70`/`:74` probe always intercepts. Missing trees are tolerated (`rl=0`) and shallow clones land on the already-enumerated "no second parent" die. Worth writing down so the next auditor reading `grep -n 'die "' prove-merge.sh` does not re-derive it. This did not reach the commit message, which was written before that ruling arrived.

Three of the five suggested remedies would have made the table *less* accurate. That is the reason #268 is a record rather than a backlog item, and why it says so in the title.

The `no-undo-audit.sh` "is unanswerable" residual is filed there too, unapplied — a synonym, not an accuracy defect, and #114's out-of-scope rule forbids widening a correction ticket.

No code change; `848e727` stands.

feigi:
## #269 filed — the code half of `survived-6`

`ci-state.mjs` guards the **parse** (`runJson`'s try/catch on `JSON.parse` → exit 2) but never the **shape**. A `gh` reply that parses cleanly and is the wrong shape reaches `:140 .filter((r) => r.headSha === prHead)` and throws — and an uncaught throw **exits 1**, the code this script uses for "not bound-green". An unanswerable query renders as a CI verdict.

`grep -n "Array.isArray" ci-state.mjs` → no hits. Both siblings have the guard: `candidates.mjs:185`, whose comment names this exact hazard ("*an uncaught throw exits 1: the code reserved for 'successful query, no survivors'*"), and `ledger.mjs:302`.

`needs-triage` rather than `ready-for-agent`: the minimum is one `Array.isArray(runs)` check, but the class fix validates every `runJson` return — and **#232** shows a partial shape guard can itself become the defect. That is triage's call, not mine.

Same class as **#111**, **#262**, **#266** and **#232** — "an unanswerable question rendered as a verdict" — with `ci-state.mjs` the highest-leverage instance, since the controller polls it per PR and acts on exit 1 without a second look.

Documented on the row in `848e727`; the code is out of scope for #114.

---

**Full deferral set for this PR:** #264, #265, #266, #267, #269 filed · #268 (`wontfix`) records the five refuted suggestions · #147 and #48 commented rather than duplicated.

PR #271, opened 2026-08-10
fix(fleet): normalize args.dimensions keys, fail loud on the unrecognised

### LINKED ISSUE #113: review-and-fix.md:49 promises args.dimensions accepts keys; review-pr.js has no normalization, so keys degrade silently
Deferred from PR #107 review.

`skills/fleet/commands/review-and-fix.md:49` documents the fan-out override as:

> Pin a subset explicitly with `args.dimensions` (**keys or dimension objects**).

The code accepts only dimension objects. There is no normalization anywhere:

```
$ git show ee81d14:workflows/review-pr.js | grep -n "explicitDimensions"
115:const explicitDimensions = args && args.dimensions; // caller override; else derived from the diff below
222:const dimensions = explicitDimensions || selectDimensions(DEFAULT_DIMENSIONS, stats);
```

`:222` assigns the caller value straight through, and every downstream use dereferences object fields:

- `:224` — `dimensions.map((d) => d.key)` in the log line
- `:236` — `` `Review PR #${pr} ... for: ${d.prompt}` ``
- `:248` — `{ label: `review:${d.key}`, agentType: d.agentType, ... }`

Pass `["correctness", "comments"]` as the doc invites and all three read `undefined`: the literal string `undefined` is interpolated into the specialist prompt, `agentType` is `undefined`, and the scratch path becomes `${scratch}/undefined/`. No throw, no warning — the fan-out runs and returns findings from misconfigured agents.

**Approach still open** — hence `needs-triage`, not `ready-for-agent`:

1. Honour the doc: map string keys through `DEFAULT_DIMENSIONS` at `:222` (and reject unknown keys loudly rather than silently dropping them).
2. Or correct the doc: drop "keys or" from `:49`, since nothing in the repo currently passes keys.

Option 1 is the larger promise; option 2 is the smaller diff. Either way the silent-`undefined` path deserves a guard — an override that names an unknown dimension should fail loudly, not review the PR with an undefined prompt.

Found independently by me and by the correctness specialist during the PR #107 fan-out; sub-threshold for that PR (pre-existing, different file, outside its diff).

feigi:
## Summary

`skills/fleet/commands/review-and-fix.md:49` has always documented `args.dimensions` as accepting "keys or dimension objects", but `workflows/review-pr.js` only handled objects — a key array (`["correctness", "comments"]`) passed straight through `explicitDimensions || selectDimensions(...)` untouched, and every downstream dereference (`d.key`, `d.prompt`, `d.agentType`) came back `undefined`. No throw, no warning: the fan-out ran specialists with an undefined prompt, an undefined agent type, and a scratch path of literal `undefined/`, returning findings from misconfigured agents.

Per triage's ruling on #113: honour the documentation rather than deleting the promise, and fail loudly on anything the workflow doesn't recognise.

`resolveDimensions(override, all)` now sits in front of the existing `selectDimensions` fallback:
- a string entry resolves against `DEFAULT_DIMENSIONS` (the workflow's own catalog) — unknown key throws, naming it
- an object entry is checked for the three fields the fan-out dereferences (`key`, `prompt`, `agentType`) — missing field(s) throw, naming them
- an override that resolves to zero dimensions throws (an empty array is truthy, so `[] || selectDimensions(...)` would otherwise pass `[]` through silently)
- a non-array override throws instead of crashing on `.map`
- no override (`undefined`/`null`) returns `null`, leaving `selectDimensions`'s size-tier logic in charge exactly as before

Object overrides behave exactly as before (same objects pass through, now just validated).

## The args-as-string precondition

The brief asked to settle, before implementing, whether the Workflow runtime can still deliver `args` as a JSON string (which would make every `args.x` read `undefined` and throw on the required-args guard before any of this matters).

**Already fixed, predating this ticket.** `decodeArgs()` was added in commit `b243b4ee` (2026-08-05, part of #194's read-rules work) and runs before any field is read:

```js
function decodeArgs(a) {
  if (typeof a !== "string") return a || {};
  try {
    return JSON.parse(a);
  } catch (e) {
    throw new Error(`review-pr: args arrived as a string this could not parse (${e.message})`);
  }
}
const A = decodeArgs(args);
```

Every downstream read in the file (`A.pr`, `A.dimensions`, `A.testCmd`, etc.) goes through the decoded `A`, not raw `args`. No decode guard needed in this change.

## Test plan

- [x] `resolveDimensions` lifted from source text (same technique as `selectDimensions`/`usableDiff` elsewhere in this file) and unit-tested directly: no override, object override unchanged, key override resolves, mixed key+object override, unknown key names the key, missing required field(s) name the field(s), empty-array override throws, non-array override throws
- [x] Mutation-tested the call-site pin: reverting `const dimensions = ...` to the old `explicitDimensions || selectDimensions(...)` form fails the new call-site test, confirming it isn't a vacuous pin
- [x] `node --test skills/fleet/scripts/*.test.mjs` — 331/331 passing (baseline 323 + 8 new)

Closes #113


feigi:
## Review applied — 4 applied, 8 deferred, 5 refuted

Pushed `26f0286`. `node --test "skills/fleet/scripts/*.test.mjs"` → **tests 333, pass 333, fail 0** (331 baseline + 2 new). No `--no-verify`; repo has no pre-commit hook.

### Applied

| # | What | Verification |
|---|---|---|
| 1 | **`if (!override) return null` → `if (override == null) return null`** (`review-pr.js:403`) | `""`/`0`/`false`/`NaN` were treated as "no override" and fell through to `selectDimensions`. Measured on a `single-file` diff: a caller pinning `["correctness","comments","types"]` with `dimensions: ""` ran `[correctness, silent-failure]` — two *different* specialists. `!explicitDimensions` is true on that path, so the log printed the size-tier line as if nothing had been passed. |
| 2 | **Test pinning the `!entry \|\| ` guard** | It had **zero** coverage — both refuters deleted the clause and the file stayed green 25/25. It is the only thing turning `[null]` into the named error instead of `TypeError: Cannot read properties of null (reading 'key')`. |
| 3+4 | **Two stale comment counts** (`select-dimensions.test.mjs:8`, `:21`) | "Both values under test are lifted" / "Both declarations end at a column-0 terminator" both undercounted once this branch added a third lift. Reworded count-free so the next lift does not restale them. |

Both new tests are **mutation-verified, not merely green**: deleting `!entry ||` turns the first red with that exact `TypeError` while a decoy mutation elsewhere in the same function leaves it green (so it discriminates, rather than being brittle); restoring `!override` turns the second red.

### Deferred and filed

| Issue | Finding | Label |
|---|---|---|
| #274 | `resolveDimensions` accepts a repeated key → two specialists, one scratch dir, double-counted `dimensionsRun` | `needs-triage` |
| #275 | `args.dimensions` validated at `:540`, after the snapshot agent at `:432` — a typo'd key pays for a full snapshot first | `needs-triage` |
| #276 | `select-dimensions.test.mjs:176` "agentType/scratch path" reads as a `d.scratch` field that does not exist | `ready-for-agent` |
| #277 | `review-pr.js:413` per-field `!entry` guard → optional chaining (now test-pinned by this PR) | `ready-for-agent` |
| #278 | `:398` says "the three fields the fan-out **actually dereferences**" — it dereferences four (`d.model` at `:551`, `:585`) | `needs-triage` |
| #279 | `:413` "object is missing required field(s)" calls a non-object an object — **fixing it requires rewriting a test pin added here** | `needs-triage` |
| #280 | `REFUTED:` hoisting the empty-override guard above the `map` — filed so the next simplifier pass finds the measurement instead of redoing it | `needs-triage` |
| #281 | `:415` missing-field error carries no **entry index** — a multi-entry override cannot say which entry failed | `needs-triage` |

Two pairs were the same finding raised twice by different specialists and are filed **once**: #274 covers `unverified-2` + `unverified-8`, #275 covers `unverified-1` + `unverified-5`.

**#278 carries a trap warning.** The imprecision is real, but the review's suggested reword — "`model` is optional — absent means inherit, see :73" — is **false for 2 of the 3 entries it describes**: `code-reviewer` (correctness) and `code-simplifier` (simplify) are pinned `model: opus` in vendored frontmatter, verified directly. It cites `:73` as authority while dropping the qualifier that is that block's entire point, and a passing test already guards the word "inherit" out of the models log. That is why it was refuted rather than applied, and the issue says so loudly so nobody re-derives the wrong fix.

### Refuted — recorded, not applied

- **`unverified-3`** (the `d.model` wording) — refuted on the fix clause; substance filed as #278 above.
- **`unverified-6`** (the "missing required field(s)" message mislabels a non-object entry) — refuted; filed as **#279** for the record. The code throws, names the arg, and aborts before any specialist dispatches, which is the *opposite* of the #113 class; `[123]` / `[["correctness"]]` are hand-fabricated, and `[{}]` — a genuine object — produces the same string, so the message is not wrong about the case it is written for. The suggested fix also turns the `[null]` test added here red for a wording nit.
- **`unverified-9`** (hoist the empty-override guard above the `map`) — refuted; filed as **#280**. Three of its four factual clauses fail under measurement: the guard is **not** dead (`select-dimensions.test.mjs:222` pins it), the "avoids per-entry work" gain is **zero** (map-callback invocations on `[]` = 0), and it is **−1** line, not −2. The one measured behavioural divergence over 43 inputs makes the hoist *worse*: an `Array` subclass with a `Symbol.species` returning `{length: 0}` throws today but would pass a non-array through `resolveDimensions(...) || selectDimensions(...)` after the hoist — exactly the silent pass-through that guard exists to stop. Pathological input, and nobody showed a real caller can produce it, but it falsifies the finding's stated ground (`resolved.length === override.length` always).
- **`refuted-2`** and **`refuted-3`** were refuted 2-0 in the review's own adversarial pass and are not reversed here. `refuted-2`'s substance is #274.

### One thing worth flagging honestly

**`refuted-1`** — the run log prints "profile unknown, full set" for an override that ran a strict subset — came back on a **1-1 split** and was not reversed. It concerns a log line's accuracy rather than correctness, and there was no evidence to overturn a tied vote. But **this PR is what makes the documented key form of the override actually usable**, so that log line will now be reached by callers who could not reach it before. If a later PR touches `review-pr.js:541-545`, re-examine it then.

Surfaced by a refuter working on a *different* finding, and explicitly disclaimed by it as not-the-claim: the missing-field error carries no entry **index**, so a multi-entry override cannot tell you which entry failed. Filed separately as **#281** rather than folded into #279, so it does not inherit that finding's refuted framing. Reachability is untested — nobody established that multi-entry overrides are used.

**#279 carries a conflict warning.** Its fix changes the exact string that this PR's new `[null]` test pins; measured, the incoming assertion goes red under it. That pin is the only coverage of the `!entry ||` clause, so it must be *rewritten* alongside, never deleted.

<sub>Reviewed against snapshot `1e1aa04`; four refuters dispatched, each with a private scratch dir and instructed to default to `refuted=true`.</sub>

Issue #262, opened 2026-08-10
fleet: the phase-3 CI monitor exhausts the GitHub REST rate limit, and the natural watcher shape reports the outage as silence


feigi:
Hit live during a `/fleet:run-team` run with 3 open PRs. `gh` began returning HTTP 403 for every Actions read; the fleet's CI gate went blind.

## Measured

```
$ gh api rate_limit --jq '{core:.resources.core, graphql:.resources.graphql}'
{"core":{"limit":5000,"remaining":0,"reset":1786376992,"used":5000},
 "graphql":{"limit":5000,"remaining":3724,"used":1276}}
$ date -u   ->  2026-08-10T15:35:43Z      (reset 15:49:52Z)
```

**The core (REST) budget was fully consumed while GraphQL was barely touched.** That asymmetry is what makes the failure confusing rather than obvious: `gh pr view` is GraphQL and kept working, so PR state, labels and heads all read fine, while `gh run list` / `gh run view` / `gh api …/compare` returned 403. The fleet looked healthy from every angle except the one the merge path gates on.

## Where the calls go

`skills/fleet/scripts/ci-state.mjs` costs **3 REST + 1 GraphQL per invocation**:

```
:83   gh pr view <n> --json headRefName,headRefOid,state,mergeStateStatus   (GraphQL)
:133  gh run list --branch <b> --workflow CI --limit 30 --json …            (REST)
:160  gh run view <runId> --json jobs,attempt,status,conclusion,headSha     (REST)
:228  gh api --hostname <host> repos/<repo>/compare/<base>...<head>         (REST)
```

`run-team/SKILL.md` instructs the controller to *"Arm a second persistent Monitor over open PRs' latest runs"* and to emit behind-count and per-job conclusions with each event — i.e. one `ci-state.mjs` per open PR per poll. It states **no cadence and no budget**. At the obvious 45s interval that is `3 REST × PRs × (60/45) × 60` per hour = **~1440 REST/hour at 3 open PRs from one monitor alone**, before the run's other `gh` traffic (candidate scans, `inflight.sh`'s three probes per ticket, issue reads, every fix-applier's `gh issue create` / `ledger.mjs check`, and each review specialist's own calls).

In this run two CI monitors were live simultaneously — one armed by the controller and one inherited — which doubled it. The skill's own text does not warn that a second monitor is redundant, and nothing in it makes the duplicate visible.

## The failure is silent by construction

The natural way to write the watcher is the one that hides the outage:

```sh
[...]

PR #273, opened 2026-08-10
fix(fleet): ci-state.mjs gets a no-CI verdict and discovers the workflow file

### LINKED ISSUE #111: fleet: the whole phase-3 merge path gates on CI, and this repo has none — every PR strands unlabelled

feigi:
## Summary

Closes #111. Scope is the two portability residuals the ticket re-scoped to after #157/#68 landed real CI on this repo — the original "this repo has no CI" stall is already fixed:

1. `ci-state.mjs` died with `cannot read <file>` for an absent workflow file, sharing exit 2 with "the question could not be answered" — no way to tell "no CI configured" from "couldn't read the workflow that does exist".
2. The workflow-file default was hard-coded to `.github/workflows/ci.yml`; no caller anywhere passed `--workflow-file`, so a repo whose workflow has a different filename made every fleet CI read exit 2.

## Changes

- **`ci-state.mjs`**: workflow file is now *discovered*, not assumed — it scans `.github/workflows/*.yml`/`*.yaml` for a file whose top-level `name:` matches `--workflow` (default `"CI"`, the same identity `gh run list --workflow` already filters by), reading off disk rather than another `gh api` round trip (relevant to the open REST-budget issue #262). `--workflow-file` remains as the explicit override for the one case discovery can't settle: two files sharing the target name (dies, exit 2, asks for the flag).
- A repo with no matching workflow gets its own **`verdict: "no-ci"`**, distinguishable from both `green`/`not-green` and from the exit-2 "unanswerable" bucket. An unreadable or malformed workflow — whether named explicitly or found by discovery — still dies with today's exit 2 unchanged (discovery fails closed on an unreadable candidate rather than silently treating it as absent, so a permission-broken workflow can never misread as no-ci).
- **Absence never means pass.** `no-ci` without `--declare-no-ci` still exits 1 (gate unsatisfied, same bucket as `not-green`). `no-ci` *with* `--declare-no-ci` exits 0 — the caller has declared this repo gates on the reviewer's own verified suite run instead. Exit-code vocabulary (0/1/2) is otherwise unchanged.
- Docs: `review-and-fix.md` (finisher gate + standalone watch step), `run-team/SKILL.md` (CI-completion event + finisher-gate paragraph), `run-merge-bot.md` (the merge-time CI check) all now name the `no-ci` verdict and the `--declare-no-ci` escape, so no reviewer has to invent it.

## The declaration shape

The brief deliberately left this to be decided in the worktree: "a flag or documented marker, whichever fits the existing call surface." Chose a CLI flag, `--declare-no-ci`, boolean via the same `has()` helper `--quiet` already uses. Reasons:

- `ci-state.mjs`'s entire existing surface is argv flags (`--pr`, `--base`, `--workflow`, `--workflow-file`, `--quiet`) — a flag is the path of least resistance and needs no new parsing.
- A repo-committed marker file was the alternative "documented marker" reading, but it's a weaker fit: it can go stale or sit uncommitted, and it's a second thing to grep for across the fleet docs versus one flag every caller already has to pass explicitly, every time, the same way `--quiet` is.
- An explicit-every-call flag is also the more conservative choice given the brief's own framing — declaring must never be inferred, and a flag can't be forgotten silently the way a file left in `.gitignore` could.

## Review response — `board.mjs`'s `mapCi()` (raised by controller)

`mapCi()` collapses `ci-state`'s verdict vocabulary to `green`/`red`/`unknown` for the cockpit and has no branch for the new `"no-ci"` string — the concern was that an unrecognised verdict silently falls through to `"unknown"` with no test either way.

Verified rather than assumed: a no-ci payload never sets `status` (`ci-state.mjs` skips the whole run-lookup section for that case, and `status` stays initialized to `null`). `mapCi`'s **first** line is `if (d.status !== "completed") return "unknown"`, so a no-ci payload is caught there — the verdict string is never even read. Confirmed by running the script end-to-end against a synthetic no-ci repo and checking the emitted JSON (`"status":null,"verdict":"no-ci"`), then feeding that exact shape through `mapCi` directly.

So the fall-through is moot by construction, not by luck — but it wasn't pinned, so I added `skills/fleet/scripts/board.test.mjs`: `mapCi: no-ci verdict (no run bound, status null) → unknown, not silently mapped`, with a comment naming what it actually guards (the status gate, not a `verdict` branch) and what would need to change in `ci-state.mjs` for this to stop being true. Did **not** touch `mapCi()` itself — out of this ticket's file list, and the existing fail-safe fallthrough to `"unknown"` is the correct cockpit behaviour for an unrecognised verdict regardless.

## Review response — documentation site audit (raised by controller)

Controller enumerated 5 sites across 2 files that state the finisher/label gate, citing a prior incident (`review-and-fix.md` fix landed without the matching `run-team/SKILL.md` update, stranding PRs on the untouched path). Re-derived against this PR's current diff:

1. `run-team/SKILL.md`, **"Monitor: CI run completes" bullet** (`- **Monitor: CI run completes** → bind it...`, check-green/no-heavy-failure condition) — **left as-is, deliberately**. No-CI is a different triggering *event*, not a variant of this one: no workflow run ever completes for a no-CI repo, so this edge never fires for it at all. Documented as its own sibling bullet immediately after ("`ci-state.mjs --pr <N>` reads `verdict: "no-ci"` → no workflow run will ever complete..."), which explicitly says "the CI-run-completion edge above never fires" — editing the check-green bullet's own text to fold in no-ci would misstate it as the same event rather than a different one.
2. Same bullet, **"do NOT gate on `ci-state --quiet` exit 0" clause** — same reasoning as (1), same sibling bullet covers it.
3. `run-team/SKILL.md`, **Reviewers section, finisher-dispatch trigger sentence** ("When the diff-validating `check` job is green **and no heavy job is in `failure`**... dispatch a finisher") — **edited**: added a forward-pointer clause, "— or `ci-state.mjs` reads `verdict: "no-ci"`, see below —", so the trigger condition is self-contained rather than relying on a reader reaching the gate paragraph 4 lines later.
4. `run-team/SKILL.md`, **"Gate on the `check` job, not on `ci-state --quiet` exit 0"** paragraph — unedited itself; immediately followed (already, before this review round) by a full paragraph on the `no-ci`/`--declare-no-ci` gate.
5. `review-and-fix.md` **step 6** (the single dense line carrying diff-check-green + deferrals-filed + the label). Per the controller's own caution, left this line unrestructured — added a new paragraph directly after it instead, same pattern as (4).

`run-merge-bot.md`'s merge-time `ci-state.mjs` check (not one of the 5 cited, but the same gate family and a direct caller) also got a `no-ci`/`--declare-no-ci` note, added in the first pass.

## Test plan

- [x] `node --test "skills/fleet/scripts/*.test.mjs"` — 330 passing (baseline 323 + 6 new in `ci-state.test.mjs` + 1 new in `board.test.mjs`), 0 fail
- [x] New `ci-state.test.mjs` tests: no `.github/workflows` directory (no-ci, exit 1, confirms `run list`/`run view` never called); `--declare-no-ci` flips the same no-ci verdict to exit 0; a differently-named workflow discovered correctly alongside an unrelated sibling workflow file (mirrors this repo's own `ci.yml` + `rebase-check-refresh.yml` layout); two same-named workflow files (ambiguous, dies asking for `--workflow-file`); an unreadable explicit `--workflow-file` target (dies, exit 2, unchanged); an unreadable discovery candidate (fails closed, exit 2, not treated as no-ci)
- [x] New `board.test.mjs` test: `mapCi` on a no-ci payload → `"unknown"`, not a silent fall-through
- [x] Ran `ci-state.mjs --pr 271` (a real open PR in this repo) with no `--workflow-file` passed — discovery found `ci.yml` automatically among its 2 sibling workflow files, both `rebase-check`/`check` jobs read, `verdict: "green"`, exit 0: current repo's behaviour is unchanged


feigi:
Review deferrals filed (from the #273 review, 6 dimensions / 33 agents):

- #284 `needs-triage` — `board.mjs` `mapCi` has no `no-ci` branch, so a no-CI repo's PRs read as `unknown` forever (dimension: correctness)
- #285 `ready-for-agent` — `run-team/references/ci-and-staleness.md` never mentions the third verdict (dimensions: correctness + comments)
- #286 `ready-for-agent` — no test pins the no-ci finisher gate in any of the three docs that state it (dimension: tests)
- #287 `needs-triage` — `verdict` and `gateSatisfied` hand-derive the same 3-state mapping twice; no live bug, fail-closed measured (dimension: types)
- #288 `ready-for-agent` — four measured simplify reductions across `ci-state.mjs`/`.test.mjs`/`SKILL.md`, plus one refuted finding recorded so it is not re-raised (dimensions: simplify + silent-failure)

Applied in `52567c9`: survived-1 through -6, plus the controller-reversed `refuted-3` (relocated not-green detectors) and `unverified-8` (vacuous `mapCi` pin), both of which survived a fresh refuter. Suite: `tests 339 / pass 339 / fail 0` from the worktree.

PR #272, opened 2026-08-10
docs(specs): sweep docs/specs for cross-file line citations and re-anchor the live ones

feigi:
Closes #117

## The sweep

```
$ grep -rnoE '(`|[A-Za-z0-9_./-]+):[0-9]+(-[0-9]+)?' docs/specs/
```

Run at `5cd42be` (`origin/main` at the time of writing): **161 hits across five
of the six specs**. `2026-07-24-fleet-cockpit-design.md` has none. The command is
now recorded in the plugin design's citation-convention marker, beside the
convention it enforces.

**The command under-reports, and the spec now says so.** It matches only a
citation's leading `path:NN`, so comma-chained continuations and bare
back-references are invisible: **2 hits for the 8 citations** on a chained line,
and **three whole sites reported zero times** at `5cd42be` (this file's `:84`,
read-rules `:149`/`:150`). `161` reproduces exactly; **`nine` did not** — the
command reports 8, the ninth being the `(:429)` it cannot see. A second command
covering the gap now stands beside it.

The check that would have cleared this was itself vacuous: filtering
`grep -rnE ':[0-9]+'` output through `` grep -vE '(`|[A-Za-z0-9_./-]):[0-9]+' ``
matches grep's own `path.md:149:` prefix, discarding every line and exiting 1
whatever the tree contains. The corrected form compares the two `file:line` sets
with `comm -13`.

Every hit was re-resolved against this worktree. **None of the numbers in #117's
inventory comment were reused** — PR #251 rewrote the file that comment measured,
so all eleven of its spec-side line numbers were stale before this ticket started.

## Classification

| file | hits | classification | action |
|---|---|---|---|
| `2026-07-22-run-team-agent-fleet-design.md` | 1 | **valid** — `~/.claude/settings.json:3` resolves exactly at `origin/main` | none |
| `2026-07-23-fleet-plugin-design.md` | 27 | **9 stale live**, 2 valid (vendored), 1 wrong sibling, 15 historical-with-note | 9 re-anchored, 1 corrected |
| `2026-07-24-fleet-cockpit-design.md` | 0 | — | none |
| `2026-07-30-fleet-trust-the-label-design.md` | 51 | **all valid at the ref the document declares** (`6eb0442`) | base line split, deleted step named |
| `2026-08-06-review-pr-cost-and-apply-policy-design.md` | 34 | **all valid at `8a84402`**, but the ref was never named for them | ref named |
[...]

Issue #258, opened 2026-08-10
run-team: the fix-applier's refuter prompt instructs no scratch isolation, so sibling refuters collide on generic filenames in one shared scratchpad


feigi:
Observed live during a `/fleet:run-team` run, reported unprompted by a refuter subagent mid-verification.

## What happened

`fix-pr-247` dispatched refuters concurrently, one per `unverified` finding. Several independently built mutation matrices in the session scratchpad using the obvious generic names (`mut`, `mutA`, `mutB`, …). One of them reported:

> Mutation discrimination (private copies; the shared scratchpad's `mut*` names are in use by other agents, my first pass was contaminated and was redone)

It noticed, moved to private copies, and redid the pass — so the verdict it returned is sound. The cost was a wasted mutation pass. The cost in the case where a refuter does **not** notice is a wrong verdict on a finding, produced by measuring a sibling's mutant, and it would read exactly like a normal refutation.

## Why the existing rule does not reach them

`skills/fleet/skills/run-team/SKILL.md` already carries the rule, under **Guards**:

> **Per-member scratchpad subdirectory.** One flat namespace, generic filenames — one agent overwrote a sibling's `package.json`.

That rule is addressed to the controller about its *members*. Refuters are not members — they are grandchildren, dispatched by a fix-applier from the verbatim refuter prompt the skill dictates:

> Try to REFUTE this finding. Default to refuted=true if uncertain. Verify by RUNNING something — compile it, run the test, apply the mutation. Do not reason your way to agreement.

`Verify by RUNNING something` and `apply the mutation` actively instruct filesystem writes, and nothing in that prompt or the surrounding fix-applier block says where to write them. So the skill simultaneously requires the risky action and omits the isolation rule it states elsewhere for the same hazard.

This is the "instruction names no mechanism" shape: the rule exists, but not on the path that needs it.

## Scope of the gap

The refuter prompt is dictated verbatim in two places, and neither carries a scratch instruction:

- `skills/fleet/skills/run-team/SKILL.md`, in the fix-applier dispatch block
- `skills/fleet/commands/review-and-fix.md`, which the fix-applier is told to read

The `review-pr` workflow path is **not** affected in the same way — it provisions `scratch` per run and passes it down — which is part of why this only surfaces on the fix-applier's own hand-dispatched refuters.

## Suggested direction, not decided

[...]

PR #248, opened 2026-08-10
docs(fleet): compress the blocks #59 added to two skill files

feigi:
`sizing-a-ticket` and `next-ticket` are both `skills/**/SKILL.md`, so caveman compression is a required acceptance criterion on anything added to them — skill prose is re-read into context on every invocation, so uncompressed wording bills forever rather than once. The blocks #59 added carry articles, filler and hedging the surrounding prose has already dropped.

Triage ruled out the issue body's second option (restating the criterion in checkable form); the criterion is a standing workspace rule, not a line invented by one design document. `docs/specs/*.md` is human-read and exempt, so the design document that states the criterion is untouched.

## Scope

A pass over the blocks #59 added, and only those — derived from `git diff 8dcddad a4b500f` on the two files, not from the issue body. Prose #59 did not touch stays untouched; a wider compression pass is a separate diff.

- `sizing-a-ticket`: the fleet-member-on-heavy entry point, the report line, the tie-break note, and the Agent-Brief red flag.
- `next-ticket`: the FIFO ranking note under step 4. #59's other additions to this file (step 1) were already replaced by `candidates.mjs`, and its one-word edits elsewhere have nothing to compress.

322 → 307 words and 1089 → 1084 words, no line reflowed outside those blocks.

## Verification

Compression removes words, never meaning. Checked with a scratchpad anchor script over whitespace-normalized text, so rewrapping is invisible to it: one regex per instruction, constraint and exception the blocks state, plus a set comparison of every backticked identifier before and after — nothing dropped, nothing added. Green after the edit; its own regexes were caught red first, which is what showed they bite. Deliberately not committed — phrase assertions on prose are too brittle to live in the repo.

Suite: `./agent-test skills/fleet/scripts` → **319 pass, 0 fail**.

Closes #66

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DFqVZoHq7F31TCZvUELRTi

PR #289, opened 2026-08-10
docs(fleet): four rules the run measured, three of them controller errors

feigi:
Mid-run tooling fixes from the 7-ticket fleet run that merged #248, #249, #247, #251, #271, #272 and #273. Each closes a defect **this run produced**, in the file this run invoked. Suite **349/349** after.

Per `run-team/SKILL.md`'s own *Fix the tooling mid-run* rule — "two members failing the same way is the instruction, not the agents; fix the file during the run, deferring loses the evidence that found it."

### 1. Scratchpad paths need two levels, and nothing outside them

The rule said **per-member**, which is the wrong axis. Refuters are *grandchildren*, and finding ids restart at 1 every review — so two fix-appliers on different PRs both reached for `<scratch>/unv1`, and one had to redo a contaminated mutation pass. Separately, a probe built a **git repo at the checkout root** instead of in its scratch dir.

Both are "where may an agent write", so one rule now covers both: two levels, and nothing outside them. Related: #258 (widened twice during the run with both instances).

### 2. Never dispatch a finisher while you still owe the fix-applier a ruling

Nothing forbade it. The controller relayed two rulings *after* dispatching #272's finisher; the applier pushed them as a new commit, and the finisher correctly halted on a diverged head. The pinned SHA is only as good as the guarantee nothing else is inbound.

### 3. A 1-1 split is not a verdict

The text said to hand a refutation over "only when you reverse it" but never said *when* to reverse. **Three tied refutations were reversed and re-examined in this run; all three findings survived** — including the one that established `ci-state.mjs` could report `no-ci` and exit 0 on a repo whose CI was red.

### 4. A test you ADD must kill its own mutant

The fix-applier gate required the suite green, never that a *new* test discriminates. A pin added on the controller's own request survived the exact mutation it was named for; the vacuity was caught by a later refuter, not by the green suite that accepted it.

---

**Three of the four are controller errors, not member errors.** The members largely did what they were told; what they were told was underspecified.

**Not touched:** `settings.json`, permissions and `CLAUDE.md` (out of bounds for a mid-run fix), and the CI-monitor cadence gap — that one stays filed as #262 with measurements rather than redesigned on no evidence.


feigi:
Review fixes applied at `15d50c3`. Suite: `node --test skills/fleet/scripts/*.test.mjs` → **350 tests, 350 pass, 0 fail**.

**Applied (4 findings, 3 defects):**

- `references/isolation.md` rewritten onto the two-level axis with both new incidents — the pointer target no longer states the per-member rule this PR replaces. One defect, reported twice (correctness + comments); fixed once.
- The never-dispatch-a-finisher-while-you-owe-a-ruling rule added to the two Phase 3 event-loop bullets the controller actually acts from. It had landed only in the narrative finisher section.
- The added mutant gate gained its second direction (a change it should *not* catch, staying green), matching `review-and-fix.md:61`.
- New pin on both Phase 3 edges in `review-path-default.test.mjs`, mutation-verified both ways: deleting either gate reddens that test and only that test; rewording an unrelated clause in the same bullet stays green.

All three `refuted` findings were reversed by the controller — two split 1-1, and the third's decisive ground (that per-member subdirectories are a subset of `pr<N>/<finding>/`) is contradicted by this PR's own body: refuters are grandchildren, not members.

**Deferred (5 findings), one refuter each, all refuted:**

| Finding | Dimension | Issue |
|---|---|---|
| `votes` missing from the documented return shape | correctness | #294 |
| scratch path template matches nothing the tooling emits | correctness | #295 |
| `unv1` is not the attested finding-id form | comments | #296 |
| two reversal counts read as one anecdote | comments | #297 |
| how a fix-applier learns its `<scratch>` root | comments | commented on #258 |

Each issue carries its refutation. #258 got a comment rather than a duplicate: it already owns the grandchild scratch-isolation question, and the refutation answers two of its open triage questions — the harness injects a scratchpad root into every dispatched agent unconditionally, and that root is shared, which is why the bullet demands two levels beneath it.

Sixth suggestion (the mutant gate) survived its refuter and is in the commit above.

PR #290, opened 2026-08-10
fix(fleet): name the pass on candidates.mjs drop lines

### LINKED ISSUE #133: candidates.mjs: the fallback path logs every spec drop twice, so the drop count reads double
Deferred from PR #116 review.

## The defect

PR #116 moved `dropSpecs` to two call sites — once for the labeled query, once inside the `--allow-fallback` block. In production the unfiltered query is a **superset** of the labeled one, so every spec dropped in pass 1 is dropped and logged again in pass 2.

Measured at head with a production-shaped pair (unfiltered ⊇ labeled), 2 distinct specs:

```
    2 candidate(s) with label:ready-for-agent
    dropped #10 — to-spec spec, not a ticket (## User Stories)
    dropped #11 — to-spec spec, not a ticket (## User Stories)
candidates: empty with label:ready-for-agent; --allow-fallback given, retrying unfiltered
    3 candidate(s) unfiltered
    dropped #10 — to-spec spec, not a ticket (## User Stories)
    dropped #11 — to-spec spec, not a ticket (## User Stories)
    #12 [ready-for-human] t12
```

**4 `dropped #` lines for 2 specs.** An operator counting drops across the run sees double.

The test suite structurally cannot catch this: its two fixtures are deliberately **disjoint**, which real `gh` could never be. That disjointness is itself load-bearing — a faithful superset gives the fallback fixture 3 rows and the stub ignores `--limit`, so the cap tests stop firing. PR #116 records that constraint in the stub's comment.

The approach is genuinely open, and one of the options is "do nothing":

1. **Accept it** — the transcript is arguably honest. Two queries ran; each really did contain and drop those specs, and the per-query counts above each block are correct. Only a cross-run `grep -c` misleads.
2. **Dedupe** by issue number across passes — costs state the script does not currently keep.
3. **Label the passes** so the second block is visibly a re-scan of a superset.

Option 1 is defensible enough that this should not be handed to an implementer as decided work.

## Severity

Low. Cosmetic and operator-facing only — no payload, exit code, or filtering behaviour is affected. Raised because the file's stated ethos is that every drop is logged so a false positive is loud rather than silent, and an inflated count is a mild step away from that.

## Blocked by

- None

feigi:
## Summary

The labeled and fallback queries in `candidates.mjs` both log dropped to-spec
specs with an identical, unqualified message. In production the unfiltered
fallback query is a superset of the labeled one, so every spec dropped in
pass 1 is dropped and logged again in pass 2 — an operator's cross-run
`grep -c` counts each spec twice.

Per triage's ruling on #133 — annotate, don't dedupe, don't silence — this
adds a `pass` label to `dropSpecs()` and appends it to the drop line. The
two passes already have distinct header lines; only the drop lines were
anonymous. Nothing is deduplicated or suppressed: a spec dropped by both
passes is still reported by both, now distinguishably.

- `dropSpecs(rows, pass)` — pass label required, not defaulted
- Labeled pass: `[label:<label>]` (or `[unfiltered]` when no `--require-label`)
- Fallback pass: `[unfiltered (fallback)]`, matching the wording `refuseIfCapped` already uses for that query
- Payload, exit codes, and filtering behaviour unchanged

## Test plan

- [x] New test pins the message shape per pass (fixtures stay deliberately
      disjoint per the file's existing constraint, so the superset case
      itself isn't reproduced — only that each pass's drop line is
      attributable to it)
- [x] `node --test skills/fleet/scripts/*.test.mjs` — 350/350 green

Closes #133


feigi:
## Review applied — 13 findings adjudicated

**Applied (4 commits):**

| SHA | What |
|---|---|
| `84451cd` | The new test's comment blamed **#12** for emptying the labeled query; **#10** does it. #12 is fallback-fixture-only and is *kept*, not dropped. (survived: `comments`, `types`) |
| `0ba04a8` | Retitled the test: `grep -c 'dropped #'` is **4 both before and after** this PR, so the old title's "must not double-count" claimed a count the change does not move. Retitled to the failure mode it really pins. (`correctness`) |
| `2dd8a03` | New test pinning the `[unfiltered]` pass tag — the one arm no test reached. Renaming it left the **full glob byte-identical** at 350/349/0/1. (`tests`, `correctness`) |
| `2dc086c` | Folded the drop-attribution asserts into the `:185` fallback test — its stderr is a strict superset, and folding discriminates *more*. Net −14 lines, one fewer subprocess. (`simplify`) |

Gate on each commit: `node --test skills/fleet/scripts/*.test.mjs` → **tests 350 / pass 350 / fail 0**, run serially from the worktree. Never `--no-verify`.

Every applied `suggestion` got one refuter dispatched against it, biased to refuse. Three returned `refuted=false`, one returned `refuted=true` and was deferred instead.

**Deferred and filed:**

- **#299** `ready-for-agent` — `die()`'s `writeSync` throws EAGAIN on a full stderr pipe, so `process.exit(2)` never runs and the process exits **1**, inverting the documented contract. Measured: pipe → exit 1, 65536 bytes, 0 refusal lines; file → exit 2, 266307 bytes, 1 refusal line. Pre-existing, and its own suggested fix argues for a separate ticket. **Not a duplicate of #132** — that is stdout payload truncation at exit 0; this is stderr plus an exit-code inversion. (`silent-failure`)
- **#300** `needs-triage` — hoist one `scope` const; the labeled-vs-unfiltered conditional is spelled three ways in 14 lines. Out of #133's scope: it rewrites `:256`/`:257`/`:269`, lines this ticket has no business in. (`simplify`, `correctness`)
- **#301** `needs-triage` — "`pass` required, not defaulted" allegedly unenforced. **Refuted**: deleting the argument at either call site fails an existing test loudly with `[undefined]`, the comment never claims a guard, and the proposed guard is provably unreachable. Recommend closing; filed so the reasoning is on record. (`silent-failure`, `comments`, `types`)
- **#302** `needs-triage` — two stderr-timing tests flake under parallel load. Not a review finding; surfaced independently by two refuters. Matters more now that `2dc086c` put the attribution pins inside one of them.

**No action:** the `types` finding at `candidates.mjs:null` reports that "casts that erase conformance" does not apply — both files are plain JS, no defect claimed, `suggested_fix: null`.

`rebase-check` is red off `behind_by=3`. That is staleness from siblings landing, not a failure of this diff — `check` is green. Not rebasing.

PR #292, opened 2026-08-10
fix(fleet): no-undo-audit reports stash as unknown, not zero, when the reflog can't be trusted

### LINKED ISSUE #147: no-undo-audit.sh:72 the stash count reads 0 for an unknown count — and git stash list exits 0 while doing it

feigi:
## Summary

`git stash list` prints nothing at rc 0 when the reflog behind `refs/stash`
is unreadable (`chmod 000 .git/logs/refs/stash`, measured), and prints
nothing at rc 1 for a corrupted stash ref (`fatal: bad object refs/stash`,
measured — the pipeline's exit status was always `wc`'s, never git's, so an
rc capture on the old pipeline would not have caught this either). Both were
indistinguishable from a genuinely empty stash, and `0` is the one value
`run-merge-bot.md` reads as nothing to look at.

Per triage's ruling on #147: cross-check `refs/stash` against the list.
If the ref resolves but the list comes back empty, the two disagree and the
count is reported as `unknown` (`stash: null` in the payload, stated on
stderr) rather than a silent zero. A genuinely empty stash has no ref
either, so it still reports zero. The stash count stays reported-only, per
the surrounding comment block, which is unchanged and was not "simplified"
into the false absolute claim that a rebase cannot reach the stash.

Named as the remaining ceiling, not closed: a reflog that is truncated but
still parses (some entries lost, the rest intact) resolves the ref and
returns a nonempty list, so the cross-check sees no disagreement and reports
the too-low count as exact.

## Changes

- `skills/fleet/scripts/no-undo-audit.sh` — cross-check `refs/stash` via
  `rev-parse --verify --quiet` against an empty `stash list`; `stash: null`
  on disagreement, `stash: 0` when genuinely empty
- `skills/fleet/scripts/no-undo-audit.test.mjs` — pins the unreadable-reflog
  case and the corrupted-ref case against the genuinely-empty one, plus that
  a dirty worktree still refuses (unknown never gates, same as before)
- `skills/fleet/commands/run-merge-bot.md` — operator now reads the stash
  list when the count is nonzero **or unknown**, not only nonzero
- `docs/specs/2026-07-23-fleet-plugin-design.md` — script-surface row
  documents the `null` value, same pattern as `ci-state.mjs`'s `behind: null`

## Test plan

- [x] `node --test skills/fleet/scripts/no-undo-audit.test.mjs` — 35/35 pass
- [x] `node --test skills/fleet/scripts/*.test.mjs` — full suite, 352/352 pass
- [x] Both traps reproduced by hand before implementing (chmod 000 reflog;
      corrupted stash object) per the ticket's instruction

Closes #147


feigi:
## Review applied — `0a6c473`

`node --test skills/fleet/scripts/*.test.mjs` from the worktree: **tests 353 / pass 353 / fail 0 / skipped 0**. No `--no-verify`.

### Applied (5 findings, 4 distinct defects)

| # | dimension(s) | what |
|---|---|---|
| 1 | correctness + silent-failure (**survived**, 2/2 refuters `refuted=false`) | `show-ref` replaces `rev-parse --verify --quiet` as the stash discriminator. An unreadable `refs/stash` file reported a confident `"stash":0` with a real entry on the stack — `rev-parse` exits 1 there for the same reason the list is empty, collapsing it onto the genuinely-empty branch. `show-ref` returns rc 1 **only** for a genuinely absent ref, so `-ne 1` separates all five states. Reproduced and re-verified independently in this tree before applying. |
| 2 | simplify (refuted=false) | Count first, then correct the one ambiguous value — 9 lines for 15, one nesting level and one variable gone, plus the discarded `\|\| stash_list=""` rc capture that read like a guard that mattered. Byte-identical stdout/stderr/exit across 9 states × 3 shells. |
| 3 | correctness + silent-failure (refuted=false) | Root guards use `return t.skip(...)`, not a bare `return` — `node --test` scores the latter as a **PASS**, so under root the new #147 coverage was vacuous green. Verified by forcing `getuid` to 0: before, `pass 35 / skipped 0`; after, `skipped 3`. Line 301's pre-existing guard went with them (same defect, same file). |
| 4 | silent-failure (refuted=false) | The payload `printf`'s comment called `$stash` "a digit count"; that slot is unquoted, so the sentence names a bare JSON token type — and it excluded the `null` the field can hold. It is the only place in the file a future editor is told this field's shape. |

`run-merge-bot.md:133` and the spec's script-surface row were reworded because fix 1 made "the stash ref resolves but the list came back empty" false for the new third state. The runbook's "read `git stash list` yourself" advice is **deliberately unchanged** — see #303.

A fourth test pins the new case; the suite passed with the hole open. Mutation-checked both ways: reverting the discriminator to `rev-parse` reddens that test **and only that test** (1 fail of 353); an unrelated at-risk mutation reddens six others and leaves it green.

`show-ref` is also right where a `test -e` on the ref path would not be — a **packed** `refs/stash` has no loose file, so packed + unreadable reflog would read as "no stash" and print the same confident zero (measured). `pop`, `drop`, `clear`, a never-stashed clone and a linked worktree all still report `0`/the count correctly.

### Deferred (5 issues)

| issue | label | dimension(s) | why deferred |
|---|---|---|---|
| #303 | `needs-triage` | correctness | **Refuted.** `git stash list` *does* answer in the object-gone state (`fatal: bad object refs/stash`), "corrupted reflog" is git's own vocabulary (`fsck`: `invalid reflog entry`), and both suggested replacements report **all-clear at rc 0** in the unreadable-reflog state. Filed with the refutation; what remains open is whether per-state advice is worth the runbook's length. |
| #304 | `needs-triage` | silent-failure | Out of #147's scope (new diagnostic capability). The suggested fix no longer applies as written — the restructure removed `stash_list` — and git emits a diagnostic in only 1 of the 3 `unknown` states. |
| #305 | `ready-for-agent` | silent-failure | Truncated-reflog ceiling is in the script comment but not the runbook. Out of #147's scope by the brief; exact one-clause edit specified. |
| #306 | `needs-triage` | tests + types | Truncated-but-parseable reflog reports a too-low count as exact. Out of scope by #147's brief; both reporters said "no action required". A truncation-safe check needs real complexity. |
| #307 | `needs-triage` | types | `stash` is `number \| null`; a future `if (json.stash)` would erase the distinction. No consumer exists to protect yet — conditional on work that does not exist. |

One reported finding (correctness, `no-undo-audit.sh:202` — "the comment's stated ceiling is wrong, the unreadable ref file is a second one") is **resolved rather than deferred**: fix 1 closes that case, so the comment naming the truncated reflog as the sole remaining ceiling is accurate again.

Each in-scope `suggestion` got one refuter dispatched against it, biased to refuse, before anything was applied — the `suggestion` band is budgeted 0 refuters by the review workflow, so it arrived unchecked. All mutation work was done on `git archive` copies outside this worktree.

feigi:
## Scope correction — `c6f2306`

`node --test skills/fleet/scripts/*.test.mjs`: **tests 353 / pass 353 / fail 0 / skipped 0**.

The `t.skip` fix was applied at three sites in `0a6c473`. Two of them (`:252`, `:305`) guard tests this PR added and stay. The third guards the pre-existing `git status` gate test — same defect, but not #147's — so it is **reverted here and filed as #308** (`ready-for-agent`, with the refuter's forced-root measurement in the body).

The test-file diff against `main` now adds three `t.skip` guards and removes none, so it touches only the tests this PR owns.

Deferrals for this PR are now **#303, #304, #305, #306, #307, #308**.

Issue #277, opened 2026-08-10
review-pr.js:413 — replace the per-field !entry guard with optional chaining (now test-pinned by #271)


feigi:
`REQUIRED.filter((f) => !entry || !entry[f])` re-evaluates a loop-invariant null-guard per field; optional chaining says it once.

**dimension:** simplify (`unverified-10`)
**file:** `workflows/review-pr.js:413`

## Finding

    const missing = REQUIRED.filter((f) => !entry || !entry[f]);

`!entry` does not vary across the three fields, so it is evaluated three times to answer one question. `REQUIRED.filter((f) => !entry?.[f])` expresses the same guard once and drops the disjunction. Optional chaining is already house style in this same file — `workflows/review-pr.js:387` uses `stats.kinds?.docs !== 0`.

## Evidence already gathered

A differential harness lifting the real `resolveDimensions` out of the source text reported `diffs=0` over 47 inputs between the current form and `!entry?.[f]`, including exactly the cases the `!entry` term exists for — `[null]`, `[undefined]`, `[0]`, `[false]`, `[NaN]` — all still yielding `review-pr: args.dimensions object is missing required field(s): key, prompt, agentType` rather than a `TypeError`, with object identity preserved for resolved keys. Applied to a snapshot copy, `node --test "skills/fleet/scripts/*.test.mjs"` matched baseline exactly.

## Note for whoever picks this up

PR #271 **added a test that pins this clause**: `select-dimensions.test.mjs`, *"a null or undefined entry stops the run instead of crashing on the field check"*, asserts `resolveDimensions([null], …)` and `resolveDimensions([undefined], …)` throw `/missing required field\(s\): key, prompt, agentType/`. It was mutation-verified — deleting `!entry ||` turns it red with `TypeError: Cannot read properties of null (reading 'key')`, and a decoy mutation elsewhere in the function leaves it green.

So this refactor is now covered: `!entry?.[f]` should keep that test green. **If it goes red, the refactor is wrong, not the test.** Before this PR there was no such coverage and the swap would have been unpinned.

## Why deferred

Style preference with no behavioural difference, on a validator that shipped in the same PR. Not worth another CI cycle on #271. **Not adversarially refuted** — the `suggestion` band gets 0 refuters by policy and this one was deferred rather than checked, so the equivalence measurement above is the specialist's, unreplicated.

Deferred from PR #271 review.

PR #293, opened 2026-08-10
fix(fleet): derive review-pr.js's testCmd from the reviewed repo

### LINKED ISSUE #142: review-pr.js: the default testCmd is claude-config-specific, so in any other repo it is a silent green
Deferred from PR #130 review. **Approach still open — triage needed.**

PR #130 changed the default at `workflows/review-pr.js:120` from `./agent-test` to `node --test skills/fleet/scripts/*.test.mjs`. That is correct for this repo and fixes #101. But `~/.claude/workflows/` is **global config**, and the fleet is portable by design — `worktree` is a caller-supplied argument, and nothing pins the workflow to this repo.

Measured on a repo without that path, `bash`/`sh`:

```
node --test skills/fleet/scripts/*.test.mjs   ->  ℹ tests 0   EXIT=0
```

**The old `./agent-test` failed loudly off-repo (exit 127).** So for foreign repos this change converts a loud failure into a silent pass — it relocates the silent-failure mode rather than removing it. The only guard is the reading rule in the prompt, which is model compliance, not mechanism.

Scope of the claim: no caller anywhere in this repo passes `args.testCmd` — grep over all tracked files at `1e158b7` finds it only at `:120` (the definition), `:249` (the prompt interpolation), and `review-and-fix.md:33`/`:43` naming it. So every real caller takes the default.

The guard added in #130 does not cover this: `review-pr-testcmd.test.mjs` cuts an archive **of this repo** and runs the default in it, so it proves the default works here and says nothing about anywhere else.

**Options, none decided:**
1. Derive the default from the target repo (read its `package.json` test script, or reuse `claim-ticket.sh`'s existing runner-inference, which already does exactly this inference).
2. Make `testCmd` required and refuse to run without it — loud, no default to be wrong.
3. Keep the default but have the workflow detect a zero-test run mechanically (see the `test_run` schema issue) so the silence becomes a failure.

Option 1 has an obvious reuse candidate and is probably right, but it changes the workflow's contract, so it wants a decision rather than a guess.

Found by the code-review and type-design specialists on PR #130, and independently by this reviewer.

feigi:
## Summary

`review-pr.js`'s `testCmd` defaulted to a literal string naming this
repo's own test path. The workflow is global config and `worktree` is a
caller-supplied argument, so in any other repository that default matched
no files and exited 0 — a silent green where the previous default
(`./agent-test`) failed loudly off-repo.

Per triage's ruling on #142 (option 1, implemented through the snapshot
agent, with refusal as the fallback):

- Extracted claim-ticket.sh's existing test-entrypoint inference (a
  manifest test script, else a direct test-file run, else refuse) into a
  standalone `skills/fleet/scripts/derive-testcmd.sh`, parameterized by
  repo + ref. `claim-ticket.sh` now calls it instead of carrying its own
  copy.
- `review-pr.js`'s snapshot agent runs `derive-testcmd.sh` against the
  repo under review (ref `HEAD`) and reports `testCmd`/`testCmdError`
  back alongside its other optional facts.
- `resolveTestCmd(explicit, snap)` prefers an explicit `args.testCmd`
  override, else the derivation, else **throws** — the review refuses to
  start rather than falling back to any default, including when the
  derivation itself can't be established or the workflow can't read it.
- Corrected two doc claims (`review-and-fix.md`, `run-team/SKILL.md`)
  that had review-pr.js "defaulting" its `testCmd` to this repo's glob
  string, now stale.

## Test plan

- [x] `derive-testcmd.test.mjs` — new, exercises the shared script
  against synthetic repos that are **not** this one (manifest test
  script, test-files-only, neither → refuse, non-repo dir, ref-scoped
  derivation), including actually running the derived command and
  reading the real test count.
- [x] `claim-ticket.test.mjs` — updated for the extraction; all prior
  cases (install + testcmd matrix, runner directory-shim behavior)
  still pass.
- [x] `review-pr-testcmd.test.mjs` — rewritten: lifts and unit-tests
  `resolveTestCmd`'s resolution order/refusal, pins its call site, pins
  that the snapshot prompt/schema carry the derivation, keeps the
  existing "specialist prompt hands testCmd over verbatim" pin.
- [x] Mutation-checked: reverting the `resolveTestCmd` call site to a
  hardcoded default fails both the call-site pin and the
  no-hardcoded-default pin.
- [x] Full suite: `node --test skills/fleet/scripts/*.test.mjs` → 365
  pass, 0 fail. Also re-run via the worktree's `./agent-test` after
  rebase — 365 pass, 0 fail.
- [x] `workflows/review-pr.js` parses (`AsyncFunction` construction from
  source, since it's neither a valid module nor script for
  `node --check`).

Closes #142


feigi:
## Review adjudication — 22 findings, pushed as `541a4e2`

**Tests:** `node --test skills/fleet/scripts/*.test.mjs` from the worktree → `tests 369 / pass 369 / fail 0 / skipped 0`. Every added or tightened pin was mutation-checked: red on the defect it names, green on an unrelated edit.

### Applied (7 findings → 6 distinct defects)

| Defect | Reported as | Fix |
|---|---|---|
| `review-pr-testcmd.test.mjs:130` schema pin vacuous against a block-commented field | `[critical] tests` | lifted `review-pr-reads.test.mjs`'s stripper into `skills/fleet/scripts/strip-comments.mjs`; both files now pin against a comment-stripped `CODE` |
| `claim-ticket.sh:70` `die "$testcmd"` prints an empty reason | **six** reports: `silent-failure:70`, `tests:70`, `comments:69`, `types:69`, `correctness:67`, `simplify:68` | `2>&1` on the capture, matching the install probe 20 lines above; test now anchors `/claim-ticket: .*pass vacuously/` |
| `derive-testcmd.sh:37` pipeline status is grep's — a dying `ls-tree` reads as "no test files" | `[important] silent-failure` | materialize the listing, `die "cannot list $ref — …"`; two new tests (unknown ref, unborn HEAD) |
| `derive-testcmd.sh:35` unparseable manifest silently degrades to `node --test` | `[important] silent-failure` + `correctness:35` | three outcomes (has script / parses-without / does not parse), refuse on the third; new test |
| `derive-testcmd.sh:33` `testfile_re` duplicated with no sync check | `[important] types` | test asserting the two literals are byte-identical |
| `review-pr.js:459` derived `npm test --` exits 127 in the `git archive` snapshot | `[important] correctness` | snapshot agent symlinks `node_modules` when the worktree has one; pinned |
| `review-and-fix.md:8` "the same command" — the derivation returns bare `node --test` | `comments:8` + `correctness:8` | "the same *effective* command", matching line 47 |

### Deferred (6 filed, all refuted)

Every in-scope `suggestion` got one refuter, biased to refuse, verifying by running. All six came back `refuted=true`, so none was applied. Refutation is in each issue body.

- #309 — `review-pr.js:705` the derived `testCmd` is not returned (*correctness*)
- #310 — `resolveTestCmd` accepts `snap.testCmd` unchecked (*silent-failure*) — the suggested allowlist is feature-breaking: the script emits `node --test\n`, trailing newline
- #311 — `resolveTestCmd` does not `typeof`-check the explicit override (*types*) — pre-existing verbatim at `HEAD~2`
- #312 — `resolveTestCmd (below)` wording (*comments*) — matches the repo's own established "(below) = downstream use" idiom
- #313 — duplicated `testCmd` tombstone comment at `:280` (*simplify*) — it is an absence marker, not a duplicate
- #314 — `snap?.` optional chaining (*simplify*) — related to #277

### Not filed

- `candidates.test.mjs:485` flake (`status 1 !== 2`, *silent-failure*) — already tracked as **#298**; this review is the third independent sighting. Pre-existing, untouched by this PR.
- `derive-testcmd.sh:27` "delete the `pkg` variable and pipe `git show` straight into node" (*simplify*) — **superseded, not deferred**. The applied parse-failure fix needs `[ -n "$pkg" ]` to tell "no manifest" from "unparseable manifest"; piping directly makes an absent `package.json` reach `JSON.parse("")` and refuse. Filing it would propose reintroducing the bug this PR closes.

PR #321, opened 2026-08-10
fix(fleet): classify finish-<n> agentType as finisher in compute-spend

### LINKED ISSUE #318: compute-spend.mjs:46 — classifyRole misses the real finisher naming convention, so finishers land in `other` and deflate every role's share
Recovered from the stale `fleet/cockpit-spend-fixes` worktree during a `/fleet:run-team` run. That branch is 340 commits behind `origin/main` and everything else on it has since landed by other routes — **this is the one fix that did not**, so it is filed rather than PR'd.

## The defect

`skills/fleet/scripts/compute-spend.mjs:46` at `origin/main`:

```js
  if (/finish pr|finisher/.test(hay)) return "finisher";
```

The controller names finishers `finish-<n>` with description `Finish-label #<n>` (`run-team/SKILL.md` — the finisher is dispatched per PR). Neither `finish pr` nor `finisher` appears in either field, so every real finisher falls through to `other`.

## Measured — by running main's own classifier

`compute-spend.mjs` extracted from `origin/main` and called directly:

```
{"agentType":"finish-436","description":"Finish-label #436"} -> other
{"agentType":"finish-315","description":"Finish-label #315"} -> other
{"agentType":"finisher","description":"finisher"}            -> finisher
{"agentType":"impl-299","description":"Implement #299"}      -> implementer
```

Only the literal string `finisher` classifies. The naming convention actually in use does not.

## Why it matters beyond a mislabelled row

`finisher` is in `ROLE_ORDER`, so the role renders — as a permanently empty or near-empty row — while its tokens are counted under `other`. Every other role's **percentage share is deflated** by the finisher spend attributed elsewhere. `run-team/SKILL.md`'s phase-2 tier guard reads `board.mjs build`'s `.spend.roles` for the implementer-vs-review split, so the number the cost guard is evaluated against is wrong in a direction that is not obvious from the board.

It fails by mis-attributing, never by throwing, so nothing surfaces it.

## The fix

One alternation, at `compute-spend.mjs:46`:

```js
  if (/^finish-|finish pr|finisher/.test(hay)) return "finisher";
```

`^finish-` anchors on the convention. Confirm what `hay` is built from before anchoring — if it is a concatenation of several fields rather than the agentType alone, `^` may not sit where you expect, and an unanchored `finish-` risks matching unrelated text.

## Acceptance criteria

- [ ] `finish-<n>` / `Finish-label #<n>` classifies as `finisher`
- [ ] The existing literal `finisher` and `finish pr` cases still classify
- [ ] A test pins a realistic `finish-<n>` meta — and **kills its own mutant**: reverting the regex must turn it red, and a decoy change elsewhere must leave it green
- [ ] `impl-<n>` and the other roles are unchanged (no new false positives)

## Provenance

The original commit `be0dd192` also claimed "28 of 52 real finishers on this machine" landed in `other`. That count is **not** re-verified here — treat it as the original author's measurement, not an established fact. The regex gap itself is verified above by execution.

## Do not

Do not resurrect the `fleet/cockpit-spend-fixes` branch to fix this. Applying it against current `main` is a 73-file, ~12000-deletion revert.

feigi:
## Summary
- `classifyRole` at `compute-spend.mjs:46` only matched literal `finish pr` / `finisher`, but `run-team` dispatches finishers as `agentType: finish-<n>` / `description: Finish-label #<n>` — every real finisher fell through to `other`, deflating every other role's percentage share.
- Fix: `/^finish-|finish pr|finisher/` — anchors on the actual naming convention.
- `hay` is `` `${type} ${desc}`.toLowerCase() `` with `agentType` concatenated first, so `^finish-` sits exactly where the convention lives; no risk of matching mid-string content in other roles since only strings where `agentType` itself starts with `finish-` can match at that anchor.

## Test plan
- [x] Added a pinned test for `{agentType: "finish-436", description: "Finish-label #436"}` → `"finisher"`.
- [x] Confirmed red pre-fix (18 pass / 1 fail — only the new test).
- [x] Applied fix → 19/19 green.
- [x] Mutant kill: reverted regex → exactly the pinned test went red (18 pass / 1 fail), nothing else. Restored → 19/19 green.
- [x] Decoy check: unrelated change to the merge-bot alternation → pinned test stayed green (19/19), proving it discriminates rather than being fragile to any diff. Decoy reverted before commit.
- [x] Existing literal `finisher` / `finish pr` cases, `impl-<n>`, `review-pr-<n>`/`fix-pr-<n>`, `merge-bot`, and other roles unchanged — full suite (19/19) still green, no new false positives.

Closes #318


feigi:
Review adjudication — applied 2, deferred 3. Pushed `73b6869`.

**Applied**
- **tests / important (`survived`)** — `compute-spend.test.mjs:133` fixture was redundant on the axis that mattered; strengthened so it discriminates `agentType` from `description`, plus a second assertion pinning the `^` anchor.
- **correctness / suggestion** — `compute-spend.mjs:16` now records that the finisher name is an *observed* convention, not one run-team fixes.

**Deferred and filed**
| # | Finding | dimension | Why deferred | Label |
|---|---------|-----------|--------------|-------|
| #325 | `board.mjs:114` empty catch labelled `/* unnamed agent */` swallows genuine `meta.json` faults | silent-failure | Pre-existing; `board.mjs` is not in this diff | `ready-for-agent` |
| #326 | `run-team/SKILL.md:34-35` documents no finisher name, but `:46` classifies on one | correctness | `SKILL.md` not in this diff; three live spellings make the canonical name a decision | `needs-triage` |
| #327 | `compute-spend.mjs:46` finisher branch ordered after the reviewer branch | silent-failure | **Refuted on measurement** — 0/127 real finishers hit the reviewer regex, hoist is a 0-change no-op over 3735 metas | `needs-triage` |

One finding needed no action: the `^` anchoring against `hay` means a description-only finisher sidecar falls through to `other`, but no such sidecar exists and the anchored form's measured miss rate is 0/3717.

**Verification** — `node --test "skills/fleet/scripts/*.test.mjs"` → `tests 376 / pass 376 / fail 0`. Mutation-checked the changed test (probes run on a private copy, not the worktree):

| Mutation | Expected | Measured |
|---|---|---|
| revert `:46` to pre-PR `/finish pr\|finisher/` | red | `fail=1` |
| `test(desc.toLowerCase())` — ignores agentType | red | `fail=1` |
| drop the `^` anchor | red | `fail=1` |
| unrelated: narrow the reviewer regex | green | `pass=1` |

Note the middle two both stayed **green** under the review's own suggested fixture, which is why the fix went further than the suggestion.

Issue #298, opened 2026-08-10
candidates.test.mjs: two different tests go red under concurrent suite runs — the fleet runs suites concurrently by design, so a real failure and a flake are indistinguishable


feigi:
Observed 2026-08-10 during one `/fleet:run-team` run, by **two independent refuters** working on separate `git archive` copies, on **two different tests**. Neither was looking for a flake; both hit one and re-ran to confirm.

## The two sightings

**1. `the refusal survives a gh stderr larger than the pipe buffer`**

Reported red on a full-glob run, green when re-run serially. The refuter attributed it to "racing my concurrent invocations".

**2. `assert.equal(stderr.match(/^\$ gh /gm).length, 2)`** — inside `a labeled queue of only specs falls back to unfiltered` (`candidates.test.mjs:185`)

```
actual 1, expected 2
```

under heavy parallel load. That refuter then re-ran **3× baseline and 3× folded, all clean**:

```
baseline, full gate   tests 350 / pass 349 / fail 0 / skipped 1   ×3
folded,   full gate   tests 349 / pass 348 / fail 0               ×3
```

Both failures are shape-consistent: an assertion about **how many `gh` invocations were observed**, or about stderr surviving a pipe, coming back short. Both involve the stub `gh` and captured stderr.

## Why this matters more here than in an ordinary repo

The fleet **runs this suite concurrently by design**. Every fix-applier gates on it before committing, and every refuter it dispatches runs it again in its own copy — this run had four refuters plus a fix-applier live at once. So the concurrent case is not an edge case, it is the normal case.

The consequence is symmetrical and bad in both directions:

- a **flake reads as a real failure**, and the standing rule is that a red suite blocks the commit — so an applier either defers a good fix or, worse, re-runs until green
- a **real failure reads as a flake**, because "it was probably a neighbour" is now an available and often-correct explanation. One refuter this run did exactly that, and it happened to be right; nothing about its reasoning would have caught the case where it is wrong

That second direction is the dangerous one. The controller's standing guidance — *"suspect a neighbour before a member's own diff"* — is correct for wrong *findings*, and it becomes a hazard when applied to a red *gate*.

## Note for whoever takes this

Sighting 2 lives in the exact test that PR #290's accepted `simplify` finding **folds another test into**. Post-fold, the same flake will surface as the folded test's failure — so a future reader will see the new assertions blamed for a pre-existing intermittent failure that has nothing to do with them.

## What is open

[...]

Issue #176, opened 2026-07-31
candidates.mjs die() duplicates gh stderr unbounded — measured 1.06 MB into the caller on one failed query


feigi:
Deferred from PR #167 review. Pre-existing on `main`, untouched by #167.

## What to build

`skills/fleet/scripts/candidates.mjs:92` interpolates the whole of `gh`s stderr into its own error message:

```js
die(`gh issue list failed: ${String(e.stderr || e.message).trim()}`);
```

Measured:

```
gh stderr ~7,700 B      -> e.stderr.len=7,500      -> die() prints 7,499 bytes
gh stderr ~1,540,000 B  -> e.code=ENOBUFS
                           e.stderr.len=1,114,112  -> die() prints 1,114,112 bytes
```

~1.06 MB, roughly 270K tokens, into whatever read the script — and the callers are markdown read by a model, so this lands in a context window.

Two things make it worth closing:

1. **It is pure duplication.** `execFileSync` with default stdio already forwards the childs stderr to the parent, so the bytes were emitted once before `die()` re-emits them. End-to-end, `gh` producing 7,700 B yielded 15,409 B of script stderr.
2. **It contradicts the files own discipline.** `candidates.mjs:119-122` deliberately names a bad row by index and never prints it, on the stated grounds that an unreduced payload carries every issue body — "~97% of what this file refuses to fetch in the first place". The header makes the same argument. This path can emit more than the payload the script declined to fetch.

## Acceptance criteria

- [ ] A failing `gh issue list` no longer re-emits the childs stderr from `die()` — dropping the interpolation is the recommended fix, since the bytes already reached the caller; cap it (e.g. `.slice(0, 2000)`) only if the message is shown to be lost otherwise
- [ ] Exit code stays 2 — this is the archetypal fail-closed path and must not change meaning
- [ ] A test covers the failed-query path, which is currently the only `die()` in `query()` with no test; `JQ_OVERRIDE: "((("` in the existing stub reaches it in one line

## Notes

Found by the `silent-failure-hunter` specialist during the #167 review; measurements above are its. The `JQ_OVERRIDE` seam that makes the test cheap was added by #167.

PR #322, opened 2026-08-10
fix(fleet): die() keeps exit 2 when stderr write throws EAGAIN

### LINKED ISSUE #299: candidates.mjs: die()'s writeSync throws EAGAIN on a full stderr pipe — exit 2 inverts to exit 1, refusal discarded
Deferred from PR #290 review. Dimension: **silent-failure**. Severity: important. Adversarially verified — both refuters returned `refuted=false`.

## What to build

`skills/fleet/scripts/candidates.mjs:33` — `die()` writes its refusal with `writeSync(2, …)` and then calls `process.exit(2)`:

```js
function die(msg) {
  process.exit(2);
}
```

On a **pipe**, once ~64 KiB of forwarded `gh` stderr is already queued, that `writeSync` throws `EAGAIN`. The throw is uncaught, so `process.exit(2)` **never runs** and the process exits **1** — the code this file reserves for "query fine, queue empty". The refusal line and every `dropped #…` line are discarded with it.

That inverts the contract stated at `skills/fleet/skills/next-ticket/SKILL.md:28` — *"Exit 1 = query fine, queue empty. Exit 2 = query broke. Different facts."* A caller reading only the exit code is told the backlog is empty when the query actually broke.

Both production callers are exactly the pipe case: `skills/fleet/skills/next-ticket/SKILL.md:15` and `skills/fleet/skills/run-team/SKILL.md:60` invoke the script from a Bash tool call.

### Measured

`gh` stub emitting ~250 KB to stderr then `exit 1`, against the script at PR #290's head:

| sink | node exit | stderr bytes captured | refusal lines |
|---|---|---|---|
| pipe (`2>&1 >/dev/null \| { sleep 2; cat; }`) | **1** | **65536** | **0** |
| file (`2>file.txt`) | 2 | 266307 | 1 |

Mechanism isolated by wrapping *only* `die()`'s `writeSync` in try/catch and re-running the identical pipe command: node then exits **2**, and the catch records `EAGAIN: resource temporarily unavailable, write`. So the exit-1 is the uncaught `EAGAIN` escaping before `process.exit(2)`.

`grep -rn EAGAIN skills/` returns nothing — nothing handles this anywhere else.

### Verified fix — keep the exit code even when the write cannot land

The exit code is the machine-readable half of the contract; the text is not.

```js
function die(msg) {
  try { writeSync(2, `\n${NAME}: ${msg}\n`); } catch { /* EAGAIN: stderr pipe full; the exit code below is the contract */ }
  finally { process.exit(2); }
}
```

Measured to restore the documented exit 2 on the pipe path.

### Explicitly out of scope

Recovering the refusal **text** as well needs a short `EAGAIN` retry loop around the `writeSync` before giving up. That is a separate decision and is **not** part of this ticket — the exit code is. File a follow-up if the text turns out to be worth the retry.

## Why deferred

Pre-existing: `die()` is untouched by PR #290's diff. PR #290 is a 4-line change to a drop **message**; a control-flow fix to the process's exit contract does not belong on it. The finding's own suggested fix says so: *"Either way this belongs in its own ticket, not as a blocker on a 4-line message change."*

Stated fairly, it does defeat PR #290's stated purpose on the large-stderr path: the same 64 KiB truncation swallows the `[label:…]` / `[unfiltered (fallback)]` attribution #133 adds, and the cross-run `grep -c` the new test exists to protect silently under-counts. That is a reason to fix it soon, not a reason to fix it in #290.

## Not a duplicate of #132

#132 is the **stdout payload** truncating at 64 KiB because `process.exit()` does not drain an async pipe write — a silently short list at **exit 0**. This is **stderr**, a *synchronous* `writeSync` **throwing** `EAGAIN`, and the symptom is an **exit-code inversion (2 → 1)**, not a short payload. Same 64 KiB pipe-buffer family, different fd, different mechanism, different fix (`process.exitCode` there, `try/catch` + `finally` here). Fixing #132 does not fix this.

## Acceptance criteria

- [ ] With the stderr pipe full, `die()` still exits **2**, not 1
- [ ] A test drives the pipe path with a large-stderr `gh` stub and asserts the exit code
- [ ] Exit codes otherwise unchanged: 0 with survivors, 1 empty, 2 broken
- [ ] The file-redirect path still emits the full refusal text

## Blocked by

feigi:
## Summary

`die()` in `skills/fleet/scripts/candidates.mjs` writes its refusal with `writeSync(2, …)` then calls `process.exit(2)`. On a pipe, once ~64 KiB of gh's forwarded stderr is already queued, fd 2 is non-blocking and that `writeSync` can throw EAGAIN. Uncaught, the throw skipped `process.exit(2)` and the process fell through to Node's default exit 1 — "query fine, queue empty" — inverting the file's own exit-code contract (`skills/fleet/skills/next-ticket/SKILL.md:28`) under exactly the large-stderr case that motivated the write in the first place.

Fix: wrap the write in `try/catch` so the exit code survives even when the message can't be written. Recovering the refusal *text* as well needs a short retry loop around the write; that's a separate decision and is explicitly out of scope here (per the ticket's respec).

## Test plan

- [x] Adjusted the pre-existing large-stderr pipe test (`candidates.test.mjs:515`, "the exit code survives a gh stderr larger than the pipe buffer") to assert only the exit code — the actual contract. Under the fix the refusal *text* can still be lost to the same EAGAIN (deliberately, per scope), so the old text-match assertion would itself have stayed flaky.
- [x] Local suite (darwin): `./agent-test skills/fleet/scripts` — 375/375 green.
- [x] The bug does not reproduce as a local-only, single-shot race on darwin (raw `writeSync` never threw EAGAIN there in extensive testing, even at 20 MB bursts) — matches the platform asymmetry the ticket's own CI sighting points at. Verified instead against real Linux (`node:22` container, this repo's CI target), where it reproduces directly:
  - Unfixed `die()`, repeated runs of `candidates.test.mjs`: 7/15 and 12/25 failed with `1 !== 2` at exactly the test the ticket names.
  - Fixed `die()`, same repeated-run protocol: 0/45 failed.
  - This is the mutant-kill: reverted the fix, confirmed the test goes red (reliably, over many runs, since the underlying race is probabilistic); restored it, confirmed green.

## #298 (suspected flake, `candidates.test.mjs` — `status 1 !== 2`)

Checked whether this fix makes it stop. The CI sighting quoted in this ticket's comments names two failures in the same run: the large-stderr pipe test above (root-caused and fixed here), and a second sighting near `a failed gh query refuses without re-emitting gh's own stderr` (small `MARKER_ZZZ(((` burst). Both go through the same `die()` call site in `query()`'s catch block, so structurally the fix covers both.

I could not independently reproduce the second test's flake locally — 0/40 failures in the Linux container both with and without the fix, likely because that test's stderr burst is much smaller than the one that reliably fills the pipe, and a single isolated test run doesn't replicate whatever cross-test stderr accumulation a full 370-test CI run has. So I can't claim to have directly confirmed #298 closes. But since the fix is in `die()` itself (used by every call site in the file, not just the one this ticket names), any inversion sharing this root cause is covered. Not closing #298 myself — leaving that to be confirmed against real CI signal.

Closes #299


feigi:
Review applied at `0d801d1`.

**Applied here**

- **The dropped pin.** This PR's rewrite deleted `assert.match(stderr, /^candidates: gh issue list failed/m)`, which turned out to be the only thing pinning the #176 fix — reverting `die()` to `console.error` passed the suite 375/375. The text assertion can't come back (under the EAGAIN this test forces, the message is legitimately allowed to be lost), so the guard's *shape* is pinned instead, through `strip-comments.mjs`.

  The obvious form of that pin is vacuous, and it was measured, not assumed: against **raw** source, a `die()` genuinely reverted to `console.error` with the good shape parked in a block comment passes **375/375** — `^…/m` matches inside the comment, the escape `strip-comments.mjs`'s header documents. The raw form is *also* over-tight, going red on a whole-line comment inserted between `try {` and the write. Only the stripped form discriminates.

- Two comment fixes in this PR's own block: a bare `//` between the two failure-mode paragraphs (the preceding break already uses one), and the `catch` comment cut to one line — it restated the block comment eight lines above.

**Mutation results** — every probe anchor-checked before running, every mutated line printed:

| mutant | result |
|---|---|
| `console.error` revert (#176) | red ✓ |
| `try`/`catch` removed (#299) | red ✓ |
| block-comment escape | red ✓ |
| reworded-comment control | green ✓ |

`node --test "skills/fleet/scripts/*.test.mjs"` → **tests 375 / pass 375 / fail 0**.

**For a macOS reviewer:** EAGAIN never fires on darwin — the pre-existing `assert.equal(status, 2)` caught a #299 revert **0 of 15** runs there. A local green is not confirmation of this fix; on darwin the new source pin is the only gate. That is now stated in the test's own comment.

**Deferred → #328** — `die()` is byte-identical and unguarded in `pr-overlap.mjs:24-27`, `diff-stats.mjs:108-111` and `ci-state.mjs:26-29`. Out of #299's scope (the `candidates.mjs` exit contract), and `pr-overlap.mjs` has no test file at all, so it is its own piece of work. `ci-state.mjs` is the sharp one: exit 1 there is a *meaningful verdict*, so an inverted exit 2 is indistinguishable from an honest "not green". Filed `ready-for-agent` — the guard is fully specified; the shared-`die()` refactor is written up as explicitly optional.

Dimensions reporting the deferred finding: `correctness`, `silent-failure`, `tests`, `comments`, `types`.

`rebase-check` is red from staleness only (four PRs merged during review); `check` is green. Not rebasing.

Issue #239, opened 2026-08-10
fleet: deferral filing is unthrottled — 85% of open issues are review deferrals, and #211 changed the apply policy, not the filing policy


feigi:
Filed by the controller during a `/fleet:run-team` run. This is a **policy question**, not a defect report — nothing here is behaving contrary to its spec.

## The measurement

Taken 2026-08-10 against the live tracker:

```
open issues total:                             108
  bodies citing "Deferred from PR":             92   (85%)
  carrying ready-for-agent:                     96
  carrying needs-triage:                        11
closed issues still carrying in-progress:       22   (that residue is #170, not this ticket)
```

This run merged **2** PRs and filed **7** issues (#229 #230 #231 #232 #233 #234 #235). Two of the seven were filed closed as `wontfix` (refuted findings, kept for rediscovery), so the net open-issue delta is +5 against 2 merges.

## Mechanism — two rules composing, both working as written

- `skills/fleet/commands/review-and-fix.md` step 2: a finding is **deferred, never dropped**. Everything not `survived` — every `suggestion`, every `unverified` — becomes a deferral.
- step 5: **each deferred finding is filed as its own issue.**

So backlog growth is a function of review fan-out width, and nothing throttles it. A six-dimension review that returns 3 survived and 18 suggestions produces up to 18 issues from one PR.

## Read this before treating it as a regression against #211

**#211 is CLOSED** and its title claims it *"replace[d] the apply/defer/file policy behind 83% of open issues"*. That is easy to misread as "this was already fixed and has regressed". It has not.

What #211 actually shipped into step 2 is a **scope split plus one refuter for in-scope suggestions** — it governs what gets **applied**. Step 5's one-issue-per-deferral filing is untouched by it. So the filing rate was never in scope, and **83% → 85% is unmoved by design, not by failure.** Spec: `docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md`.

## The open question

Should deferral → issue stay 1:1? Options, none obviously right:

1. **A severity floor** — file `important` and above, record the rest on the PR. Cheap; loses the `suggestion` band, which is where several genuinely load-bearing findings in this run lived (an unpinned newline path, a dead test line, a false comment).
2. **Batch per review** — one issue per PR review carrying all its deferrals. Keeps everything, collapses the count; makes each issue a poor unit of agent work, which is the thing that made them grabbable.
[...]

PR #324, opened 2026-08-10
docs(fleet): the merge-bot runbook calls the stash count exact — a truncated reflog makes it a floor

### LINKED ISSUE #305: run-merge-bot runbook: a nonzero stash count is a floor, not a proof — the truncated-reflog ceiling is in the script comment but not in the text the operator reads
**Deferred from PR #292 review** (implements #147). **Dimension: silent-failure.** Severity: suggestion. Reported at `skills/fleet/commands/run-merge-bot.md:133`.

## The finding

The remaining truncated-reflog ceiling is documented in the script's own comment but not in the operator-facing runbook, so a silently-low count still prints as an exact number to the only reader who acts on it.

Evidence given (fixture `fx-badlog`: three stashes, then the oldest reflog line's new-oid overwritten with 40 zeros):

```
$ (cd fx-badlog/w && git stash list; echo "rc=$?")
stash@{0}: WIP on fix/1-thing: 67daad8 feat
stash@{1}: WIP on fix/1-thing: 67daad8 feat
rc=0                      <- 2 of 3, no error
$ no-undo-audit.sh fx-badlog/w fix/1-thing
    stash entries (repo-global, not gated): 2
```

`skills/fleet/scripts/no-undo-audit.sh` names this gap in-line ("reports the (too-low) count as exact. That gap is the remaining ceiling"), but `run-merge-bot.md:133` — the text the merge-bot operator actually reads — carries only the nonzero/unknown rule, so a `2` that is really a `3` reads as a complete account.

## Why deferred

#147's brief explicitly puts the truncated-but-readable reflog **out of scope**, requiring only that it be named as the remaining ceiling. The script comment does name it; PR #292 kept that sentence intact and accurate. Carrying it into the runbook is a separate docs change, so it was deferred rather than folded in.

## Why `ready-for-agent`

The approach is settled — the finding names the exact edit and the exact line, and it is one clause of prose with no code change and no design choice:

> One clause on the runbook line: a number is a floor, not a proof — a partially-parseable reflog drops entries silently, so `git stash list` is worth reading whenever the audit reports any nonzero count, not only an unexpected one.

The sentence to extend is in `skills/fleet/commands/run-merge-bot.md:133`, currently ending:

> Read `git stash list` yourself when the count is nonzero **or `unknown`** — an entry naming this branch may be a dead member's only copy — and never pop, drop or apply an entry you did not create.

Note PR #292 reworded the neighbouring `unknown` sentence on the same line (the old text said "the stash ref resolves but the list came back empty"; it now says "the list came back empty while `refs/stash` is not absent"). Anchor on current text, not on the finding's quotation.

Consider whether the same clause belongs in the spec's script-surface row at `docs/specs/2026-07-23-fleet-plugin-design.md:180`, which is machine-checked by `no-undo-audit.test.mjs` for field names only.

feigi:
`no-undo-audit.sh` names the remaining truncated-reflog ceiling in its own comment, but `skills/fleet/commands/run-merge-bot.md:133` — the text the merge-bot operator actually reads — carried only the nonzero/unknown rule. A silently-low count printed as an exact number to the only reader who acts on it.

One sentence added to the stash passage: the number is a floor, not a proof, plus where the floor comes from.

**Deliberately not the ticket's proposed remedy.** The ticket suggested the clause end with "so `git stash list` is worth reading whenever the audit reports any nonzero count". Two problems, both checked against the tree:

- The line already says exactly that — `Read `git stash list` yourself when the count is nonzero **or `unknown`**`.
- It would not close the gap anyway. The audit count *is* that list's length (`stash=$(git -C "$wt" stash list 2>/dev/null | wc -l ...)`), so a truncated reflog shortens the operator's own read by the same entries. Recommending it as the remedy would ship a new wrong claim.

The added sentence says the audit and the operator's own read come up short by the same entries and neither reports it.

## Spec row: not changed

The ticket's `Consider` asks whether the same clause belongs in the script-surface row of `docs/specs/2026-07-23-fleet-plugin-design.md`. Declined. That row documents the script's contract surface — field names and exit codes — and truncation produces no distinction there: `stash` is a number in both the honest and the too-low case. The ceiling is a caveat on how to *read* the number, which is the runbook's job, and it is already stated for implementation readers in the script comment. Adding it would put the same prose in a third place.

(The row is also not at `:180` as the ticket says — see below.)

## Line numbers re-measured

- `skills/fleet/commands/run-merge-bot.md:133` — correct as cited.
- `docs/specs/2026-07-23-fleet-plugin-design.md:180` — **stale**, the `no-undo-audit.sh` row is at line 184. Not cited in the diff, so nothing shipped on it.
- Anchored on the live `unknown` wording ("the list came back empty while `refs/stash` is not absent"), not the finding's pre-#292 quotation.

## Test

`./agent-test skills/fleet/scripts/no-undo-audit.test.mjs` — tests 36, pass 36, fail 0. Includes the spec-row field-name check, which is unaffected.

Closes #305


feigi:
**Review adjudication (fix-applier)** — `correctness` + `comments`, docs profile: 0 survived, 0 refuted, 3 unverified (all `suggestion`, a band the workflow budgets 0 refuters). Adjudicated here: **0 applied, 0 filed. HEAD unchanged at ae808a0.**

- **[U3] comments — positive confirmation, no action.** The added sentence is accurate against `no-undo-audit.sh` runtime behaviour and matches the pre-existing comment at `no-undo-audit.sh:213-218`. Its own point (4) kills the "add a tracking-issue pointer" idea: `run-merge-bot.md` cites no issue numbers anywhere, so adding one would be inconsistent, not an improvement.
- **[U1] correctness — real second undercount path, not applied.** `refs/stash` removed while `.git/logs/refs/stash` keeps entries → `show-ref` rc1, which the script's measured table (`no-undo-audit.sh:195-215`) defines as genuine absence, so the audit prints `0`. Pre-existing script behaviour, untouched by this diff, and it needs manual removal of the ref file (`git update-ref -d refs/stash` deletes the reflog too). #305 is a `class=correction` ticket with a one-clause ceiling; widening the clause is how a correction ships a new wrong claim. The sentence's headline ("a floor, not a proof") already covers this case truthfully.
- **[U2] correctness — no regression test pins the truncated-reflog undercount. Deliberately not filed.** The ceiling is already documented at the point of change: a reader editing the stash block cannot miss the 24-line comment ending "That gap is the remaining ceiling." The only failure mode is safe-direction doc rot (the runbook stays over-cautious after a hypothetical improvement), not operator harm. Against #239 — deferral filing unthrottled, 85% of open issues are review deferrals — a ~20-line test pinning a known, exotic-to-trigger deficiency is the marginal case that issue exists to stop. Ledger `verdict: clean` for both U1 and U2 (tracker reached, `ok: true`); nearest rows both resolve to closed #147. Recorded as an agent-brain workspace memory instead of an issue, per the `not-filed` precedent from PR #320.

`node --test "skills/fleet/scripts/*.test.mjs"` from the worktree: **375 pass, 0 fail** (not a `tests 0` run).

Issue #184, opened 2026-07-31
release-ticket.test.mjs: the three 0o000 permission fixtures leak on failure, and go vacuous under euid 0


feigi:
Deferred from PR #171 review. **Pre-existing** — #171 copied an established pattern; it did not introduce either half.

Two independent weaknesses in the same three tests: `release-ticket.test.mjs:722`, `:757`, and `:330` (`a stray worktree the script may not stat keeps the hand-release remedy`, added by #171).

## 1. The restore is not failure-safe

The shape is `chmodSync(parent, 0o000)` → `release(r, c)` → `chmodSync(parent, 0o755)`, with the restore deliberately placed before the first assertion — that part is correct and commented.

But `release()` itself can throw before it returns: `release-ticket.test.mjs:126` runs `JSON.parse(res.stdout)` unguarded. A regression that emits a malformed payload *in one of these fixtures* throws inside `release()`, skips the restore, and the `t.after` cleanup then fails on the `0o000` directory.

Measured:

```
rmSync on 0o000 parent: THREW ENOTEMPTY, Directory not empty: …
```

So the real assertion is masked by an `ENOTEMPTY`, and a temp dir leaks. The failure mode is worst exactly when something is already broken.

**Fix once, not three times:** wrap the body of `release()` (`release-ticket.test.mjs:117-129`) rather than each call site — all three tests then inherit it.

## 2. `0o000` denies nothing under euid 0

These fixtures depend on `chmod 000` actually denying the search. Running as root it does not: `-e` succeeds, the stray reads as present, and `:330` would pass **for the wrong reason** — it would also pass under the very mutation it exists to catch (`gone "$stray"` → `[ ! -e "$stray" ]`), silently ceasing to be the discriminator it was written to be.

Not a live problem: `.github/workflows/ci.yml` is `runs-on: ubuntu-latest` with no `container:` key, so CI runs as `runner`, non-root. The hazard is that moving the suite into a container — an ordinary thing to do — turns three tests into no-ops with nothing failing.

**Fix:** make it loud instead of silent.

```js
if (process.geteuid?.() === 0) return t.skip("chmod 0o000 denies nothing under euid 0");
```

## Why this matters more than it looks

`:330` is the only case in the file that distinguishes the reused ancestor-walk predicate from a naive `[ ! -e ]` — verified by mutation during the #171 review: that swap reddens `:330` and nothing else. A test carrying that much weight should not be able to pass vacuously.

## Provenance

[...]

Issue #179, opened 2026-07-31
release-ticket.sh: the stray guard claims a branch mismatch it never observed when git cannot resolve the worktree's HEAD


feigi:
Deferred from PR #171 review.

## Problem

`skills/fleet/scripts/release-ticket.sh:184-188`:

```sh
if gone "$stray"; then
  block "worktree $stray is this claim's and its directory is gone — git worktree prune to clear the registration"
else
  block "worktree $stray is this claim's but is not on $branch — release it by hand"
fi
```

`gone` returns false for **two** different states: the directory is *present*, and the script *could not stat it* (unsearchable prefix). The `else` collapses them into one message.

The script owns a vocabulary for the third state and uses it on the sibling path — `:256` says `die "cannot tell whether $wt exists, so whether it holds uncommitted work is unknown"` — but the stray guard has no such branch. An operator whose worktree sits under a `chmod 000` prefix is told to "release it by hand", with nothing indicating the script never managed to look.

## Not a behaviour bug

The direction is safe and deliberate. `release-ticket.test.mjs:330` (`a stray worktree the script may not stat keeps the hand-release remedy`) pins it, and pruning on an unmeasured absence is exactly the data-loss the ancestor walk exists to prevent — verified in review: `git worktree prune` on that state unregisters a worktree whose `precious.txt` is still on disk. Nothing here should change that outcome.

What is arguably wrong is only that the message asserts a state the script never measured — the same shape as the #81 defect it just fixed, narrowed to the unstattable case.

## Approach is open — hence `needs-triage`

Two candidates surfaced in review, and they are not equivalent:

1. **A third branch.** `elif` on "exists" vs "could not stat", with its own wording. Costs a new test and a third `block` string.
2. **Four words on the existing message.** `— release it by hand (its directory could not be stat'ed)`. No new branch; but `release-ticket.test.mjs:359` asserts `match(/release it by hand/)`, which still passes, so the change is nearly free.

There is also a real case for **doing nothing**: "release it by hand" is sound advice in the unknown state — it is precisely *not* a prune — and the extra clause may be noise on a path an operator hits rarely.

Wants a maintainer's call on whether the distinction is worth any code at all before an agent picks it up.

## Provenance

Raised independently by two reviewers on PR #171 (silent-failure-hunter, rated MEDIUM; corroborated by type-design-analyzer's "invariant expression" finding, which locates the same gap at the `gone` contract rather than at the message).

feigi:
[...]

PR #330, opened 2026-08-10
fix(fleet): release-ticket.sh fails closed on an unreadable worktree registry

### LINKED ISSUE #84: release-ticket.sh: an unreadable .git/worktrees makes `worktree list --porcelain` omit linked worktrees at rc 0 — the release deletes the branch and drops the label
Deferred from PR #72 review. Pre-existing.

**Rewritten in triage on 2026-07-31.** The original body asked whether to guard `wt_list=$(git worktree list --porcelain)` with a `|| die` for uniformity, or close it as unreachable. That premise was tested and does not hold: the command does not fail, and a guard on its exit status catches nothing. What is at the same read is worse, and measured.

## Problem

`release-ticket.sh` locates the claim's worktree by parsing `git worktree list --porcelain`, and treats the absence of a matching entry as "this claim has no worktree of ours".

When the repository's worktree registry (`.git/worktrees`) cannot be read, git **exits 0** and lists **only the main checkout**. Every linked worktree silently disappears from the output.

## Measured — git 2.50.1

Fixture: a normal claim — branch `fix/9-x`, worktree at `.worktrees/9-x` holding one untracked file.

```
$ chmod 000 .git/worktrees
$ git worktree list --porcelain ; echo rc=$?
worktree /.../w
HEAD d1e4fa71...
branch refs/heads/main
                                  <- the linked worktree is simply not listed
rc=0

$ sh release-ticket.sh 9 x fix --apply
$ gh issue view 9 --json labels
$ git branch -d fix/9-x
$ gh issue edit 9 --remove-label in-progress
{"issue":9,"branch":"fix/9-x","worktree":"","label":true,"released":true,"applied":true,"blockers":[]}
exit=0

branch fix/9-x: DELETED
worktree dir:   SURVIVES
precious.txt:   SURVIVES        <- the member's untracked work, now orphaned
```

Both the worktree lookup and the stray-directory lookup come back empty. The "no branch and no worktree" refusal does not fire, because the **branch** still exists. The dirty check is skipped entirely — there is no worktree to check, as far as the script can see. So the run deletes the branch, drops `in-progress`, and reports a clean release of a claim that is still standing on disk. Once permissions come back the registration is still there, now naming a branch that no longer exists (`HEAD 0000000...`).

This is the fail-open direction the 0/1/2 contract exists to prevent, and the same class as #95 and #82: a degraded read that returns rc 0 with the answer missing. `bb6203a` closed exactly this shape for the dirty check — "establish absence rather than infer it" — but the read that *locates* the worktree still infers.

## Why the original framing was dropped

`|| die` guards the exit status, and the exit status is 0. Adding it would leave every case above unchanged while reading as though the hole were closed — worse than not adding it.

## What triage decided

`bug`, `ready-for-agent`, on the rewritten finding. Brief follows.

feigi:
## Summary
- `git worktree list --porcelain` reads `.git/worktrees` to enumerate linked worktrees, and exits 0 while silently dropping any entry — or the whole directory — it cannot read (measured, git 2.50.1).
- `release-ticket.sh` treated the resulting empty listing as "this claim never had a worktree", so an unreadable registry made `--apply` delete the branch and drop `in-progress` while the worktree — still holding the member's uncommitted work — survived on disk with a stale registration.
- Establish the registry's readability directly (the `worktrees/` directory itself, and each entry inside it) before trusting an absence the listing reports — the same "establish absence, don't infer it" rule the dirty check already follows. An unreadable registry now exits 2 with nothing mutated; a genuinely empty one (worktree actually removed and pruned) still releases at exit 0.

Closes #84

## Test plan
- [x] `./agent-test skills/fleet/scripts/release-ticket.test.mjs` — 47/47 pass, including three new cases: registry directory unreadable, a single entry inside it unreadable, and the regression case (worktree removed + pruned still releases at exit 0)
- [x] `./agent-test .` — full suite, 379/379 pass
- [x] Manually reproduced the ticket's exact repro (`chmod 000 .git/worktrees`, `--apply`) against a scratch repo: exits 2, branch and worktree both survive


feigi:
## Review applied — `2958bb9`

The review's headline critical was real: **this PR did not close #84.** The guard tested read+execute on each `.git/worktrees/<id>` *directory*, but git reads *files* inside it, so `chmod 000` on an entry's `gitdir` passed every check the PR added and still reproduced the fail-open end to end — branch deleted, label dropped, `"released":true`, uncommitted work orphaned. Four dimensions found it independently and both refuters reproduced it.

Fixed by not predicting which files git needs: count registry entries on disk against the worktrees git reported, refuse on a mismatch.

### Applied

|---|---|---|
| entry-directory guard misses `gitdir` (#84 unclosed) | correctness, silent-failure, types — all `critical` | count-based check replaces the per-entry stat loop |
| refusal misattributes another claim's unreadable entry to `#$issue` | silent-failure `important` | refusal now reports what was observed, not a conclusion about this ticket |
| `&&` → `\|\|` on the parent guard survives the suite | tests `important` | that test now also runs at `0o400`; the mutant reddens it |
| `[ -x "$entry" ]` refuses on a stray plain file | correctness `suggestion` | entries counted with `[ -d ]`, plus a test |
| `wt` cited as `:77`, moved by this diff | comments `critical` | cited by construct instead of line |
| the `gone` sentence reads as a call to `gone()` | comments `suggestion` | reworded |
| inline branch lookup / `readdirSync` in the new tests | simplify `suggestion` ×2 | `artefacts()` reused; `readdirSync` dropped |

Each `suggestion` above went to a refuter biased to refuse before being applied.

### Verification

Full suite **380 pass / 0 fail** (`env -u NODE_TEST_CONTEXT -u NODE_TEST_WORKER_ID node --test "skills/fleet/scripts/*.test.mjs"`; 376 on `main`, 379 before this commit).

Mutation matrix, each mutation confirmed applied before running:

| mutant | result |
|---|---|
| parent guard `&&` → `\|\|` | reddens `an unreadable worktree registry…` alone |
| entry `-d` → `-e` | reddens `something that is not a registry entry…` alone |
| count check deleted | reddens `an entry git cannot read INSIDE…` alone |
| swap `-r`/`-x` order (semantics-preserving) | all green |
| rename `registered` → `on_disk` (semantics-preserving) | all green |

Both #84 acceptance cases hold: a genuinely pruned worktree still releases at exit 0, and a clean claim is unchanged. Also checked: `.worktrees` deleted by hand still releases; registry at `0o000` and `0o400` both refuse.

### Deferred

- **#334** (new) — three measured-unreachable hardening nits, two of them **refuted** by their refuters: `\|\| die` on `worktree list` (17/17 registry mutations exit 0; every rc=128 state already dies at the `--git-dir` guard) and `[ -n "$common" ]` (every empty-output state exits 128, behind that same guard). Plus the `--git-path worktrees` simplification, sound but with a sibling-consistency argument against it.
- **#129** (commented) — four cross-file line citations in `inflight.sh` / `inflight.test.mjs` that this PR moved again; three are recurrences of rows already in that issue. **Not editable from here:** those two files are open PR #332's diff, and the numbers must be re-derived after both PRs merge.
- **#184** (commented) — euid-0 vacuity and leak-on-failure in the `0o000` fixtures; this PR takes the set from three to five.
- **#179** (commented) — a third state reaching the same `else`: with `HEAD` unreadable, git lists the worktree with no `branch` line, so the count check does not cover it and the blocker still claims a different-branch state it never measured. Verified still present at `2958bb9`.

Not filed: `simplify`'s "the parent check and the entry loop are the same test written twice" — the fix dissolved the premise (the loop now counts rather than tests permissions), and the mutation matrix shows the parent check is **not** redundant with the count. The `types` finding refuted in review targeted the entry-directory test, which this commit replaced.

Issue #288, opened 2026-08-10
ci-state.mjs/.test.mjs + SKILL.md simplify pass: 4 measured reductions from the #273 review (plus 1 refuted, recorded)


feigi:
**Deferred from PR #273 review** (ticket #111). Four `simplify` findings plus one `silent-failure` finding that was **refuted** and is recorded here so it is not re-raised. All severity `suggestion`.

Each item below was applied and measured by the specialist that found it. **Re-measure before applying: PR #273's fix-apply pass changed both files, so two of the five are now partly stale — the notes say which.**

---

### 1. `ci-state.test.mjs` — two hand-rolled unreadable-file tests duplicate `run()` (`dimension: simplify`)

The two tests at the end of the file each hand-roll ~30 lines of tmpdir / gh-stub / env setup that `run()` already does, differing only in whether `--workflow-file` is passed.

Measured: collapsing both onto `run()` plus a 2-case loop took the file 221 → 166 lines with `tests 6 / pass 6 / fail 0`, and coverage was proven intact by mutation (replacing the discovery-time `die()` with `continue` turned the discovery-scan case red while the explicit-path sibling correctly stayed green — it dies later, in `expectedJobs()`).

**Now easier than when filed:** PR #273's fix pass added an `unreadable: []` option to `run()` (chmod 0o000 before spawn, original mode restored after) for exactly this purpose, and `git init`s the fixture. The collapse no longer needs a new harness — just move the two tests onto the existing option.

### 2. `ci-state.test.mjs` — dead override flexibility in `run()` (`dimension: simplify`)

`run()`'s `prView` / `runList` / `runView` options and the `content === null` branch of `fixtureFile` were documented as "pass `null` to make that gh subcommand fail" — with no call site passing any of them. Deleting all three params and the null branch kept `tests 6 / pass 6 / fail 0`.

**Partly stale after #273's fix pass:** `runView` is now used, by the four new not-green detector tests. `prView`, `runList` and the `content === null` branch are still unreached — check before deleting.

### 3. `ci-state.mjs` — `expected` declared in a branch that never reads it (`dimension: simplify`)

`const expected = noCi ? [] : expectedJobs(workflowFile);` plus the `if (!noCi)` guard on its `vlog` exist only to give `expected` a value in the branch that never reads it. `expected` is read at exactly one site, inside the non-no-ci branch; the payload carries `missing`, not `expected`.

[...]

Issue #240, opened 2026-08-10
docs/specs/2026-07-22-run-team-agent-fleet-design.md:201 names `--label ready-for-agent` — a flag candidates.mjs now refuses, and the string is hand-copied across three files


feigi:
Deferred from PR #228 review (issue #173).

**dimension:** `code-reviewer` (relevance-refuted, unanimous)

## The finding

`docs/specs/2026-07-22-run-team-agent-fleet-design.md:201` still spells the mandatory candidate-scan flag `--label ready-for-agent`:

```
   list; they are ~97% of the payload. **`--label ready-for-agent` is mandatory
   here, and there is no fallback.** `next-ticket` drops the label filter and
```

`candidates.mjs` has never accepted `--label`. After #173 it refuses it outright at exit 2, so this line now names a flag that is not merely ignored but rejected.

## Why deferred

Out of scope. PR #228 fixes the two *live* carriers of the wrong spelling — `skills/fleet/skills/run-team/SKILL.md:60` (the rule a controller actually executes) and `docs/fleet-rule-manifest.txt:5`. This third copy is in a dated design spec, a historical record rather than an instruction anything follows, and #173's scope does not reach `docs/specs/`. Both refuters agreed the finding is real and out of scope for that PR.

## The generalisable part

**The string is hand-duplicated across at least three files** and nothing keeps them in sync. Measured at `e840bc5`:

```
skills/fleet/skills/run-team/SKILL.md:60                  --require-label ready-for-agent   (fixed by #228)
docs/fleet-rule-manifest.txt:5                            --require-label ready-for-agent   (fixed by #228)
docs/specs/2026-07-22-run-team-agent-fleet-design.md:201  --label ready-for-agent           (this issue)
docs/specs/2026-07-30-fleet-trust-the-label-design.md:88  --require-label ready-for-agent   (correct)
docs/specs/2026-07-30-fleet-trust-the-label-design.md:311 --require-label ready-for-agent   (correct)
skills/fleet/skills/next-ticket/SKILL.md:15               --require-label ready-for-agent   (correct)
```

PR #228 added a test pinning `run-team/SKILL.md`'s copy (`candidates.test.mjs`, `phase0Step1()`), and nothing pins the other five. Whoever picks this up should decide whether the spec copies are worth pinning at all, or whether the manifest/spec copies should point at the skill instead of restating it.

[...]

Issue #213, opened 2026-08-02
fleet spend cleanup: distilled brief in dispatch, caps 2/3, finisher+merge bot on haiku, run-merge-bot references split


feigi:
After #210, #211, #212 — file collisions on `run-team` SKILL.md, and item 2's rationale reads their measured effect.

Four small spend cuts, deliberately one PR:

1. **Distilled brief in the implementer dispatch prompt** (`run-team` phase 2). The controller already read title + Agent Brief + `Out of scope` in phase 0 step 4; paste that distillate into the prompt. Member fetches the full issue only when the brief is insufficient — replacing the unconditional full `gh issue view` re-read per member per refill. The isolation rationale ("re-reads its Agent Brief from scratch", run-team-agent-fleet-design:168) is preserved: the member still gets the brief verbatim, it just doesn't re-pay the fetch.
2. **Queue-depth default 2/5 → 2/3** (`run-team` Queue depth). With #210/#211 landed, reviews are cheaper and faster; re-derive the review-bottleneck ratio from the first post-#211 wave's spend rollup and set the documented default accordingly — 2/3 if the data supports it, otherwise state the measured ratio.
3. **Finisher and merge bot on `model: "haiku"` / low effort** (`run-team` phase 3 + merge-bot dispatch). Both are checklist work — audit worktree, check per-job CI state, label, one merge pass — and the merge gate catches errors downstream.
4. **Split `run-merge-bot.md` war stories into `references/`** — ~5.1k tokens on-invoke per bot per wave (fleet-plugin-design baseline table); same treatment `run-team` got. Move, don't duplicate; keep every rule reachable per `docs/fleet-rule-manifest.txt`.

## Out of scope

- Always-on prose (~302 tok, already solved per fleet-plugin-design).
- Any isolation/SHA/ledger guard.

## Acceptance

- Each item independently verifiable; manifest patterns still resolve after item 4.
- Item 2 cites the measurement it was derived from.

Origin: fleet token-spend analysis, 2026-08-02.

PR #331, opened 2026-08-11
fix(fleet): dependency scan finds refs in the forms tickets are actually published in

### LINKED ISSUE #58: candidates.mjs dependency scan misses to-tickets' blocking edges — every chain reaches the fleet with d: []

feigi:
## Summary

`candidates.mjs`'s `d` scan required a blocking phrase immediately before a
`#N` ref (`(?i)(?:depends on|blocked by|requires|after)\s+#\d+`), so it
missed every form `to-tickets` actually publishes: a `## Blocked by` heading
with list-item refs, and a bold `**Blocked by:**` label. Every chain reached
the fleet with `d: []`, and the consumers' "open blocker → drop" step had
nothing to act on.

- `d` is now built by a small per-line state machine, not one regex: a
  heading line toggles a "blocking section" flag on/off by whether its text
  names one of the four phrases (off again at the *next* heading, blocking or
  not); independently, any line carrying a bold/colon-optional blocking label
  followed by one or more `#N` refs is read too — the original bare inline
  phrasings (`depends on #5`) are the label-only case of this same match. A
  ref elsewhere in the body, under neither, is not collected.
- `d` now holds **plain issue numbers**, not matched phrase strings — both
  consumers only ever checked "is `d` non-empty".
- RE2 (gojq, gh's embedded engine) has no lookahead, so "capture refs up to
  the next heading" can't be one pattern — hence the line walk.
- Native sub-issue/dependency links stay explicitly out of scope. The
  script's own FIFO-sort comment and `next-ticket/SKILL.md`'s step 4 no
  longer claim blanket coverage — both now name that ceiling.

## Verification

- New `candidates.test.mjs` cases cover every form in the issue's acceptance
  criteria: heading + bulleted refs, multiple refs under one heading, the
  bold-label form, the original bare inline forms, "None — can start
  immediately", a reference in unrelated prose (not collected), several refs
  on one inline line, and a heading section correctly closing at the next
  heading.
- Per #63 (this repo's own note that its stub execs system jq/Oniguruma, not
  the gojq/RE2 gh actually applies `--jq` with): added a `findGojq()` helper
  that picks up a real `gojq` binary when present on `PATH` (`go install
  github.com/itchyny/gojq/cmd/gojq@latest`) and runs the dependency-form
  fixtures through it instead of system jq — skips loudly, not silently, when
  unavailable. Ran locally against real gojq 0.12.19: all forms pass under
  the actual engine gh applies, not just a same-family stand-in.
- Live-ran the fixed script against `feigi/claude-config` itself
  (`node candidates.mjs --require-label ready-for-agent`) — no crash, and
  real chains resolve correctly, e.g. `#213 d:[210,211,212]`, `#240 d:[173]`,
  `#288 d:[273]`, `#64 d:[59,60]`, none of which the old regex could see.

## Test plan

- [x] `./agent-test skills/fleet/scripts/candidates.test.mjs` — 41/41 green,
      including the gojq-verified block
- [x] `./agent-test .` — full suite, 385/385 green
- [x] Live sanity run against the real repo (see above)

Closes #58


feigi:
Review fixes applied in `11eb72b`. Suite: `node --test "skills/fleet/scripts/*.test.mjs"` -> **tests 392 / pass 392 / fail 0 / skipped 0** (385 before).

**Applied — 7 `survived` findings + 4 `suggestion`-band findings that survived a refuter**

| what | why |
| --- | --- |
| in-section collection restricted to list-item lines | an open `## Blocked by` swept every later `#N` to EOF; `"None — can start immediately"` + notes invented a blocker and the consumer dropped the ticket |
| `after` dropped from the HEADING alternation (kept inline) | `## After the migration` is narrative; 5 of 7 `after`-heading bodies over-fired. `## After #12 lands` and `must land after #4` still resolve |
| `findGojq()` guards `.stdout` | `spawnSync("go", …).stdout` is `undefined` on ENOENT, so on any machine without Go the module threw at import and all 41 tests vanished. Now falls through to the loud skip the comment promised |
| `findGojq()` checks the binary IS gojq; bad `GOJQ_BIN` throws | it accepted anything answering `--version`, so it could resolve to system jq and report green under a name claiming RE2 |
| gojq test gained an engine discriminator | deleting the STUB's `JQ_BIN` plumb left it green, re-running all seven assertions on system jq. Now asserts a U+00A0 fixture, `[12]` under Oniguruma vs `[]` under RE2 — so that deletion is red |
| gojq test asserts exit status first | a gojq REFUSAL reported `Cannot read properties of undefined (reading 'd')`, naming neither the engine nor the refusal it printed |
| leading `\*{0,2}` removed from the inline regex | inert — 262k adversarial + fuzzed bodies, zero cases where its presence changed the result under either engine |
| `(. // "")` + two single-use bindings collapsed | `//` fires only on null/false, both already consumed by the sole call site's `(.body//"")` |
| 4 new coverage tests | mutation found live code no test touched: the `and` separator, the asterisk group closing BEFORE the colon (`**Blocked by**: #12`), and `## Requires` / `## Depends on` headings |
| line-locality pinned + documented | the removed regex's `\s+` crossed newlines; the narrowing is deliberate, named in the `depnums` comment and in `next-ticket`'s step 4 ceiling |

Every added test was mutation-verified: break what it pins, confirm it and only it goes red, restore — then a no-op change (renaming two jq bindings) confirmed green.

**Deferred — filed**

- #337 `silent-failure` — CI provisions no gojq, so the engine gate never runs there. Out of #58's scope; **sequence against #163**, a separate staged change to the same workflow file.
- #338 `correctness` — "gojq harness is #63's work". Refuted by #58's own acceptance criteria; #63 is closed.
- #339 `correctness` — `**Blocked by:**` alone on a line with refs bulleted below yields `[]`. Not a `to-tickets` form (both templates checked directly), but plausible hand-written.
- #340 `tests` + `comments` — gojq test omits three phrase words / comment overstates. Both refuted by a 65-fixture two-engine run and an 8-mutant matrix.
- #341 `simplify` — collapsing the two `+` terms. Refuted: as filed it reinstates the phantom sweep.
- #342 `simplify` — template literal for the `JQ` constant. Out of scope.
- #343 `simplify` — redundant `d holds plain issue numbers` test. Out of scope.
- #344 `simplify` — one `FORMS` table for both loops. Out of scope; a naive merge would destroy the new discriminator.
- #345 `simplify` — drop `GOJQ_BIN`. Refuted: it now carries a hard-failure contract.

**Not filed, adjudicated:** the version-pin mismatch (`@v0.12.19` in the comment records what was verified; `@latest` in the skip hint is install guidance — both correct in role, and the finding's own verdict was "not a defect"), and the `to-tickets` verification boundary (the reviewer could not see the gitignored skill; checked directly — `SKILL.md:75` writes `**Blocked by:**` and `:99-101` writes `## Blocked by` + list, so the comment is accurate).

**Related, not touched here:** #204 (`\s`/`\d` class divergence between the engines) now has its first executable pin — the U+00A0 assertion in the gojq test. #205 (dependency scan has zero test coverage) looks closable by this PR.

PR #332, opened 2026-08-11
fix(fleet): inflight's remote probe can prompt or hang forever

### LINKED ISSUE #92: inflight.sh: an unattended probe can hang on a credential or host-key prompt — GIT_TERMINAL_PROMPT is never set

feigi:
## Summary
- `skills/fleet/scripts/inflight.sh`'s probe 2 (`git ls-remote --heads origin`) never set `GIT_TERMINAL_PROMPT`, so a missing credential or unknown host key could block on a prompt an unattended fleet never answers.
- Separately, a stalled transport blocks identically with or without prompt suppression (measured in the issue: killed at 8s, rc 142 either way) — prompting and hanging are different failures, and `timeout(1)`/`gtimeout(1)` are absent by default on this darwin host, so nothing bounded it either.
- `GIT_TERMINAL_PROMPT=0` is now unconditional (not gated on stdin being a tty, per triage's respec — this is a machine probe with a 0/1/2 contract). The bound comes from git's own transport knobs: `BatchMode=yes` + `ConnectTimeout`/`ServerAlive*` for ssh, `http.lowSpeedLimit`/`http.lowSpeedTime` for http. A user's own `GIT_SSH_COMMAND`/`core.sshCommand` is honoured — the new options land on top of it, not in place of it.

## Test plan
- [x] New test: a real TCP listener that accepts the connection and never answers — the probe still terminates with exit 2 (measured ~10.5s, bounded by the fix's own `ConnectTimeout`).
- [x] New test: an existing `GIT_SSH_COMMAND` is preserved verbatim in the invoked command, with `BatchMode=yes`/`ConnectTimeout=10`/`ServerAliveInterval=5`/`ServerAliveCountMax=2` appended on top.
- [x] Both new tests verified red against the pre-fix script (mutation-tested): one via the spawnSync backstop firing (proves the hang reproduces without the fix), the other via the missing bound options.
- [x] Existing probe-2 regression tests (unreachable origin, no origin configured, reachable origin free/taken) still green — the fix doesn't change those paths.
- [x] Full suite: `./agent-test .` → 378 passed, 0 failed (376 baseline + 2 new).

Closes #92


feigi:
## Review fixes applied — 4cea683

**Suite: 380 pass / 0 fail** (`env -u NODE_TEST_CONTEXT -u NODE_TEST_WORKER_ID node --test "skills/fleet/scripts/*.test.mjs"`), against base `0fa08cb`. This PR was at 378; +2 new tests. Note `main` has since moved to `f59c3b9` (#330 merged, +4 tests), so a rebase makes this **384**. Shellcheck clean at CI's `-S warning`.

### Applied (5 survived + 1 suggestion that survived its refuter)

| --- | --- | --- |
| `inflight.test.mjs` substring option assertions | `tests` (critical) | whole-word comparison — `ServerAliveCountMax=25` no longer satisfies an assertion written for `=2` |
| `inflight.sh` ConnectTimeout/ServerAlive attribution | `comments` (critical) | reworded to what was measured; shipped option values unchanged |
| `inflight.sh` https unbounded | `correctness` | comment now states the real ceiling; residual filed as #346 |
| `inflight.sh` GIT_SSH silently dropped | `silent-failure` | added as the third fallback tier, quoted (it is a path, not a command line) |
| `inflight.sh` core.sshCommand tier untested | `types` | one test per tier; each kills its own mutant and only its own |
| `inflight.test.mjs` no-op `createServer` handler | `simplify` | deleted (refuter confirmed inert) |

**Mutation matrix** — every new assertion kills its own mutant, plus a control that must stay green:

| mutation | result |
| --- | --- |
| `ServerAliveCountMax=2` → `=25` | 3 red (was **0 red** before this commit) |
| `ServerAliveInterval=5` → `=50` | 3 red (was 0) |
| `core.sshCommand` tier deleted | 1 red — only that test |
| `GIT_SSH` tier deleted | 1 red — only that test |
| **control:** reorder the four options | **36/36 green** — order-independent, as intended |

### Deferred and filed

| # | label | what |
| --- | --- | --- |
| #346 | `needs-triage` | https origin still unbounded — `http.lowSpeed*` only bounds a transfer already under way; needs a bound outside git. Folds in the credential-helper and connect-stall findings. |
| #347 | `needs-triage` | every unattended git network call outside this probe still carries #92's hazard (out of scope; `release-ticket.sh` just merged via #330) |
| #348 | `ready-for-agent` | `ci.yml` Shellcheck comment enumerates two info-level hits, there are three (SC2015). Pre-existing. **⚠️ PR #335 is open against the same file — sequence against it.** |
| #349 | `needs-triage` | "the one network call in the script" — raised twice, refuted 2/2, filed so the judgement is not lost |
| #350 | `needs-triage` | "options land on top of" vs ssh's first-wins `-o` semantics |
| #351 | `wontfix` (closed) | four in-scope `suggestion`s refuted by measurement, kept for rediscovery |

### Two claims appeared in more than one bucket — reconciled

**`inflight.sh:155`** was `refuted` from `comments` **and** sitting unverified as a `suggestion` from `correctness`. Same assertion about the same sentence; the `correctness` copy quotes only the fragment, dropping the parenthetical that both refuters identified as the scoping clause. Its extra evidence (`next-ticket/SKILL.md:44` says "three network round-trips") does not defeat the refutation — that doc counts *all* round-trips, the comment counts git-transport ones, and both are true at once. **Not applied**, filed as #349.

**`inflight.sh:182`** was `survived` from `correctness` **and** `refuted` from `types`. These assert genuinely different things: `correctness` says the bound does not hold for **https** (measured >90s and >120s); `types` says the http half has **no test coverage**. The `types` refuter's counter-measurement used a plain **http** origin (bounded at 10.09s) and never touched https, so it cannot contradict the correctness claim. Treated separately: the correctness comment fix is applied, the coverage claim stays refuted.

feigi:
## Follow-up — 78dcccd

**Suite: 382 pass / 0 fail** against base `0fa08cb` (380 at 4cea683). Shellcheck clean at `-S warning`.

The `simplify` refuter that killed the "flatten to one substitution" claim surfaced a wider gap than the finding it was checking: **the suite pinned only the tier where the caller supplies its own ssh command**, so every tier where the script picks the program itself was unpinned — which is why a form that hands git an *empty program* scored a clean 34/34. My first commit covered tier 2 (`core.sshCommand` set) and the new `GIT_SSH` tier but not these two.

Added:

- **nothing configured at all** → the fallback must name plain `ssh`
- **`core.sshCommand` set but EMPTY** → `git config --get` exits **0** with empty output for an empty value, so a fallback written as an exit-status test reads "configured" and produces a command line starting with `-o`; real git dies `-o: command not found`. This is what the `-n` guard exists for.

Neither was observable before: every existing test supplies its own ssh command, so none could see an empty program, and "could not look" is exit 2 whether git failed on a bad command line or a bad host. The stub is named `ssh` in the fixture's `bin`, already first on PATH.

| mutation | result |
| --- | --- |
| drop the final `\|\| base_ssh=ssh` guard | **exactly these two red** |
| the refuted flattened form, applied in full | **only the empty-`core.sshCommand` test and the `GIT_SSH` tier red** — the no-config tier stays green, because `\|\| echo ssh` *does* fire for an unset key |
| **control:** reorder the four bound options | **38/38 green** |

That second row is the discrimination that matters: the new test is what stands between that form and merge, and it is not a blunt instrument — three of the five tier tests are untouched by it.

Issue #100, opened 2026-07-31
agent-test: node silently drops argv entries it cannot resolve, so a typo'd path reads as a pass


feigi:
Deferred from PR #75 review.

The runner counts files with `find`; node decides separately which of them to run. The two are never reconciled, and node's discards are silent whenever at least one other argument resolves.

```
$ ./agent-test t/a.test.mjs t/typo.test.mjs
ℹ tests 1  ℹ pass 1  EXIT=0
# occurrences of "Could not find" in that output: 0
```

A member believes two files ran. Alone, the typo is loud (`Could not find`, exit 1); mixed with anything valid, it vanishes. The same mechanism silently drops `node_modules` paths (#98) and, before PR #75's escape, bracketed paths.

This is the pre-existing passthrough branch (`claim-ticket.sh:151`, the `else set -- "$@" "$arg"` arm), not the directory expansion — the shim's `[ -n "$files" ]` guard proves `find` matched, and proves nothing about execution.

## Why the approach is open

There is no cheap POSIX fix. Options, none obviously right:

- Stat each non-directory argument before passing it on — cheap, but rejects the quoted-glob form deliberately (`./agent-test 't/*.test.mjs'`), which currently works.
- Parse node's summary and refuse when the count of files it reports is below the count handed in — node does not report that number.
- Accept it and record the ceiling in a `ponytail:`-style comment naming what the guard does and does not prove.

The third is the honest minimum if nobody wants to pay for the first two, but it is a decision, not a fix.

Verified by the reviewer on Node v26.5.0.

feigi:
> *This was generated by AI during triage.*

**Verified against `main` (`9199314`), Node v26.5.0:**

```
./agent-test t/b.test.mjs t/typo.test.mjs   ->  tests 1, pass 1, EXIT=0   (no "Could not find")
./agent-test t/typo.test.mjs                ->  Could not find 't/typo.test.mjs', EXIT=1
```

**#125 is absorbed here and closed.** Its measured case — a vendored *directory* sharing argv with a real one — is closed at the `find` level by #123's ruling to prune (measured: both spellings then expand to nothing and are refused). What survives is its general point, which is this issue: an argument that passes the runner's guard and is then silently discarded by node. That still reaches this issue through a **file** argument under a vendored directory, which bypasses the directory branch entirely.

## Agent Brief

**Category:** bug
**Summary:** the runner must refuse an argument it can tell node will discard, instead of greening on the ones that survived

**Current behavior:**
[...]

PR #333, opened 2026-08-11
fix(fleet): a bare agent-test invocation runs the same suite as "."

### LINKED ISSUE #97: claim-ticket.sh: a bare `./agent-test` runs node's own discovery, which disagrees with the emit guard's test-file shape

feigi:
## Summary
- `for arg do` (no `in` clause) iterates `"$@"` — on an empty argv the loop body never ran, so a bare `./agent-test` skipped both the directory expansion and its zero-match refusal and fell through to bare `node --test`, node's own default discovery. That discovery does not recognise the `.spec.` form or the `x` extension variants the emit guard certified the repo against, so a bare invocation could report green over a subset of the suite (or none of it).
- `[ $# -gt 0 ] || set -- .` in the emitted runner template routes a bare invocation through the same expansion, test-file shape, and `node_modules` prune as every other invocation.
- The blocker named in the issue body (defaulting to `.` sweeping vendored tests) was already discharged by #109's `node_modules` prune — verified here too, on the defaulted path specifically.

## Test plan
- [x] New test: bare invocation and explicit `"."` report the same pass count on a fixture mixing `.test.` and `.spec.` forms (mutation-verified: fails pre-fix, since node's default discovery misses the `.spec.` file)
- [x] New test: a repo whose only test file uses the `.spec.` form is not reported green without running it under a bare invocation (the issue's own repro; mutation-verified: fails pre-fix)
- [x] New test: a bare invocation excludes vendored tests under `node_modules`
- [x] `./agent-test skills/fleet/scripts/claim-ticket.test.mjs` — 40 passing, 0 failing
- [x] `./agent-test .` — full suite 379 passing, 0 failing (baseline 376 + 3 net new tests)

Closes #97


feigi:
## Review applied — `c71d4e9`

14 findings across 6 dimensions: 4 `survived`, 9 `unverified` (all `suggestion`, 0 refuters budgeted), 1 `refuted`.

### Applied

| Finding | Dimension(s) | What changed |
|---|---|---|
| The vendored bare-invocation test is a vacuous pin | correctness, tests, types, simplify | `claim-ticket.test.mjs:332` — fixture now `{ ...SUITE, "t/d.spec.mjs": PASSES }` asserting `pass 7`, vendor moved to `t/node_modules/` |
| Comment cites a test as "above" that sits 127 lines below | comments | `above` → `below` |
| Comment overstates the emit-guard guarantee | correctness | last sentence of the `:322` block replaced with the measured statement |

The four `survived` findings were one defect, fixed once. Mutation matrix on the changed test — it now goes red for the fix **and** for the prune it claims to confirm, and stays green under a semantics-preserving rewrite:

```
base                                    pass 3 / fail 0
#97 guard deleted                       RED (was green — the defect)
node_modules prune deleted              RED, both siblings green
guard rewritten `if [ $# -eq 0 ]; …`    pass 3 / fail 0
```

### Deferred — out of scope

- **#352** — a bare invocation carrying only flags still falls through to node's own discovery. Pre-existing; `[ $# -gt 0 ]` guards a literally-empty argv only. Approach open: defaulting on "no non-flag operand" reorders argv ahead of node's flags and needs its own measurement.
- **#353** — the defaulted `.` never trips the `case "/$arg/" in */node_modules/*)` guard, so a bare run from *inside* `node_modules` reports vendored results. Byte-diff-verified as not a regression; two competing fixes.

### Deferred — refuted during the fix pass

Each in-scope `suggestion` got one refuter biased to refuse ("verify by RUNNING something; default to refuted if uncertain"). Two died:

- **#355** — "the bare form widens the find-readability refusal, add a test". The pre-fix runner refuses identically on the `.` spelling, so nothing is new; a 5-leg mutation matrix on the suggested test showed it kills only mutants `:186`, `:302` and `:316` already kill.
- **#356** — "the `.spec.` test at `:316` is dominated by `:302`, delete it". Two mutations in `derive-testcmd.sh` redden `:316` alone. Under one of them `:316` is the **only test in the repo** that catches drift in that script's *use* of `testfile_re` — the existing sync test compares the declarations, not their use-sites.

### Reconciliations

1. **`correctness`/`types` vs `tests` vs `simplify` on `:332`** — three different prescriptions for one defect: mix in a `.spec.` file (`pass 7`); move the vendor fixture to a nested path; or delete the test. They are not alternatives. The `.spec.` file is what discriminates the **#97 fix**; nesting the vendor is what discriminates the **prune**. Applied both; the matrix above shows the test now pins each independently, which makes `simplify`'s deletion moot.
2. **`comments` vs `simplify` on `:322`** — "fix the direction word" vs "delete the block". The block carries a load-bearing claim, so the checked one-word fix wins; with the `correctness` overstatement fix landing in the same block, `simplify`'s stated harm ("misdirects the reader") is gone rather than deferred. Not filed — say so if you want it on the backlog anyway.
3. **`#100` is cross-referenced, not fixed.** Node dropping unresolvable argv entries is why the old root-level fixture proved nothing. That is #100's ticket; widening this PR into it would be scope creep.

### Not defects

- `comments` on `:328` — the `#109` attribution could not be checked from the sandbox. Checked against the tracker: #109 *is* the `node_modules` exclusion. Reworded "prune" → "exclusion", since #109 shipped `-not -path '*/node_modules/*'` and the prune came later.
- `types` on `claim-ticket.sh:160` — a positive confirmation that the invariant is code-enforced, not a finding.
- The one `refuted` finding (bare runs from a subdirectory run a subset) needs no action.

`379/379` green against base `0fa08cb` — six behind `origin/main`, which has since gained #330's and #331's tests.

feigi:
## Follow-up — `245d9c6`

The refuter that survived on the `:322` comment turned up one more thing: after correcting that block, the corrected text asserted a state **nothing pinned**. The only `no test files under` assertions in the file were `:159` (an explicit directory argument) and two `doesNotMatch` — nothing covered the defaulted path.

A comment asserting unpinned runtime behaviour is the same defect class as the vacuous fixture `c71d4e9` fixed, so it got the same treatment: one test, vendored fixture (the symlink shape needs a mode 120000 entry `apply()` cannot express; both end in the same guard).

Mutation matrix — every mutated line printed and read back before running, and each mutant built by line number rather than pattern substitution:

```
base                                     41/41
#97 guard deleted (claim-ticket.sh:160)  RED (with the three siblings)
guard rewritten `if [ $# -eq 0 ]; …`     41/41   <- control leg
derive-testcmd.sh:62 stops counting      RED, and ALONE — 379/380 repo-wide,
  vendored files toward the emit           with derive-testcmd.test.mjs 13/13
  decision                                 green
```

That last mutant is what makes it earn its place rather than duplicating `:302`/`:316`: it is the **only test in the repo** that reddens when the emit guard and the runner's walk stop disagreeing. The comment and its pin now move together.

It also clears the same bar #355 was rejected against — a test whose only unique kill is an alternative implementation buys nothing; this one's unique kill is a plausible future edit to `derive-testcmd.sh`.

Side effect: the free-floating comment block now has a test attached, which is the half of the `simplify` finding on `:322` that the direction-word fix did not address.

`380/380` green against base `0fa08cb`.

PR #335, opened 2026-08-11
ci: read node version from .nvmrc instead of a literal pin

### LINKED ISSUE #163: ci.yml pins node-version 22, a version the suite had never been run against
Deferred from PR #157 review.

`.github/workflows/ci.yml` pins `node-version: 22`. Nothing in the repo declares a node version — no `.nvmrc`, no `.node-version`, no `package.json`, no `.tool-versions` (`git ls-tree -r HEAD --name-only` confirms). So CI picked a version the suite had never been run against.

That pin is not cosmetic. It caused two of the three failures on this PR's first CI run: `node --test`'s default reporter is **spec** on a TTY and on newer node but **tap** when older node writes to a pipe, and two assertions matched only the spec prefix `ℹ`. The suite is developed on v26.5.0 (a comment in `review-pr-testcmd.test.mjs` says so outright).

Those assertions are now reporter-agnostic, so the immediate breakage is fixed. The underlying question is not: **which node does this repo actually target?**

- If the answer is "whatever the user runs" (v26.5.0 today), CI on 22 is testing a runtime nobody uses, and can go green on code that breaks in production or red on code that works.
- If the answer is "22 is the supported floor", that is a real decision and should be recorded, with CI as its enforcement.

## Why deferred

Choosing the version is the maintainer's call, and it changes what CI means. Suggested shape once decided: commit `.nvmrc` and switch to `node-version-file: .nvmrc`, so the pin has a single source of truth instead of being a literal in a workflow. A matrix over both is also viable if the floor matters.

feigi:
## Summary
- `ci.yml` pinned `node-version: 22` as a bare literal, and nothing in the repo declared a target runtime — no `.nvmrc`, `.node-version`, `package.json`, or `.tool-versions`.
- That mismatch already caused two of three failures on PR #157's first CI run: `node --test`'s default reporter differs between old and new node when writing to a pipe (`spec`/`ℹ` vs `tap`/`#`), and assertions pinned to one reporter's output broke on the other. Those assertions are already reporter-agnostic on `main` (`grep -n "ℹ" skills/**/*.test.mjs` shows every hit accepts both `ℹ` and `#`) — unaffected by this PR, no further work needed there.
- Adds `.nvmrc` (`26.5.0`) — the version the suite is actually developed and run against (`node -v` here, and `skills/fleet/scripts/candidates.mjs:108`'s own comment: "measured on node v26.5.0") — and points `actions/setup-node` at it via `node-version-file` instead of a literal, so the version has one source of truth.
- No package manifest added (out of scope per the issue), no version matrix (no supported floor has been claimed for this repo — a bare literal was never that claim either).

## How I established "changing the version file changes what CI runs" (not just a green run)
A green CI run alone doesn't prove the workflow is deriving its version from the file rather than coincidentally matching. `actions/setup-node`'s `node-version-file` handling logs the file it read and the version it resolved from it (`Resolved .nvmrc as 26.5.0`) before installing. Once this PR's CI run completes I will pull that run's log for the `check` job's `Setup Node` step and confirm:
1. the logged resolved version is `26.5.0`, matching `.nvmrc`'s content exactly, and
2. the installed/active node version for the rest of the job (visible in the Tests step, e.g. via node's own version in any diagnostic output) is 26.x, not 22 — i.e. the actual runtime CI executes against changed as a direct result of this file, not the workflow's own literal.

Will paste the relevant log lines here once the run is up.

## Test plan
- [x] `./agent-test .` locally: 376 passing, 0 failing (matches `origin/main` baseline — same count before and after this change)
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — parses
- [x] `actionlint .github/workflows/ci.yml` — no findings
- [x] CI green on this PR, with the `Setup Node` log line confirmed to read `26.5.0` from `.nvmrc` (see comment below)

Closes #163


feigi:
CI run [31446836473](https://github.com/feigi/claude-config/actions/runs/31446836473) confirms the version is actually being derived from the file, not just matching by coincidence — the `Setup Node` step logs:

```
node-version-file: .nvmrc
Resolved .nvmrc as 26.5.0
Acquiring 26.5.0 - x64 from https://github.com/actions/node-versions/releases/download/26.5.0-28944078721/node-26.5.0-linux-x64.tar.gz
```

`.nvmrc`'s content (`26.5.0`) is what `setup-node` reports resolving and installing — previously the workflow always installed the hardcoded `22` regardless of any file content. That's the causal link: edit `.nvmrc`, the acquired runtime changes; nothing else in the workflow specifies a version anymore.

CI: green. Suite: 376/376 (unchanged from baseline).

PR #336, opened 2026-08-11
feat(fleet): fleet-tick — the queue-depth reconcile as executable code

feigi:
**Scope: items 1 and 3 of #3 — the executable reconcile, and its invocation on
the fleet's merge-side edges. This PR carries no closing keyword for that ticket:** item 2, the periodic resync
heartbeat, is deliberately not here. It is the only trigger that survives a
**fully drained** queue emitting no events at all, and arming a persistent timer
in a turn-based fleet is a design call rather than an implementation — filed as
#357 (`needs-triage`). #3 stays open until that lands.

run-team's Phase 3 is an event loop and implementer refill was wired as an
**edge** — "a member finished → refill its slot". That edge dies the moment
live implementers reach 0, because 0 implementers emit no completion event, and
"free capacity + a non-empty pool" is a **level** condition an edge-triggered
loop cannot observe once it stops changing. Measured on a live run: implementers
sat at 0/target with pool 1 and ~57 `ready-for-agent` in supply until a human
asked why nothing was being implemented.

`skills/fleet/scripts/fleet-tick.mjs` makes the deficit **computed, not
remembered**. One invocation prints, per role, `actual/target` and an explicit
ACTION, with run-team's queue-depth guards applied in code:

```
implementers 6/2 → AT CAP      (pool=0 supply=82 review-backlog=4)
reviewers    0/5 → DISPATCH 4  (review-backlog=4)
merge-bot    0/1 → IDLE OK     (merge-queue=0)
```

(real output against this repo, mid-fleet-run)

## The contract decision — how the tick learns live counts and targets

#3 left this open and it is the one design choice in the ticket, so, stated
plainly:

**Live member counts and the pool are REQUIRED args. Everything derivable from
the repo the script reads itself.**

Args-vs-env-vs-ledger-parse was the genuine choice. **Ledger parse loses on
correctness, not on effort.** A ledger row records a *dispatch* — `impl-332` —
and that token outlives the member's death, its bail before implementing, and
the merge; no field distinguishes them. Any liveness derived from it is a guess,
and **both** of its error directions are the failures this script exists to
prevent: over-count → HOLD, which is the #3 stall itself, and under-count →
DISPATCH past the cap, breaking a stated invariant. Env vars carry the same
staleness with less visibility at the call site. The controller dispatched the
members and receives their reports, so it is the only component that knows —
it states them.

**Required, never defaulted, for the same reason.** A defaulted `--implementers`
turns a forgotten flag into one of those two failures silently: 0 dispatches a
full cap, the cap holds forever. Refusing at exit 2 matches `candidates.mjs`'s
existing flag discipline.

What the script reads for itself: **review backlog** and **merge queue** from
one `gh pr list` by label, **supply** from `candidates.mjs` — the same shortlist
phase 0 uses, so the tick and the maintainer count the same queue. Every failed
read refuses and prints **no** reconcile line: backlog 0 + merge-queue 0 is a
perfectly plausible tick, so an unread pipeline printed as an idle one is the
silent stall all over again.

Two smaller calls that follow from it:
- **Target = the configured cap for every role**, and work availability lives in
  the ACTION. A reviewer target that shrank to the backlog would print `0/0 →
  idle` and read as correctly sized for exactly the drained state that is the
  defect.
- **The `review-backlog ≥ 2` gate short-circuits the pool/supply table**, rather
  than being a modifier on its top row. Re-shortlisting to enable a dispatch you
  are holding buys nothing, and a `/triage` suggestion under a hold is noise the
  controller would act on.

## Deliberate ceiling

`review-backlog` counts every open PR without `ready-to-merge` — wider than
run-team's own "PRs verified and queued with no reviewer slot", because a PR
already reviewed and merely waiting on CI counts too. Narrowing it needs per-PR
review state that lives in the controller's head, not in the repo. So it can
hold the refill **earlier** than the definition, never later. Marked with a
`ponytail:` comment naming the upgrade path (a `--review-backlog <n>` override,
on the same "the controller states what only it knows" contract) if the
over-count is ever measured to throttle implementers in practice. Flagging it
because it interacts with the standing preference to keep the implementer pool
full — the threshold and the derivation are each a one-line change.

## Scope — the heartbeat is NOT here

Per the brief, the **periodic resync heartbeat is deferred** and not built:
arming a timer in the controller loop is a run-team mechanics decision for the
maintainer. This PR is the executable reconcile and its **edge-triggered**
invocation only.

**So this does not by itself fix a fully drained queue with no incoming
events.** With zero live members and zero open PRs, nothing wakes the controller
at all and no edge fires. It cures the common case — attention captured by a
merge cascade while the implementer side silently drains — by piggybacking the
level-check onto the merge-side events the controller is already handling. The
skill says so in the same words, so a reader cannot mistake it for the whole
cure.

## Prose

Both merge-side edges (merge-bot wave done, CI run completes) now end with the
reconcile, and the Queue depth section points at it as the enforcement rather
than carrying a second hand-run copy of the guards. Two latent prose bugs fell
out of encoding the table: the pool-0 rows overlapped on supply 0 (`< cap`
sitting next to `0` — two answers for one input, which the code had to
disambiguate), and the review-backlog definition now states what is actually
counted.

## Tests

`420 passing, 0 failing` on `./agent-test .` after rebase on `origin/main`
(baseline 376 + 4 landed in #330 + 40 new).

Prose pins were **mutation-tested both ways**: 11 breakages go red — including
one that removes the phrase from the pinned bullet *and plants it in a
neighbouring bullet*, proving slice size is what anchors and not the regex — and
4 unrelated in-slice rewords stay green. The mutation harness also refuses any
run that did not execute all 9 prose tests, so a vacuous green cannot pass for a
red.


feigi:
## Review applied — `eab4606`

Suite **422/422** against base `f59c3b9` (was 420/420; +2 tests). Command run verbatim from the worktree:
`env -u NODE_TEST_CONTEXT -u NODE_TEST_WORKER_ID node --test "skills/fleet/scripts/*.test.mjs"`

### Applied

| what | dimension |
|---|---|
| **`supply()`'s exit 1 is Node's exit 1 too.** Any exit 1 from `candidates.mjs` was read as its documented "query fine, queue empty", but Node uses 1 for `ERR_MODULE_NOT_FOUND`, syntax errors and every uncaught throw — so a broken `candidates.mjs` made *this script* print `supply=0` and the exact #3 stall line at exit 0. Three breakages measured byte-identical to a healthy run. Gate is now on the payload: `r.status === 1 && r.stdout.trim() === "[]"`. | `silent-failure` (critical) |
| **The refusal dropped `r.signal`** — a `candidates.mjs` killed by OOM refused with `exited null`, naming nothing. A reintroduction: already fixed once at `candidates.mjs:170-178`, and the `gh` path 37 lines up in this same file already carried the clause. Reused verbatim. | `silent-failure` |
| **The bad-value loop's `-1` case pinned nothing.** With a space, `parseArgs` rejects a leading dash as ambiguous *before* the guard runs, so `status === 2` was satisfied by the parser — widening the guard to `/^-?\d+$/` survived the whole suite. Now `--implementers=-1`, and every case asserts the guard's own stderr. | `tests` |
| **The cap defaults bypassed the validator beside them.** Threaded through `int()` as a `dflt` param, a default of `"two"` reached the reconcile math; declared in `OPTIONS` it refuses. | `simplify` |
| **`...{ mergeQueue, reviewBacklog }`** spread an object literal identical to its own fields — braces dropped, named fields kept. | `simplify` |
| **`assert.doesNotMatch(r.action, /triage/)`** under an `assert.equal` from `node:assert/strict`, which already pins identity and type. Verified to catch nothing under any mutation, including the `String`-wrapper case. | `simplify` |

Two pins added for the first two rows (a `candidates.mjs` stubbed with a throw and with a syntax error → exit 2, empty stdout; a signal-killed one → `killed by SIGKILL`), plus the benign empty queue kept pinned so the refusal cannot be bought by breaking the case that must still work. **Nothing in the suite distinguished the two meanings of exit 1** — patching the fix in alone left it at 420/420.

Every added pin is mutation-verified: it kills its own mutant and *only* it, and stays green against an equivalent rewrite of the same line.

Two incidental finds while doing it, both fixed here:
- `runCli`'s temp dir needed `realpathSync` — on macOS `tmpdir()` is `/var → /private/var`, and a script **copy** under the unresolved path never runs its own `main()`: `import.meta.url` resolves the symlink and `process.argv[1]` does not, so it exits 0 having printed nothing. Same shape as a passing tick.
- The pool-cap test ran at `supply: 0`, where a bare `DISPATCH 1` is indistinguishable from one that also suggests re-shortlisting. At `supply: 57` it now pins run-team's table row 1 as *silent* — a mutant that previously survived.

### Deferred

| issue | what |
|---|---|
| **#357** `needs-triage` | The periodic resync heartbeat — **item 2 of #3**, and the only `fleet-tick` trigger that survives a fully drained queue emitting no events. Arming a persistent timer in a turn-based fleet is a design call, so it is triaged rather than agent-ready. |
| **#358** `needs-triage` | The label-only `review-backlog` proxy — raised independently by **four** dimensions and refuted by all four. Recorded because four converging readers is signal about legibility even when the design is right. |
| **#359** `needs-triage` | The whole `suggestion` band, checked. 3 applied (above), 7 refuted with the evidence that sank each, 1 unverified. |

### Scope

The body no longer says `Closes #3`. This PR is items **1 and 3** — the executable reconcile and its invocation on the merge-side edges. Item 2 is #357, and #3 stays open until that lands. Merging this as `Closes #3` would have closed the ticket with its actual cure unimplemented.

`rebase-check` is red purely from being behind `main`; that is staleness, not a diff failure, and is the merge bot's to clear.

Issue #186, opened 2026-07-31
claim-ticket.sh: the emitted runner runs vendored tests when the argument is a symlink into node_modules


feigi:
Deferred from PR #177 review.

PR #177 fixed `find "$arg"` → `find "$arg/"` so the shim stops disagreeing with `[ -d ]` about a symlinked directory. That is correct and should ship. It has one narrow side effect worth tracking separately.

## The regression

Point `agent-test` at a symlink whose **target is inside a `node_modules` tree**. Measured on `/usr/bin/find` (what the emitted `#!/bin/sh` runner gets):

```
$ ln -s vend/node_modules/pkg vendlink
BASE  ./agent-test vendlink  →  status=1  agent-test: no test files under vendlink
HEAD  ./agent-test vendlink  →  status=0  ✔ VENDOR-GREEN   ℹ pass 1
```

`-not -path '*/node_modules/*'` matches the path find **prints**. Printed paths under `vendlink/` read `vendlink/test/v.test.mjs` — no `/node_modules/` substring — so the filter cannot see it. Node's own refusal does not fire either: it only rejects an argv path whose relative form *starts* with `node_modules/`, and the symlink rewrites that away.

The result is a green suite whose entire content is third-party code — the outcome `claim-ticket.sh:147-154` says the filter exists to prevent, and which base refused loudly.

## Bounding it — this is narrow

- **Not a new blind spot, only a new route to it.** A bare relative path is already admitted by *both* forms:
  ```
  find node_modules/pkg  -type f -not -path '*/node_modules/*'  →  node_modules/pkg/n.test.mjs
  find node_modules/pkg/ -type f -not -path '*/node_modules/*'  →  node_modules/pkg/n.test.mjs
  ```
  Pre-existing; see the closed #125, which recorded the direct-argument form.
- **The ordinary shape still works.** A symlink to a directory *containing* `node_modules` is still correctly excluded — verified, only `projlink/ok.test.mjs` comes back.
- Reaching it requires deliberately pointing the runner at a symlink to a vendored subtree. Not a plausible accident.
- This repo has **zero tracked symlinks** (`git ls-tree -r <head> | awk '$1=="120000"'` → 0), so nothing hits it today.

## Why needs-triage rather than ready-for-agent

[...]

Issue #246, opened 2026-08-10
ledger.mjs read: console.log + process.exit(0) truncates the payload at 64 KiB on a pipe — the cockpit silently shows an empty ledger


feigi:
Hit live at the start of a `/fleet:run-team` run, launching the cockpit.

## Measured

`skills/fleet/scripts/ledger.mjs:117-119`:

```js
if (cmd === "read") {
  console.log(JSON.stringify(data));
  process.exit(0);
}
```

Against this repo's own `.fleet/ledger.md`:

```
$ node skills/fleet/scripts/ledger.mjs read | wc -c
   65536
$ node skills/fleet/scripts/ledger.mjs read | cat | wc -c
   65536
```

Exactly 64 KiB, exit 0. `process.exit()` does not wait for an async stdout write to drain on a pipe, so the JSON is cut mid-string.

## Why it is worse than a short read

`board.mjs:227-228` consumes it through a pipe and swallows the parse failure into a default:

```js
const ledgerJson = tryRun("node", [join(scriptDir, "ledger.mjs"), "--file", ledgerFile, "read"]);
const ledger = tryParse(ledgerJson, { rows: [], filed: [], ruled: [] }, "ledger read");
```

Observed on both cockpit launches this run:

```
board: ledger read parse failed: Unterminated string in JSON at position 64766 (line 1 column 64767)
board: cockpit on http://localhost:8124  (interval 15s)
```

The server then comes up **HTTP 200 with zero ledger rows** — no rows, no filed list, no ruled list. A cockpit that renders an empty board is indistinguishable from a run that has not started, and the failure line scrolls past in a backgrounded process.

The second consumer is the controller itself. `run-team/SKILL.md` names `ledger.mjs read` as the recovery path after a compaction — specifically to recover a ticket's `class=`, which is a raw-row field the cockpit does not parse. A controller recovering from compaction against a truncated read silently loses every row past 64 KiB, and the documented consequence of a lost class is that the ticket dispatches at the cheaper tier.

## Same defect, already fixed in a sibling

`candidates.mjs` had exactly this and was fixed to `process.exitCode` (its `:291` comment states the reason: `process.exit()` drops async stdout on a pipe). `#132` is the ticket for that class. `ledger.mjs` was not swept — every one of its subcommands still ends `console.log(...)` + `process.exit(...)`:

```
$ grep -nE 'process\.exit\(' skills/fleet/scripts/ledger.mjs
[...]

PR #360, opened 2026-08-11
fix(fleet): arg() reads a trailing flag as absent in four more scripts

### LINKED ISSUE #169: fleet scripts: arg() returns undefined for a trailing flag in four more scripts — the defect PR #167 fixed in candidates.mjs only

feigi:
Closes #169

## Summary

Same defect PR #167 fixed in `candidates.mjs`'s copy: `process.argv[i + 1]` is `undefined` when a flag is given with no value, and every caller reads that falsy result as "the flag was absent" instead of a refusal. Four other scripts carried the identical (or, for `board.mjs`, an inlined arrow-function) copy.

Per triage's ruling on the issue: fixed in place at all four sites, no shared module extracted — these are standalone scripts invoked directly by agents and markdown instructions, and a load-path seam for four other callers isn't worth it to dedupe three lines.

## Consequence analysis (the actual work here)

- **`ci-state.mjs` — examined first, the fleet's gate on PR green.** `base`/`workflow`/`workflow-file` are all read as `arg(name) || default`. A trailing `--base` (or `--workflow`) previously read as omitted and silently compared against the DEFAULT base/workflow — `ci-state.mjs --pr 5 --base` returned a real, wrong verdict at exit 0/1 with **no refusal at all**. This is the priority site named in the #167 PR review (the `type-design-analyzer` specialist) and the same fail-open class #61 fixed in `candidates.mjs`.
- **`board.mjs`.** `ledger`/`prev`/`spend-since`/`port`/`interval` are all read the same way. A trailing `--spend-since` silently widened the spend panel to all-time instead of the requested window; a trailing `--ledger` silently read the default ledger file instead of the one asked for. Same silent-widening class.
- **`diff-stats.mjs` / `pr-overlap.mjs`.** Already saved from silent widening by their own `if (!x) die(...)` guards immediately below the flag reads (`undefined` is falsy, so these already refused). Not silent. What was still missing: the refusal named the flag generically ("usage: ...") rather than pinpointing which flag was malformed, and `pr-overlap.mjs`'s `--a` consumed as `--b`'s value (`pr-overlap.mjs --a --b 5`) fell through to a confusing downstream `gh pr diff --b` error instead of a clean rejection. Hardened for both.

All four keep their own `die()`/exit-code convention rather than sharing one — `ci-state.mjs`/`diff-stats.mjs`/`pr-overlap.mjs` use `writeSync`-based `die()` at exit 2 (per #176), `board.mjs` uses its own `console.error`-based `die()`, also exit 2. Nothing here changes any script's exit-code contract or normal-invocation behavior.

## Widened search

Required by the ticket's acceptance criteria before the list could be claimed complete. Scope:

- **Repo-wide `*.mjs`/`*.js`** (excluding `node_modules`, `.worktrees`): only `skills/fleet/scripts/*.mjs` has this `arg()`-style trailing-flag reader. `ledger.mjs` has its own inline `--file` reader, safe for a **trailing** `--file` only — `if (!file) die(...)` catches that shape and nothing else. It is **still exposed to `--file <next-flag>`**: `file` becomes the next flag's own text (truthy, so the guard never fires) and the following `argv.splice(fileIdx, 2)` then eats that flag's token, so `--file --require-file check "<subject>"` silently loses `--require-file` and answers "safe to file" at **exit 0** where a correct invocation exits 2. That is the same value-eats-a-flag shape this PR fixes in `pr-overlap.mjs` (`--a --b 5`). Filed as **#362** rather than fixed here: `ledger.mjs` is outside this diff, PR #361 already modifies that file, and triage scoped #169 to the four `arg()` copies. An earlier revision of this section called it "already safe"; that was measured false and is corrected here. `fleet-tick.mjs`, `compute-spend.mjs`, `compute-board.mjs`, `strip-comments.mjs`, `workflows/review-pr.js`, `hooks/statusline.js` read no CLI flags at all.
- **Repo-wide `*.sh`** (excluding `node_modules`, `.worktrees`): 11 files, none use `--flag value` CLI parsing — every one validates positional args (`$1`, `$2`, ...) via a `case ... in ''|*[!0-9]*) die ...` guard, a different mechanism already safe against this class.
- `skills/fleet/scripts/claim-ticket.sh` was reached by this search and is positional-arg based like its siblings (so not in this defect class regardless) — left untouched per the note that it's #186's subject, staged separately in this run's pool.

## Tests

One `node --test` file per script pins the trailing-flag case (`pr-overlap.test.mjs` is new — no test file existed for that script before). Each new test:
- confirms the mutant (temporarily reverting the four/five sites' `arg()` back to the pre-fix body) turns exactly those new tests red and nothing else — verified by hand for all four scripts, restored after
- confirms a legitimate invocation with real flag values stays green (also verified by hand — `pr-overlap.mjs --a 5 --b 6` exits 0)

Full fleet suite: 458/458 green, before and after rebase onto `origin/main` (branch was already up to date).

## Review round (PR #360 review, applied in `3a181de`)

- **`arg()`'s other two branches were unpinned.** Mutation-measured: deleting `|| value.startsWith("--")` left board 16/16, ci-state 18/18, diff-stats 10/10 green; stripping `value.trim() === "" || ` from all four left the four suites 47/47 green. Added the next-flag case to board/ci-state/diff-stats and the empty-value case to all four — 9 new tests, each re-verified to kill its own mutant and stay green under the other's.
- **`board.mjs serve`**: a `--port` that could not be used falls back to 8123, and the `EADDRINUSE` message reported that substituted default as if the caller had chosen it — telling someone who *did* pass `--port` to "pass `--port <n>`". The message now marks the default as one. Refusing the bad value outright is out of #169's scope: **#366**.
- **Two comment/import cleanups** in the new test files, each checked by a refuter that ran the mutation rather than reasoning.

Suite: **467/467** green (458 before, 9 added).

Deferrals filed from this review: #362, #363, #364, #365, #366, #367, #368, and one refuted-and-recorded finding (see the PR comment below).


feigi:
## PR #360 review — deferrals filed

15 findings adjudicated: **4 survived**, **11 suggestions** scope-split, with one refuter dispatched per in-scope suggestion. Applied in `3a181de`; everything else is filed below, none dropped.

**Applied (6):** the two unpinned `arg()` branches pinned across the four suites (survived ×2), the `serve` bind error no longer naming a substituted default as the caller's choice (survived), and two comment/import fixes in the PR's own new test files (suggestions, each survived a refuter that ran the mutation). Suite 467/467.

**Filed (7 tickets + 1 record):**

| # | Label | What | Why not here |
|---|---|---|---|
| #362 | `bug` `ready-for-agent` | `ledger.mjs --file` swallows the next flag → `--require-file` silently off, exit 2→0 | outside this diff; PR #361 already edits that file; different hand-rolled reader |
| #363 | `bug` `needs-triage` | `board.mjs`: a single child emitting >64 KiB into a piped stderr may swallow `die()`'s refusal — the `console.error` half of the #246/#328 family | out of #169's scope; general form of the claim already refuted by gradient, narrow case unmeasured |
| #364 | `needs-triage` | `has()` is blind to `--flag=value`, so a boolean flag written that way reads as absent | different function, fail-*closed*; refusing vs accepting `--flag=true` is an open CLI decision |
| #365 | `needs-triage` | a misspelled flag reads as absent and takes the default — no unknown-flag sweep | different defect class; the #167 reference fix does not handle it either |
| #366 | `needs-triage` | `--port` / `--interval` accept a non-numeric value and silently default | invalid value, not absent value; pre-existing. **`--spend-since` is excluded — already fixed on PR #361** |
| #367 | `needs-triage` | extract one shared `arg()`/`has()` module — **four** findings, one vehicle | triage ruled "fix the four copies in place. No shared module." Filed to be re-ruled with post-#360 evidence, not implemented on this ticket |
| #368 | `needs-triage` | consolidate the per-file CLI-guard tests into one table-driven file | repo-wide test layout across five scripts, one outside this PR |
| #369 | `documentation` | **Record of the whole suggestion band** — disposition of all 11, the mutation matrix that refuted the `--flag=value` position-blindness claim, and one checked-but-deliberately-not-applied item | not an open defect; close once read |

The PR body's widened-search section claimed `ledger.mjs` was "already safe". That was measured false for `--file <next-flag>`; **the body has been corrected** and now links #362.

PR #361, opened 2026-08-11
fix(fleet): ledger check widens the term-selection stoplist, pins it with tests

### LINKED ISSUE #153: ledger.mjs check: longest-first term selection drops the most distinctive word from the tracker query
Deferred from PR #150 review.

## The finding

`skills/fleet/scripts/ledger.mjs:283-285` picks the three query terms by **length, longest first**, after dropping a 20-word stopword list and everything shorter than 3 characters. Since `gh` ANDs search terms, a bad pick can only subtract.

Measured against this tree:

| subject | query actually sent |
|---|---|
| `postgres connection pooling exhausted under sustained load` | `connection exhausted sustained` — **`postgres` dropped** for a generic adjective |
| `the guard should never fail open on a fork` | `should guard never` |
| `CI is red but the PR still merged` | `merged still red` — `CI` and `PR` erased by `length >= 3` |
| `the near-miss cap silently drops rows that should have surfaced` | `silently surfaced should` |

These queries return nothing, so the script prints *"tracker search … found no related issues"* with `ok: true`, `hits: []`, **exit 0** — a confident clean bill from a query that never had a chance, on the path its own test calls *"the only path that reads safe"*.

Live demonstration while filing these very deferrals: the subject *"review-and-fix.md files issues without running ledger.mjs check"* produced the query `without running review`.

## Why the approach is open

Three candidate directions, none obviously right:

- widen `STOP` with modals, negations and temporals (measured by the silent-failure specialist as zero-fallout, 209/208/0)
- prefer identifier-shaped tokens (`candidates.mjs`, `CI`, `gh`) over long ordinary words — but `norm()` has already destroyed the punctuation that identifies them
- real frequency weighting, which the code comment at `:270` already names as the upgrade path

Lowering `length >= 3` interacts with all three: the code-simplifier measured that `>= 3` → `>= 1` currently reds **no test**, so term selection has no regression gate in either direction.

## Why deferred

The comment at `ledger.mjs:264-272` already discloses longest-first as a crude proxy with a named upgrade path, and the feature works for the case #145 was filed about (#114 is caught at score 1.0). Tuning an admittedly-crude heuristic that no test covers, on the final PR of a run, is churn — and picking among the three directions is a design call.

feigi:
## Summary

Triage's ruling on #153 (pin term selection, then widen the stoplist with modals/negations/temporals — nothing more; reject identifier-shaped preference and frequency weighting):

- `ledger.mjs check`'s term selection had no regression gate in either direction. `STOP` now excludes modals/negations/temporals ("should", "never", "still", ...), which were displacing real content words in both the tracker-search query and near-miss scoring — the two consumers share the same `STOP` set. `"the guard should never fail open on a fork"` now queries `guard fail open`, not `should guard never`.
- Six new tests exact-pin the selected terms/scores for representative subjects, including the four the issue measured, so a stoplist/floor/sort/cut change now reds a specific test. Verified by mutation: reverting the widening reds exactly the four term-selection tests it should; reverting the wording change (below) reds exactly the wording test; an unrelated change (`gh --limit`) reds neither.
- A zero-hit tracker search now reports `"returned no matches"`, not `"found no related issues"` — a 3-term heuristic query coming up empty establishes that the query found nothing, not that the tracker has nothing. The JSON `verdict` contract from #152 is unchanged; only the human-readable prose changed.
- The code comment records what stays broken on purpose, per triage's "state the ceiling honestly": longest-first still prefers a long ordinary word over a short distinctive one (`postgres` still loses to `exhausted`/`sustained`; `cap`/`near-miss` still loses to `silently`/`surfaced`) — that's the frequency-weighting upgrade path, not attempted here — and the `>= 3` floor still erases `CI`/`PR` outright, since `norm()` has already destroyed the punctuation that would mark them as identifiers. Lowering that floor was explicitly out of scope.

## Test plan

- [x] `./agent-test skills/fleet/scripts/ledger.test.mjs` — 24/24 pass
- [x] `./agent-test` (full suite) — 454/454 pass
- [x] Mutation-tested the new tests: reverting the `STOP` widening reds exactly the 4 term-selection tests that should catch it; reverting the wording change reds exactly the 1 wording test; an unrelated change (`gh --limit 5` → `6`) reds nothing

## Also carried: cockpit spend-panel rescue (orphaned `be0dd19`)

This branch now additionally carries a rescue of orphaned commit `be0dd19`, a second round of cockpit spend-panel fixes that never reached `main`. It was authored five minutes **after** its own PR #10 had already merged, on branch `fleet/cockpit-spend-fixes`, and never got a PR of its own. Five of its six fixes were still missing from `main`; it touches only the cockpit files (`board.*`, `compute-spend.*`, the cockpit design doc) and is disjoint from the ledger work above.

- `findSubagentsDir` ranks on the newest `*.jsonl` mtime, not the subagents **directory's** mtime. A directory's mtime moves only when an entry is created or removed, never when a transcript inside it is appended to — so ranking on it tracked the last agent *spawn*, not activity. The cockpit launches in run-team phase 0, before the first agent exists, so the live session had no dir yet and a *previous* session won: the board rendered a prior run's spend as this run's, then silently switched.
- `computeSpend` reports every bucket that collected something, iterating `byRole` rather than `ROLE_ORDER`. Roles outside that array were dropped from the table while their tokens stayed in totals, so the percentage column silently stopped summing to 100.
- `encodeProjectDir` covers every non-alphanumeric character, not just `/` and `.`.
- `gatherSpend` distinguishes its three outcomes instead of collapsing them into one bare `null`: `{ error }` (a bug — the UI now says so), `null` (nothing yet — panel stays hidden), or the model. Collapsing them is what let the path bug live unnoticed. Related: one unreadable transcript no longer discards every other agent's numbers, and errors reach the browser rather than a stderr nobody is watching.
- `--spend-since` validates instead of `Number(x) || null` silently meaning "no filter at all", which turned a typo — or the natural seconds-instead-of-milliseconds mistake — into a panel rendering confidently over the whole session.

**Dropped during the rescue:** `be0dd19`'s sixth fix, the finisher classification, had already landed independently on `main` via `2e2531d`. `main`'s version is kept verbatim — its `/review pr|review-pr-|fix pr|fix-pr-/` also catches `fix-pr-<n>`, which `be0dd19`'s narrower line missed. Its accompanying `^finish-` comment block was dropped as duplicate: `main` already carries a finisher comment, and the classifier line itself was already identical on both sides.

One test-setup conflict in `board.test.mjs` resolved in favour of `be0dd19`'s per-file `utimesSync` stamps, which is what the new ranking actually reads; `main`'s directory-level stamp is dead under it. No assertion from either side was dropped — `main`'s tie-break rationale is carried forward as a comment.

### Test plan (rescue)

- [x] `./agent-test skills/fleet/scripts/board.test.mjs` — 17/17 pass
- [x] `./agent-test skills/fleet/scripts/compute-spend.test.mjs` — 22/22 pass
- [x] `./agent-test` (full suite) — 460/460 pass, 0 fail

Closes #153

https://claude.ai/code/session_016pZCzSMbRBet2Lp4ebdMuj


feigi:
Review findings applied in `234e668`. **27 findings: 11 survived, 16 suggestions.**

**Applied (18):** all 11 survived, plus 7 of the suggestion band.

**Deferred, filed:**
- **#371** — `board.html`'s `renderSpend` has zero test coverage, and the tri-state `spend` contract it duck-types is prose only. `needs-triage`; the approach is genuinely open (reshape to `{ ok }` vs. extract the branch logic as a testable pure function). Deferred on scope: #153 has no business in a cross-file contract reshape. Note the SURVIVED #8 fix shipped into that untested file.
- **#372** — record of the whole suggestion band: 7 applied, 5 refuted with the evidence inline, 1 deferred on scope. `documentation`, close once read; nothing in it is an open defect.

**The one that mattered most.** The two tests claiming to pin "rank on newest transcript mtime, not directory mtime" did not discriminate that fix — reverting `board.mjs` to `m: statSync(d).mtimeMs` left both green, and all 17 tests in the file with them. The fixtures made the two rankings agree by construction. Rebuilt so the newest *directory* is not the one holding the newest *transcript*, and mutation-verified: the reverted mutant now reds it.

`tests 466 / pass 466 / fail 0` (baseline 460). Mutation matrix: 7 mutants killed, 2 controls green, every mutation asserted to have actually changed the file before the run.

feigi:
Addendum — head is now `a5a68b9`, not `234e668`.

`234e668` left the PR **CONFLICTING**, so no `pull_request` run started at all. Cause: PR #360 (merged) adds `spawnSync`/`createServer`/`fileURLToPath` to `board.test.mjs`'s import block, and my new CLI tests added the same imports there. Two adjacent inserts at one point conflict regardless of content — matching main's exact positions still left main's one-sided `node:net` line colliding.

Resolved without a rebase, by moving the tests rather than the imports: the `--spend-since` cases now live in `skills/fleet/scripts/board-cli.test.mjs`, which they earn on their own — they need a stub `gh` on PATH and a fake `HOME` (gather() reaches this guard only after its gh reads), where #360's `--ledger`/`--port` cases die before any gh call and need neither. `board.test.mjs` is back to main's exact import block.

- `mergeable=MERGEABLE`
- `check: success` on `a5a68b9` (run 31479041429, bound to head). `rebase-check: failure` is `behind=3` staleness — the merge bot rebases at merge time.
- Suite on the branch: `tests 466 / pass 466 / fail 0`.
- Suite on the **simulated merge with current main**: `tests 485 / pass 485 / fail 0` — so main's hardened `arg()` and the new range guard coexist.

PR #373, opened 2026-08-11
docs(agents): issue-tracker carries the Agent Brief and Respec rules it dropped

### LINKED ISSUE #79: docs/agents/issue-tracker.md:34-37 drops the Agent Brief and Respec sentences from the issue-read block it otherwise copies verbatim
## Problem

`skills/fleet/skills/run-team/SKILL.md:149-154` carries the implementer issue-read prompt block. It is **four** sentences:

1. `Read the issue with `gh issue view <N> --json title,body,comments --jq '…'`.`
2. `` Not bare `gh issue view <N> --comments` — … so the loss is silent. ``
3. `` The `## Agent Brief` comment is authoritative over the issue body. ``
4. `` Honor its `Respec` block — it may explicitly rule out hypotheses the body raises. ``

`docs/agents/issue-tracker.md:34-37`, under `## When a skill says "fetch the relevant ticket"`, reproduces sentences **1 and 2 word-for-word** (differing only in line wrapping) and **omits 3 and 4**.

```
$ git grep -n -I -F 'Not bare' <head>
docs/agents/issue-tracker.md:35
skills/fleet/skills/run-team/SKILL.md:150

$ git show <head>:docs/agents/issue-tracker.md | sed -n '34,37p'   # ends at "the loss is silent."
$ git show <head>:skills/fleet/skills/run-team/SKILL.md | sed -n '152,154p'
> body either way, exit 0, so the loss is silent. The `## Agent Brief` comment
> is authoritative over the issue body. Honor its `Respec` block — it may
> explicitly rule out hypotheses the body raises.
```

## Why it matters

This is not cosmetic duplication. The two dropped sentences are the **only** statements of the Agent-Brief-outranks-body rule and the `Respec` rule in this file, and `docs/agents/issue-tracker.md` is where the global `CLAUDE.md` sends every agent for issue operations ("Issues live in `feigi/claude-config` GitHub Issues, via the `gh` CLI. See `docs/agents/issue-tracker.md`").

So an agent that reads the tracker doc — the documented entry point — gets the command and the `--comments` caveat but never learns that the `## Agent Brief` comment outranks the issue body, or that a `Respec` block can rule out hypotheses the body raises. That is the exact failure the block exists to prevent, and it is silent: the copy looks complete because the part it does carry is verbatim.

**Already diverged.** The other sites carry paraphrases (`next-ticket/SKILL.md:41`, `sizing-a-ticket/SKILL.md:11`) that do mention `Agent Brief` / `Respec`. `issue-tracker.md` is the one that dropped them outright.

## Desired behavior

Either is acceptable:

- **Point at the block** — replace `:34-37` with a reference to `run-team/SKILL.md:149-154` as the authoritative text, or
- **Restore the two sentences** so the copy is complete.

Prefer whichever matches how the rest of `docs/agents/` handles shared text.

## Provenance and scope

Surfaced by the `comment-analyzer` specialist during the **PR #71** review (issue #13, the block-count correction), independently corroborated by `code-reviewer`, and re-measured before filing.

Pre-existing at `origin/main` — PR #71 neither caused nor worsened it. #71 removed the *spec's* duplicate copy of the prompt blocks; blocks 1 and 3 are now unique repo-wide, and block 2 is the one whose text survives elsewhere. #71 deliberately did not widen to fix this.

Two corrections to what was reported, for whoever picks this up:

- The specialist described this as "the first four sentences verbatim, dropping the last two". Re-measured, the block is **four** sentences total and the copy carries **two**. The *substance* — that the `Agent Brief` and `Respec` sentences are the missing ones — is correct.
- Cross-referenced on **#23**, which tracks the related but distinct question of the divergent `--comments` caveat wording across three *skill* files. #23 owns "which wording should the skills share"; this ticket owns "the tracker doc silently lost two rules". Whoever takes either should read both — #23's table lists three sites and does not include `issue-tracker.md`.

Deferred from PR #71 review

feigi:
## What

`docs/agents/issue-tracker.md`, under `## When a skill says "fetch the relevant ticket"`, reproduced the issue-read command and its bare `--comments` caveat verbatim from run-team's implementer prompt but carried neither the rule that the `## Agent Brief` comment outranks the issue body nor the `Respec` clause. Both sentences are restored word-for-word; the only difference from the source is where lines wrap.

Those two sentences were stated nowhere else in the file, and the global `CLAUDE.md` routes every agent here for issue operations — so the copy read as complete while both rules were absent.

## Which fix shape, and why

The ticket named two acceptable shapes and tie-broke on "whichever matches how the rest of `docs/agents/` handles shared text". The tree settles it toward **restore**, not a pointer:

- Every file under `docs/agents/` is self-contained, and each one directs skills to defer to *it*: `triage-labels.md` ("When a skill mentions a role … use the corresponding label string from this table"), `domain.md` ("Where an installed skill defers to `/domain-modeling`, do the glossary or ADR edit inline instead"), and this file already overrides `/wayfinder`'s `SKILL.md` outright ("this file overrides it; skip the label step").
- Authority in this tree runs `CLAUDE.md` → `docs/agents/` → skills. Pointing the tracker doc at `skills/fleet/skills/run-team/SKILL.md` would invert that, and run-team's block is a verbatim subagent prompt that cannot itself reference anything.
- A cross-file line citation would drift. The ticket's own reference to `run-team/SKILL.md:149-154` resolves to `:238-245` at this head.

## Tests

Two prose pins added in `skills/fleet/scripts/issue-tracker-prose.test.mjs`, sliced to the one section so they cannot be satisfied by text elsewhere in the file. Mutation-tested four ways: dropping either sentence reddens that pin and only that pin; relocating both out of the section reddens both (the slice anchors); rewrapping the section leaves both green.

Suite: 467 → 469, `fail 0`, `cancelled 0`.

## Out of scope

- **#23** — which `--comments` caveat wording the three *skill* files should share. Untouched here; it is sequenced after this ticket. No skill file is modified.
- run-team's block carries a further sentence this section still lacks (read the issue **before touching code**, bail if undecided). That is a different rule from the two this ticket names, so it is left alone rather than widened into.

Closes #79


feigi:
## Review findings applied — `799c7ad`

**Applied (3):**

- **[S1] critical · comments** — the header comment's claim that *"the global `CLAUDE.md` routes every agent here for issue operations"* is false at this head. `CLAUDE.md` has no match for issue/ticket/tracker/gh pr in 93 lines; it routes to `docs/agents/triage-labels.md` and `docs/agents/domain.md` by the `See docs/agents/X.md` pattern and carries no such line for this file. `git grep -l issue-tracker` matches exactly one path in the whole tree — the test file making the claim. **This claim was inherited from #79's own body**, so the ticket is not usable as a citation for it.
- **[S2] critical · comments** — the claim that this section *"defines the phrase 'fetch the relevant ticket' that the fleet skills defer to"* is false: no skill file contains that phrase. Replaced with what greps settle — the rules reach agents through three independent SKILL.md copies (`run-team` word-for-word inside its verbatim implementer prompt, `next-ticket` and `sizing-a-ticket` paraphrased), none of which mentions this doc. The doc is the catch-up copy, not their source.
- **[U1] suggestion · correctness** — refuter returned **refuted=false**. With `## Wayfinding operations` gone the fetch section becomes the file's last, `indexOf("\n## ")` returns -1, and a *correct* doc reported 2 failures.

  **Took a stricter fix than the finding suggested.** The literal suggested fix (EOF fallback on `\n## ` alone) flipped the rules-relocated-under-a-later-`#`-h1 mutant from **red to green** — it widens the slice past an h1, and an h1 closes the section. Since slice size is what does the anchoring in a prose pin, that is a real weakening. The applied version stops at `#` or `##` and falls back to EOF.

**Deferred (1):** [U3] · `dimension: tests` — the same two rules are triplicated across three SKILL.md files this PR does not touch and this test does not cover. Out of scope (pre-existing, not worsened here). **Not filed as a new issue:** #23 already owns unification across those exact three files and is `ready-for-agent`, so a rival ticket would collide on the same lines. Recorded there instead → #23 (comment), with the routing facts above, since #23 may be assuming a dependency direction that does not exist yet.

**No action (1):** [U2] · `dimension: tests` — affirmative finding (both tests discriminate on every mutation shape tried; the one limitation is self-disclosed in the header). Its own suggested fix is "none needed". The ceiling paragraph it depends on is preserved verbatim in the rewritten header.

**Verification.** Mutation matrix of 11 variants run against the *applied* file, not a candidate string: red on each rule deleted separately, section deleted, rules relocated under a later `##`, under a later `#`, and each of those with the section last; green on pristine, on rules nested under a later `###`, on an unrelated edit elsewhere, and on a correct doc whose fetch section is last. **0 mismatches.** Suite `node --test skills/fleet/scripts/*.test.mjs` → **469/469 pass, 0 fail**; `./agent-test` on the changed file → 2/2.

The restored doc prose was confirmed accurate and is **untouched** — the diff is the test file only. `93b90b5`'s commit body carries the same false routing claim; `799c7ad`'s body is the retraction, as amending a pushed commit was not worth a force-push.

Issue #204, opened 2026-07-31
candidates.mjs --jq uses \s and \d, whose classes differ between gh's gojq and the system jq the suite tests with — a U+00A0 in a heading leaks a spec


feigi:
`candidates.mjs:70-73` builds the `--jq` reduction with `\s` and `\d`. `gh` applies that expression with its **embedded gojq** (Go's RE2), while every test in `candidates.test.mjs` applies it with the **system jq** (Oniguruma). The two classes are not the same:

- Oniguruma `\s` / `\d` are Unicode-aware.
- Go's RE2 `\s` is `[\t\n\f\r ]` and `\d` is `[0-9]` — ASCII only.

Both engines *accept* the pattern, so nothing errors. They disagree on the **input**.

## Measured

`gh` 2.96.0, `dep github.com/itchyny/gojq v0.12.17` (`go version -m "$(which gh)"`); system jq `jq-1.7.1-apple`. The literal char in both commands below is U+00A0 (no-break space):

```
$ jq -n '"## User Stories "|test("(?m)^##\\s+User Stories\\s*$")'
true      # system jq — what the suite runs
$ gh api repos/feigi/claude-config --jq '"## User Stories "|test("(?m)^##\\s+User Stories\\s*$")'
false     # gojq — what production runs
```

Same split in the `d:` scan, with an Arabic-Indic digit:

```
$ jq -n '"depends on #١٢"|[scan("(?i)(?:depends on|blocked by|requires|after)\\s+#\\d+")]|length'
1
$ gh api repos/feigi/claude-config --jq '"depends on #١٢"|[scan("(?i)(?:depends on|blocked by|requires|after)\\s+#\\d+")]|length'
0
```

Ruled out as the cause: `(?m)` behaves identically in both (both anchor per-line with it, both anchor at string bounds without it, and neither reads `m` as dot-matches-newline — measured). Key ordering is harmless: `candidates.mjs:111-118` type-checks properties, not order.

## Why it matters

Production is the **stricter** matcher. A to-spec spec whose `## User Stories` heading carries a U+00A0 reduces to `spec:false` under `gh`, so `dropSpecs` never fires and the spec ships as a claimable ticket. `candidates.mjs:155` states the stakes:

> This is the ONLY line of defence — nothing downstream catches a spec.

The failure is silent in both directions: the suite is green because it runs the permissive engine, and production drops nothing because it runs the strict one.

[...]

PR #377, opened 2026-08-11
fix(fleet): widen candidates.mjs's spec predicate to any heading depth, bound its separator to horizontal whitespace

### LINKED ISSUE #65: candidates.mjs: the to-spec heading predicate has two undecided shapes — nested h3 leaks, split heading false-positives
## What to build

Two real body shapes get surprising verdicts from the `## User Stories` predicate, and neither is a deliberate decision:

- `### User Stories` is **kept** — a spec nested one level deeper would leak into the queue.
- A body whose line is exactly `##`, with the next line starting `User Stories`, is **dropped** — `\s` spans newlines, so the separator crosses a line break.

The split-heading false positive is a straightforward fix (bound the separator to horizontal whitespace). The h3 case is a genuine question: does to-spec ever nest its template under a parent heading? If yes the predicate wants `#{2,}`; if no, current behaviour is correct and should be recorded as intended rather than left accidental.

Near-miss tests added in #59 pin todays behaviour, so they must be updated deliberately by whatever this decides.

## Acceptance criteria

- [ ] Separator bounded so a heading cannot span a line break
- [ ] The h3 question answered either way, with the answer stated in the comment
- [ ] Near-miss fixtures updated to match the decision
- [ ] A mutation that loosens the predicate still fails at least one test

## Blocked by

feigi:
## Summary

`candidates.mjs`'s to-spec spec-detection predicate had two undecided shapes, per #65's triage ruling:

- `### User Stories` (nested one level deeper) was **kept** as an ordinary ticket — the only line of defence against a leaked spec stepped over silently, since the predicate matched exactly `##`.
- A body whose line is exactly `##`, with `User Stories` starting the next line, was **dropped** as a spec — `\s` spans newlines, so the separator crossed the line break.

Per triage's ruling: widen the depth (to-spec's own `<spec-template>` is flat — every heading is `##`, nothing nests — so widening costs nothing against real output, and the asymmetry between a silent/expensive leaked spec and a loud/logged false drop decides it), and bound the marker/text separator to horizontal whitespace only (`[ \t]+`), so a heading can never span a line break.

- `spec:` predicate: `^##\s+User Stories\s*$` → `^#{2,}[ \t]+User Stories[ \t]*$`
- The decision and its reasoning are recorded in the comment above `dropSpecs`.
- Near-miss fixtures updated: nesting (`### User Stories`) moved from "kept" to "dropped"; the split-heading fixture moved from "dropped" to "kept".
- New tests for depth ≥2 detection and trailing-whitespace-still-matches.
- New gojq-checked test (`#63` — the RE2 engine gh actually applies) confirming both decisions hold under the production engine, not only system jq.

## Test plan

- [x] `./agent-test skills/fleet/scripts/candidates.test.mjs` — 51/51 pass, including the new gojq-backed test (gojq available locally)
- [x] Repo-wide: `node --test skills/fleet/scripts/*.test.mjs` — 470/470 pass
- [x] Mutation-tested both changed clauses: reverting `#{2,}` → `##` fails exactly the two depth tests and nothing else; reverting `[ \t]+` → `\s+` fails exactly the two split-heading tests and nothing else — each mutant confirmed and reverted

Closes #65


feigi:
## Review applied — `db5a833`

`node --test skills/fleet/scripts/*.test.mjs` → **tests 469, pass 469, fail 0, skipped 0** (470 at `afa08cc`; the delta is the one test folded away below, not a regression).

### Two defects, applied

**CRLF regression, introduced by this PR** (`correctness`/`silent-failure`/`types`). Narrowing the trailing class from `\s*$` to `[ \t]*$` broke `## User Stories\r\n` — under `(?m)` the `$` matches before the `\n`, so the line still ends in `\r`. GitHub's web textarea submits CRLF, so that spec leaked into the queue silently, past the only line of defence. Restored `\s*` at the trailing position only; the separator stays `[ \t]+`, which is what the PR's rationale was actually about. Pinned in **both engines**.

**Depth floor unpinned — #65's acceptance criterion unmet** (`correctness`/`tests`). Loosening `#{2,}` to `#+` passed the entire test file. Now red. Depth also bounded above to `#{2,6}`, matching `depnums`' own `#{1,6}` and CommonMark — a 7-`#` line is not a heading and was being dropped as the signature.

Mutation results, final tree:

| mutation | expected | result |
|---|---|---|
| `\s*$` → `[ \t]*$` (this PR's head) | red | ✖ depth test **and** ✖ gojq test |
| `#{2,6}` → `#+` (#65's named criterion) | red | ✖ near-miss test |
| `#{2,6}` → `#{1,6}` / `#{2,}` | red | ✖ near-miss test (each) |
| `[ \t]+` → `[ ]+` | red | ✖ depth test |
| `\s*$` → `[\r]*$` | red | ✖ depth test |
| `#{2,6}` → `##{1,5}` (same language) | **green** | ✔ 3/3 |

Without the two new near-miss fixtures, `#+` is green — the hole was real.

### Also applied — six `suggestion`-band findings, each dispatched to a refuter, none refuted

`types` tab separator (narrowing `[ \t]` to `[ ]` left all 470 green) · `simplify` fold of the absence-only trailing test into the depth test's run, making it positive (it passed against an early `process.exit(2)`) · `simplify` removal of the redundant `dropped #N` stderr assertion · `simplify` hoist of the duplicated gojq skip option · `simplify` removal of two gojq assertions flat across both predicates and both engines · `simplify`/`comments` de-duplication of the restated rationale. `correctness`/`types` depth-ceiling findings are subsumed by `#{2,6}`.

Comments falsified by the fix were corrected: the depth is no longer unbounded, and the gojq block no longer claims the predicate has no engine-divergent position — the trailing `\s*` is one (U+00A0: jq `true`, gojq `false`), which is #204 and unchanged here.

### Deferred

- **`comments`** — `candidates.test.mjs:476` cites `candidates.mjs:239` for the #133 rationale that lives at `:332`. Pre-existing, not introduced here → added to **#129**, the open aggregator for this exact class: feigi/claude-config#129 (comment). No new issue filed — #129 already tracks four instances and asks for the structural fix.
- **`docs`** — `docs/specs/2026-07-30-fleet-trust-the-label-design.md:235` quotes the old regex. Not filed: it is a dated design record (`Status: implemented in PR #59`, pinned to `6eb0442`) carrying an explicit disclaimer that its references describe the files as they stood then. A document that says it is historical is not misdescribing current code, and this PR does not touch it.

Issue #334, opened 2026-08-11
release-ticket.sh: three measured-unreachable hardening nits deferred from PR #330 review


feigi:
Deferred from PR #330 review (closes #84). All three are `suggestion` severity, all three were checked by a refuter biased to refuse, and **none is a live defect**. Filed for the record so they are not rediscovered; the maintainer call is whether any of them is worth code at all.

## 1. No `|| die` on `wt_list=$(git worktree list --porcelain)` — REFUTED

**Dimension:** correctness. **Site:** `skills/fleet/scripts/release-ticket.sh`, the `wt_list=` assignment (cite by construct, not line — PR #330 moved it and #332 will move it again).

**Claim:** the assignment has no `|| die`, so a failure exits via `set -e` at 128 with git's raw fatal and no `release-ticket:` prefix, unlike `inflight.sh`'s `if ! worktrees=$(git worktree list --porcelain); then die ...`.

**Why refuted (measured, git 2.50.1):**

- The registry cannot make the command exit non-zero. 17/17 registry mutations — empty/removed/garbage/binary `gitdir`, garbage/empty/removed `HEAD`, garbage/removed `commondir`, entry-as-regular-file, entry emptied/removed, checkout deleted, symlink loop, `gitdir` as a directory, and `chmod 000` on the registry / an entry / `gitdir` — all exit **0**. That is precisely what the script's own comment asserts, so the `|| die` would be dead code for every registry fault.
- The states that *do* produce rc=128 (broken repo, `repositoryformatversion=99`, missing `.git/objects` or `.git/refs`, outside a repo, deleted cwd) all fail the script's earlier `git rev-parse --git-dir || die "not inside a git repository"` first — verified end to end, rc=2 with the prefix.

The consequence claim is accurate if it were ever reachable (a PATH shim forcing rc=128 does print an unprefixed `fatal:`), but the residual is a TOCTOU race that applies identically to every other git call in the script and is not registry-driven.

**Citation confirmed:** `inflight.sh:202-204` really does carry that wording — but re-derive the number, #332 is open against that file.

## 2. No `[ -n "$common" ]` after `git rev-parse --git-common-dir` — REFUTED

**Dimension:** silent-failure. **Site:** `release-ticket.sh`, the `common=` assignment.

**Claim:** an empty `common` at exit 0 makes `wtroot` become `/worktrees`, `[ -e "$wtroot" ]` false, and the whole fail-closed guard a silent no-op.

[...]

Issue #208, opened 2026-07-31
release-ticket.sh: `done_wt` records the exit code, not what landed — `worktree remove` can clear the registration and still fail


feigi:
Deferred from PR #203 review (closes #85). Found by the silent-failure-hunter specialist, reproduced independently against bare git.

## Problem

`skills/fleet/scripts/release-ticket.sh:376-379`:

```sh
if ! err=$(git worktree remove "$wt" 2>&1); then
  halt "git worktree remove refused $wt: ..."
fi
done_wt=true
```

`done_wt=false` means "the command returned non-zero". It does not mean "the removal had no effect". git deletes the admin dir `$GIT_DIR/worktrees/<id>` even when the subsequent directory removal fails — `remove_dir_recursively` failing does not skip `delete_git_dir`.

## Measured

git 2.50.1 (Apple Git-155), no script involved. Clean worktree, one subdirectory at mode 555:

```
--- registry BEFORE: 2
error: failed to delete '.../wt-x': Permission denied
rc=255
--- registry AFTER: 1
--- admin dir: ls: .git/worktrees: No such file or directory
--- dir still on disk: wt-x
```

The registration is gone, the directory and its files are not.

## Why it matters now

Before #203 this path printed `PARTIALLY RELEASED`, which was accurate by luck. After #203 it prints:

```
release-ticket: #N HALTED mid-release — nothing landed: git worktree remove refused ...
    worktree removed: false, branch deleted: false, in-progress: still on the issue
```

`nothing landed` is a positive assertion and it is false — the registration landed. #203 promoted an already-wrong detail value into the headline. This is not the wording question #85 settled; the wording is right, the boolean behind it is not a measurement.

## Knock-on (pre-existing, not introduced by #203)

Next run sees no `wt` and no `stray`, deletes the branch, drops the label, exits **0** with `"released":true,"blockers":[]` over an orphaned directory — which then makes the next `claim-ticket.sh` for that slug fail.

## Bounding

git runs `check_clean_worktree` first, so this is only reachable on an already-clean worktree: no member's uncommitted work is lost. Reaching it needs a mid-delete permission failure.

## Open question (why needs-triage, not ready-for-agent)

The obvious 3-line fix is to recheck the registry after a refusal rather than trust the rc:

```sh
git worktree list --porcelain | grep -qx "worktree $wt" || done_wt=true
```

[...]

Issue #135, opened 2026-07-31
release-ticket.sh: a .git naming a gitdir whose core.worktree points elsewhere still makes the dirty check answer about another repo at rc 0


feigi:
PR #115 added a `.git` linkage guard to `release-ticket.sh`'s dirty check, closing #74 (absent `.git`) and, in review, an empty `.git` **directory**. Both leaked the same way: `git -C "$wt" status --porcelain` walks UP to the enclosing repository and reports the **parent's** status at rc 0.

One input in that class survives the shipped predicate. A `.git` file naming a gitdir whose `core.worktree` points at some **other** directory is a well-formed regular file, so `[ -f "$wt/.git" ]` is true, and git then answers happily — about the other working tree.

Measured, git 2.50.1:

```
$ printf 'gitdir: %s\n' "$D/other/.git" > "$wt/.git"
$ git -C "$D/other" config core.worktree "$D/other"

$ [ -f "$wt/.git" ] && echo TRUE
TRUE                                  # shipped predicate passes

$ git -C "$wt" status --porcelain; echo "rc=$?"
rc=0                                  # empty

$ git -C "$wt" rev-parse --show-toplevel
/…/other                              # a completely different working tree
```

So the dirty check returns a positive "this claim is clean" derived from an unrelated repository, with the claim's own uncommitted files sitting untouched in `$wt`.

## Severity: low, and nothing is destroyed today

`git worktree remove` refuses independently, the script turns that into a `halt`, and the label is dropped **after** the worktree and branch — so the claim survives intact. The cost is the same one PR #115's commit message objects to: the run reaches `halt`, exits 2 announcing a PARTIALLY RELEASED claim that was never touched, and the tracker was already asked. A second, unrelated guard re-validating what the first was supposed to establish.

It is also an exotic input — it takes a hand-written `.git` or a hand-set `core.worktree` to reach. That is why PR #115 deliberately did **not** widen to cover it, and why the guard's comment there was changed to claim only what it delivers:

```sh
# Existence, NOT "points at this worktree", which is deliberately not claimed:
# a .git naming a gitdir whose core.worktree is some other directory passes
# this and still answers about that other tree at rc 0 (measured). Stopped
# downstream by `worktree remove` today, and left to its own ticket rather
# than widened into here.
```

This is that ticket.

## Why the approach is still open

[...]

Issue #94, opened 2026-07-31
release-ticket.sh:190 dirty check reads CLEAN at rc 0 when the worktree directory is unreadable — untracked-only work goes invisible


feigi:
Deferred from PR #72 review. **Pre-existing** — verified identical on base `73d7faf` and on the post-fix head. Surfaced by `comment-analyzer`, reproduced independently.

## Problem

`skills/fleet/scripts/release-ticket.sh:190`:

```sh
if ! dirty=$(git -C "$wt" status --porcelain); then
  die "cannot read the status of $wt, ..."
fi
```

When the worktree directory is **readable-less but executable** (`chmod 111`), `git status --porcelain` does **not** fail. It exits **rc 0**, writes `warning: could not open directory ...: Permission denied` to **stderr**, and prints nothing on stdout. The script deliberately does not fold stderr into `$dirty` (`:189`, and correctly so — folded-in stderr would be counted as changes), so `$dirty` is empty and the check reports **no uncommitted work**.

The `die` never fires, because nothing failed.

## Measured — git 2.50.1, untracked-only work

```
$ echo "PRECIOUS UNTRACKED WORK" > .worktrees/9-x/precious.txt
$ chmod 111 .worktrees/9-x

$ [ -d .worktrees/9-x ]                      -> TRUE   (so the check runs)
$ git -C .worktrees/9-x status --porcelain ; echo rc=$?
warning: could not open directory ".": Permission denied      <- stderr only
rc=0                                                          <- stdout empty
```

The guard that exists to answer *"does this worktree hold uncommitted work?"* returns **clean** for a worktree it was never able to look inside.

**Untracked-only is the bad case.** A *modified tracked* file still shows (` M f.txt`) because that comes from the index, so it blocks. Work that exists only as untracked files — the common shape for scratch output, a new file not yet `git add`ed — is invisible.

## What stops it today, and why that is not a fix

```
$ sh release-ticket.sh 9 x fix --apply
exit 2  PARTIALLY RELEASED — git worktree remove refused ...:
        warning: could not open directory ".": Permission denied
        error: failed to delete ...: Directory not empty
$ ls .worktrees/9-x/precious.txt   -> survives
$ branches left: fix/9-x main      -> survives
```

[...]

PR #380, opened 2026-08-11
fix(fleet): release-ticket.sh's dry run predicts a locked/non-directory worktree the same as --apply

### LINKED ISSUE #86: release-ticket.sh: the dry run reports `"released":true` exit 0 where `--apply` exits 2 — the `:188` / `:241` predicate asymmetry
Deferred from PR #72 review. Introduced by that PR; surfaced by `type-design-analyzer`, reproduced independently.

## Problem

PR #72 changed the dirty-check guard at `:188` to `[ -n "$wt" ] && [ -d "$wt" ]`, while the removal at `:241` still runs on `[ -n "$wt" ]` alone. The two now disagree about which worktrees they act on, and the dry run predicts from the first while `--apply` executes the second.

## Measured — a gone directory whose worktree is locked

```
$ git worktree lock .worktrees/9-x && rm -rf .worktrees/9-x

DRY:   exit 0   {"released":true,"applied":false,"blockers":[]}
APPLY: exit 2   PARTIALLY RELEASED — git worktree remove refused .../9-x
```

A regular file at `$wt` produces the same divergence.

The dry runs entire job is to predict the apply, and here it promises a clean release that the apply then refuses. Nothing is lost — the apply fails closed — but the prediction is wrong, and `--apply` is gated on operators trusting it.

## Still present after the #72 follow-up

Reproduced against `1e54bac`, which fixes the unreadable-path fail-open but does not change the `:188` / `:241` asymmetry.

Whether the dry run *can* predict this without attempting the removal. Both divergent cases hinge on what `git worktree remove` will refuse, and knowing that requires the same `stat` (plus the lock state) the guard already could not settle. Options: teach the dry run to check the lock file and the file-type explicitly, accept the divergence and document it, or have the dry run say "removal may be refused" rather than `released:true`. None is obviously right.

Verified on git 2.50.1.

feigi:
## Summary
- The dirty check (old `:188`) only opened when the claim's worktree path was a directory, while the removal step (old `:241`) ran whenever the path was non-empty. A locked worktree or a regular file sitting at the worktree path cleared every guard, so the dry run reported `"released":true` and `--apply` then hit `git worktree remove`, which refused it (rc 128) mid-flight — exit 2 `PARTIALLY RELEASED` where the dry run had promised a clean exit 0.
- Both cases are now decidable up front, without a second `git` call: lock state is read from the `locked` line in the porcelain worktree listing the script already captured for `wt`/`stray`, and "exists but is not a directory" is a plain `-e`/`-d` test. Both become ordinary blockers, so the existing blocked-report path (exit 1, `"released":false`, nothing touched) now fires identically in dry run and under `--apply`, before anything is attempted.

## Test plan
- [x] New test: a locked worktree (directory present) blocks the same way in a dry run and under `--apply`
- [x] New test: a locked worktree whose directory was removed by hand still blocks (the exact `#86` repro), instead of `--apply` reaching a mid-flight refusal
- [x] New test: a regular file at the worktree path blocks instead of promising a release
- [x] Mutant-killed: reverted the `.sh` fix, confirmed exactly the 3 new tests go red and nothing else moves; restored, then applied an unrelated mutation (wording in a different blocker) and confirmed the 3 new tests stayed green
- [x] `./agent-test skills/fleet/scripts/release-ticket.test.mjs` — 51/51 pass
- [x] `node --test skills/fleet/scripts/*.test.mjs` — 488/488 pass

Closes #86


feigi:
## Review actions applied — `8c83502`

**Applied (5), all mutation-verified:**

- **The guard followed symlinks, so #86's split survived this PR** (`correctness`, `silent-failure`) — `-L` added ahead of the `-e`/`-d` pair. Both shapes now block in dry run and under `--apply`; a real `git worktree add` directory does not trip it.
- **The lock probe was unpinned beyond one worktree** (`tests`, `types`) — sibling-lock regression case added, numbered **99** on purpose (see below).
- **A locked `$stray` was handed a `git worktree prune` that clears nothing** (`silent-failure`) — probe moved into a shared `locked()` helper, lock named first on both paths with `git worktree unlock`.
- **The awk `-v` assignment processed escapes in the path** (`correctness`) — a literal backslash in the repo path made the lock comparison never match, a reachable fail-open. Now via `ENVIRON`.
- **`wt_locked` was single-use state** (`simplify`) — gone; the helper move covers it. Note for anyone citing that finding: `has_branch` is read **four** times, not three.

**Deferred and filed** — all `needs-triage`, each with its `dimension` in the body:

| # | Finding | Why deferred |
|---|---|---|
| #385 | A worktree containing a submodule is the same parity break (`correctness`) | No submodules in this repo; the finding itself argues YAGNI |
| #386 | cwd inside the worktree being released partially releases it (`correctness`) | Verified unreachable — every documented caller runs from the repo root |
| #387 | A `die` after any `block` discards every accumulated blocker (`silent-failure`) | Pre-existing ordering, and two materially different fixes are on the table |
| [#334 (comment)](https://github.com/feigi/claude-config/issues/334#issuecomment-5253634825) | The lock probe reads an awk *failure* as "not locked" (`silent-failure`) | **Refuted** — measured unreachable, and the filed fix fails 31/51. Recorded on the existing measured-unreachable aggregate for this file rather than as a fifth ticket |

`ledger.mjs check` reported `tracker-hit` for #386 on five issues (#85, #86, #94, #135, #208). All five read and confirmed to be about something else — none mentions cwd or `$PWD` — so it was filed as new.

**Not filed:** the three staleness findings against "four preconditions" (`SKILL.md:752`, `reaping.md:34`, `release-ticket.test.mjs:4`), and the `locked [<reason>]` comment wording — refuted. The disputed sentence is gone anyway: the `locked()` move deleted it, including its "right after its `branch` line" clause, which a refuter separately measured as false for a detached worktree (there is no `branch` line). The helper now looks for `locked` anywhere in the record, and a stray is detached by definition.

**Fixture ordering — worth knowing before writing another of these.** `git worktree list --porcelain` orders entries **lexicographically**, so a `10-other-claim` sibling sorts *before* `9-release-ticket`. Against that order a sticky-`cur` mutant (`{if(...)cur=1}`, set but never reset) passes **57/57** — the sibling's `locked` line has gone by before anything sets the flag. The pin is vacuous. Renamed to `99-`, the same mutant fails 1/57.

**Tests:** `node --test skills/fleet/scripts/*.test.mjs` → **tests 494 / pass 494 / fail 0** (488 before). `shellcheck -s sh` unchanged: the one pre-existing SC2015 at `:220`, nothing new.

Issue #219, opened 2026-08-06
run-team/SKILL.md restates review-and-fix.md steps 2-3 in a file that says 'do not restate it' — pick one


feigi:
Deferred from PR #216 review. Dimension: `simplify` (code-simplifier).

`skills/fleet/skills/run-team/SKILL.md:351-386` now carries the apply/verify split and the commit gate verbatim from `skills/fleet/commands/review-and-fix.md:7-8`. The same file states the opposite convention three times:

- `:376-378` — CI facts are a **pointer**: "The CI facts in that file apply to you."
- `:396-398` — "`review-and-fix.md` states them; the prompt only has to say they apply."
- `:655` — "Specialist tree isolation is `review-and-fix.md`'s job — **do not restate it**."

Who reads what, traced:

| Path | reads `review-and-fix.md` | reads run-team's prompt |
|---|---|---|
| `/fleet:review-and-fix` standalone | yes (it is the command) | no |
| Fleet fix-applier | yes — `SKILL.md:345` orders it | yes |
| Fallback hand-dispatched reviewer | yes — `SKILL.md:444-455` | no |

No agent reads the prompt without also being told to read `review-and-fix.md`, so the duplication is not reachability — it is defence against a skimmed file read.

The counterargument is real and is why this is `needs-triage`, not `ready-for-agent`: commit `2c88aa2` created this duplication deliberately ("gaps between two files that must say the same thing because different agents read each"), prompt text is guaranteed in context where a file read is an instruction an agent can skip, and `review-path-default.test.mjs:246-337` pins both copies.

Decide one way and make the file consistent: either collapse `:351-368` and most of `:380-386` to the pointer form already used at `:376-378` (~28 lines off a prompt sent once per reviewed PR), or amend `:396-398` and `:655` to carve out the apply/commit rules. Keep regardless: the `<testCmd>` substitution lead-in at `:337` and the "read your refuter's report yourself" block, both fleet-only.

feigi:
> *This was generated by AI during triage.*

## Triage: rejected — the duplication is deliberate and stays

Verified against `origin/main`: all three convention lines are still there, the restated blocks are still there, and `review-path-default.test.mjs` still pins both copies. So the finding is accurate. It is being refused on substance, not on staleness.

**Why the duplication wins the trade this issue frames:**

[...]

Issue #220, opened 2026-08-06
review-pr.js comment blocks and the #211 spec restate each other and the tests — ~70 lines of triple-stated rationale


feigi:
Deferred from PR #216 review. Dimension: `simplify` (code-simplifier).

PR #216 added 78 lines to `workflows/review-pr.js`, **58 of them comment** (74%) for 20 lines of executable code. Several rationales now exist in three places — the code comment, `docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md`, and `skills/fleet/scripts/select-dimensions.test.mjs`.

Candidates, none of which change behaviour:

- `review-pr.js:73-84` (12 lines) — model-tier rationale, also at spec §115-147 and `select-dimensions.test.mjs:126-130`. Could be ~4.
- `review-pr.js:164-171` (8 lines) — 5 of them a parked "UNVERIFIED, confirm on the first run" note whose own text concludes "a wrong answer costs nothing", already in the spec twice.
- `review-pr.js:239-254` — the mixed-prose story duplicates spec §169-175 and `select-dimensions.test.mjs:107-111`. The INTERSECT-not-early-return paragraph is load-bearing and should stay.
- `review-path-default.test.mjs` — the "a loose match is vacuous" lesson written five times (`:103-113`, `:234-238`, `:273-282`, `:290-293`, `:312-315`), ~25 lines, on top of the header block at `:20-28` that already states it generally. Keep `:275-277` verbatim — that one is measured mutation evidence, not restatement.
- `select-dimensions.test.mjs:68-124` — eight near-identical `assert.deepEqual(dimensionKeys(...), ...)` cases that map 1:1 onto the spec's matrix; a table would be ~20 lines instead of ~57 and would make the correspondence visible.
- Spec `:258-280` — a 23-line per-line changelog of a diff that is now in git history; `git log --stat` is the durable form.
- Spec `:160-167` — a literal copy of `review-pr.js:199-203` + `:251-252`, untestable as a copy.

`needs-triage` because the repo's agent-facing prose is deliberately dense and the owner may want it that way; this is a judgment call, not a defect. Note the asymmetry if it is actioned: agent-facing markdown is executable instruction, so cutting a clause there can change behaviour, whereas the `review-pr.js` comments and the spec cannot.

[...]

Issue #211, opened 2026-08-02
review-pr.js + review-and-fix.md: tier specialist models, size-scale the fan-out, and replace the apply/defer/file policy behind 83% of open issues


feigi:
## Problem

Two independent cost holes in `workflows/review-pr.js`, one PR:

1. **Finders run at top tier.** Review-phase `agent()` calls pass no `model`/`effort` — every specialist inherits the session model at default effort. Verifiers already got the treatment (`verifierEffort` defaults `"low"`, `review-pr.js:125`); finders didn't. The adversarial refute pass exists precisely to kill plausible-but-wrong findings, so a cheaper finder's failure mode is already caught downstream.
2. **`selectDimensions` (`review-pr.js:146`) trims by content type only** — `docsOnly`/`hasSrc`/`hasTests` booleans. A single-file 5-line source fix pays the same six-dimension fan-out as a 400-line feature. `diff-stats.mjs` already computes file and line counts; the classifier just never reads them.

## Change

1. Review-phase specialists: `model: "sonnet"` by default, overridable via `args.specialistModel`. Effort inherited (unchanged). Verify pass untouched.
2. Size tier in `selectDimensions`: an affirmatively-reported profile (same strict guards as the existing branches) with **1 changed source file OR < 50 changed source lines** → `correctness` + `silent-failure` only. `docsOnly` branch unchanged; unknown/empty profile → full set, unchanged.
3. `log()` the trim and the model choice — no silent caps.

## Coordination

- #118 (selectDimensions untestable in place) — if it lands first, add unit tests for the new tier; if not, don't block on it.
- #113 (`args.dimensions` key normalization) touches the same override path; keep changes disjoint.

## Acceptance

- `args.specialistModel` override works; defaults documented next to `verifierEffort`.
- Unknown/unparseable profile still returns the full set.
- Thresholds named once, as constants, not repeated in prose.

Origin: fleet token-spend analysis, 2026-08-02. Review layer is the dominant spend; this cuts cost-per-specialist and specialist count for the modal small PR.

feigi:
## Scope widened — brainstormed 2026-08-06

Design: `docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md` (base `8a84402`, all line refs verified). The body above remains the origin record; it describes changes 1 and 2 only.

[...]

PR #382, opened 2026-08-11
docs(out-of-scope): record the duplication rejections from #219 and #220

feigi:
Triage closed issues #219 and #220 as rejected enhancements. Per the `.out-of-scope/` convention, the reasoning lands in the KB so the next reviewer does not re-derive it.

**New concept — `fleet-prompt-doc-duplication.md` (#219).** `run-team/SKILL.md`'s prompt restating `review-and-fix.md`'s apply/commit rules is deliberate: prompt text is guaranteed in context where a file read is skippable, `2c88aa2` created it for that reason, and both copies are pinned by `review-path-default.test.mjs`. The observed drift runs the other way — a fix landing in one file and not the other.

**Appended to `review-pr-micro-refactors.md` (#220)** rather than given its own file. That concept already refuses "deleting comments that look duplicated", which covers #220's `review-pr.js` half; the added paragraphs extend it to comment-*density* findings and record what is **not** covered — the test-side consolidations in `review-path-default.test.mjs` and `select-dimensions.test.mjs`, which remain open as their own change.

#220's other half is already fixed on `main`: the spec's `plugins/cache/.../<sha>/agents/` cite it complained about now explicitly explains why that path is deliberately not cited.

Docs only — no code, no tests touched.


feigi:
Review findings applied and deferred.

**Applied** (2 commits, HEAD `291b1cf`):

- `1d9bcc6` — **S1 + S2**, both `important`/`correctness`, both survived 2 refuters. The doc called the "read your refuter's report yourself" block fleet-only with "no counterpart in `review-and-fix.md`"; it has a near-verbatim counterpart in that file's **Specialists** rule, and `run-team/SKILL.md` self-labels it as such. Since "no counterpart" is the doc's own keep-regardless criterion, that was the one-directional drift its third argument exists to prevent. Moved to the deliberately-duplicated list; the `<testCmd>` lead-in stays fleet-only. Also added the rebuttal of the two SKILL.md lines the doc never reached — the `A caller has to restate what the callee should say itself` Trigger and the unscoped `delete the copy you superseded` rule — which were the strongest textual basis for #219.
- `291b1cf` — **U2 + U3**, both `correctness`/`suggestion`, each dispatched one refuter, both returned `refuted=false`. Dropped the hand-derived "eight" from the `select-dimensions.test.mjs` pointer (the #211 spec's item 12 explicitly refuses a count for this file, naming "nine" as the hand-derived number that motivated the policy), corrected "map 1:1 onto the spec's matrix" (part of the matrix is pinned by `assert.equal`/`assert.ok`), and excluded the `kinds`-missing case, which hand-builds its stats blob and cannot become a table row — mutating `!== 0` to `> 0` fails that case alone. Replaced "keep the copy carrying measured mutation evidence" (~15 restatements exist, 4 carry a measurement — it singled out nothing) with the three greppable evidence strings.

**Deferred**: #392 — `needs-triage`, `dimension: correctness`. The doc scare-quotes a "pointer" wording absent from `run-team/SKILL.md`. Refuted: the word is genuinely absent, but the adjacent anchors are greppable and the proposed replacement citation pointed at merge mechanics rather than the CI-facts line it claimed. Filed for the record; `wontfix` plausible.

**Not applied**: the `docs/specs/…-specialist-read-rules-design.md:437` finding, refuted 2-0, in a file this PR does not touch.

Citations throughout use greppable file + quoted string, not `file:line` — measured: zero `file:line` citations exist across all five `.out-of-scope/` docs, and line numbers into a live 1036-line `SKILL.md` rot.

Gate: `node --test skills/fleet/scripts/*.test.mjs` → **487 tests, 487 pass, 0 fail**, run before each commit. Docs-only, suite unchanged.

feigi:
Correction to the comment above, and the final state.

`1940e2d` adds the **U1 residue**, which the earlier comment listed as deferred. On the ruling, the refuted finding's smaller true part — the scare-quoted `"pointer"` quoting text absent from `run-team/SKILL.md` — is a six-word deletion, not worth a separate round while the file was already open. Line 34 now names the convention with no line number.

**#392 is closed as superseded, not filed.** Nothing is deferred from this review; **the ledger is empty**.

Final: **`1940e2d`**, three commits — `1d9bcc6` (S1+S2), `291b1cf` (U2+U3), `1940e2d` (U1 residue). Gate re-run before each: **487 tests, 487 pass, 0 fail**.

One convention worth not re-deriving, measured during this review: **`.out-of-scope/` docs cite by greppable file + symbol, never `file:line`** — `git grep -nE "\.(md|js|mjs)[`'\"]?:[0-9]+" -- .out-of-scope/` returns zero across all five. Every citation added here follows it, including the two places the review's own suggested wording carried a line number (`SKILL.md:537`, `spec:362`); both are quoted-string references instead, which stay greppable when the line drifts.

Issue #367, opened 2026-08-11
fleet scripts: arg()/has() is now five copy-pasted copies that have already drifted — extract one shared module? (triage ruled no for #169; re-ruling with post-#360 evidence)


feigi:
Deferred from PR #360 review. **Four findings, one fix vehicle** — filed as one issue rather than four duplicates. Dimensions: **silent-failure**, **types**, **simplify**, **simplify**. All four are state **unverified** (the `suggestion` band is budgeted 0 refuters by policy).

## Triage already ruled against this once — that ruling is why this is a ticket and not a patch

Triage's ruling on #169 was explicit: **"fix the four copies in place. No shared module."** PR #360 followed it. The rationale recorded in the PR body: *"these are standalone scripts invoked directly by agents and markdown instructions, and a load-path seam for four other callers isn't worth it to dedupe three lines."*

Four independent specialists in the #360 review nonetheless arrived at the extraction from four different angles, so it is filed here for a maintainer to re-rule with the post-#360 evidence — **not** to be implemented on the strength of this issue alone. If the ruling stands, close as `wontfix` and this ticket becomes the record of why.

## The four findings

**1 (silent-failure, `board.mjs:25`)** — shipping `arg()` as a fifth byte-identical copy is *why* `ledger.mjs` was left behind and why the numeric half of the guard is missing in exactly one of the five. `grep -n 'const value = process.argv\[i + 1\]' skills/fleet/scripts/*.mjs` (non-test) matches `board.mjs:31`, `pr-overlap.mjs:41`, `diff-stats.mjs:126`, `candidates.mjs:75`, `ci-state.mjs:46`. `ledger.mjs:41-47` is a sixth, divergent, unfixed variant — now filed as **#362**.

**2 (types, `board.mjs:25`)** — the copies have **already drifted after one PR**. md5 of the `arg()` body: `ci-state.mjs`, `diff-stats.mjs`, `pr-overlap.mjs` all hash to `004fd91e...` (byte-identical); `board.mjs` hashes to `1df1cbf...` — same logic, different style (arrow fn + param `n` vs `function`/`name`). And `candidates.mjs` is the only one of the five carrying the extra `Number.isInteger` guard, which the other four did not inherit when this PR copied the base pattern (see **#366**). No module owns the contract, so it is enforced only by whoever remembered to paste the current version.

[...]

PR #384, opened 2026-08-11
fix(fleet): inflight.sh probe 3 fails closed on an unreadable refs dir or worktree registry

### LINKED ISSUE #95: inflight.sh probe 3: git's degraded reads exit 0 with output missing, so a real branch still reads as "free"

feigi:
## Summary

Probe 3's two lookups (`git for-each-ref refs/heads` and `git worktree list --porcelain`) exit 0 while silently dropping what they cannot read. `chmod 000 .git/refs/heads` or the worktree admin dir makes a ticket with a live branch or worktree read as `taken=false` — the wrong "free" #76 exists to rule out, one probe down. The exit-status guard #76 added only catches a lookup that could not run at all (rc 128); it never sees this.

- `refs/heads` gets a directory-level read+execute preflight before the lookup runs. A single loose ref git skips as corrupt (rc 0, stderr warning) stays undetected — the Agent Brief puts that out of scope, so the ceiling comment is narrowed to name what's still uncovered rather than claim it's closed.
- the worktree registry matches `release-ticket.sh`'s fix for the same defect in its own worktree-registry read (#84), rather than inventing a second convention: a directory-level test alone isn't enough there (a `gitdir` file chmod'd 000 *inside* an entry passes every test on the entry itself while git still drops it), so registry entries on disk are counted against what git reported.
- a repository that never had a linked worktree has no registry directory at all, and that absence is not a permission problem — pinned explicitly against a naive "refuse whenever missing" over-fix (measured: that mutant reddens 17 tests).

Closes #95

## Test plan

- [x] 5 new cases added to `inflight.test.mjs`, each guard mutation-killed individually (broke it, confirmed only its own test reddened, restored) and the naive-over-fix control case checked too
- [x] `./agent-test skills/fleet/scripts/inflight.test.mjs` — 43/43 pass
- [x] `node --test skills/fleet/scripts/*.test.mjs` (full repo suite) — 490/490 pass


feigi:
## Review fixes applied — `1cdafb2`

**The headline: this PR's guard did not close its own ticket for the realistic case.** It tested only the top of `refs/heads`, and no fleet branch lives there — `claim-ticket.sh` builds `branch="$type/$issue-$slug"`, so the ref is `refs/heads/fix/95-…`. `chmod 000 .git/refs/heads/fix` left the top at 755, the guard passed, and the probe answered `taken=false` for a live branch. Fixed, with the subdirectory case pinned.

One correction to the review's proposed replacement, measured: `find` alone does **not** work. BSD `/usr/bin/find` — what `sh` resolves on macOS — never evaluates the expression for a starting point it cannot open, so `refs/heads` at 000 or 400 gives empty stdout, identical to healthy. The review's measurement used `bfs`, which does print it. The builtin pair is kept and `find` added below it; measured across all five permission states.

Also applied: the stray-**directory** miscount, the direction-blind mismatch message, `grep -c … || true` → one awk, and a recount before refusing (the ~10 ms window between the registry count and git's read — 1.99% spurious aborts at 2 sibling mutations/s, 0.00% with the recount).

Tests: 7 new, each mutation-verified to kill its own mutant **and only its own**. 50/50 in `inflight.test.mjs`, 499/499 repo-wide.

### Deferred, not dropped

| finding | disposition |
|---|---|
| 7 wrong cross-file `file:line` citations in comments (`inflight.sh` ×3, `inflight.test.mjs` ×4) | **#129** — one comment on the existing aggregator, with each correct target. Pre-existing, in prose this diff does not touch; correcting seven line numbers in a refs-guard PR widens the diff past its ticket. The comment also argues the case for file + symbol over `file:line`, per PR #382's measurement. |
| the same three registry defects are still live in `release-ticket.sh:96-124` | **#395** (`ready-for-agent`) — near-verbatim copy, same test blind spots. Different file, out of #95's scope. |
| the block is a duplicate of `release-ticket.sh` — extract a shared module? | **#367** — commented with the measurement (15 lines, 13 byte-identical) and today's drift event. Not acted on; #367 is the open re-ruling. |

### Refuters (4 dispatched, all in-scope `unverified` findings)

| finding | verdict |
|---|---|
| die message blames the listing for a prunable leftover | **refuted** — git still lists a prunable entry (`prunable` is an extra line, not a dropped one), so the die never fires; measured across 5 registry states |
| registry/listing race aborts spuriously | **survived** — window ~10 ms by two independent methods, 1.99% of probes at λ=2/s. Fix applied and pinned deterministically |
| line-19 `rev-parse --git-dir` guard is subsumed by line 269 | **refuted** — line 269 is unreachable for a non-repo (probe 2 dies first); deleting the guard costs the root-cause diagnostic and makes the script do two live GitHub round-trips first, 88× slower |
| the absent-registry test adds no coverage (16 tests already cover it) | **refuted** — all 16 fail *incidentally* on fixture shape; neutralising the shape under the same mutant turned all 16 green and left the new test the only red |

feigi:
**`ready-to-merge` was removed by the fleet controller, not by a human reviewer** — recording that here because the timeline shows it as `feigi` (every fleet member writes through the same credentials, so no tracker write is attributable after the fact, and `merge-bot-7` reasonably read it as the author withdrawing sign-off).

**It was not the rebase.** The rebase to `c691853` is good and is kept: byte-identical `merge-tree` proof, behind=0, CI green on the new head, worktree clean on-branch.

**The reason:** while writing up the state table for #395, the fix-applier found a defect in `1cdafb2` itself. The stray-entry guard keys its skip on the absence of a `gitdir` file, which cannot distinguish an operator's stray `mkdir` from **a registry entry whose `gitdir` was deleted**. Git drops the latter, so the guard skips it, the counts agree, no refusal fires — and `inflight.sh` answers `taken=false` for a ticket that may have a live checkout. That is the wrong-"free" class issue #95 exists to close, reintroduced one layer inside its own fix.

**Why no gate caught it:** the suite is 50/50 green with the buggy guard *and* 50/50 green with the correct one — the shipped S3 test uses an empty `mkdir`, which passes either way. CI green, local suite green, both finisher mutations discriminating correctly, and the label legitimately granted were all consistent with the defect being present.

A follow-up commit is landing now: the guard keyed on emptiness instead, plus the missing discriminator (`gitdir` deleted → rc 2), pre-verified to redden against the shipped guard and pass against the fix. The PR will be re-gated by a fresh finisher against the new head; the label granted to `1cdafb2` deliberately does not carry over.

Issue #125, opened 2026-07-31
claim-ticket.sh: the runner's guard passes arguments node then silently discards, so a mixed argv greens having run less than asked


feigi:
Deferred from PR #109 review. #98 recorded this as "related, same root cause" and #109 closes #98 without it, so it needs its own ticket or it is lost.

The runner's directory branch refuses a directory that expands to nothing. It has no answer for a path that expands fine, passes the guard, and is then **discarded by node**. Measured post-#109, Node v26.5.0:

```
$ ./agent-test t node_modules
EXIT=0   tests 3
```

`t`'s three tests ran. `node_modules` produced `node_modules/dep/x.test.mjs` — no leading component, so `*/node_modules/*` does not match it — the guard saw a non-empty expansion and appended it, and node then dropped it without a diagnostic. Two arguments in, one honoured, exit 0, silence.

Contrast the same target spelled differently:

```
$ ./agent-test ./node_modules
agent-test: no test files under ./node_modules
EXIT=1
```

Here `./` makes the path match the filter, the expansion is empty, and the guard refuses correctly. **The same directory, opposite verdicts, decided by how the caller spelled it.**

Node refuses an argv path only when its relative form begins with `node_modules/`, and only exits non-zero when it refuses *every* argument:

```
$ node --test ours.test.mjs node_modules/dep/x.test.mjs   ->  tests 1, pass 1, EXIT=0
$ node --test node_modules/dep/x.test.mjs                 ->  Could not find ..., EXIT=1
```

## Severity

No vendored code executes, so this is not the #98 defect. It is the *other* vacuous-pass shape the guard at `:157` exists to refuse: you asked for two things, got one, and a green. It also makes #109's PR-body claim ("a bare `node_modules` argument is still refused, never a false green") true only in isolation and false the moment it shares argv with anything resolvable.

## Why the approach is open

The runner cannot currently tell what node accepted. Options, none settled: refuse any argument whose expansion node will reject (requires replicating node's first-segment rule, a fourth consumer of the same shape — see #97); compare a count of passed paths against node's reported file count; or invoke node once per argument. Each trades a different thing.

Related: #97 (the emit guard, the expansion and node's own discovery already disagree on what a test file is), #123 (the filter's spelling).

Surfaced by the interface-design reviewer on PR #109; re-measured independently.

feigi:
> *This was generated by AI during triage.*

Absorbed and closed — split across the two issues that now own its halves.

[...]

Issue #230, opened 2026-08-10
agent-test: a suite whose absolute path has a node_modules ancestor is refused, while the relative spelling of the same suite runs


feigi:
Deferred from PR #224 review (`dimension: silent-failure`, surfaced while checking a finding against `skills/fleet/scripts/claim-ticket.sh:177`).

**Pre-existing, not introduced by #224** — but #224 is the first change to state the behaviour explicitly, so it should be tracked rather than left implicit.

## The defect

A package developed in place inside someone's `node_modules` — worktree and all — is legitimate code with a legitimate suite. The emitted runner refuses it when the argument is spelled absolutely and runs it when the same directory is spelled relatively.

Measured, checkout relocated to `/tmp/anc-XXXX/node_modules/@acme/plugin/wt`:

```
                                                    shipped base   PR #224 head
arg=t                                               0  pass 3      0  pass 3
arg=./t                                             0  pass 3      0  pass 3
arg=<abs>/node_modules/@acme/plugin/wt/t            1  REFUSED     0  pass 3
arg=<abs>/node_modules/@acme/plugin/wt              1  REFUSED     0  pass 6
```

Same directory, two spellings, opposite verdicts. Under the shipped base the refusal is silent about its cause — `agent-test: no test files under <abs path>` — which reads as a discovery bug rather than an exclusion.

## Why it is inherent to both mechanisms

Any exclusion that tests a path *substring* has this property:

- `-not -path '*/node_modules/*'` (base) matches every printed path when the walk root is below a `node_modules`, so the whole suite is filtered out.
- `case "/$arg/" in */node_modules/*)` (head, added by #224) matches the argument itself for the same reason.

Only PR #224's intermediate state — the prune with no argument guard — ran this case, and it did so as a side effect of the regression #224 then fixed: `-prune` fires only where the walk descends *through* a `node_modules` dirent, so a walk starting below one prunes nothing and prints everything, vendored or not. Getting this case right and getting vendored exclusion right are the same knob pulled in opposite directions.

## Why it was not fixed in #224

Out of scope for #123, and the behaviour is identical to the shipped base — measured, all four rows above. #224 changes only the message, from `no test files under <arg>` to `<arg> is under node_modules — excluded from the run, not missing`, which at least names the cause.

## Open design question (hence `needs-triage`, not `ready-for-agent`)

[...]

PR #394, opened 2026-08-11
fix(fleet): agent-test refuses an argv entry node would silently drop

### LINKED ISSUE #100: agent-test: node silently drops argv entries it cannot resolve, so a typo'd path reads as a pass

feigi:
Closes #100.

## What changed

The emitted runner's passthrough branch (`skills/fleet/scripts/claim-ticket.sh`, the `else set -- "$@" "$arg"` arm) handed non-directory argv entries straight to `node --test`. The runner counts files with `find`; node decides separately which of them it actually runs, and the two were never reconciled. Node drops an argument it cannot resolve and exits non-zero only when it refuses *every* argument — mixed with anything resolvable, the discard was silent, and the runner exited 0 having run less than it was asked.

Per the Agent Brief on #100, the passthrough branch now validates what it can before handing an argument to node, and refuses loudly rather than dropping silently:

- an argument containing shell glob metacharacters (`*`, `?`, `[`) passes through unvalidated — only node can expand it, and this is the deliberately supported quoted-glob form (`./agent-test 't/*.test.mjs'`)
- an argument whose relative form starts with `node_modules/` is refused, mirroring node's own exclusion rule (measured directly, Node v26.5.0) rather than guessing at it — this is #125's surviving case: a vendored *file* argument bypasses the directory branch (`[ -d ]`) entirely
- anything else that does not exist is refused as a typo, naming the argument

The residual ceiling is written into the code as a comment: a glob that matches nothing is still discarded by node silently, since the runner cannot expand it to check and node does not report what it ran.

## Class enumeration

Ticket named two cases (a typo'd path, a file under a vendored directory). The broader class is "an argv shape node accepts and then silently discards, that the runner's own `find`-based count never sees." Enumerating the rest before implementing, per the fleet brief's process requirement:

**Covered by this PR:**
- nonexistent path / typo — refused, existence-checked before passthrough
- file argument whose *relative* form starts with `node_modules/` — refused, matching node's own rule (measured: this is the only vendored spelling node actually drops silently)

**Deliberately left, per the Agent Brief's stated ceiling:**
- a glob that matches zero files — node expands it internally and runs nothing; the runner cannot expand glob syntax itself, and node's summary output does not report which files it ran, so there is no cheap way to detect this in advance. Recorded as a comment at the point the guard steps aside for glob arguments.

**Not members of this class — measured and ruled out, not guessed:**
- an *absolute* path under `node_modules`, or a vendored file reached through a *deeper* segment (e.g. `t/node_modules/pkg/x.test.mjs`) — measured directly against Node v26.5.0 and both **run fine**; node's own exclusion is relative-and-top-level only. Refusing these would over-refuse arguments node would have honored, which is a different (and worse) bug than the one this issue reports.
- a directory argument resolving to zero test files, an unreadable directory, or a vendored directory argument spelled any of six ways — already handled by the pre-existing directory branch (`[ -d ]`) and its own guards; untouched by this PR.
- a syntactically invalid test file, a duplicate argument, a file node's own naming convention wouldn't discover via bare discovery — none of these are *silent* drops; node either runs them, errors loudly, or (for direct file args, as opposed to directory discovery) ignores its own naming filter entirely.

**Explicitly out of scope (per the issue):**
- the vendored-directory filter's spelling — #123
- the bare, argument-less invocation — #97
- reconciling node's own execution count — not available; node does not report it

## Tests

Added to `skills/fleet/scripts/claim-ticket.test.mjs`:
- a typo'd path mixed with a valid one refuses and names the typo
- a typo'd path alone still refuses
- a file path under a vendored directory, mixed with a valid one, refuses
- a quoted glob argument (unexpanded by the shell, matching real files) still runs
- a mixed argv of files and directories still runs everything

Mutation-tested: reverting the new validation block to the old bare passthrough turns exactly these three refusal tests red (and only these); an unrelated one-line change elsewhere in the script leaves all 46 tests in the file green.

## Test plan

- [x] `sh -n skills/fleet/scripts/claim-ticket.sh` — syntax check
- [x] `node --test skills/fleet/scripts/claim-ticket.test.mjs` — 46/46 pass
- [x] `node --test skills/fleet/scripts/*.test.mjs` — 503/503 pass (full repo suite)
- [x] `./agent-test skills/fleet/scripts/claim-ticket.test.mjs` — 46/46 pass
- [x] Mutation check: reverted fix → exactly the 3 new tests fail; unrelated change → all 46 stay green
- [x] Rebased onto latest origin/main, re-ran full suite post-rebase


feigi:
## Review actions applied — `327b326` + `b134407`

Seven findings, two defects, one ordering fix. Full suite **507 pass / 0 fail / 0 cancelled** (`node --test skills/fleet/scripts/*.test.mjs`); `claim-ticket.test.mjs` 50/50 via `./agent-test`.

### Applied

| # | Finding | Resolution |
|---|---|---|
| S1/S3/S7 | existing bracketed path classified as a glob, passed through unescaped, silently dropped | existence is now settled **before** glob classification; an existing path gets the directory branch's own `sed 's/\[/[[]/g'`. `t/a.test.mjs br[a]cket/g.test.mjs` → `pass 2`, was `pass 1`/exit 0 |
| S5/U1 | vendored path holding a metacharacter bypassed the vendored refusal | vendored check no longer sits behind the glob classification |
| S2 | **regression**: every `node --test` flag refused as a missing file | `-*` passes through to node — the only thing that can judge a flag. `--test-name-pattern=ok`, `--test-only`, `--test-reporter=tap`, `--test-concurrency=1` and POSIX `--` all back to exit 0; a typo'd flag stays loud (node exits 9) |
| S4 + U2 | vendored comment claimed to mirror node's rule while matching two literal spellings | pattern widened to the rule it claimed: node excludes an argv entry when its **normalized** relative form starts with `node_modules/`, so `t/../node_modules/pkg/v.test.mjs` was excluded by node and silently dropped through the guard. Now resolved, not prefix-matched |
| R3 | ceiling comment ("the only residual is a glob that matches nothing") was false while an existing bracketed file was also dropped | true as written now that existing paths are handled; restated as "a glob matching nothing and a typo holding a metacharacter are the same string to this runner" |
| S6/U8 | vendored test pinned one spelling of two | loops four: `node_modules/…`, `./node_modules/…`, `t/../node_modules/…`, `node_modules/pkg/*.test.mjs` |

Two tests were added beyond the findings, both guarding against the fix going too far: one pins the spellings node **runs** (`t/node_modules/…`, absolute) so the guard cannot be widened into them, and one pins that the rule is anchored at the **cwd** — the same two files swap verdicts depending only on where the runner is invoked from, exactly as node does it.

### Deferred

**U3** (absolute vendored file argument runs, while the directory branch refuses that same spelling) — **refuted** as out of scope for #100: node runs it, counts it and names it, so there is no silent drop. It is a policy inconsistency between the two branches, and it is the open design question on **#230**, which `ledger.mjs check` surfaced on all three `near` rows. Measurements posted there as a comment rather than filed as a duplicate: https://github.com/feigi/claude-config/issues/230#issuecomment-5254955802

No other deferrals; **U4–U7** were affirmative `tests` findings needing no action.

### Mutation results

Each mutation printed the line it changed, so a substitution that failed to apply could not read as a survived test.

| Mutation | Red |
|---|---|
| drop the bracket escape | bracketed-path test **only** (48/49) |
| drop the `-*)` arm | flag test **only** (48/49) |
| normalizer → literal prefix | vendored-spellings test **only** (48/49) |
| widen to the directory branch's `*/node_modules/*` | "spellings node runs are not refused" **only** (48/49) |
| anchor at `git rev-parse --show-toplevel` instead of cwd | cwd-anchoring test **only** (49/50) |
| revert the whole block to `0cbfba0` | vendored + bracketed + flags (46/49) |
| benign rename of the `argdir` local | **none — 49/49 green** |

PR #404, opened 2026-08-11
fix(fleet): candidates.mjs exits 3 when filtering empties an otherwise non-empty queue

### LINKED ISSUE #64: candidates.mjs: "every candidate was filtered out" is byte-identical to "the queue is empty"

feigi:
## Summary

`candidates.mjs` reserved exit 1 for "successful query, no survivors" and exit 2 for "the query broke" — a contract a caller can read without parsing stderr. After the to-spec filter landed (#59/#60), a third outcome had no signal: the query returned rows and every one was dropped by the filter. That outcome was byte-identical to a genuinely empty queue on both stdout (`[]`) and exit code (1); only stderr told them apart, which the file's own contract says callers must not have to read.

Practical consequence: run-team's phase 0 candidate scan runs with no fallback, so a `ready-for-agent` queue holding nothing but to-spec specs reported "no work" instead of "the only labeled items are specs — run to-tickets".

**Fix:** exit 3 means the query succeeded, returned at least one row, and filtering removed every one. Tracked by capturing the raw row count immediately before each `dropSpecs()` call and comparing it to the post-filter count. The resulting flag is reassigned wholesale per query pass, never OR'd or summed across the labeled and fallback attempts — so exit 3 always describes the *final* attempt whose rows reached the filter, per the ticket's named edge case: a labeled all-specs query that falls back to a genuinely empty unfiltered query still exits 1, not 3.

Full outcome set: 0 = survivors, 1 = query succeeded, no rows at all, 2 = query broke, 3 = query succeeded, rows returned, all filtered out. Documented in the file header next to the existing 0/1/2 contract.

Exit 0/1/2 behavior and stdout shape (bare JSON array in all cases) are unchanged; verified by the full pre-existing suite staying green, including both cap-before-drop ordering tests.

Out of scope (per ticket): how the fleet controller or next-ticket *react* to exit 3 — this PR only produces the signal.

## Test plan

- [x] `./agent-test skills/fleet/scripts/candidates.test.mjs` — 53/53 pass (50 pre-existing + 3 new)
- [x] New tests pin: no-fallback all-filtered → 3; fallback pass all-filtered despite genuinely-empty labeled pass → 3; the ticket's named edge case (labeled all-specs → fallback → genuinely-empty unfiltered) → 1, not 3
- [x] Mutation-tested: reverting the exit line to `rows.length === 0 ? 1 : 0` turns exactly the two new exit-3 tests red and nothing else; the edge-case test correctly stays green under that same mutant (true negative)
- [x] Existing cap-before-drop tests (labeled and fallback) still pass unmodified

Closes #64


feigi:
Review applied at d956632839326d5144530bdba9de8a5064a309bb — `node --test skills/fleet/scripts/*.test.mjs` → **tests 541 / pass 541 / fail 0**.

**Applied (8 findings):** the `fleet-tick.mjs` exit-3 regression (`correctness`, `silent-failure`, `types` — 3 critical + 1 important, all `survived` with 2 refuters each), the discarded-stderr refusal path, and three comments that asserted things false about this PR's own code (`comments`, 3 critical + 1 important). Two new `fleet-tick.test.mjs` cases pin the regression and the stderr surfacing; each was mutation-checked to kill only its own mutant and to stay green under a real out-of-scope change.

Worth flagging: the stale mutation citation at `candidates.test.mjs:862` claimed the flip fails "this test and only this one". Re-verified against the **whole** suite it fails **three** — the third is in `fleet-tick.test.mjs`, which the review's own refuters missed by mutating against `candidates.test.mjs` alone. The corrected comment names all three.

**Deferred, filed with evidence:**

| Issue | Label | Subject |
|---|---|---|
| #407 | `ready-for-agent` | Exit 3 is documented nowhere outside the script — all three consumer-facing locations (`next-ticket/SKILL.md:28`, `docs/specs/2026-07-23-fleet-plugin-design.md:180`, `run-team/SKILL.md:60-64`) still name only exits 1 and 2. Location list independently re-enumerated rather than taken from the review. |
| #409 | `needs-triage` | How the `#64` counter/flag invariant should be expressed — three `suggestion`-band findings, none applied: one refuted by mutation, one unchecked, one proven behaviourally equivalent over 844 differential cases but whose two stated justifications (house style, "4 lines to 2") both turned out false under checking. |

One `important` `comments` finding on `candidates.mjs:19` was **refuted** by both its refuters (the header's "can say" is modal, not a claim that run-team already branches on exit 3); the controller accepted that refutation, so it is neither applied nor filed. The remaining `suggestion`-band items were confirmations that asserted no defect and needed no change.

Issue #234, opened 2026-08-10
REFUTED: the five hits-undefined assertions on the tracker failure arms are not near-duplicates


feigi:
Deferred from PR #225 review (ticket #152). **Dimension:** `pr-test-analyzer` (attribution per controller relay).

**Filed as a record, not as work. A refuter was dispatched biased to refuse and returned `refuted: true`. Labelled `wontfix` and closed — the conclusion is "change nothing", and it is recorded so the same deletion is not proposed again.**

## The finding as raised

PR #225 adds five `assert.equal(r.json.tracker.hits, undefined)` assertions. Four of them land on the same `tracker = { ok: false, query, error }` literal inside the single `catch` in `ledger.mjs`, so they were called near-duplicates pinning one line rather than four independent pieces of coverage.

## The mechanical premise is confirmed

- **Mutation A** — put `hits: []` back in the catch literal: `tests 18, pass 14, fail 4`. Exactly the four tests, each dying on **its own** new assertion (verified via AssertionError messages, not exit codes).
- **Mutation B** — put `hits: []` back in the `terms.length === 0` literal: `tests 18, pass 17, fail 1`. That arm is independently covered.

## Why the redundancy conclusion is still wrong

**The four entry paths are four different error objects**, instrumented at the catch:

| test | probe |
|---|---|
| ghFails | `code=undefined status=1 name=Error` |
| gh:false | `code=ENOENT status=null name=Error` |
| ghGarbage | `code=undefined status=undefined name=SyntaxError` |
| wrong-shape | `code=undefined status=undefined name=Error ("gh returned JSON that is not an issue list")` |

Two do not even originate in `execFileSync` — the `SyntaxError` comes from `JSON.parse(out)`, the last from the script's own shape-guard `throw`. They converge on one handler today; they are not one failure.

**Mutation C — the refactor the finding ignores.** Split the catch per error class and regress only the ENOENT branch to `hits: []`:

- with all four assertions present → `fail 1`, and only `gh missing from PATH entirely degrades the same way` catches it
- after deleting just that one assertion (what the remedy implies) → `tests 18, pass 18, fail 0`. The regression ships silently, exit code and `verdict` both unchanged.

**`verdict === "unverified"` does not rescue it either.** Under mutation A every verdict assertion stayed green — `!tracker.ok` short-circuits before `hits` is read. The `hits, undefined` assertion is the only thing catching that regression. Two claims per test, not one repeated.

[...]

PR #405, opened 2026-08-11
fix(fleet): bind ledger.mjs check's tracker query to the ledger's own repo

### LINKED ISSUE #155: ledger.mjs check queries the tracker of the cwd repo, which need not be the repo the ledger belongs to
Deferred from PR #150 review.

## The finding

The tracker query at `skills/fleet/scripts/ledger.mjs:294-298` passes no `--repo`, so `gh` resolves the repository from the **cwd**. The ledger, by contrast, is resolved either from `--file` or from `defaultLedgerPath()`'s `git rev-parse --git-common-dir`.

When those disagree, the guard searches the wrong tracker and says nothing. Measured:

```
cwd = a different repo (remote cli/cli), --file = claude-config's ledger
$ ledger.mjs --file <claude-config ledger> check "candidates.mjs row states the opposite of its code"
rc=0
ledger: not previously filed; tracker search 'candidates opposite states' found no related issues
```

`tracker.ok: true`, `hits: []`, exit 0 — a confident clean bill of health from a tracker that was never the right one. Compare the cwd-is-not-a-git-repo case, which correctly degrades to `TRACKER NOT CHECKED`.

## Reachability

Low. In the documented flow a member runs `check` with no `--file`, so ledger and tracker both derive from cwd and cannot disagree. It needs an explicit `--file` pointing outside cwd's repository.

## Why the approach is open

`defaultLedgerPath()` exists precisely so a worktree member reads the run's ledger rather than a cwd-local one; the tracker query has no equivalent binding. But the obvious fix — derive `--repo` from the same git dir the ledger came from — has no defined answer when `--file` is passed explicitly, since an arbitrary path has no repository. Options: pin `--repo` when the ledger was resolved from a git dir and warn otherwise; refuse when they diverge; or accept and document. Needs a decision.

Sibling finding in a different script: #135 (`release-ticket.sh` answering about another repo via `core.worktree`).

feigi:
## Summary

`ledger.mjs check`'s tracker query resolved "the repository" from gh's ambient cwd — the caller's process cwd, never the ledger's own location. In the documented flow (no `--file`, run from inside the repo) the two happen to agree, but an explicit `--file` naming a ledger outside the caller's repo diverged silently: the query searched the wrong tracker and reported a confident, empty result (`tracker.ok: true`, `hits: []`, exit 0) — indistinguishable from a genuinely clean tracker.

Fix: resolve the ledger's own repository (`git -C <ledgerDir> rev-parse --show-toplevel`) and run `gh` from there instead of leaving its cwd implicit — no `--repo` flag needed, gh's own remote-based resolution does the rest once pointed at the right directory. A ledger path outside any git repository has nothing to bind to, and now degrades through the **existing** unchecked-tracker state (same `WARNING — TRACKER NOT CHECKED`, same payload shape, same exit code) rather than silently querying whatever repository the caller's cwd happened to be in — no second "unchecked" representation invented.

## Enumerate the class

Every place this script resolves a repository or path and could answer about the wrong one:
- **Explicit `--file`, ledger inside a different repo than cwd** — covered (this is the fix; pinned by a new test)
- **`defaultLedgerPath()`'s git-common-dir derivation (no `--file`, documented flow)** — covered; unchanged and pinned as a "must ACCEPT" regression test (tracker still gets checked, no new warning fires)
- **Relative vs absolute `--file`** — covered; `resolve(file)` normalizes both before the repo lookup
- **`--file` inside a worktree rather than the main checkout** — covered; a worktree shares `.git` (commondir) and remote config with its main checkout, so `git rev-parse --show-toplevel` there and gh's own resolution land on the same repo either way
- **`--file` inside a repo with no `origin`** — covered by the pre-existing generic degrade path (gh itself fails to determine the repo, same as today when cwd has no `origin`); not a new failure mode, no new test needed since the mechanism is untouched
- **`--file` outside any git repository entirely** — covered; new pre-flight check, pinned by a new test
- **cwd not a repository at all, no `--file`** — pre-existing `defaultLedgerPath()` fallback, left untouched (explicitly out of scope); pinned by a new test that it still degrades exactly as before

Left deliberately out of scope (per the ticket): changing how the ledger path itself is resolved, making the repository configurable independently of the ledger, and #153/#152's tracker-query term selection and payload semantics.

## Test plan

- [x] `./agent-test skills/fleet/scripts/ledger.test.mjs` — 29/29 green (25 pre-existing + 4 new)
- [x] `./agent-test` (full suite) — 540/540 green
- [x] Mutation-tested all three new behaviors by hand (temporarily reverted, confirmed exactly the intended test(s) went red, restored):
  - removing the `cwd: ghCwd` binding → only the divergent-repo test fails
  - falling through to gh with an unbound cwd on repo-check failure → only the outside-any-repo and cwd-not-a-repo tests fail
  - inverting the repo-check condition (over-refusal) → the documented-flow "must ACCEPT" test fails, along with every test that depends on gh actually running

## Release label

`patch` — bug fix only, no interface or payload-shape change (`tracker`'s JSON contract is identical; only which repository it queries changes), fully backward compatible with the documented flow.

Closes #155

https://claude.ai/code/session_01CYmKcbwLAGbdzmMWimEDT2


feigi:
## Review applied — `99f8ab630527505cbc204d5f1cdc2fcd3accb428`

`node --test skills/fleet/scripts/*.test.mjs` → **tests 542, pass 542, fail 0** (base was the 540-pass run; +2 new tests).

### Applied (8)

The headline was a **regression this PR introduced**: the new probe ran `git -C <ledgerDir>`, but `.fleet/` is gitignored and created lazily by `save()`, so on any fresh clone or worktree it does not exist and `git -C` exits 128 — read as "not a repo". In the *documented* flow the tracker query was dropped entirely, `verdict` flipped `clean` → `unverified`, and the warning asserted the ledger path "is not inside a git repository" while it demonstrably was. Fixed by climbing to the nearest existing ancestor.

| # | dimension | what |
|---|---|---|
| 1 | silent-failure (`critical`) | the `.fleet`-missing regression above |
| 2 | correctness | same defect — `git -C` needs an existing dir |
| 3 | silent-failure | `repoCheck.stderr`/`.error` never read: four distinct causes collapsed into one fabricated sentence. Now carries git's own reason, capped, like the sibling gh catch (#176) |
| 4 | silent-failure | cwd binding does not survive an inherited `GIT_DIR`/`GIT_WORK_TREE`; the "can never disagree" claim was false. Both scrubbed, claim rewritten |
| 5 | tests | deleting the guard was an equivalent mutant (29/29 passed). Now pinned |
| 6 | comments | the `status !== 0` branch comment named a cause it does not fire for |
| 7 | simplify | same root cause as 1 |
| — | comments | `always` → `by default` on the `#155` section header (one-word tightening) |

Two in-scope `suggestion`s survived their refuters and were applied:

- **`GH_REPO` defeats the cwd binding outright** — gh reads it *before* consulting git. Measured against a live tracker with the binding in place: `GH_REPO` pointing at another real repo returned `{"ok":true,"hits":[],"verdict":"clean"}` — #155 verbatim, in the documented flow, no `--file` divergence needed. This repo's own `.github/workflows/rebase-check-refresh.yml:41` exports it job-wide.
- **`gitRepo: false` on the last new test** — applied on the minimal form only; the finding's own stated end state dropped `noFile: true` and measured RED.

Also fixed the harness: under an ambient `GIT_DIR` the fixture's `git init` exits 0 and creates **nothing** in the target, so the fixture was silently not a repository and the status assert could not see it. Suite was 26 pass / 3 fail under an ambient `GIT_DIR`; now 31/31.

### Mutation verification

The named mutant — delete the guard branch so the `else` always runs — **survived 29/29 on `dce8837`** and is now **killed** (30/31, exactly the strengthened test). Five more, each killing exactly one test: revert the ancestor climb; drop the env scrub from the probe; from the gh call; the `GH_REPO` scrub. Two changes that should *not* be caught (a comment reword, a local variable rename) both stayed **31/31**.

### Deferred (5) — all refuted `suggestion`s, filed as records

#410 · #411 · #412 · #413 — `wontfix`, closed, each carrying its refutation and dimension. Plus a comment on #234, which already owns the mirrored-`hits`-assertion question; the PR-#405 proposal was its inverse and was refuted on #234's own established grounds.

One `simplify` suggestion (restrict the probe to the `--file` arm) overlapped the applied fix and was folded in rather than filed; its residual is a ~5 ms subprocess saving that would require reshaping `defaultLedgerPath()`'s return, not applied.

Issue #93, opened 2026-07-31
run-team/SKILL.md: the in-flight step says "any hit = taken" and never names inflight.sh's exit 2, which PR #76 makes far more reachable


feigi:
## Problem

`skills/fleet/skills/run-team/SKILL.md:54` (line number at committed `9de019bc`; the file currently has uncommitted working-tree changes that shift it) describes the in-flight check as a two-state decision:

```
3. **In-flight check** — `~/.claude/skills/fleet/scripts/inflight.sh <N>` per
   candidate; any hit = taken. It runs all three probes (open/closed PRs, remote
   heads, local worktrees + branches) so a partial one cannot read as free.
```

"any hit = taken" never names **exit 2**, so a controller following this file literally has no instruction for "the question could not be answered". In `sh`, both `cmd || handle` and `if cmd; then` treat 1 and 2 alike, so the natural reading silently files an unanswerable probe under "taken".

`skills/fleet/skills/next-ticket/SKILL.md:36` documents the same script correctly:

```
~/.claude/skills/fleet/scripts/inflight.sh <N>   # 0 free, 1 taken, 2 unanswerable
```

## Why now

Pre-existing, but PR #76 makes exit 2 **materially more reachable**: probe 2 previously reported a failed `git ls-remote` as "no remote branch" (exit 0) and now correctly exits 2. Conditions that reach it include no `origin` configured, an unavailable SSH key or agent, a suppressed host-key prompt, and a stale remote URL after a repository rename — none of which require GitHub to be down.

Treating that as "taken" is the *safe* direction (one skipped ticket, versus two agents on one ticket), so this is a documentation gap rather than a live wrong-free. But it is undocumented behaviour that the reader has to infer.

## Suggested fix

Name the three states, matching `next-ticket/SKILL.md:36`'s wording. Something like: `exit 1 = taken, exit 2 = unanswerable — treat as taken and say so in the log, never as free.`

## Sequencing

**Do not edit this file yet.** PR #71 and PR #73 both own `run-team/SKILL.md` in the current wave, and it has uncommitted changes in the working tree. This should land after both.

Related: #45 covers the same class of error in `docs/specs/` (two design docs describing `inflight.sh`'s payload and exit contract wrongly) but does not cover this file — different document class, and this one is executed by agents every wave rather than kept as a historical record.

Deferred from PR #76 review.

feigi:
[...]

PR #406, opened 2026-08-11
fix(fleet): inflight.sh accumulates probe results instead of dying on first failure

### LINKED ISSUE #96: inflight.sh: a failed probe dies immediately, discarding a sufficient hit another probe already found

feigi:
## Summary

`inflight.sh` runs three probes that form a monotone disjunction — any single
hit is sufficient to answer "taken" — but each probe's failure path called
`die` immediately, so a probe that could not look discarded a hit an earlier
probe already found (or prevented a later probe from ever finding one). An
open linked PR was announced on stderr and the run still exited 2 with no
payload at all.

Each probe (`pr`, `remote`, `local`) is now wrapped in its own function. A
step inside it that fails now calls `add_unknown "<probe>" "<message>"` and
`return 1` instead of `die` — the exact same diagnostic text as before, just
no longer fatal — and the run continues to the next probe. All three probes
always run to completion. The verdict:

- any hit → taken, exit 1 (checked first, unconditionally — a hit outranks
  an unknown regardless of what else failed)
- else any unknown → unanswerable, exit 2
- else → free, exit 0

Exit 2 from a probe failure now carries the same payload shape as 0 and 1,
plus a new `"unknown":[...]` field naming which probes could not look — so
exit 2 stops meaning "no stdout at all" for that case.

Also updated `docs/specs/2026-07-23-fleet-plugin-design.md`'s script table
row for `inflight.sh` to reflect the new payload field and the split between
"probe failed" (payload) and "nothing established yet" (no payload).

## Enumerate the class

**Every `die` call site in `inflight.sh` (24 total), and where it landed:**

Converted to accumulate (probe records itself unknown, run continues) — 18
sites:
- Probe 1 (`pr`, 4 sites): `gh issue view` network/auth failure, `gh pr
  list` failure, the PR-filter python3 call, the PR-count python3 call
- Probe 2 (`remote`, 2 sites): `git ls-remote` failure, the remote-branch
  awk filter
- Probe 3 (`local`, 12 sites): git-common-dir resolution, refs-dir
  read/exec check, the `find` sweep of refs subdirectories, a bad refs
  subdirectory, `git for-each-ref`, the local-branch awk filter, the
  worktree-registry read (inside `count_registry`), `git worktree list`,
  the worktree-listed count, both directions of the listed-vs-registered
  mismatch, the worktree-path awk filter

Left as a hard, payload-less `die` — 6 sites:
- 3 true preconditions, unchanged: arg count, arg not numeric, not inside a
  git repository — nothing has been established yet
- 1 special case, unchanged: "issue does not exist" — not a probe that
  could not look, it's the premise every probe depends on being false, so it
  stays terminal like the preconditions above it (distinct from the
  adjacent "gh issue view network failure" branch two lines away, which
  *is* now accumulated)
- 2 payload-construction-time failures, unchanged: the JSON-escape step
  (`jstr`/`jrewritten`) and the final `printf` write — these run after all
  three probes finish, so a failure here means the payload itself cannot be
  safely emitted, not that a probe couldn't look

**Probe-pair orderings** (hit in probe X, failure in probe Y, X≠Y — 6 total
combinations across the 3 probes): the two the brief measures — (pr hits,
remote fails) and (local hits, remote fails) — are pinned as new tests,
plus all-three-unknown and genuinely-free. The other four orderings ((pr,
local), (remote, pr), (remote, local), (local, pr) as hit/fail pairs) are
deliberately left unpinned: the fix is probe-agnostic by construction (the
same `probe_X() { ... } ; probe_X || :` shape for all three, no branching
on which probe is which), so a fault that broke one ordering and not its
symmetric twin would have to live inside one probe's own body — a fault
class each probe's own extensive pre-existing failure tests already cover.

**What this could wrongly REFUSE** (false-positive class, since a new
accumulate path could start emitting a payload somewhere it never had one):
pinned that argument/repository-precondition failures still exit 2 with
*no* payload (two new tests), and that a genuinely free ticket still exits
0 with an empty `unknown` list (one new test).

## For #93 (not touched here — staged separately per the brief)

`run-team/SKILL.md` doesn't currently name exit 2 at all. Once this lands,
it needs to say exit 2 now carries a payload naming which probe(s) could
not answer (`unknown: [...]`), distinct from the two cases that still exit
2 with no payload (a bad argument, or running outside a git repository).

## Test plan

- [x] `./agent-test skills/fleet/scripts/inflight.test.mjs` — 57/57 passing
- [x] 6 new tests added: both hit-survives-failure directions, all-three-unknown,
      genuinely-free-with-empty-unknown, bad-argument-still-no-payload,
      not-a-git-repo-still-no-payload
- [x] 13 pre-existing tests updated from "exit 2 → no payload" to "exit 2 →
      payload with the right `unknown` entries" (unchanged stderr/exit-code
      assertions kept intact)
- [x] Mutation-tested: reverting probe 2's accumulate back to `die` kills
      exactly the 3 new tests that depend on probe 2 surviving (both
      hit-survives directions + all-three-unknown), while genuinely-free and
      both precondition tests stay green; removing the `unknown`-elif from
      the verdict kills only the all-three-unknown test; disabling the
      not-a-git-repo precondition kills only that one test. Each mutant
      restored after confirming red, full suite re-verified green after.

Closes #96


feigi:
## Review applied — `cc60e73`

`node --test skills/fleet/scripts/*.test.mjs` from the worktree: **`tests 545 / pass 545 / fail 0`** (was 542 at `865129e`).

### Applied (7 findings, 3 code changes)

| Finding | What |
|---|---|
| `survived[2]` `[3]` `[4]` | **The headline: accumulation was per-PROBE, not per-SIGNAL.** A hit already computed inside a probe was still discarded when a later stage of the *same* probe failed. Each signal now commits where it is established. |
| `survived[1]` `[5]` | Header comment named one payload-less exit-2 cause where the spec row lists four, and justified it with "nothing has been established yet" — false for two of them. |
| `refuted[5]` (reversed by the controller) | The accumulate-tests preamble claimed a fault inside a probe body was covered by existing tests. Mutation refuted it — six named internal branches survived the whole suite. Replaced with what is actually pinned. |
| `unverified[6]` | The nonexistent-issue test never checked stdout. Added the pin. |
| `unverified[9]` | Four comments still explained a `|| die` at sites #96 converted to `add_unknown … return 1`. |

Measured before/after on the headline case — local branch `fix/77-slug`, `.git/worktrees` at mode 000:

```
before: exit 2  {"taken":false,"hits":[],"unknown":["local"],   …"localBranch":"fix/77-slug"…}
after:  exit 1  {"taken":true, "hits":["local"],"unknown":["remote","local"],…}
```

### Mutation verification — whole 545-test suite, one test red per mutant

| Mutant | Result |
|---|---|
| revert the `probe_pr` half | kills **only** `accumulate: a PR hit outlives a later failure in its OWN probe` |
| revert the `probe_local` half | kills **only** `accumulate: a local-branch hit outlives a later failure in its OWN probe — the worktree half` |
| silent `return 0` at `could not filter PR search results` — one of `refuted[5]`'s six, previously surviving | now killed by `probe 1: a PR filter that could not run is unknown, never free` |
| nonexistent-issue `die` → `add_unknown; return 1` — previously surviving | now killed by `a nonexistent issue dies 2, distinctly from a network failure` |

Neither new accumulate test catches the other's mutant, so they discriminate rather than merely both going red. Script restored byte-identical after every leg (`cp` backups, never `git checkout -- path`, which writes the index and makes a revert read clean).

### Two corrections found *in* the corrections

- The `unverified[6]` fix as suggested (`assert.equal(r.stdout.trim(), …)`) **throws** — this suite's `inflight()` helper returns no `stdout`. Used `assert.equal(r.json, null, …)`.
- My own first draft of the `probe_pr` comment claimed an unguarded count would leave `$raw` unset and trip `set -u`. Measured: it is set to the **empty string** and nothing aborts — the tally just prints a hole. Comment corrected before commit.

### Deferred and filed

| Issue | Findings | Why deferred |
|---|---|---|
| #415 | `refuted[6]` + `unverified[1]` + `unverified[3]` | Same latent `set -e` exemption raised three times. Real mechanism, no live fail-open (full guard inventory in the body). Three materially different remedies, none settled. |
| #416 | `unverified[7]` | `add_unknown`'s probe-name argument duplicated at 17 sites. Out of #96's scope; collides with #415. |
| #417 | `unverified[8]` | Test-assertion helper collapse. Out of scope, and `cc60e73` changed how many sites actually match. |
| #418 | `unverified[10]` `[11]` `[12]` | Three one-to-seven-line simplify nits, grouped — independent, judge each separately. |
| #419 | `unverified[2]` + `unverified[4]` | `hits[]` `remote-branch` vs `unknown[]` `remote`. **Checked and refuted** by a dispatched refuter: the `hits` token predates the PR, the added token already matches `evidence.remote`, and the two are unreachable in one payload. Recorded so it is not re-derived a third time. |

`refuted[1]` + `refuted[2]` (`taken:false` on exit 2) — refutations accepted, not applied. Their real residual is that `run-team/SKILL.md` carries no exit-2 rule at all, which is #93's territory; recorded there as [a comment](https://github.com/feigi/claude-config/issues/93#issuecomment-5257315097) rather than edited from this PR. `refuted[3]` and `refuted[4]` accepted, no action. `unverified[5]` was a confirmation, not a defect.

`rebase-check` red is staleness from #404 merging (`behind=3`), not a failure — not rebased.

feigi:
**Correction to my previous comment:** I wrote `behind=3` and attributed the staleness to #404. Measured: `behind_by=6` — #404 and #405 both merged after this branch's base (`cbecf7e`), three commits each. The conclusion is unchanged and did not depend on the number: `rebase-check` red at any depth is staleness, not a failure, and this branch was not rebased. Flagging it because a wrong figure in a status comment is still a wrong figure.

Issue #191, opened 2026-07-31
no-undo-audit.test.mjs: dead fixture push, an unused parameter default, and an assertion message that overclaims


feigi:
Deferred from PR #180 review — test hygiene only, no behaviour change. Three one-line items, measured.

### 1. Dead push in `nestedWorktree`

`git(c.w, "push", "-q", "origin", c.branch);` pushes the `.gitignore` commit to `origin/fix/1-thing`, a ref none of the tests using this fixture consults — the audit runs against `origin/main` and `origin/fix/9-nested`. Deleting the line leaves the suite green.

**Do not delete the three lines above it.** The `.gitignore` write + add + commit are load-bearing: without them the enclosing repo is not clean, so the leak never manufactures `clean` and the refusal fixtures pin nothing. Measured — removing them turns the refusal tests red for the wrong reason.

### 2. Unused parameter default

`function nestedWorktree(t, branch = "fix/9-nested")` — `branch` is never overridden at any call site. Inline the constant and drop the parameter. The file's own precedent argues for this: `repo(t, branch, prefix)` carries defaults because a caller really does override them.

### 3. Assertion message overclaims

```js
assert.ok(existsSync(join(c.w, ".git")), "fixture must leave the linkage intact");
```

`existsSync` proves presence, not intactness, and not that `.git` is a **file** — which is what the test title asserts. The following `git status` assertion does the real work. Cheap fix: `assert.ok(statSync(join(c.w, ".git")).isFile(), ...)`.

All three in `skills/fleet/scripts/no-undo-audit.test.mjs`. Approach settled, hence `ready-for-agent`.

Deferred from PR #180 review.

feigi:
Five more items in the same file, deferred from the **PR #408** review (closes #189). Filed here rather than as a new issue because they are the same kind of one-line, measured test hygiene this issue already collects — item 1 below is literally the sibling of this issue's own item 1, one fixture over.

Line numbers are as of PR #408's fix commit, which added tests above some of them.

### 1. `nestedWorktreePair`'s sibling push is dead too — `:172` · dimension `simplify`

[...]

PR #408, opened 2026-08-11
fix(fleet): no-undo-audit.sh verifies whose admin dir answered, not just whose root

### LINKED ISSUE #189: no-undo-audit.sh: a .git file naming another worktree's admin dir passes the linkage guard, and clean is computed against a foreign HEAD
Deferred from PR #180 review. **Verified by construction; approach still open, hence `needs-triage`.**

PR #180 replaced the linkage guard with `git -C "$wt" rev-parse --show-prefix` + `[ -z "$prefix" ]`, which establishes *"the worktree root git resolved is `$wt`"*. That closes the whole walk-up class. It does **not** establish *"the gitdir git resolved belongs to `$wt`"*, and those are different claims.

### Repro

`$wt` is a linked worktree whose `.git` **file** is rewritten to name a *sibling* worktree's admin dir:

```sh
printf 'gitdir: %s/.git/worktrees/8-y\n' "$parent" > "$wt/.git"
```

For a `.git` **file** git takes the worktree root from the file's own location, so:

```
git -C $wt rev-parse --show-toplevel -> <parent>/.worktrees/9-x     (i.e. $wt — correct)
git -C $wt rev-parse --show-prefix   -> []                          (empty — guard ADMITS)
```

The guard passes, correctly. But `status --porcelain` now compares `$wt`'s files against the **sibling's** HEAD and index. Measured leak (from the PR #180 silent-failure review): `$wt` holds an uncommitted `f.txt` whose content the sibling's branch happens to have committed, so:

```
git -C $wt status --porcelain -> []
no-undo-audit.sh $wt <branch> -> exit 0, {"clean":true,...}
$wt/f.txt on disk             -> PRECIOUS UNCOMMITTED WORK   (exists nowhere else)
```

`"clean":true` for a worktree holding uncommitted work — the same outcome class as #126, reached through gitdir identity rather than worktree-root identity.

### Why the approach is open

- Needs the admin dir's `gitdir` back-pointer checked against `$wt/.git`, i.e. a **second** identity claim on top of the one PR #180 added. Whether that belongs in this script or in a shared helper is the open question.
- `release-ticket.sh:229-233` already documents this exact residual against its own guard and explicitly defers it *"to its own ticket"* — this is that ticket, and it should close for both scripts at once.
- `core.worktree` redirection is a related variant, but PR #180's `--show-prefix` **already refuses** that one (measured), so it is out of scope here.

### Reachability

Requires a hand-edited or copied `.git` file **and** a content collision between the uncommitted work and the foreign HEAD. Contrived — which is why this is deferred rather than folded into #180 — but the script gates an irreversible action, so "contrived" is not "harmless".

`skills/fleet/scripts/no-undo-audit.sh` (linkage guard), `skills/fleet/scripts/release-ticket.sh:229-233`.

Deferred from PR #180 review.

feigi:
Closes #189.

## The gap

`--show-prefix` (PR #180) establishes one identity claim: *"the worktree root git resolved is `$wt`"*. For a `.git` **file**, that root comes from the file's own location — so a `.git` rewritten to name a **sibling** worktree's admin dir still yields an empty prefix (the guard admits) while every git command below, `status` included, answers against the sibling's HEAD and index. When the uncommitted content happens to match something already committed on the sibling's branch, `status --porcelain` reads empty and the audit reports `{"clean":true}` for a worktree holding the only copy of that work.

## The fix

A second identity claim, `no-undo-audit.sh` only, beside `--show-prefix` (`skills/fleet/scripts/no-undo-audit.sh:150-183`):

- `git -C "$wt" rev-parse --git-dir` names which admin dir answered. The literal string `.git` is the main worktree — its `.git` *is* the repo, nothing to cross-check, the root claim already covers it.
- Any other value is a linked worktree's admin dir. Every such dir carries its own `gitdir` back-pointer file (written once by `git worktree add`), pointing at the worktree `.git` file it belongs to. Resolved, it must name `$wt`'s own `.git` — anything else refuses (`die`, exit 2, never a warning beside a clean payload).
- Canonicalized via `cd "$wt" && pwd -P` on `$wt`'s side (not `--show-toplevel`, which the existing comment above the guard already rules out as a string-compare target) — matches the resolved form git itself writes into the back-pointer at `worktree add` time, so a legitimate worktree reached through a symlinked path still compares equal.

## Enumerate the class

**Covered (new tests in `no-undo-audit.test.mjs`):**
- `.git` file naming a sibling's admin dir by absolute path — the ticket's own repro, with a real content collision (sibling commits the exact bytes `$wt` holds uncommitted) — refused
- Same spoof via a **relative** `gitdir:` path — refused (git resolves both spellings to the same absolute admin dir; the fix isn't defeated by the spelling)
- Admin dir's own `gitdir` back-pointer file **missing** — refused
- Admin dir's own `gitdir` back-pointer file **unreadable** (chmod 000) — refused
- A legitimate back-pointer file with **trailing whitespace** — still passes (the trim exists so a hand-edited or copy-touched back-pointer doesn't cause a false refusal; pinned both directions)
- A dangling `.git` **symlink** — refused, but via the *existing* `--show-prefix` walk-up guard, not new code (measured: nested inside a real repo, a dangling symlink walks up exactly like a deleted `.git`)
- Legit linked worktree reached through a **symlinked path** — passes (previously-uncovered gap, closed)
- Legit linked worktree, direct — passes (pre-existing coverage)
- Main (non-linked) worktree, `.git` is a directory — passes (pre-existing coverage, dozens of tests)
- Empty `.git` directory — refused, pre-existing coverage, unaffected (dies at the prior `--show-prefix` check before the new code runs)

**Deliberately left out, with reason:**
- `core.worktree` redirection — out of scope per the brief; `--show-prefix` already refuses it (measured, PR #180)
- The `#84` unreadable-`gitdir` listing drop at `release-ticket.sh:86` — a different script, different mechanism, out of scope per the brief
- A shared shell helper for both scripts — out of scope per the brief ("two scripts and a few lines; keep it in place")
- Trailing whitespace on `$wt`'s **own** `.git` file (the `gitdir:` line, not the admin dir's back-pointer) — not given a dedicated test. This fix never reads that file itself; it asks git for `--git-dir`, and git's own gitfile parser owns trimming that line. A test here would pin git's parser, not this script's logic.

Every new test was mutation-tested: guard removed → exactly the targeted tests go red (and only those); guard restored → green. The whitespace-trim line was separately mutated (trim removed) → exactly its own test goes red.

## Does `release-ticket.sh` need the same claim?

**No — measured, not inferred from the stale cross-reference.**

`release-ticket.sh`'s own linkage check (`:463-503`) is structurally different and weaker: `[ -f "$wt/.git" ]` plus `git -C "$wt" status --porcelain`, with no `--show-prefix` root claim at all. Its own comment (`:473-477`) already documents that a `.git` naming a wrong gitdir (there, specifically `core.worktree`) passes this check and answers about the wrong tree at rc 0 — the same class of gap, and the same as this ticket's spoof would produce if aimed at `release-ticket.sh`'s block-check in isolation.

But the block-check isn't what performs the irreversible action. The actual delete is `git worktree remove "$wt"` (`:551`), invoked as a **bare path argument** — not `-C "$wt"` — from the main repo. That form looks up `$wt` in the **main repo's own** worktree registry (unaffected by whatever `$wt/.git` has been rewritten to say) and independently validates the linkage before removing anything. Measured just now, git 2.50.1:

```
$ git worktree remove ../link
fatal: validation failed, cannot remove working tree: '.../link' does not point back to '.git/worktrees/link'
```

rc 128 — and `release-ticket.sh:552` already halts on that refusal (`halt "git worktree remove refused $wt: ..."`), same as the `core.worktree` variant its own comment already defers to `worktree remove`. So the same spoof aimed at `release-ticket.sh` produces a wrong "clean" report from the pre-check, but not data loss: git's own remove validation is the actual backstop, and it catches this spoof the same way it already catches `core.worktree`. `no-undo-audit.sh` has no equivalent downstream backstop — its `clean:true` is the final answer a caller (a rebase) consumes directly — which is why the claim belongs there and not in `release-ticket.sh`.

## Tests

`node --test skills/fleet/scripts/no-undo-audit.test.mjs` — 43/43 green (was 42; +1 net after two iterations of fixture fixes, see below).

## Release label

`patch` — bug fix to an existing guard's behavior on a previously-mishandled dangerous input; no new interface, no behavior change for any legitimate worktree shape (all pinned passing above).


feigi:
Review deferrals filed:

- **#420** `ready-for-agent` — `run-merge-bot.md:141` + `docs/specs/2026-07-23-fleet-plugin-design.md:184` carry the same false exit-2 parenthetical (*"git walks up and reports the enclosing repo"*). Dimension `comments`. The **enumeration** is fine — both new die paths fall inside existing clauses; the **mechanism clause** is what is false, and both files must be tightened together. Neither is in this PR's diff.
- **#421** `needs-triage` — `release-ticket.sh:60` / `claim-ticket.sh:29` still carry the bare `rev-parse --git-dir` guard, weaker than this script's pre-#189 one. Dimension `types`. Triage: #189's own body asked for both scripts while triage scoped `release-ticket.sh` out.
- **#191** (comment) — five test-hygiene items in `no-undo-audit.test.mjs`, dimensions `simplify` ×4 and `tests` ×1, added to the existing issue for this file rather than duplicated.

PR #422, opened 2026-08-11
fix(fleet): judge agent-test's vendored-directory guard by resolved path

### LINKED ISSUE #125: claim-ticket.sh: the runner's guard passes arguments node then silently discards, so a mixed argv greens having run less than asked

### LINKED ISSUE #186: claim-ticket.sh: the emitted runner runs vendored tests when the argument is a symlink into node_modules

feigi:
## Summary
- agent-test's directory-argument branch judged vendored-ness by matching `$arg`'s own spelling against `*/node_modules/*` — a symlink whose target lies inside a vendored tree carries no `node_modules` in its own name, so the check missed it, and `-prune` only fires on a dirent named `node_modules` met DURING the walk, which the symlink's target has already passed. The result: `agent-test <symlink-into-node_modules>` exited 0 reporting a pass over third-party code.
- Fix: judge the argument by its RESOLVED directory (`cd "$arg" && pwd -P`, following the symlink and any inside its path) instead of its spelling, before the vendored check runs. Closes this route and the pre-existing bare-relative-path route (#125) with the same mechanism, per triage's ruling on #186.
- A symlink to a directory that merely *contains* a vendored tree (not one whose own resolution ends inside it) resolves outside `node_modules` and is untouched — `-prune` still excludes its real vendored contents once the walk reaches them.

## Enumerate the class
Ticket named three routes into this blind spot:
- **Covered**: a symlink into a vendored tree (directory argument) — this PR's fix and its RED test.
- **Covered** (already, pre-fix, unaffected by this change — verified by grep there's only one copy of this find+prune walk in the repo): a vendored subdirectory named relatively (`node_modules/pkg`, `t/node_modules/pkg`, etc.) — pinned by the existing `runner: a vendored directory argument refuses however it is spelled` test.
- **Covered** (already, pre-fix): the direct-argument form from the closed #125 — same test as above.
- **Deliberately left**: a symlink to a *file* (not directory) whose target resolves inside a vendored tree, passed directly as an argument (falls into the `else`/file branch, not `[ -d ]`). The ticket's own "Current behavior"/"Desired behavior" narrative and its three named routes are all about the directory branch's find-based filter; the ticket's Out of scope list points at "Arguments node discards silently — #100", which owns that branch's existing validation machinery. Reachability is the same as the fixed case (no tracked symlinks in this repo, deliberate-only), and it's a real sibling of the same class — worth a follow-up ticket if it matters, not folded into this one.

## Accept-case guard
New test `runner: a symlink to a directory containing a vendored tree still runs its own tests` — a symlink to a directory that merely contains (not resolves into) a vendored subtree must still run the directory's own tests and exclude only the real vendored ones. Already green pre-fix (find's own prune handles it once inside); pinned so the resolved-path guard can't regress it by over-refusing every symlinked directory with `node_modules` anywhere underneath.

## Test plan
- [x] `./agent-test skills/fleet/scripts/claim-ticket.test.mjs` — 52/52 pass, including the two new tests
- [x] Watched the fix's own test RED before implementing (exit 0, vendored test counted as a pass) and GREEN after

Closes #186


feigi:
Review deferrals filed: #428 (dimension `silent-failure` — refusal message names the caller's spelling; refuted, and its suggested fix measured wrong twice) and #429 (dimension `types` — guard checks `$resolved` while `find` walks `$arg`; out of #186's scope).

`unverified-1` was already filed as #424 — confirmed it covers the symlinked-*file* case, not re-filed.

Refuted with no defect to track, recorded rather than filed: the projlink test is not redundant (a symlink-scoped widening kills it alone), its throwing fixture is not redundant (one mutation leaves it RED while `PASSES` stays GREEN), and the new comment block's `-prune` mention is the premise of a conclusion the earlier two never draw — all five of its factual assertions were measured true against the emitted runner.

Issue #128, opened 2026-07-31
reap.sh:53 and worktree-audit.sh:36 read a worktree with a deleted .git file as CLEAN at rc 0 — the same parent-repo leak as #74, in two more callers


feigi:
Same root cause as #74 (fixed for `release-ticket.sh` by PR #115) and #126 (`no-undo-audit.sh`), in two more callers. Neither establishes that `$wt` is the worktree it is about to describe, so when the worktree's `.git` file has been deleted outright — directory and every uncommitted file still on disk — `git -C "$wt" status --porcelain` walks UP to the enclosing repo and reports the **parent's** status at rc 0. `.worktrees/` is gitignored, so with a clean parent that answer is **empty**.

Measured, git 2.50.1, clean parent, `.worktrees/` gitignored:

**`reap.sh:53`**
```sh
if [ -n "$(git -C "$wt" status --porcelain 2>/dev/null || echo dirty)" ]; then
  keep "$b" "dirty worktree $wt"
```
```
$ v=$(git -C .worktrees/9-x status --porcelain 2>/dev/null || echo dirty)
$ printf '[%s]\n' "$v"
[]                       -> -n is false -> reads CLEAN, falls through to the remove
```
The `|| echo dirty` fallback shows the intended fail-safe direction, but it only fires on a **non-zero** rc. This path exits **0**, so the fallback is bypassed by the one input it was written for.

Blast radius is contained, but only by luck downstream: the worktree is under `.worktrees/`, so the ignored-file `keep` is skipped by the `*/.worktrees/*` case, and the run reaches `git worktree remove "$wt"` — which independently refuses (`fatal: validation failed … '.git' does not exist`), so `keep "$b" "worktree remove refused"` fires and the branch survives. That is a second, unrelated guard re-validating what the first was supposed to establish — exactly the situation PR #115 describes for `release-ticket.sh`. The dirty check still manufactured a "clean" it never measured, and the reported reason is the wrong one.

**`worktree-audit.sh:36`**
```sh
if ahead=$(git -C "$wt" rev-list --count "$base"..HEAD 2>/dev/null) \
   && status_out=$(git -C "$wt" status --porcelain 2>/dev/null); then
```
```
$ git -C .worktrees/9-x status --porcelain; echo "rc=$?"
rc=0                     -> status_out empty -> audit REPORTS the worktree as clean
```
[...]

Issue #421, opened 2026-08-11
release-ticket.sh/claim-ticket.sh: the git-dir identity invariant #189 put in code is only prose here — and #189's own body asked for both scripts


feigi:
Deferred from PR #408 review (`unverified[4]`). **Dimension: `types`** — invariants expressed in code vs merely documented. `needs-triage` because the scope question below is open, not because the gap is unverified.

### The gap

PR #408 (closing #189) put the identity invariant *"the git dir that answered must belong to `$wt`, not merely resolve a root at `$wt`"* into **code** in `skills/fleet/scripts/no-undo-audit.sh`. `skills/fleet/scripts/release-ticket.sh` names the identical gap in **prose** and defers it:

> a `.git` naming a gitdir whose `core.worktree` is some other directory passes this and still answers about that other tree … left to its own ticket rather than widened into here

— while its own guard is a bare `git rev-parse --git-dir >/dev/null 2>&1 || die "not inside a git repository"` (`release-ticket.sh:60`), with no `--show-prefix` and no back-pointer check. `claim-ticket.sh:29` carries the same bare form. Both are weaker than even `no-undo-audit.sh`'s **pre-#189** guard.

Search scope for that claim: grepped `gitdir|git-dir|show-prefix|show-toplevel` across `release-ticket.sh`, `claim-ticket.sh`, `inflight.sh` — `release-ticket.sh:60` and `claim-ticket.sh:29` are the only guards, and neither checks either identity claim.

### Why this is triage, not ready-for-agent

**The scope is contested in the record, and needs a human to settle it:**

- Ticket #189's own body says: *"`release-ticket.sh:229-233` already documents this exact residual against its own guard and explicitly defers it 'to its own ticket' — this is that ticket, and it should close for both scripts at once."*
- Triage nonetheless scoped `release-ticket.sh` **out** of #189 on the grounds that the cross-reference had gone stale, and PR #408 was implemented against that narrower brief.

So #189 shipped for one script while its own text asked for two. Someone has to decide whether that was right.

Also open, whichever way that lands: whether the now-hardened check becomes a **shared helper** or is re-derived per script. `no-undo-audit.sh`'s version is not a drop-in — it leans on `--show-prefix` having already established the root claim, and it admits git dirs that record no worktree (submodule, `--separate-git-dir` target) rather than refusing them. A caller with a different threat model may want the opposite.

### Not a defect this PR introduced

[...]

PR #423, opened 2026-08-11
fix(fleet): release-ticket.sh confirms .git linkage resolves to this worktree

### LINKED ISSUE #135: release-ticket.sh: a .git naming a gitdir whose core.worktree points elsewhere still makes the dirty check answer about another repo at rc 0

feigi:
## Summary

The dirty-check's `.git` linkage guard in `release-ticket.sh` established only that a `.git` file exists, not that it points back at the worktree it names. A hand-written `.git` (or a hand-set `core.worktree`) naming a gitdir whose working tree is some *other* directory is a well-formed regular file, so the guard passed — and `git -C "$wt" status --porcelain` then answered about that other tree at rc 0, a positive "this claim is clean" derived from an unrelated repository while the claim's own uncommitted files sat untouched.

Widened the guard: compare `git -C "$wt" rev-parse --show-toplevel` against `$wt`, both canonicalised via `cd "$wt" && pwd -P` (POSIX, no `realpath`). Canonicalising both sides is required — a worktree reached through a symlinked parent has a non-canonical recorded `$wt` (`worktree list --porcelain` echoes the admin file verbatim) while `--show-toplevel` always answers canonical, so comparing the raw strings would false-refuse a healthy worktree.

The new check is gated on the existing `-f "$wt/.git"` guard having actually found a file, so an unsearchable `$wt` still falls straight through to the status die below, preserving git's own "Permission denied" rather than a message this fix invented.

## Enumerate-and-declare

Members of the ".git shape passed to the dirty check" class, and how each is handled:

- **Absent `.git`** — refused today (`-f` false), unchanged.
- **Empty `.git` directory** — refused today (`-f` false, since `-e` is true but `-f` requires a regular file), unchanged.
- **Dangling `.git` symlink** — refused today (`-f` false through a dangling link), unchanged.
- **`.git` we cannot stat (unsearchable `$wt`)** — deferred to the existing status die, git's own "Permission denied" preserved; the new linkage check does not run for this case.
- **`.git` naming a gitdir whose linked repo cannot itself be resolved** (e.g. points at a missing `.git/worktrees/<name>` entry) — refused, via the new `git -C "$wt" rev-parse --show-toplevel` failing the same way `status` does; message still says "whether it holds uncommitted work is unknown". Covered by the existing "an unreadable worktree" test.
- **`.git` naming a gitdir whose `core.worktree` is some other directory** (this ticket's class, including the sibling-worktree-swap shape) — now refused, not leaked. New test: "a `.git` linkage naming another repository's worktree refuses instead of leaking that repo's clean status".
- **Healthy `.git` at a canonical path** — unaffected; covered by the full existing suite (57 pre-existing tests still pass; one of them — the unsearchable-`$wt` case — was **tightened** in the review fix `6ae59b3` to pin which die fires, since excluding only one wrong wording left the gap below invisible).
- **Healthy `.git` reached through a symlinked parent (non-canonical recorded path)** — must still release normally; this is the false-refusal the fix must not introduce. New test: "a healthy worktree reached through a symlinked parent still releases normally".

Deliberately left (out of scope per the ticket):
- The same root cause in `reap.sh`, `worktree-audit.sh` and `claim-ticket.sh` — #128.
- The unreadable-worktree input class — #94.
- Keying on git's `prunable` annotation — a different question, would skip the check on a worktree still holding work.

## Test plan

- [x] `./agent-test skills/fleet/scripts/release-ticket.test.mjs` — 59/59 green (57 pre-existing + 2 new).
- [x] Full suite (`./agent-test`) — 570/570 green.
- [x] Both new acceptance-criteria cases measured against real git 2.50.1 before writing the fix.

Closes #135


feigi:
## Review fixes applied — `6ae59b3`

13 findings adjudicated. **One code change**, because the four `survived` findings were one defect seen from four dimensions.

### Applied

**`survived[1-4]`** (critical/comments + important × correctness, silent-failure, types) — all four reported the same thing: the comment asserts the linkage block is *"Gated on `-f` … this block must not run"* for an unsearchable `$wt`, but the block was ungated, so `cd "$wt"` fired first and the run died with `cannot resolve $wt` — the message the comment says it exists to avoid — losing git's own `fatal: cannot change to '…': Permission denied`.

Closed by **adding the gate the comment describes** rather than rewriting the comment, on the evidence of intent the reviewers pointed at: the pre-existing `[ ! -x "$wt" ]` fallback on the guard three lines above deliberately declines to die for exactly this case in order to leave the answer to git, and the ungated block undid that.

**`unverified[5]`** applied as the pin. The existing test excluded only *one* wrong wording (`doesNotMatch(stderr, /has no \.git/)`), which is why it stayed green through the defect; it now also pins **which** die fires.

**`unverified[6]`** was the same defect a fifth time — covered by the same fix. Took the minimal wrap rather than its if/else restructure, which would have rewritten the pre-existing guard too.

**Mutation-verified both directions:** neutralising the gate reddens that test and only that test, on the new assertion; rewording the unrelated mismatch die reddens a *different* test and leaves it green.

`node --test skills/fleet/scripts/*.test.mjs` → **tests 570 / pass 570 / fail 0 / skipped 0.**

### No-ops

**`unverified[2]`, `unverified[3]`** — confirmations that this PR's two new tests genuinely discriminate. No defect, no issue.

### Deferred

Each in-scope `suggestion` got one refuter, biased to refuse and required to verify by running. **Nothing survived**, so nothing else was applied.

| finding | dimension | outcome | filed |
|---|---|---|---|
| `refuted[1]` GIT_WORK_TREE | silent-failure | framing refuted, residue real | **#427** (new, `needs-triage`) |
| `unverified[7]` second `--show-toplevel` spelling | simplify | out of scope (#189's shape) | comment on **#421** |
| `unverified[4]` no test for the `cd` failure arm | tests | measured-unreachable, TOCTOU-only | comment on **#334** |
| `unverified[1]` gitdir pointing back at `$wt` | correctness | **refuted** | comment on **#421** |
| `unverified[8]` "26 lines of comment overstate" | simplify | **refuted** | comment on **#421** |

Commented rather than filed where an open ticket already owned the subject (#421 the guard-coverage/shared-helper class, #334 the measured-unreachable hardening nits in this same file).

### Two things worth flagging beyond the fix

**A suggested fix was actively unsafe.** `unverified[1]` proposed closing its shape with `[ "$(git -C "$wt" rev-parse --absolute-git-dir)" = "$wtroot/$(basename "$wt")" ]`. Measured false-refusals on healthy worktrees with no sabotage: the admin-dir name is de-duplicated at `worktree add` time (`9-dup` → `9-dup1`) and frozen across `git worktree move`, so it is not a valid identity for the git dir. It would have reintroduced the permanent-exit-2 class this PR's own canonicalisation exists to prevent.

**`unverified[8]`'s rewording would have shipped a false claim.** It asserted `git worktree remove` refuses these shapes too; measured false — with `extensions.worktreeConfig` + `config.worktree` the back-pointer stays intact, so pre-PR a **clean** redirected worktree **fully releases at rc 0** (`"released":true`, worktree removed, branch deleted, label dropped). That also means this guard averts more than an untidy halt, which is the opposite of what that finding argued. Filed against #135.

Issue #242, opened 2026-08-10
PR #227 review: nine findings refuted, recorded with their refutations


feigi:
Deferred from PR #227 review (issue #146, branch `fix/146-escape-control-bytes`). Nothing here was applied. Filed so a future reader can see these were **examined and ruled non-defects**, and on what evidence — not merely unnoticed.

**Dimension attribution is incomplete.** The six review dimensions ran under the controller; only `file:line` + claim + verdict were relayed to the fix-applier, not the per-finding `dimension`. The three at the bottom carry theirs because the fix-applier dispatched them.

## Refuted during the review (6)

Four of these split **1-1** between their two refuters — one refuter judged the finding real. They come back refuted, but they were **contested, not unanimous**.

| # | Site | Claim | Verdict |
|---|---|---|---|
| 1 | `skills/fleet/commands/run-merge-bot.md:136` | the `conflictsRewritten:false` paste-safety paragraph is wrong | refuted, split 1-1 |
| 2 | `skills/fleet/scripts/no-undo-audit.sh:226` | a `\001` path aborts with a "contains a newline" message | refuted, split 1-1 |
| 3 | `skills/fleet/scripts/inflight.sh:238` | `"same pipeline as release-ticket.sh:115"` now points at the wrong line | refuted, unanimous |
| 4 | `skills/fleet/scripts/inflight.sh:150` | stale cross-reference to `release-ticket.sh:173` | refuted, split 1-1 |
| 5 | `skills/fleet/scripts/inflight.sh:273` | the `*Rewritten` invariant is enforced only by prose | refuted, split 1-1 |
| 6 | `skills/fleet/scripts/release-ticket.sh:181` | `blockers[]` lacks a `*Rewritten` flag | refuted, split 1-1 — this was the scope cut the implementer declared in their own report, and the refuters ruled it acceptable, so that call stands |

### Finding 4 — the refutation is wrong, measured

`inflight.sh:150` reads *"That matters less here than in release-ticket.sh:173, whose pushed-branch lookup tests `$remote` raw"*. Measured at PR head `ebcb5c1`:

- `release-ticket.sh:173` held `echo "    worktree removed: $done_wt, …"` — not a pushed-branch lookup.
- The pushed-branch lookup, `if ! remote=$(git ls-remote --heads origin "refs/heads/$branch"); then`, was at **`:281`**.

[...]

Issue #190, opened 2026-07-31
no-undo-audit.sh: jstr failure inside the printf argument list is invisible to set -e, so the payload can carry an empty worktree name


feigi:
Deferred from PR #180 review (found by the silent-failure sweep; pre-existing, not introduced by that PR).

The final payload interpolates two command substitutions directly into `printf`'s argument list:

```sh
printf '{"worktree":"%s","branch":"%s","clean":%s,"stash":%s,"conflicts":[%s],"atRisk":[%s]}\n' \
  "$(jstr "$wt")" "$(jstr "$branch")" "$clean" "$stash" "$conflicts_json" "$at_risk_json" \
  || die "could not write the audit for $branch"
```

A command substitution in a **simple command's argument list** has its exit status discarded by POSIX, so `set -e` cannot see it. The `|| die` covers `printf` only. If the `tr` or `sed` inside `jstr` fails, the substitution yields the empty string and the script emits an otherwise well-formed payload reading `"worktree":""` — at whatever exit code the audit had reached.

Contrast the array path, which is already correct: `conflicts_json=$(printf ... | jarr)` and `at_risk_json=$(...)` run in **assignments**, where `set -e` does see the status.

### Fix

Same shape as the `jarr` lines — assign first, then interpolate:

```sh
wt_json=$(jstr "$wt") || die "could not escape the worktree name"
branch_json=$(jstr "$branch") || die "could not escape the branch name"
printf '...' "$wt_json" "$branch_json" ... || die "could not write the audit for $branch"
```

### Severity

Low — reachability is marginal (`tr`/`sed` do not ordinarily fail). Filed because this script's stated standard is that an unanswerable question is exit 2 and never a payload, and this is the one path that can emit a *silently degraded* payload instead. The approach is settled, hence `ready-for-agent`.

`skills/fleet/scripts/no-undo-audit.sh` — the closing `printf` block.

Deferred from PR #180 review.

feigi:
Closing as **already fixed** — verified against `origin/main` @ `fca74fd` during a `/fleet:run-team` phase-0 liveness check, before claiming.

## The cited construct no longer exists

This issue's defect is the two command substitutions spliced directly into `printf`'s argument list, where POSIX discards their exit status so `set -e` cannot see them:

```sh
printf '{"worktree":"%s",...}\n' \
  "$(jstr "$wt")" "$(jstr "$branch")" ... || die "could not write the audit for $branch"
```

[...]

PR #425, opened 2026-08-11
fix(fleet): inflight.sh nulls an unrenderable evidence field instead of dying

### LINKED ISSUE #120: inflight.sh: a broken sed/tr silently reports a claimed ticket as FREE
Deferred from PR #110 review. **Pre-existing** — reproduced identically at the merge base `a56c01a` and at PR #110's head `d663193`, so #110 neither caused nor worsened it.

## The wrong-decision bug

Probe 2 filters its branch list through a pipeline ending in `|| true` (`skills/fleet/scripts/inflight.sh:157-158`):

```sh
remote=$(printf '%s\n' "$heads" | awk '{print $2}' | sed 's#refs/heads/##' |
         grep -E "(^|[/-])$n([-/]|$)" | paste -sd, - || true)
```

A pipeline reports its *last* command's status, and `|| true` discards even that. So any failure of `awk`, `sed` or `grep` — a missing binary, a fork failure under process-table pressure — yields an empty `$remote`, which is indistinguishable from *no branch matched*.

Measured. Bare origin holding `refs/heads/fix-55-thing`, so the correct answer is **taken**:

```
$ PATH=<dir with a sed that exits 127>:$PATH sh inflight.sh 55
exit=0  {"issue":55,"taken":false,"hits":[],"evidence":{...all empty}}
```

**Exit 0 is "free".** Two agents then claim the same ticket — the one outcome the script's own header says it must never invent ("a wrong 'free' puts two agents on the same one"). Probe 3 has the same shape at `:189-193` (`awk | while read | awk | paste -sd, - || true`).

This is a strictly worse instance of the class than anything in #110: it moves the **decision**, not the evidence.

## The related evidence-only case introduced by #110

`inflight.sh:240-242` — the four `$(jstr ...)` calls sit in the argument list of the `printf` whose `|| die` is supposed to catch failure. A command substitution's status is discarded there, so neither `set -e` nor `|| die` can fire. With `jstr` forced to fail:

```
exit=1  {"issue":42,"taken":true,"hits":["local"],"evidence":{"pr":"","remote":"","localBranch":"","worktree":""}}
```

Internally contradictory — `taken:true` with `hits`, every field blank — and silent. The exit code stays correct, so this is evidence corruption only. Also `jstr` is itself a pipeline (`printf | tr | sed`) returning *sed's* status, so a dead `tr` is invisible even if the call were hoisted; that half is inherited from `release-ticket.sh:115` and fixing it there fixes both copies.

## Why triage rather than a decided fix

`set -o pipefail` is not POSIX and these are `#!/bin/sh`. The guarded-capture idiom the script already uses (`if ! heads=$(git ls-remote --heads origin)`, `:155`) works for a single command but not for a filter pipeline whose `grep` legitimately exits 1 on no match — the whole point is that *looked, found nothing* and *could not look* must stay distinct.

And the obvious fix for the `jstr` half — hoist to locals with `|| die` — is probably **wrong**: `die` exits 2, converting a correctly-computed verdict into "unanswerable" because a formatter broke. Worth deciding deliberately.

Suggested direction, not settled: capture each stage's status explicitly, or drop `|| true` and treat `grep`'s exit 1 as the only tolerated non-zero.

feigi:
## Summary

`inflight.sh`'s evidence formatter (`jstr`/`jrewritten` for the `pr`,
`remote`, `localBranch`, `worktree` fields) used to sit in one long `&&`
chain ending `|| die`. Any one of the eight escaping calls failing threw
away an already-correct, already-established verdict and reported exit 2
("unanswerable") instead of the free/taken answer the three probes had
already decided — converting a correct answer into a wrong one over a
formatter, not a decision.

## Fix

- `jstr`/`jrewritten` now short-circuit on empty input (no fork at all) —
  a field a probe legitimately found nothing for never calls sed/tr, so it
  can never observe an escaper outage.
- Each of the four evidence fields is now rendered independently via
  `render_evidence`. A field that cannot be escaped becomes JSON `null`
  (never `""`, which already means "probe looked and found nothing"), with
  a stderr line naming which field. `$taken`, `hits`, `unknown`, and the
  exit code are untouched.
- The final verdict `printf`'s own `|| die "could not write the verdict"`
  is unchanged — a failure to write the payload itself still exits 2, as
  before.

## Enumerate-the-class

Ticket's own class: all four evidence fields (`pr`, `remote`,
`localBranch`, `worktree`) go through the same `jstr`/`jrewritten` pair in
one call site (`inflight.sh:621` area, pre-fix). All four are covered —
`render_evidence` is called once per field, independently, so a failure on
one cannot suppress or null a sibling that rendered fine.

Left deliberately out of scope, per the issue's own Respec:
- `release-ticket.sh` has the exact same `&&`-chain-into-`die` pattern
  (`branch_j`/`wt_j`/`blocker_j`) — same bug class, but the issue's Respec
  explicitly assigns the shared `jstr` helper and its duplication across
  scripts to #119, "not separately here."
- `jstr`'s own internal pipeline-status flaw (its `sed | tr` pipe reports
  only `tr`'s status, so a broken `sed` can be masked by a working `tr`) is
  also #119's, not #120's — confirmed while writing the regression test:
  forcing `sed` to fail alone does NOT make `jstr` report failure, only
  forcing `tr` (the pipeline's actual last stage, shared by both `jstr` and
  `jrewritten`) does. The test targets `tr` for exactly this reason.
- The "no such issue" / "not a git repo" / "bad argument" dies are a
  different class (preconditions, not evidence formatting) — unchanged.

## Test plan

- [x] `./agent-test skills/fleet/scripts/inflight.test.mjs` — 61/61 green,
      including the rewritten regression test (previously pinned the old
      die-on-format-failure behavior; now pins null-field + unchanged
      verdict) and a new "must ACCEPT ordinary evidence" positive case
- [x] `./agent-test skills/fleet/scripts/release-ticket.test.mjs` — 57/57
      green (untouched file, confirms no cross-file regression)
- [x] `./agent-test skills/fleet/scripts/ledger.test.mjs` — 31/31 green
- [x] `./agent-test skills/fleet/scripts/no-undo-audit.test.mjs` — 48/48
      green

Closes #120


feigi:
## Review findings applied — deferrals filed

Head is now `1a2f59f` (two commits on top of `aea3deb`).

**Applied (6):**
- **`survived-1`** — `jrewritten` now returns its `tr` failure explicitly (`|| return 1`). Its last command is an always-0 AND-OR list, so the failure previously reached the caller only via `set -e`, which is exempted inside the `if` condition at the call site — and whether the exemption reaches the assignment is the shell's choice. Measured: `dash`, Apple's `/bin/sh` and bash 3.2.57 in POSIX/sh mode abort correctly; **bash 5.3 in every mode (including as `sh` and `--posix`)**, bash 3.2.57 outside POSIX mode, and zsh 5.9 run on and emit `"localBranchRewritten":true` about a name holding no control bytes, with no warning. That covers every distro whose `/bin/sh` is bash 5.x.
- **`survived-2`** — the deleted `trFailWhenArgsHave: "-d"` case is restored and now runs under **both** `sh` and `bash`, because `sh` cannot see the above on either platform this suite runs on. Its mirror is new: `"\037 "` isolates jstr (its `tr` argv ends `\037` + space; jrewritten's ends at `\037`), so the `&&`'s two operands are pinned independently.
- **`refuted-1`** (controller reversal) — the two non-`-d` `tr` calls are inside `add_unknown` messages, not `die` messages.
- **`unverified-3`** — the stderr line now names the escaper, `... as JSON (jrewritten) — reported as null`. Its suggested fix was rejected: verbatim it reddens the suite, because `inflight.test.mjs:905` pins the existing wording.
- **`unverified-4`** — the header's unconditional null guarantee is scoped. A broken `sed` leaves jstr **succeeding** with empty output (pipeline reports only `tr`'s status), so a real branch renders as `""` with no null and no stderr line. The same over-claim sat in `render_evidence`'s own comment — the load-bearing copy, which the finding had missed — and is fixed too. Masking itself stays #119's.
- **`unverified-9`** — the four `render_evidence` calls became an `$evidence` accumulator, the idiom `add_hit`/`add_unknown` already use. Differential against `ff0ec73` over four shapes: stdout and exit **byte-identical**; stderr differs only by the intended escaper name.

**Deferred and filed (2 issues, 6 findings):**
- **#431** — `no-undo-audit.sh:421` still retracts an established audit over a formatter failure, the same class #120 fixes here. Out of #120's scope. `needs-triage`, not `ready-for-agent`: the payload schema is an open decision and it collides with open #190, whose proposed fix is what introduced the shape.
- **#432** — the five refuted `suggestion`-band findings with their refutations, plus the review's own `refuted-2`/`refuted-3`. Same record pattern as #242/#297. None should be actioned.

**Verification.** `node --test skills/fleet/scripts/*.test.mjs` → **tests 572, pass 572, fail 0**. Every mutant was reverted and its mutated line printed after applying:

| mutation | result |
|---|---|
| strip jrewritten's `\|\| return 1` | only the `bash` case red (1 of 64) |
| jrewritten failure undetected | both `-d` cases red (2 of 64) |
| jstr failure undetected | the mirror case red (1 of 64) |
| equivalent `if ! raw=…` rewrite | all 64 green — pins behaviour, not spelling |

Before this branch, the last two mutants passed all 61 tests in the file.

**One correction for the record.** The review's `survived-1` claimed the split was "bash >= 4.4"; that was corrected in triage to "bash-invoked-as-`sh` vs Apple's `/bin/sh`". **Both are wrong.** The corrected version rested on a symlink named `sh32` — bash enters POSIX mode off `basename(argv[0]) == "sh"`, so that probe never tested sh-mode. Measured here: bash 3.2.57 symlinked as `sh` *aborts* (correct), while bash 5.3 masks in **every** mode including `--posix`. The version boundary is real, but only within POSIX mode. The finding's practical claim — a bash-5 `/bin/sh` hits this — survives intact.

CI on `ff0ec73`: `check` **success**, `rebase-check` failure at `behind_by=6` — staleness, not a diff failure, so not rebased.

feigi:
Follow-up: head is now `d7256b7` (was `1a2f59f`).

One further commit, applying the **residual** the `unverified-5` refuter uncovered — the finding itself was refuted and is recorded in #432, but its measurement exposed a real, narrower gap that the controller authorised separately.

jrewritten's empty short-circuit returns `printf false`, and that return **value** is asserted nowhere: every other `*Rewritten` assertion in the file sits on a non-empty field. So flipping it to `printf true` emits `"remoteRewritten":true` about a string no escaper ever looked at, and passes all 61 tests the file had before this branch. Same class as #120 itself — a disclosure bit contradicting what actually happened — one slot over.

One assertion, on `probe 2: a reachable origin with no matching branch still reports free`, where the field genuinely is empty. Mutation-checked both ways: the `printf false` → `printf true` flip reddens exactly that test (1 of 64); respelling `[ "$raw" = "$orig" ] && printf false || printf true` as an if/else with the operands swapped stays green, so it pins behaviour and not spelling.

Deliberately **not** placed at the "quote in a remote branch" test the finding proposed — that remote is non-empty, so the assertion there cannot reach the empty path (measured: mutation + that assertion stays green).

`node --test skills/fleet/scripts/*.test.mjs` → tests 572, pass 572, fail 0.

Issue #265, opened 2026-08-10
reap.sh:102 — a failing `git worktree prune` exits 1 with no payload, after the branches are already deleted


feigi:
**dimension:** correctness · **severity:** critical · Deferred from PR #251 review (ticket #114)

`skills/fleet/scripts/reap.sh:102`

```sh
[ "$apply" = true ] && git worktree prune

printf '{"applied":%s,"reaped":[%s],"kept":[%s]}\n' ...
```

Under `set -eu` the prune is the **last command of the AND-OR list**, so `-e` applies to it. A failing `git worktree prune` aborts the script **before the `printf`**: exit **1**, no JSON payload — after the branches have already been deleted. The caller loses the `reaped[]`/`kept[]` record of work already done.

Exit 1 is the code `docs/specs/2026-07-23-fleet-plugin-design.md`'s new preamble reserves for "a verdict the caller must act on and never an error", from a script the same table says has no verdict.

## Reproduction

```
# real git
$ sh skills/fleet/scripts/reap.sh --apply
    REAPED feature
{"applied":true,"reaped":["feature"],"kept":[]}          rc=0

# PATH shim failing only `git worktree prune`
$ PATH=$W/bin:$PATH sh skills/fleet/scripts/reap.sh --apply
    REAPED feature
stub: worktree prune failed (e.g. unwritable .git)        rc=1, no JSON line
```

Shell rule confirmed separately: `sh -c 'set -e; true && false; echo no'` → rc=1, nothing printed.

## Two viable approaches — pick one at triage

1. Guard in place: `[ "$apply" = true ] && { git worktree prune || die "worktree prune failed"; }` — makes exit 2 correct, but still emits no payload.
2. Print the payload first, prune after — the caller keeps the record of what was deleted regardless.

Both change the script's exit surface, which is why this is filed rather than fixed: #114's brief is *"Changing any script's exit codes — this is a documentation audit; if a code looks wrong, file it, do not fix it here."*

PR #251 corrects the spec row to state the real exit-1 behaviour. This issue is the code half. **No test covers `reap.sh`.**

feigi:
> *This was generated by AI during triage.*

**Triage verification:** claim confirmed live at `974a3b8` — the prune is still the last command of the AND-OR list, ahead of the payload `printf`. **Decision taken at triage: approach 2, plus a loud refusal** (the body left the choice open).

## Agent Brief

**Category:** bug
[...]

Issue #391, opened 2026-08-11
reap.sh: `worktree remove` refusal is read from the exit code, so a cleared registration reports as "refused" — and git's reason is discarded


feigi:
> *This was generated by AI during triage.*

Filed during the `/triage` grill of #208, which rules on the same root cause in `release-ticket.sh`. Same defect, second site, different consequence — filed separately because the two scripts share no helper and #208 already carries dependency edges.

## Problem

`skills/fleet/scripts/reap.sh`, in the `[gone]`-branch loop:

```sh
git worktree remove "$wt" 2>/dev/null || { keep "$b" "worktree remove refused"; continue; }
```

A non-zero exit from `git worktree remove` does **not** mean the removal had no effect. Git deletes the admin entry even when the subsequent directory removal fails — `remove_dir_recursively` failing does not skip `delete_git_dir`.

Two measured shapes produce it, both on git 2.50.1 (Apple Git-155):

1. **Mid-delete permission failure.** A clean worktree with one subdirectory at mode 555: `error: failed to delete '…': Permission denied`, rc 255, registration gone, directory and files still on disk.
2. **A symlink standing in for the worktree directory.** `error: failed to delete '…': Not a directory`, rc 255, and the porcelain no longer lists the entry afterwards.

## Why this site is arguably worse than #208's

`release-ticket.sh` at least halts and tells the operator the claim is unfinished. Here the branch is **kept** — `keep "$b" "worktree remove refused"` — so the run reports the branch as deliberately retained and moves on. The result reads as a safe no-op while an orphaned directory is left behind, and the branch that would have been the clue is still standing.

## Second defect on the same line

`2>/dev/null` discards git's own reason. The operator gets `worktree remove refused` with no cause — no permission error, no `Not a directory`, nothing to distinguish a locked worktree from a symlink from a genuine dirty tree. `release-ticket.sh` quotes git's stderr back verbatim on the equivalent path; this one throws it away. Worth fixing in the same change: the reason is the only thing that tells an operator which remedy applies.

## Vocabulary

Per `CONTEXT.md` (PR #390), the state produced here is **Deregistered** — registration cleared, directory on disk — and the directory it leaves is an **Orphaned worktree directory**, invisible to any check reading git's registry.

## Why `needs-triage`

[...]

PR #426, opened 2026-08-11
fix(reap): a dead git cherry no longer authorizes deleting a [gone] branch

### LINKED ISSUE #264: reap.sh:44 — a git cherry that DIES reads as "merged", so --apply deletes a [gone] branch whose commits exist nowhere else, at exit 0
**dimension:** silent-failure · **severity:** critical · Deferred from PR #251 review (ticket #114)

`skills/fleet/scripts/reap.sh:44`

```sh
if git cherry "$base" "$b" 2>/dev/null | grep -q '^+'; then
  keep "$b" "unmerged commits"
  continue
fi
```

The pipeline's exit status is **grep's**, never `git cherry`'s. A `git cherry` that dies (exit 128 — a single unreadable loose object is enough) emits nothing on stdout, so grep exits 1, and "the probe could not answer" is indistinguishable from "no unmerged commits". `--apply` then runs `git branch -D` on a `[gone]` branch whose commits exist **nowhere else**, at **exit 0**, with git's error suppressed by the `2>/dev/null` on the same line.

The `git branch -D` site's own comment says the deletion "is authorized by the cherry check above and by nothing else" — so when the cherry check fails open, nothing authorizes it.

## Reproduction

Two independent reproductions, one with **real corruption and no fault injection**:

**A — real corruption.** Bare origin + clone, `main = base`, `feature/unmerged` with two commits pushed then remote-deleted (upstream `[gone]`). `chmod 000` on an ancestor loose object:

```
$ git cherry origin/main feature/unmerged
error: unable to open loose object b257d786...: Permission denied
fatal: revision walk setup failed                      (exit 128)

$ sh skills/fleet/scripts/reap.sh --apply
$ git fetch --prune origin
    REAPED feature/unmerged
{"applied":true,"reaped":["feature/unmerged"],"kept":[]}   rc=0

$ git branch --list 'feature/*'
(empty — branch gone, commits unreachable)
```

**B — PATH wrapper failing only `git cherry`.** Same result; `git log --oneline --all` afterwards shows only the base commit.

**Control on the same uncorrupted repo:** `KEEP feature/unmerged — unmerged commits`, `kept[]` populated, rc=0.

## Bounding fact — this fleet run was NOT affected

The `reap.sh --apply` in this run is verified unaffected. It reaped `docs/66-compress-59-blocks`, whose tip `bcf854a7` is an ancestor of `origin/main`; and in the **same invocation** it KEPT `feat/212-implementer-model-by-class` for `unmerged commits` — a verdict reachable only when `git cherry` actually emits `+`. One healthy KEEP proves cherry was answering in that pass. The defect is real and latent, not something that has already destroyed work here.

## Suggested direction (not settled — hence `needs-triage`)

Capture the probe before judging it, and fail closed the way the `status` check on line 53 already does:

```sh
if ! cherry=$(git cherry "$base" "$b" 2>&1); then
  keep "$b" "cherry failed — cannot tell if merged: $cherry"
  continue
fi
case $cherry in *'+'*) keep "$b" "unmerged commits"; continue;; esac
```

`prove-merge.sh:39-43` applies the same rule to `merge-base --is-ancestor`. Same trap `worktree-audit.sh:31-34` documents for `wc -l`.

**No test covers `reap.sh`** — a regression test belongs with the fix.

Out of scope for #114 (a documentation audit: *"if a code looks wrong, file it, do not fix it here"*). PR #251 documents the behaviour in the `reap.sh` spec row — it does not fix it.

feigi:
## Summary

`reap.sh`'s merge probe piped `git cherry` into `grep -q '^+'`. A pipeline's exit status is the *last* command's, never `git cherry`'s — so a `git cherry` that dies (exit 128, one unreadable loose object is enough) prints nothing, `grep` sees empty stdin and exits 1, and "the probe could not answer" was indistinguishable from "no unmerged commits." Under `--apply`, `reap.sh` then force-deleted a `[gone]` branch whose commits existed nowhere else, at exit 0, with git's own diagnostic thrown away by the same line's `2>/dev/null`.

Fix captures the probe's output *and* status before judging it — same fail-closed shape the worktree `status` check just below it already uses, and the shape `prove-merge.sh`'s `is_ancestor()` uses for `merge-base --is-ancestor`. A non-zero probe now `KEEP`s the branch, with a reason naming the failure and carrying git's own message (flattened to one line — git's real error is multi-line, and an embedded newline would break the JSON payload).

- [x] A `[gone]` branch whose merge probe exits non-zero is KEPT, reason names the probe failure + git's own message
- [x] `--apply` never deletes on a failed probe; run still exits 0, branch recorded in `kept[]`
- [x] Uncorrupted behavior unchanged: a probe emitting `+` keeps with "unmerged commits"; a clean probe still reaps
- [x] Test file added for the reaper (none existed), covers the failed-probe case via a PATH shim failing only `git cherry`, plus both control cases
- [x] New test fails against the pre-fix pipeline form — mutation-checked by running the suite against the unmodified script

## Class enumeration (per ticket)

`reap.sh` has other places that discard or take-the-tail-of a pipeline's status. Covered vs. deliberately left:

**Fixed:** the named `git cherry | grep -q '^+'` probe (line 44).

**Considered, left as-is:**
- `git for-each-ref | awk` branch enumeration — same command-substitution status-discard shape, but fails in the *safe* direction: a failure yields empty enumeration, the loop never runs, nothing is deleted, exit 0 with an empty payload. Opposite failure direction from the named bug.
- `git worktree list --porcelain | awk` worktree lookup — same shape, but git's own `branch -D` refuses to delete a branch checked out in a live worktree, so a failed lookup can't lead to losing that worktree's uncommitted work; the script's `|| keep` on the `-D` failure catches it regardless.
- `printf | awk | paste` ignored-files formatting — pure string reformatting of data already validated by the fail-closed check immediately above it; no realistic trigger for `printf`/`awk`/`paste` failing on a fixed in-memory string.

**Confirmed, explicitly out of scope (#265):** `[ "$apply" = true ] && git worktree prune` (now line ~112) is still present unchanged. It's the last command of an `&&` list, so under `set -eu` a failing `prune` trips `-e` and the script exits 1 with **no JSON payload**, even though branches were already deleted earlier in that same loop. Verified live at this SHA. Not touched here — same script, no test file today, but the two fixes stay separable per the ticket's brief.

## Test plan

- [x] `./agent-test skills/fleet/scripts/reap.test.mjs` — 3/3 pass
- [x] Mutation check: ran the same suite against the pre-fix script — the failed-probe test fails there (branch wrongly reaped) and passes only with the fix

Closes #264


feigi:
## Review applied — 27 findings adjudicated

Two commits: `7985892` (survived findings) and `141ad17` (the one in-scope suggestion that survived refutation). Suite: `node --test skills/fleet/scripts/*.test.mjs` → **573 pass, 0 fail** (571 baseline + 2 new pins). Both added tests were mutation-tested in both directions — each kills its own mutant and only its own, and stays green under a behaviour-preserving rewrite.

### Applied

| Finding | Fix |
|---|---|
| `survived-1` + `survived-2` (one defect) | The spec's script-surface `reap.sh` row asserted, present-tense, that the merged check "reads a `git cherry` that failed as 'no unmerged commits' and reaps the branch" — verbatim the bug this PR fixes. Row rewritten to the fail-closed contract |
| `survived-2`, second half | Nothing pinned that row, which is why 571 tests passed over it. New pin, modelled on `no-undo-audit.test.mjs:1037` and derived from a real run, so a phrase typed into the doc cannot drift from the reason the script emits |
| `survived-3` | The new comment had the ambiguity backwards. A dead cherry and a genuinely clean one both make grep exit 1 — so **"merged"** is what a failed probe is indistinguishable from, never "unmerged", which always keeps |
| `survived-4`, prose half | `keep()` now carries a comment naming the unescaped-splice limitation and pointing at #119, so a reader meets it in the code rather than in a corrupt payload |
| `unverified-1` + `unverified-4` (one defect) | The `^` anchor was dropped when `grep -q '^+'` became `case $cherry in *'+'*`, and the capture folds stderr in via `2>&1` — a `+` in any diagnostic read as a commit line. Anchored form restored, plus a test shimming a successful-but-noisy cherry |
| `unverified-3`, first half | `reaping.md` gained the fail-closed bullet its own enumeration was missing |

### Deferred, with the measurement, to existing tickets

No new issues — every deferral had a home already, and the ledger's tracker query caught the near-duplicate.

- **#119** (`fleet scripts: five more emit unescaped JSON, and jstr is now duplicated`) — `survived-4`, `unverified-2`, `unverified-17`, and the refutation of `refuted-3`. A refuter reverse-applied this diff and ran a quote-named-branch fixture against the reconstructed pre-PR script: **byte-identically broken JSON, exit 0, no line from this diff involved**. So this PR adds a fifth instance of a pre-existing defect and is not the introducing site. Left to #119 rather than patched here because `jstr` already exists twice (`no-undo-audit.sh:46`, `inflight.sh:493`) and a third divergent inline escape makes the duplication that ticket exists to fix strictly worse.
- **#391** (`worktree remove` refusal read from the exit code, git's reason discarded) — `refuted-1`'s surviving residue and `unverified-5`. Two more sites on the same axis: the `git worktree list | awk` swallow at `reap.sh:70` and the `git branch -D` catch-all at `:115`, both reporting `branch delete failed` while git's actual reason goes to `/dev/null`. Measured as **not** a data-loss vector — `git branch -D` consults the same registry, so readable → it refuses → KEEP, unreadable → the ref may go but the worktree and its uncommitted files stay. Also recorded there: the whole worktree-removal path has zero test coverage, before or after this PR.
- **#250** (`reap.sh` guards arity only) — `unverified-12`. No test exercises the dry-run path, which is exactly the branch a fix for #250 changes.
- **#265** — `unverified-13` and `unverified-16` are that ticket (a failing `git worktree prune` under `set -eu` exits 1 with no payload after the branches are already deleted). Confirmed present, correctly out of scope here.

### Refuted by measurement — recorded so they are not re-raised

- `refuted-1` — the claimed data-loss vector is unreachable; only the mislabeled reason survives, and it is deferred to #391 above.
- `refuted-2`, `refuted-3` — the JSON corruption is real but pre-existing and reproduces identically with this diff reverse-applied. Real `git cherry` corruption output carries no quote characters, making the new call site the least likely of the five to fire.
- `refuted-4` — `derive-testcmd.sh:31`'s citation is past-tense history and its pointer resolves: `reap.sh` names the exact shape in prose directly above the fixed line.
- `unverified-18` — dropping `runReap`'s `args` parameter. Refuted: this repo's two other `--apply` script helpers (`drop-merged-label.test.mjs:84`, `release-ticket.test.mjs:117`) both keep apply variable and exercise dry run heavily; hard-coding it makes this the one `--apply` script test that cannot reach dry run, and #250's fix needs that branch.
- `unverified-19` — dropping the `r.stdout.trim() ? … : null` guard. Refuted on a measurement: all three tests `assert.equal(code, 0)` **first**, so the null is never dereferenced. With the guard an empty-stdout run reports `2 !== 0` at the test's own line; without it, `SyntaxError: Unexpected end of JSON input` inside the helper. A diagnostic regression, not a wash.
- `unverified-3`, second half (a `Check:` pointer line) — refuted: n=1 repo-wide, so it is the exception, not a convention to extend.
- `unverified-7` through `unverified-11`, `unverified-15` — confirmations rather than defects; no action.

CI reads `rebase-check=failure` / `behind=6`. That is staleness from a merge wave, not a test failure, and is not rebased away here.

PR #434, opened 2026-08-11
fix(fleet): shape-guard every gh JSON reply ci-state.mjs dereferences

### LINKED ISSUE #269: ci-state.mjs guards the parse but not the shape — a valid-JSON/wrong-shape gh reply throws and exits 1, the code reserved for the not-green CI verdict
**dimension:** types · **severity:** critical · Deferred from PR #251 review (ticket #114)

`ci-state.mjs` guards the **parse**, never the **shape**. `runJson()` try/catches `JSON.parse` and dies with exit 2, but a `gh` reply that parses cleanly and is the wrong shape flows on unchecked until the first use of it throws:

```
ci-state.mjs:140
  .filter((r) => r.headSha === prHead)
   ^
TypeError: runs.filter is not a function
```

**An uncaught throw exits 1 — the code this script uses for "not bound-green".** So an unanswerable query renders as a **CI verdict**. That is the precise failure mode the file's own `runJson` comment (`:62-65`) says it exists to prevent.

`grep -n "Array.isArray" ci-state.mjs` → **no hits.** Nothing checks `runs`, `view.jobs`, or `prInfo.headRefName`.

## Reproduction

Fake `gh` on PATH: `pr view` returns valid PR JSON; `run list` exits **0** printing `{"error":"rate limited"}` — valid JSON, but an object where `--json databaseId,headSha,...` produces an array.

```
$ PATH=$PWD/bin:$PATH node ci-state.mjs --pr 1; echo "EXIT=$?"
    branch=b head=abc state=OPEN mergeState=CLEAN
    expected jobs (2): a, b
$ gh run list --branch b --workflow CI --limit 30 --json databaseId,headSha,status,conclusion,event,createdAt
TypeError (uncaught)
EXIT=1
```

Reproduced independently by two reviewers.

## Both siblings already have the guard

`candidates.mjs:183-185` — and its comment names this exact hazard:

```js
// throws, and an uncaught throw exits 1: the code reserved for "successful
// query, no survivors". Fail closed, same as the failed-query path above.
if (!Array.isArray(rows)) die("gh output is not an array — the --jq reduction did not apply");
```

`ledger.mjs:302`:

```js
if (!Array.isArray(parsed) || parsed.some((h) => !h || typeof h.number !== "number")) {
  throw new Error("gh returned JSON that is not an issue list");
}
```

Two sibling sites, one carrying a comment that describes this defect in the abstract, are the strongest argument for the guard: the project already treats this as a real and distinct failure mode, and `ci-state.mjs` was simply never swept.

## Class

This repository keeps closing variants of "an unanswerable question rendered as a verdict":

- **#111** — the phase-3 merge path gates on CI in a repo that has none
- **#262** — the CI monitor exhausts the REST rate limit and the natural watcher shape reports the outage as silence
- **#266** — `verify-sha.sh:31` reports `reachable:false` when the probe exited 128
- **#232** — `ledger.mjs check`'s shape guard validates only `h.number`, so a partial row prints `TRACKER HIT` with `undefined`

`ci-state.mjs` is the highest-leverage of these: the controller polls it for every PR, and exit 1 is acted on as "not green" without a second look.

Minimum is the `candidates.mjs` guard applied to `runs` before `:140`. The open question is scope: a single `Array.isArray(runs)` check, or shape validation on every `runJson` return (`prInfo`, `view.jobs`, `runs`) — the latter is the class fix, and #232 shows a partial guard can itself be the defect.

**No test file covers `ci-state.mjs`** — confirmed by directory listing. A regression test belongs with the fix.

## Scope note

Out of scope for #114 (documentation audit). PR #251 documents the behaviour on the `ci-state.mjs` row — exit 1 now states that a wrong-shaped reply throws and lands there rather than on 2. This issue is the code half.

Review findings `survived-6` (dimension `types`) and, on the doc side, `survived-1/4/5`, `refuted-5`, `unverified-8/12` — all one row, fixed once in #251.

feigi:
## Summary
- `ci-state.mjs`'s `runJson()` checked a `gh` reply was parseable JSON but never that it was the expected *shape* — a valid-JSON/wrong-shape reply (an error object where the runs array is expected, a run view missing its jobs, a PR view missing its head fields) flowed on unchecked until the first dereference threw, and an uncaught throw exits 1: this script's code for "not bound-green". So an unanswerable query rendered as a CI verdict — the exact failure the file's own `runJson` comment already says it exists to prevent, one layer in.
- `runJson()` now takes an optional shape validator, checked immediately after parse, so each call site declares what it expects instead of hand-rolling a check — matching `candidates.mjs:256` and `ledger.mjs:338`, the two siblings that already carry this guard.
- Covers every payload the script dereferences: the PR info's `headRefName`/`headRefOid`, the runs list (array + each row is an object), and the run view's `jobs` array (array + each job entry is an object). A wrong shape now dies with exit 2, naming the query and the field — never exits 1.

## Enumerate-and-declare
Ticket named three payloads explicitly (runs list, run view + jobs, PR info's branch/head) plus #232 as the caution against a *partial* guard. Covered:
- PR info missing `headRefName`/`headRefOid` → exit 2, field named.
- `run list` returning a non-array (the ticket's own reproduction: `{"error":"rate limited"}`) → exit 2.
- `run list` row that isn't an object (`null` in the array) → exit 2 — the #232 extension: a guard that checks the array but not its elements is itself partial, since `r.headSha` is read unguarded per row.
- `run view` missing its `jobs` array entirely → exit 2, never silently read as "zero jobs" (which would misreport an unanswerable query as "expected jobs absent").
- `run view.jobs` entry that isn't an object (`null`) → exit 2 — same #232 extension, `j.name` is read unguarded per job.

Deliberately left out (false-positive guard, not a gap): job/run-row element checks are **object-ness only**, not per-field type checks. A job's `conclusion` is legitimately `null` while in progress — validating it as a required string would make the guard itself refuse a well-formed reply. Pinned by a dedicated accept-case test.

Also out of scope, matching the ticket: rate-limit/outage handling (#262), sweeping the same guard through other scripts, and the three-way exit contract itself — all unchanged.

## Test plan
- [x] `./agent-test skills/fleet/scripts/ci-state.test.mjs` — 26/26 pass (20 pre-existing + 6 new: 5 refusal cases, 1 acceptance case for the false-positive class)
- [x] Mutation-checked: reverted the guard, reran — the 5 refusal tests failed as expected (crash → exit 1 instead of the expected exit 2), the acceptance test still passed
- [x] Well-formed replies unchanged: full pre-existing suite (green/not-green verdicts and reasons) still passes byte-identical

Closes #269

https://claude.ai/code/session_01CYmKcbwLAGbdzmMWimEDT2


feigi:
Review deferrals from this PR are recorded on **#288** (the existing ci-state.mjs/.test.mjs quality-pass ticket, `ready-for-agent`) rather than as a new issue — same two files, same `suggestion` band, and #288 already collects exactly this class from the #273 review: https://github.com/feigi/claude-config/issues/288#issuecomment-5278591238

Two coverage gaps deferred (both guard clauses and the reported row/job index are unpinned — four surviving mutants, measured on `10bad45`), one simplify finding refuted and recorded so it is not re-raised. Everything else the review surfaced is applied in this PR.

PR #435, opened 2026-08-11
fix(fleet): board.mjs refuses a non-numeric --port/--interval instead of silently defaulting

### LINKED ISSUE #366: board.mjs: --port and --interval accept a non-numeric value and silently fall back to the default
Deferred from PR #360 review (dimension: **correctness**). Finding state: **unverified** — the `suggestion` band is budgeted 0 refuters by policy. The measurements below are the reporting specialist's own.

**Scope, read first — this ticket is `--port` and `--interval` ONLY.**

- The `--spend-since` half of this finding is **already fixed on PR #361** and is not part of this ticket. That branch replaces `Number(arg("spend-since")) || null` with a real refusal:
  ```js
  const sinceRaw = arg("spend-since");
  const sinceMs = sinceRaw == null ? null : Number(sinceRaw);
  if (sinceRaw != null && !Number.isFinite(sinceMs)) die(`--spend-since wants epoch milliseconds, got ${sinceRaw}`);
  ```
- A *separate* finding in the same review claimed PR #360's added comment overclaims on this subject. That one was **refuted and adjudicated closed** by the review controller — the PR's comment scopes itself to the trailing-flag case, which is genuinely closed. Not to be re-raised.

## The behaviour

`arg()` now refuses a *missing* value but returns any non-empty, non-`--` string unchanged. `serve()`'s two numeric flags then run it through `Number(...) || default`, so a **garbage value** silently becomes the default (`skills/fleet/scripts/board.mjs`, in `serve()` — anchor on the code, line numbers move between branches):

```js
interval = Number(interval ?? arg("interval")) || 15;
```

and the same shape for `port` (now `Number(port ?? arg("port")) || null` feeding a `?? 8123`, after PR #360's review round). `Number("abc")` is `NaN`, so `--interval abc` silently becomes 15 and `--port abc` silently becomes 8123, with no refusal on either.

Verified against PR #361's fold: both lines are byte-identical to `main` there, so neither branch closes this.

## Precedent already in the tree

Two now, pointing the same way:

- `candidates.mjs:89-90` — `const limit = Number(arg("limit") ?? 500); if (!Number.isInteger(limit) || limit < 1) die(...);`
- PR #361's `--spend-since` guard, quoted above — same script, same class, already decided.

`--port`/`--interval` are the two that did not get it.

## Related, already applied in PR #360

The **survived** half of the original finding is fixed: `board.mjs`'s `EADDRINUSE` message no longer reports a substituted default as if the caller had chosen it — it now marks the default as one (`port 8123 (default) in use`). That is a message fix. It does not stop `--port abc` from becoming 8123 in the first place.

## Why `needs-triage`

The *what* is clear; the *contract* is not, and two implementers would build materially different things:

- `Number.isInteger` plus a range, or just finiteness? A port of `0` is meaningful to `listen()` (bind an ephemeral port); an interval of `0` is not.
- Whether `serve()`'s programmatic callers — `serve({ port })` is exported and called with an already-typed value — should be validated on the same path, or only the CLI one.

Follow PR #361's wording (`--<flag> wants <unit>, got <value>`) if this is taken, so the three guards read alike.

Invalid value, not absent value — a different class from #169, and pre-existing rather than introduced. Triage's ruling on #169 was "fix the four copies in place. No shared module."

feigi:
## Summary

`serve()`'s `--port`/`--interval` and `gather()`'s own `--interval` fallback (the build payload — a second, independent read site) all ran the raw argv string through `Number(x) || default`. `Number("abc")` is `NaN`, which is falsy, so a garbage value silently became the default with no refusal — same class already fixed for `--spend-since` in #361 and `--ledger`/`--prev` in #360.

Fixed per triage's contract on #366:
- `--port`: refused unless an integer `0`-`65535`. **`0` stays legal** (`listen(0)` binds an ephemeral port) — the old `Number(...) || null` idiom read `0` as falsy and would have silently substituted `8123` for an explicit `--port 0`, which is the false-refusal case the ticket calls out by name.
- `--interval`: refused unless finite and `> 0`.
- Both guards live in one place each (`argPort()`/`argInterval()`), called from every CLI-argv read site — `interval` is read from argv in two places (`serve()` and `gather()`'s build payload) and both now share one guard instead of two coercions that could drift.
- CLI path only: `serve({ port })` called with an already-typed value bypasses the guard entirely via `??` — validating a caller's in-process value was explicitly out of scope.
- Wording follows #361: `--<flag> wants <unit>, got <value>`.

## Enumerate-and-declare

Ticket named two numeric flags plus (in passing, then ruled out of scope) `--spend-since`:
- **In scope, covered:** `--port` (CLI, `serve()`), `--interval` (CLI, both read sites — `serve()` and `gather()`'s build payload).
- **Out of scope, left alone:** `--spend-since` (owned by #361, already merged); the `EADDRINUSE` default-marking message (already applied in #360); `serve({ port })`/`serve({ interval })` called programmatically with an already-typed value (ticket: "a caller passing a bad number in-process is a different class of bug and not this ticket's").
- **Not touched:** `arg()`/`has()`/`die()` themselves, and the other flags in this file (`--ledger`, `--prev`, `--open`) — none are numeric. #367 (extracting the shared CLI-helper module) is still open, unmerged on `origin/main` as of this branch — verified by grep, no shared module exists yet — so the guard lives locally in `board.mjs` per the ticket's own fallback instruction.

## What must NOT be refused

`--port 0` is the named case. A test pins it end-to-end through the real CLI (`serve --port 0`), asserting the process is *not* refused (status ≠ 2, no `--port wants` in stderr) and does reach `listen()` (`cockpit on http://localhost:0` in stderr) rather than hanging forever unobserved — it only exits because the test kills it after a short timeout, which is the expected shape of success since `serve()` otherwise blocks until SIGINT/SIGTERM.

## Test plan

- [x] `--port abc`, `--port -1`, `--port 70000`, `--port 1.5` each exit 2 naming the flag and value
- [x] `--port 0` is accepted (not refused, reaches `listen()`)
- [x] `--interval abc`, `--interval 0`, `--interval -5` each exit 2 (both CLI entry points: `serve` and `build`)
- [x] Both `interval` read sites covered — `board.test.mjs` pins `serve()`, `board-cli.test.mjs` pins the build payload
- [x] Fixed the one existing test that assumed a garbage `--port` fell through to `EADDRINUSE`'s default-marking message — it now dies before ever reaching `listen()`, so that test now drives the true "flag not given" case instead
- [x] `588/588` — full `skills/fleet/scripts` suite green after rebase onto `origin/main`

Closes #366


feigi:
## Review applied — `58bbd6a`

All 21 findings adjudicated. Suite: **590/590** (`node --test skills/fleet/scripts/*.test.mjs`, run bare from the worktree), up from 588.

**Three defects fixed** (the 6 `survived` findings collapse into these):

- **`--port 0` was unusable, and this PR is what made it reachable.** `argPort()` deliberately permits 0 so `listen(0)` binds an ephemeral port, but `serve()` echoed the *requested* port into both the startup line and `--open` — so the operator was told `http://localhost:0` (curl exit 7, connection refused) while the board sat on a kernel-picked port nobody was told. Now reports `server.address().port` from inside the listen callback. The PR's own test asserted `/cockpit on http:\/\/localhost:0/`, pinning the broken message as expected; it now pins the opposite.
- **`--interval` had no upper bound** while its sibling `argPort()` bounds both ends — the asymmetry #366 exists to remove. Above 2147483s the ms delay overflows `setInterval`'s 32-bit field and Node clamps to 1ms, so `--interval 3000000` (34 days) became a spin loop: 489 stub-`gh` invocations in 8s, 589 in 25s, announced by nothing but a Node warning.
- **`argPort()`'s upper bound was untested** — mutating `n > 65535` to `n >= 65535`, which rejects the legal maximum port, survived the whole suite. Both boundaries of both flags are now pinned.

Every added test was mutation-verified: reverting the bound-port fix, both boundary off-by-ones, and removing the interval ceiling each go red; an unrelated mutation (the `--spend-since` floor) leaves all three green, so they discriminate rather than just fail.

**Also corrected two comments this PR itself added:** `serve()`'s claim that `??` over `||` is what lets a caller-chosen 0 survive as `portGiven` (nothing verifies it — triage scoped the in-process path out, and on the argv path the two operators are identical), and `board-cli.test.mjs`'s reference to `gather()'s own arg("interval")`, which this same PR moved to `argInterval()`.

**Suggestion band (13 findings, 0 refuters by policy) checked and recorded in #438** — 2 applied here, 2 refuted by measurement, 3 answered by the comment narrowing, 6 deferred. Worth flagging from the refutation: the suggested `/^\d+$/` string-shape gate for `Number()`'s hex/exponent acceptance would silently regress `--interval 1.5` while leaving the suite green — `argInterval()` uses `Number.isFinite`, not `isInteger`, deliberately. Do not re-apply it.

`rebase-check` is red at behind=14 — merge-wave staleness, not this diff. Not rebased.

Issue #400, opened 2026-08-11
Every verbatim dispatch blockquote in run-team/SKILL.md is unpinned — deleting one leaves the suite fully green


feigi:
## What

The three blockquotes `run-team/SKILL.md` requires the controller to paste **verbatim** into every implementer prompt are anchored by no test at all. Any of them can be silently deleted, reworded, or hollowed out and the suite stays green.

## Evidence

Measured while gating PR #399 (head `0bfdd2d`), which added the third of them:

- Deleted all 30 lines of the enumerate-the-class block (`sed -i '' '251,280d'`), ran the full suite: **498/498 pass, 0 fail** — byte-identical to the unmutated run.
- `grep` for `Commit incrementally` and for `gh issue view` across `skills/fleet/scripts/*.test.mjs` returns **no pin** for the commit-incrementally block or the read-the-issue block either.
- The nearest existing slice, `implementer-model-tier.test.mjs:45`, ends at `"One named member per ticket"` — it stops immediately **before** the verbatim blockquotes begin.

## Why it matters

These are not prose about the fleet; they are the fleet's actual behaviour. `SKILL.md` itself says the commit-incrementally block "is not optional" and records that a member going idle mid-task leaves its diff unrecoverable — observed twice in one run. A rule whose whole value is that it reaches the member's prompt intact, with nothing asserting it is still there, rots the first time anyone edits around it.

The repo already pins verbatim prompt text elsewhere — `review-path-default.test.mjs` for the fix-applier prompt, `fleet-tick-prose.test.mjs` for the reconcile block — so this is an inconsistency, not a missing convention.

## Constraint on any fix

A positive regex over a whole markdown section is a **vacuous pin**: it passes no matter what the section says. Slice size does the anchoring. Any pin added here must be mutation-tested **both ways** — alter the pinned sentence and confirm red, make a harmless nearby edit and confirm still green.

## Out of scope

Not fixed in PR #399. The gap is pre-existing and wider than the block that PR adds — pinning one of the three and leaving the other two would be the same missed-sibling shape that block exists to prevent.

feigi:
From the PR #440 review (issue #172), which pins two of these blocks — the identity block and the step 7 PR handoff. Full record in #445; this is the half that bears on **how** #400 should be fixed, measured rather than argued.

## The pins that exist have a ceiling, and it survives whatever you add here

[...]

Issue #257, opened 2026-08-10
fleet-trust-the-label spec:212 quotes sizing-a-ticket's pre-compression wording verbatim — mark it dated, or rewrite it?


feigi:
Deferred from PR #248 review. **dimension:** `comments` (severity `suggestion`, unverified — the workflow budgets that band 0 refuters).

**file:line:** `docs/specs/2026-07-30-fleet-trust-the-label-design.md:212`

## Finding

The spec's rationale text at `:212` reads *"the skill no longer decides admissibility for anyone"*. That was an exact quote of `skills/fleet/skills/sizing-a-ticket/SKILL.md`'s own sentence. PR #248 compresses that sentence to *"no admissibility call here, for anyone"*, so the spec's quote no longer matches the file it quotes.

Verified against the PR head snapshot:

```
$ grep -rn "no longer decides admissibility" .
docs/specs/2026-07-30-fleet-trust-the-label-design.md:212
```

The phrase now exists **only** in the spec, not in the skill it describes.

**Meaning is preserved — this is verbatim-quote drift, not a factual error.** Cross-checked `skills/fleet/skills/run-team/SKILL.md:89-92` (untouched by PR #248): it independently states `sizing-a-ticket` picks the *process path* only, so the spec's underlying claim stays accurate.

## Why deferred

Out of PR #248's scope by #66's Agent Brief, which lists as out of scope: "Compressing the design specs, which are human-read and exempt". #248's acceptance criteria also require "No prose outside #59's additions is modified" — the spec is not a #59 addition. Editing it to chase the quote would have violated the ticket #248 was implementing.

The compression itself is what #66 asked for and is correct; only the spec's copy of the old wording is now stale.

## Why `needs-triage` and not `ready-for-agent`

Defect verified, **approach open** — and the open question is a policy one about these specs, not an implementation detail:

- if `2026-07-30-fleet-trust-the-label-design.md` is meant to stay a literal changelog of the skill's wording, update the quoted phrase to match; or
- if it is a dated historical design record (which is how these specs mark superseded content elsewhere), leave it — it correctly quotes the skill *as it stood on 2026-07-30*, and rewriting it falsifies the record.

The finding's own `suggested_fix` opens with "No action required for correctness." Same question #117 and #48 raise for the line-number citations in this same file: **mark, or rewrite?** Settle it once for the class rather than per-quote.

feigi:
[...]

PR #440, opened 2026-08-13
fix(fleet): a member is told it is unattended, and gets the sizing rule, verbatim

### LINKED ISSUE #172: run-team phase 2: the member's route to sizing-a-ticket is paraphrase-dependent, and nothing tells it it is a fleet member
Deferred from PR #168 review. **All pre-existing** — verified not introduced or worsened by that PR, whose own change is confined to `:170-172` and `:184-188`. Two independent specialists and the reviewer each reached this separately.

PR #168 fixed exactly this class of defect for the bail rule (moved it into the pasted block). The same defect remains one sentence later, and the commit's own rationale is the argument for fixing it: *"only the block reaches the member as written; the paragraph reaches it only if the controller paraphrases."*

## The problem

`skills/fleet/skills/run-team/SKILL.md:159` says the controller carries the `>` blocks **verbatim**. There are three (`:161-163`, `:165-172`, `:174-176`). The instruction to run `sizing-a-ticket` is **not** in any of them — it is at `:184`, in controller-facing prose:

> A member that does not bail runs `sizing-a-ticket` for the process path and proceeds on **either row** …

So the member reaches `sizing-a-ticket` only if the controller paraphrases that line. The alternate route is effectively closed: `next-ticket` §6 (`skills/fleet/skills/next-ticket/SKILL.md:70-72`, "Size the ticket, then pick a path") does say "Run `sizing-a-ticket`", but `run-team:190` sends the member to `next-ticket` **step 7**, not step 6.

## Why it bites

`skills/fleet/skills/sizing-a-ticket/SKILL.md:18` is **conditioned**:

```
**Fleet member on heavy: enter at `superpowers:writing-plans`.**
```

and `:21-22` routes the other way — *"Solo session has a user: run the full path"* → `superpowers:brainstorming` → its `<HARD-GATE>` → waits on maintainer approval no unattended member can reach → stall.

Nothing in run-team's three verbatim blocks tells the member it is a fleet member or that it is unattended. The only signal is the incidental word *controller* at `:175`. A member that self-identifies as a solo session takes the branch that stalls.

This became load-bearing when PR #168 (correctly) deleted run-team's local copy of the heavy-row rationale and delegated to `sizing-a-ticket`. Before, the copy was in the same non-pasted prose, so the member never got it verbatim either way — **net change zero, which is why this is not #168's regression** — but the delegation now rests entirely on a condition that may not fire.

## Also in the same region

`run-team/SKILL.md:188` — "Selection and claiming are done; the member starts there." **"there" has no named antecedent.** Since #168 removed `superpowers:writing-plans` from this paragraph, the nearest place-like phrase is "the fleet's heavy-row entry point" (`:186`) — a binding that would wrongly send *light*-row members to `writing-plans`. The intended target is most likely `next-ticket` §6, since §1-5 are exactly selection and claiming. The pre-image had the same hazard (bare "start there" with `writing-plans` three sentences back), so again **not a regression** — but it is now unfixable by inference alone, and the two reviewing specialists read it differently, which is itself the evidence that it is ambiguous.

The defect is verified; the **approach is not decided**. Open choices:

- add an explicit "you are an unattended fleet member" line to a verbatim block, **or** make `sizing-a-ticket:18` unconditional for the fleet, **or** have the controller pass the row it already knows;
- move the `sizing-a-ticket` instruction into a pasted block vs. leave it and fix the condition;
- name step 6 explicitly at `:188` vs. restructure the handoff so only one place points into `next-ticket`.

Note `run-team:184` and `next-ticket` §6 both independently say "run `sizing-a-ticket`" — arguably a third instance of the duplication #62 was opened about, which may argue for one route rather than a patch.

Deferred from PR #168 review.

feigi:
`run-team` phase 2 carries a fixed set of `>` blocks to each member verbatim; everything else in that section reaches the member only if the controller paraphrases it. The instruction to run `sizing-a-ticket` lived in that non-carried prose, and nothing anywhere told the member it was a fleet member at all.

That mattered because `sizing-a-ticket`'s fleet entry point is conditioned on the reader being one. A member that reads itself as a solo session takes `superpowers:brainstorming`, whose `<HARD-GATE>` withholds every implementation action until a human partner approves — and parks on an approval no unattended member can obtain.

## What changed

`skills/fleet/skills/run-team/SKILL.md`, +30/-11, one file plus a new test:

- **an identity block** — unattended fleet member, no maintainer reachable, no approval gate will clear
- **the sizing rule and the `next-ticket` step 7 handoff moved into a carried block**, so they arrive as written rather than by paraphrase
- **the handoff names `next-ticket` step 6**, replacing "the member starts there", whose nearest place-like antecedent was the heavy-row entry point a light-row member must not follow

`sizing-a-ticket` is untouched: its branches stay conditioned, and a solo session still runs the interactive path. The demote half of the old sentence stays with the controller — a member that bails demotes nothing itself.

## Enumerating the class

The class is *member-facing instructions in phase 2 that reach the member only by paraphrase*.

| Member of class | Call |
|---|---|
| Member's identity | covered |
| Sizing instruction + "either row" | covered |
| Handoff with no antecedent | covered |
| `next-ticket` step 7 handoff — named in passing by the body, same paraphrase dependence, and it carries "never merge" | covered |
| "bail and demote" — `demote` is the controller's action | covered by splitting: member bails and names the cause, controller demotes |
| `sizing-a-ticket:18`'s conditioned entry | deliberately left — unconditioning it is the rejected fix; a test pins it stays conditioned |
| `next-ticket` §6's duplicate sizing instruction | deliberately left — #62 owns it; both copies stay correct |
| Fix-applier prompt blocks, which also lack an identity line | deliberately left — a fix-applier never runs `sizing-a-ticket`, so the stalling branch is unreachable for it |

## Verification

`skills/fleet/scripts/member-prompt-prose.test.mjs` pins **location**, not vocabulary: moving these words back out into prose is the defect, and every word would still be in the file. Its first assertion is that no non-blank line in the carried region lacks a `>`.

Seven mutations, each red on the pin named for it: sizing rule unquoted, identity block deleted, destination back to a pronoun, destination bound to the heavy-row entry, never-merge dropped, sizing entry made unconditional, solo interactive path deleted.

The accept side, since a pin that only sees input it must reject pins nothing: an unrelated legitimate edit to phase 2's controller prose stays green, and the last test asserts `sizing-a-ticket` still carries both its fleet condition and its solo path — it fails if #172 is ever "fixed" by deleting the distinction.

`./agent-test skills/fleet/scripts` — **609 pass, 0 fail**, re-run after rebase on `origin/main` (`09f9914`).

Closes #172


feigi:
## Review findings applied — `0ed2b1b`

**Applied (1).**

- **`SURVIVED-1`** *(important, `tests`, verified 2–0)* — `member-prompt-prose.test.mjs`: tests 4 and 5 asserted `**step 6**` and `**step 7**` were present *independently*, so swapping the two step numbers in `run-team/SKILL.md` left the whole suite green while telling a member to run the PR/merge steps at the sizing checkpoint and vice versa. Each number is now bound to its adjacent action in one contiguous span.
- **`REFUTED-1` residue** — one shared file-header comment naming the presence-pin ceiling, covering both this site and `UNVERIFIED-3`'s.

**Deferred (5) — recorded in #445, none of it open work here.**

| Finding | dim | Outcome |
|---|---|---|
| `REFUTED-1` | `tests` | refuted as a defect (the location control fails, `pass 5, fail 1` — the test does pin what it was written to pin); residue applied above |
| `UNVERIFIED-1` | `correctness` | refuted — its only proposed action was a PR-body callout, and this body already carries it twice |
| `UNVERIFIED-2` | `correctness` | refuted — same; the "bail and name the cause" rewording is already called out in prose and in the class table |
| `UNVERIFIED-3` | `tests` | refuted — the proposed `doesNotMatch` tightening does not kill its own mutant, four synonyms walk around the broadened form, and it reds on legitimate prose. Ceiling recorded on **#400** |
| `UNVERIFIED-4` | `comments` | out of this PR's diff (`docs/specs/2026-07-30-…:202`); same class as **#257**, commented there rather than duplicated |

The three in-scope `suggestion`-band findings each got one refuter, biased to refuse and required to verify by running; all three came back refuted with running evidence.

**Verification.** `node --test skills/fleet/scripts/*.test.mjs` → **`tests 609, pass 609, fail 0`**. Mutation-verified both ways: the step-number swap reds exactly the two tightened tests (`pass 4, fail 2`) where the previous bare pins gave `pass 6, fail 0`; two unrelated legitimate rewordings inside the same block stay green. Committed with hooks enabled, no `--no-verify`.

PR #441, opened 2026-08-13
fix(fleet): rebase server-side before merge, drop local force-push

### LINKED ISSUE #149: merge bot cannot rebase: git push --force-with-lease is denied by the permission classifier, so every rebasing wave merges a stale head
Hit twice in one `/fleet:run-team` session, on two different merge bots, with identical symptoms. This is a blocker for the merge path, not a defect in any script.

## Symptom

`git push --force-with-lease` → **"Blocked by classifier"** (Claude Code auto mode). Not a lease race, not a stale-lease rejection — a local permission denial before the push is attempted.

Consequence: the bot cannot land its rebase, so `gh pr merge` merges the **stale reviewed head**, and `prove-merge.sh` correctly reports failure:

```
{"preIsAncestor":true,"postIsAncestor":false,
 "secondParent":"32dc63e…","firstParent":"4e9538e…",
 "parentCount":2,"proofPath":"rebase","headWasCurrent":true,"proved":false}
```

The second parent is the pre-rebase head because the rebased commit never reached the remote.

## Frequency and scope

| wave | PRs | behind at merge | push | result |
|---|---|---|---|---|
| 4 | #130, #136 | 22, 18 | denied | both merged unrebased, `proved=false` |
| 5 | #131 | 13 | denied | merged unrebased, `proved=false` |

Waves 1–3 merged five PRs with `proved=true` — those either needed no rebase or landed one before the block appeared. **Not universal:** PR #131's *reviewer* force-pushed that same branch successfully earlier the same day, so the denial is not a blanket rule on the command.

## Why it matters

`run-merge-bot.md` treats the ancestry proof as the artifact that makes "nothing in main was undone" checkable after the fact. Without a landed rebase that artifact is structurally unobtainable, and the merge-time currency gate degrades from *proved* to *argued*.

## What the bots did instead — and it held, twice

Both substituted, and both flagged the deviation rather than burying it:

1. rebase locally onto `origin/main`, confirm no conflicts and 0 behind after
2. review `git diff origin/main...HEAD` hunk by hunk — only the PR's own changes, nothing of main's undone
3. suite green on the rebased head
4. `merge-tree` clean, `MERGEABLE`/`CLEAN`, label present at the merge instant

Wave 5 added a stronger check worth adopting regardless: **merged `origin/main` compared byte-identical to the rebased tree it had tested** (`git diff --stat origin/main <rebased-sha>` empty). That is content-level proof the graph proof is missing.

The controller verified both waves independently afterwards — merged main green (182/181 then 194/193), and specific earlier gains confirmed intact (`jstr` from #110, `dropSpecs` from #116, the stash-gate paragraph in `run-merge-bot.md` that #131 prepends to rather than replaces). **Nothing was undone in either wave.** The risk was real but did not land.

Wave 5's bot also declined to reach for an alternate flag spelling to get past the classifier, on the grounds that it would be *"evading the intent, not accomplishing the task."* That judgement is correct and should stay correct — the fix belongs in configuration, not in a workaround.

## Options

1. **A Bash permission rule for `git push --force-with-lease`** on the fleet's branches. Smallest change, restores the artifact. Maintainer's call — an agent must not edit permission config, and this ticket does not ask one to.
2. **Rebase-free merge path.** If the bot cannot push, have it merge and then verify content equivalence (wave 5's byte-identical check) as the *documented* gate rather than a fallback. Honest, and removes the dependency — but loses the graph proof permanently.
3. **Refuse to merge when the push is denied and a rebase is needed.** Safest, and stalls the queue on a config issue; would have blocked 3 of 9 PRs this run.
4. **`--force` with a freshly re-derived lease check in the script.** Rejected as written — `--force-with-lease` exists precisely to make the race safe, and dropping it to dodge a classifier trades a real safety property for a permission workaround.

Option 1 plus adopting wave 5's byte-identical check as a standing step is the combination I would suggest; option 2 alone is the honest description of what the fleet does today.

## Where to record the fallback

Whichever is chosen, `run-merge-bot.md` currently has no branch for "the rebase could not be pushed". Both bots invented the same substitution independently and correctly, which is the signal it should be written down rather than re-derived.

feigi:
## Summary

`run-merge-bot.md` step 1 rebased a behind PR locally and pushed with
`git push --force-with-lease`. That push has no `autoMode.allow` entry, so
the auto-mode classifier judges it per invocation and intermittently denies
it (2 denials / 3 successes measured on identical commands in one session,
per the issue thread) — the rebased head never reaches the remote, and
`gh pr merge` lands the stale one, which `prove-merge.sh` correctly reports
as unproved.

Step 1 now rebases **server-side first**: `gh pr update-branch <pr> --rebase`
is an API call, not a push, so the classifier is never consulted for it.
`prove-merge.sh` needed no change — verified directly (see below), not
assumed: it reads only the object graph (ancestry + merge parents), never
how a head reached the remote, so the existing rebase proof path already
covers this shape.

Where the server-side rebase can't run (a real conflict, or
`allow_update_branch` off), step 1 falls back to a local, **unpushed**
rebase used purely for verification — no-undo audit, hunk-by-hunk diff,
suite green, mergeability recheck at the merge instant — then merges
whatever's actually on the remote and reports the deviation as
`rebase-fallback-#<pr>`. That merge is *expected* to disprove
(`prove-merge.sh` correctly returns `proved=false`, pinned by the existing
`ATTACK: un-rebased head that merged cleanly still proves false` test);
step 4 now says so explicitly and requires reporting it as **argued**, never
as **proved**.

Also documents a real, previously ad-hoc-corrected mistake (PR #408): after
a server-side rebase, a stale local worktree is proven safe with
`git cherry origin/<branch> HEAD`, never `origin/main` — `main` never
contains an unmerged PR's commits either way, so it trivially reads `+` and
proves nothing.

## What was verified, not assumed

- Ran `prove-merge.sh` directly against a local fixture that mimics the
  platform-performed-rebase shape (one clone plays the caller and never
  pushes; a second clone plays GitHub, independently rebasing the PR branch
  onto an advanced `main` and pushing the result — the caller only polls and
  later merges what's on the remote). Result: `{"proved":true,
  "proofPath":"rebase","headWasCurrent":true,"preIsAncestor":false,
  "postIsAncestor":true}`, exit 0.
- Ran the fallback shape against the same fixture technique — matches the
  pre-existing `ATTACK: un-rebased head that merged cleanly still proves
  false` case in `prove-merge.test.mjs`: `proved=false`, exit 1, by design.
- Verified the `git cherry` upstream claim directly: from a stale
  pre-rebase worktree, `git cherry origin/<branch> HEAD` → `-` (nothing
  unique); `git cherry origin/main HEAD` on the identical HEAD → `+`
  (proves nothing, reads like divergence).

## Deliberately out of scope

- `next-ticket.md` step 7's own `git push --force-with-lease -u origin HEAD`
  — a sibling force-with-lease call site, but structurally different: no PR
  exists yet at that point (so `gh pr update-branch` can't apply), and
  there's no ancestry-proof/currency-gate concept to protect (single-author,
  unreviewed branch). The brief's Key Interfaces scope this ticket to "the
  merge bot's rebase-and-merge sequence," not PR creation.
- A separate, unrelated classifier-denial risk on `gh run watch` (step 2's
  CI-wait command) found in an unrelated memory during this work — different
  command, different root cause, not named anywhere in this issue's thread.
  Left untouched.
- Changing `autoMode.allow` / permission config itself — explicitly out of
  scope per the Agent Brief.
- Weakening the ancestry proof or the currency gate, or merging without the
  label — none of the above touch those.

## Test plan

- [x] `./agent-test skills/fleet/scripts/*.test.mjs` — 608/608 pass, 0 fail
- [x] Two new prose tests pin the new step 1 content and step 4's
      "expected to disprove" language, mirroring the file's existing
      convention (`run-merge-bot-prose.test.mjs`)
- [x] `prove-merge.sh` itself run against fixtures for both the primary
      (server-side rebase) and fallback shapes — see above

Closes #149


feigi:
## Review applied — `6b7bde5`

**Applied (4):**

- **`run-merge-bot.md:54` — step 1's poll loop could not terminate** *(critical, correctness)*. `until [ "$(gh pr view … headRefOid)" != "$pre" ]; do sleep 5; done` never checked `gh pr update-branch`'s exit status and had no cap — so the head *not* moving, which is exactly the failure case, spun forever and the doc's own escapes ("`pre` and `post` are simply equal", "Either → the fallback below") were unreachable. Now checks `rc` before polling, caps at 60 tries, and names which of the three outcomes routes to the fallback. Verified with a stub `gh`, snippet extracted verbatim from the doc: exit 1 → 1 `gh pr view` call total, no polling, terminates in 0.12s; exit 0 with a head that never moves → hits the cap and stops; head moving on the 2nd poll → breaks early. The old snippet was still spinning when killed in both of the first two states.
- **`run-merge-bot.md:49` + `run-merge-bot-prose.test.mjs:83` — "no `autoMode.allow` entry" was false against this repo's own tree** *(critical, comments)*. An entry for exactly `git -C <worktree> push --force-with-lease origin <branch>` sits at `settings.json:197`, added by `1dadc5d` (2026-08-11). That is the damaging kind of false claim: it sends a future fixer to add an entry that already exists. Corrected at both sites. Note the dates, since they narrow the fix: the 2-allowed/2-denied measurement is **2026-08-10**, so it *predates* the entry — the honest statement is that the classifier judges the command per invocation and an allow entry is not the missing fix, not that the denials happened despite one.
- **`run-merge-bot.md:77` — the fallback could not reach a mergeable state here, and never said so** *(important, correctness)*. It leaves the remote head behind `origin/main` on purpose, which is the exact condition `rebase-check` fails on, and `rebase-check` is an expected job. Verified live on this PR rather than reasoned: `behind=4`; `ci-state.mjs --pr 441` → expected jobs `rebase-check, check`, `rebase-check: completed/failure`, `verdict=not-green`; `gh pr view 441 --json mergeable,mergeStateStatus` → `MERGEABLE`/**`BLOCKED`** — never the `CLEAN` the fallback's own gate demands. The doc now says the path ends blocked and handed back, and that step 4's merge is reachable from it only where no expected job gates on currency.
- **`run-merge-bot-prose.test.mjs:128` — the new pin matched the whole 205-line file**, which this file's own header forbids ("a positive match on a whole file is a vacuous pin"). Bounded to the Report line. Mutation-checked three ways: deleting the token from the Report line → red; deleting it **and** planting the identical literal elsewhere → still red (the old whole-file pin went **green** on exactly that); an unrelated reword in step 2 → green.

Suite from the worktree: `node --test skills/fleet/scripts/*.test.mjs` → **tests 608, pass 608, fail 0**.

**Deferred (7), filed as #447 and #448.**

#447 is the one live defect: the `git cherry` probe at `run-merge-bot.md:60` gives no instruction for a `+` result on the server-side path, and the `worktree-diverged-#<pr>` STOP now lives only in the fallback. `needs-triage` — the approach is genuinely open, because the primary path may be structurally unable to act on a `+` at all. The stronger version of that finding (that the check is vacuous without a `git fetch`) was **refuted by measurement** and is recorded as such so it is not re-raised.

#448 aggregates the `suggestion` band, which the review workflow budgets 0 refuters by policy. All three in-scope items were dispatched to a refuter biased to refuse and **all three came back refuted** — two of them with suggested fixes that would have made the file worse: one would have broken 4 tests for no correctness gain, and one would have **inverted** the ticket attribution it claimed to fix. One live item remains there: `settings.json:197`'s allow-entry justification ("a merge bot uses it only to land a rebase it just performed") no longer matches a workflow whose only local rebase is verification-only. Not touched here — config is out of this PR's diff and off limits to the fix-applier.

Two further findings claimed 5 suite failures in `skills/fleet/scripts/member-prompt-prose.test.mjs`. **That file does not exist in this PR's tree** — they are review-snapshot contamination, already filed as #446, and the 608/608 run above confirms it.

**Not rebased.** The branch is 4 behind after #440 merged, so `rebase-check` is red from staleness, not from this diff. `check` is green. The merge bot establishes currency once, at merge time.

PR #442, opened 2026-08-13
fix(fleet): candidates.mjs header names gh, not the API, as what applies --jq

### LINKED ISSUE #201: candidates.mjs:6 calls the --jq reduction "server-side" — the same file says at :105 it runs inside gh
`skills/fleet/scripts/candidates.mjs:6` (header comment):

> The body is reduced **server-side** to dependency references only.

That is false, and the same file contradicts it 99 lines later at `:105`:

> The reduction runs **inside gh, not here**, so nothing at this end guarantees it applied — an older gh, an expression it rejects, an error object from a proxy.

`--jq` is applied by `gh` in its own process, with an embedded gojq — the GitHub API never sees the expression. Measured:

```
$ go version -m "$(which gh)" | grep gojq
	dep	github.com/itchyny/gojq	v0.12.17
```

Why it is not cosmetic: `:105`'s fail-closed check exists *because* the reduction is client-side and can silently not apply. A reader who believes `:6` has no reason for that check to exist, and the header is the first thing read.

## Scope

- `skills/fleet/scripts/candidates.mjs:6` — untouched, this ticket.
- `skills/fleet/scripts/candidates.test.mjs:6` carried the same word; **PR #200 (issue #63) removes it there**, so this ticket must not re-fix that file. Verify at merge time rather than trusting this line.

Scope searched: `grep -rn 'server-side' skills/fleet/scripts/` — two hits, both named above.

Found by `impl-63` while implementing #63; controller reproduced `:6`, `:105` and the grep independently before filing.

feigi:
Closes #201

## What

`candidates.mjs`'s header called the body reduction *server-side*. It is not: `gh` applies `--jq` in its own process with an embedded gojq, and the GitHub API never receives the expression. The same file already stated this correctly where it checks the reduction's result, so the header contradicted its own guard — and the header is what a reader reaches first.

The corrected header keeps the causal link the guard depends on. A reader of the header alone now expects the rows to be checked, rather than finding the check unmotivated.

## Enumerate the class

The ticket's acceptance grep is `grep -rn 'server-side' skills/fleet/scripts/`, which sees exactly one hit. Repo-wide, the word appears five times. All five were checked against the tree, not the issue body:

**Fixed**

- `skills/fleet/scripts/candidates.mjs` header — the ticket. False: the reduction is client-side.
- `skills/fleet/skills/next-ticket/SKILL.md` red-flag bullet — the sibling, one directory outside the acceptance grep. It read `server-side --search + --jq`, one adjective governing both conjuncts. Only `--search` is server-side.

**Left, verified true**

- `next-ticket/SKILL.md`, the "Script owns the query" line — `server-side` attaches to the label exclusion alone, and the label really does travel as a `label:` qualifier inside `--search`.
- `docs/specs/2026-07-22-run-team-agent-fleet-design.md` — "the server-side label exclusion and the `--jq` body reduction": the adjective sits inside the first noun phrase, the second carries its own determiner, so it does not distribute.
- `docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md` — describes a `--search` substring search, a different subject entirely.

Also considered and left: the "never fetches raw bodies" line above the corrected one. Its subject is this script, which genuinely never receives them, and the row-shape refusal corroborates that reading with "what this file refuses to fetch in the first place". It is true in the post-image and the ticket did not raise it.

The header names `gh` and the locality but deliberately not gojq. The engine is #204's ground and already documented where the regex restriction needs it.

## Verification

- `grep -rn 'server-side' skills/fleet/scripts/` → no hits, exit 1.
- Full suite via the isolation runner, `./agent-test skills/fleet/scripts`: **603 pass, 0 fail, 0 skipped**. Run with both commits in the tree.
- No test file touched; no runtime behaviour, `--jq` expression, or guard changed. Comments only.

No new predicate is introduced, so this change has no false-positive class to pin — the regression it invites is the word returning, which the acceptance grep already catches.


feigi:
Review adjudication — **no-op, HEAD unchanged at `a19a6421`.** Nothing survived; nothing applied.

| Finding | file:line | dim | Outcome |
|---|---|---|---|
| REFUTED-1 | `candidates.mjs:5` | correctness | refuted 2-0, upheld — `fetch` means *into this process* here, a convention predating this PR (`candidates.mjs:281`, untouched) |
| UNVERIFIED-1 | `next-ticket/SKILL.md:18` | correctness | **refuted** — `server-side` there attaches to label exclusions, which are server-side (measured in the GraphQL variables); the sentence opens `Script owns the query:`, so 2/2 cold readers bound `never fetched` to the script |
| UNVERIFIED-2 | `next-ticket/SKILL.md:98` | correctness | **refuted** — #201's `Scope searched:` is a grep receipt, not a prohibition list; `:98` was a literal instance of the claim #201 names |
| UNVERIFIED-3 | `candidates.mjs:5` | silent-failure | **refuted** — duplicate of REFUTED-1; its `:18` half fails with U1 |
| UNVERIFIED-4 | `candidates.mjs:7` | comments | **refuted** — the header carries *what/where*, the guard at `265-269` carries *why that `die()` exists*; the proposed shortening deletes the file's only correction and regresses toward `origin/main`'s `server-side` wording |

U1 and U2 proposed **mutually exclusive** remedies for one file, so they went to a **single** refuter. Verdict: **NEITHER** — merge as-is. `:18` is a narrower residue than #201's, it changes no action for that file's reader, and #201's actual harm (fail-closed row checks reading as redundant) lives in `candidates.mjs`, where lines 6-8 measurably closed it: 2/2 control readers answered wire = YES, subject = the script.

**Every proposed `suggested_fix` was measured worse than the untouched text** and none was applied — one mints a false clause (`never returned`; GitHub does return them) and contaminates the following clause, one inverts the binding it meant to fix, one regresses this PR's own correction.

Residue recorded on #444 (kept open, `needs-triage`, premise corrected) with a verified replacement line, its +31% cost, and the finding that nothing tests that file: [#444 comment](https://github.com/feigi/claude-config/issues/444#issuecomment-5280547317). No new issues filed.

`node --test skills/fleet/scripts/*.test.mjs` from the worktree at `a19a6421`: **`tests 603 / pass 603 / fail 0`** — plus 603/603 in three scratch mutants (each proposed wording, and `SKILL.md:18` deleted outright), which is how "nothing tests that file" was established.

Issue #381, opened 2026-08-11
reap.sh leaves the worktree behind when a server-side rebase detached it — the branch is reaped, the directory is not


feigi:
Found live during a `/fleet:run-team` merge wave, immediately after PR #373 merged. Not hypothetical — it happened, and the controller cleaned up by hand.

## What

`reap.sh` removes a merged PR's worktree **and** its `[gone]` branch. Worktree removal is keyed on the worktree being associated with that branch. A worktree at **detached HEAD** has no such association, so the branch is reaped and **the directory is silently left on disk**.

Measured, same invocation shape as every other wave:

```
$ git worktree list
.worktrees/79-issue-tracker-brief-respec  dc992bc (detached HEAD)

$ reap.sh                      # dry run
    would reap docs/79-issue-tracker-brief-respec
{"applied":false,"reaped":["docs/79-issue-tracker-brief-respec"],"kept":[]}
```

Compare the same command one wave earlier, on a worktree still on its branch — it names both:

```
    would remove worktree /Users/chris/.claude/.worktrees/370-out-of-scope-kb
    would reap docs/out-of-scope-kb
```

No `would remove worktree` line, no `kept` entry with a reason. The worktree is not refused, it is **not considered at all**, so the silent-cap rule the sibling scripts follow does not fire either.

## Why the worktree was detached — this is now the normal path, not an oddity

The merge bot has switched to `gh pr update-branch <n> --rebase` (server-side rebase, issue #149: it avoids the `--force-with-lease` the permission classifier denies). That rewrites the branch **on the server**, so the local worktree must be moved to the new head. `git reset --hard` is denied by the auto-mode classifier; `git checkout --detach <sha>` is permitted and is what the bot used.

So: **every PR merged via the server-side path leaves a detached worktree**, and every one of them will be missed by `reap.sh`. Two of this run's three merges used that path.

## Why it matters

This is the silent queue-shrink the fleet already knows about, reached from a new direction. A stale worktree still answers `git worktree list`, and `inflight.sh` probe 3 reads a worktree as a live claim — so an already-merged ticket reads as **taken** in the next phase-0 scan, and the candidate queue quietly shrinks. It compounds per wave, and nothing reports it.

## What was verified before cleaning up by hand

```
git -C .worktrees/79-issue-tracker-brief-respec status --porcelain   -> 0 lines
git merge-base --is-ancestor dc992bc origin/main                     -> YES
[...]

PR #450, opened 2026-08-13
fix(fleet): release-ticket's stray guard no longer claims a branch mismatch it never observed

### LINKED ISSUE #179: release-ticket.sh: the stray guard claims a branch mismatch it never observed when git cannot resolve the worktree's HEAD

feigi:
## Summary

When git cannot resolve a stray worktree's HEAD, `release-ticket.sh`'s stray
guard fell into the branch-mismatch `else` and told the operator the
worktree was on some other branch — an assertion it never observed, and one
that is false whenever the worktree is in fact on the claim's branch.

Four distinct routes produce the identical porcelain shape (null OID, no
`branch` line), measured on git 2.50.1 (Apple Git-155): `chmod 000` on the
admin `HEAD`, garbage content in it, a dangling symlink in its place, and a
directory in its place. Two of those four also print a spurious `detached`
line, so the fix keys on **null OID together with the absence of a `branch`
line** — never on `detached` — per the two false candidates the ticket's
triage history ruled out.

- New arm: `locked` → `gone` (directory-gone) → **`unresolved_head`
  (new)** → the existing branch-mismatch `else`, reading only the porcelain
  the script already holds (no extra `git` call).
- `block`, never `die`, matching every sibling arm.
- `gone()`'s contract is untouched.

## Enumerate-and-declare

**Covered** (all four routes to the null-OID/no-branch state, one built as a
fixture, three named in comment per the brief's economy directive):
- garbage content in the admin `HEAD` (fixture)
- `chmod 000` on the admin `HEAD` (named in comment; smoke-tested by hand, not built as a suite fixture)
- dangling symlink in `HEAD`'s place (named in comment; smoke-tested by hand)
- `HEAD` replaced by a directory (named in comment; smoke-tested by hand)
- directory-gone-AND-HEAD-unresolvable ordering (gone must still win) — tested
- locked-outranks-everything ordering — untouched, still first arm

**Deliberately left / must-ACCEPT, not swept into the new arm** (both named
explicitly in the ticket's final brief):
- a genuine detached checkout (real sha + `detached` line) — still falls
  through to the unchanged branch-mismatch `else`; both pre-existing pinned
  tests for this are untouched
- an unborn branch (`git worktree add --orphan`: null OID **with** a real
  `branch` line) — new test added; must not be swept into the new arm since
  it's a legitimate worktree, not a fault

**Out of scope per the brief** (unchanged by this PR):
- the original unstattable-directory framing this issue started from — ruled
  no-change in the ticket's final triage: both existing fixtures for it are
  genuinely detached, so `is not on $branch` is true there and stays as-is
- `gone()`'s contract
- #208's `Indeterminate` state (the `blocked_by` edge was dropped by the
  ticket itself; #208 owns that now)

## Test plan

- [x] `node --test skills/fleet/scripts/release-ticket.test.mjs` — 62/62 green (4 new)
- [x] `./agent-test skills/fleet/scripts` — full suite, 617/617 green
- [x] Hand-verified against all 4 measured HEAD-broken shapes plus the 2
      must-ACCEPT shapes (genuine detached, unborn branch) by running the
      script directly against constructed fixtures — matches the acceptance
      table in the ticket exactly

Closes #179


feigi:
## Review deferrals filed

All findings from this PR's review are accounted for. Applied here: `survived[1]` (comment miscount), the merged `survived[2]`+`survived[4]` sibling test, and three `simplify`/`comments` suggestions whose refuters survived. Everything else is filed.

| Finding | Dimension | State | Where it went |
|---|---|---|---|
| `survived[3]` | silent-failure | survived (2 refuters) | **#453** `needs-triage` — a MOVED worktree with an unresolvable HEAD releases silently, rc 0. Pre-existing; the fix adds a new blocking path, which #179 puts out of scope. |
| `unverified[1]` | correctness | own refuter **died mid-run**; claim independently measured by another | **#458** `ready-for-agent` — lock-vs-unresolved-HEAD precedence unpinned in both directions. |
| `unverified[2]` | correctness | suggestion, never measured end to end | **comment on #381** — `reap.sh`'s branch-keyed lookup shares the fault class, same silent skip, different trigger. |
| `unverified[3]` | silent-failure | suggestion | **#454** `needs-triage` — `unresolved_head` in `elif` position cannot tell a dead helper from an honest false. |
| `unverified[4]` | silent-failure | suggestion | **#455** `needs-triage` — the blocker names `$wtroot` but not which entry; the cheap `${stray##*/}` variant is measurably wrong after `git worktree move`. |
| `unverified[5]` | tests | positive finding | No action — confirms the three new tests discriminate. Nothing to file. |
| `unverified[6]` | tests | suggestion | **#456** `ready-for-agent` — the symlink-over-HEAD and directory-in-HEAD shapes are named in the comments but never built as fixtures. |
| `unverified[7]` | comments | suggestion | **#457** `needs-triage` — `reaping.md:48` frames the stray remedy as a binary; there are four arms. |
| `unverified[8]` | types | suggestion, no known trigger | No action — the finding's own verdict is "nothing here blocks the PR". |
| `unverified[9]` | simplify | **REFUTED by measurement** | **#459** `wontfix` — four mutations red exactly one test with those negatives and survive all 63 without them. Do not delete them. |
| `unverified[10]` | simplify | refuter survived | Applied — `artefacts(r, c).branch`, one line. |
| `unverified[11]` | simplify | refuter survived | Applied — the test comment stops carrying a second copy of the four-route list. |

`node --test skills/fleet/scripts/*.test.mjs` → `tests 618 / pass 618 / fail 0` at `24411c5`.

feigi:
## Update — two controller rulings applied in `fda2041`

Amends the deferral table above.

**`unverified[1]` — no longer deferred. Applied and its issue closed (#458).** #179's brief lists "a locked stray worktree still gets the unlock remedy first" as an acceptance criterion, and nothing could detect it being violated:

```
MUTATED LINE 393:   if locked "$stray" && ! unresolved_head "$stray"; then
  -> tests 63 / pass 63 / fail 0   SURVIVES the whole suite
```

Fixture added — the stray is locked AND its admin HEAD is garbage, directory in place. The lock has to win: `git worktree prune` skips a locked entry at rc 0 and `remove` rejects one, so repairing a HEAD file resolves nothing until the operator also unlocks. Verified `64 / 63 / 1`, only the new test, plus a control mutation it correctly does not catch.

**`unverified[9]` — still refuted (#459 stands), but one of its five lines was right.** The `doesNotMatch(/could not read its HEAD/)` in the unborn-branch test is genuinely redundant: the sibling test carries the same assertion. Deleted, after confirming the sibling test still reds under the mutation that needs it:

```
MUTATED LINE 400:     block "... but is not on $branch and git could not read its HEAD — release it by hand"
  -> ✖ a SIBLING's unresolvable HEAD is not this claim's unresolvable HEAD
```

The other four stay — each is the sole detector of a mutation that changes *what remedy* an arm names while leaving *which arm fired* correct, which is all the positives pin.

`node --test skills/fleet/scripts/*.test.mjs` → `tests 619 / pass 619 / fail 0` at `fda2041`.

Issue #383, opened 2026-08-11
candidates.mjs: should `\s`/`\d` be explicit ASCII classes? — the engine-portability half of #204, undecided


feigi:
> *This was generated by AI during triage.*

Split out of #204 during triage. #204 named two materially different builds and could not be actioned while it owned both. This issue owns **option (1)** — the behaviour decision. #204 keeps **option (2)**, the harness extension, which is now decided and briefed.

## The undecided question

`candidates.mjs` builds its `--jq` reduction with `\s` and `\d`. `gh` applies that expression with its embedded **gojq** (Go RE2 — `\s` is `[\t\n\f\r ]`, `\d` is `[0-9]`, ASCII only), while the suite applies it with the **system jq** (Oniguruma — both Unicode-aware). Both engines accept the pattern; they disagree on the input.

Measured (from #204 and its later comment):

```
"## User Stories<U+00A0>"      jq -> spec:true    gojq -> spec:false
"Blocked by:<U+00A0>#12"       jq -> [12]         gojq -> []
```

Production runs the strict engine, so the divergence is real in both predicates and fails in opposite directions: the spec predicate **fails to drop** a spec (it ships as a claimable ticket), and the dependency scan **fails to detect** a blocker (a blocked ticket is admitted).

**What nobody has decided: should a no-break space count as separating whitespace here at all?**

Two coherent answers, and the choice is not obviously either:

1. **Yes — make the expression engine-portable.** Replace `\s`/`\d` with explicit classes so both engines agree, and pick the Unicode-aware semantics deliberately. A `## User Stories` heading padded with U+00A0 is still a to-spec spec by any human reading, and to-spec's template is authored text — a no-break space is what several editors and paste paths produce around headings. Under this answer, today's gojq behaviour is the bug.
2. **No — pin to ASCII and make the tests say so.** Production's behaviour becomes the specified behaviour. Simpler classes, one engine's semantics to reason about, and the spec-leak becomes a documented input requirement rather than a matcher bug. Under this answer, the fixture is the bug.

Note the answer may differ per predicate. The dependency scan's parity test currently asserts `[]` for the U+00A0 case — that is gojq's answer, adopted as the pin because it is production's, **not** because anyone ruled the behaviour correct. Answering (1) would flip that assertion.

`\d` is the easier half: an Arabic-Indic digit in an issue reference is not a case anyone is trying to support, so `[0-9]` is very likely right regardless. `\s` is the real question.

## Why this is `needs-triage`

[...]

Issue #345, opened 2026-08-11
candidates.test.mjs: drop the GOJQ_BIN escape hatch — refuted, it now carries a hard-failure contract


feigi:
**Deferred from PR #331 review** (issue #58). Dimension: `simplify`. Severity: suggestion.

**Finding:** `process.env.GOJQ_BIN` in `findGojq()` is config for a value nothing ever sets — no other file reads or documents it, and the `$GOPATH/bin` branch is what actually locates gojq — so the candidate loop reduces to a single `spawnSync("gojq", …)` probe.

**REFUTED by a change that landed in the same review, and not applied.** A separate `survived` finding measured that the old `findGojq()` accepted any binary answering `--version`, so `GOJQ_BIN=/usr/bin/jq` resolved to system jq and the parity test reported green having run on Oniguruma under a name claiming RE2; and a bad explicit `GOJQ_BIN` was silently discarded in favour of the GOPATH binary. `11eb72b` fixes both: the probe now checks the binary identifies itself as gojq, and an explicit `GOJQ_BIN` that is not gojq is a hard failure —

```js
if (named && !isGojq(named)) throw new Error(`GOJQ_BIN=${named} is not gojq`);
```

— because a caller who names the engine asked for that engine. `GOJQ_BIN` is now a documented contract with a failure mode, not an unused knob. Removing it would delete the only way to point the parity test at a specific gojq build, which is what the reviewers themselves used to verify the engine divergence.

Filed for the record so the ruling is auditable. **Recommend closing as wontfix.**

feigi:
> *This was generated by AI during triage.*

**Closing — refuted, record retained.**

No longer dead config: an explicit `GOJQ_BIN` that is not gojq is now a hard failure. The old probe accepted any binary answering `--version`, so `GOJQ_BIN=/usr/bin/jq` reported the parity test green having run on Oniguruma under a name claiming RE2.

Recorded in `.out-of-scope/gojq-parity-harness.md` so the same proposal is not re-litigated from scratch by the next reviewer.

The refutation in the body is the point of this issue, and closing does not remove it. Reopen only on evidence that contradicts the measurement — not to re-raise the finding.

Issue #338, opened 2026-08-11
candidates.test.mjs: gojq harness claimed as scope creep for #63 — refuted by #58's own acceptance criteria


feigi:
**Deferred from PR #331 review** (issue #58). Dimension: `correctness`. Severity: suggestion.

**Finding:** the gojq harness added by PR #331 — the STUB's `JQ_BIN` plumb, `findGojq()`, and the engine-parity test — implements #63 rather than #58, and drags in the import crash and the vacuous pin the same review flagged. #58 needs none of it to land.

**REFUTED, and not applied.** #58's own acceptance criteria say: *"Tests cover each form above and are written against the engine `gh` actually applies — see #63."* The harness is required by the ticket, not smuggled in. #63 is itself CLOSED (2026-07-31), so removing the harness would leave the gap with no owner at all.

The two defects the finding attributes to the harness were real and are both fixed on PR #331 (`11eb72b`): the unguarded `spawnSync("go", …).stdout` that killed the whole file at import on a machine without Go, and the parity test that stayed green when the `JQ_BIN` plumb was deleted. Neither is an argument against the harness existing.

Filed for the record so the ruling is auditable. **Recommend closing as wontfix.**

feigi:
> *This was generated by AI during triage.*

**Closing — refuted, record retained.**

The harness is required by #58 own acceptance criteria, which call for tests written against the engine `gh` actually applies. #63 is closed, so removing it would leave the gap with no owner. Both real defects it arrived with are fixed in `11eb72b`.

Recorded in `.out-of-scope/gojq-parity-harness.md` so the same proposal is not re-litigated from scratch by the next reviewer.

The refutation in the body is the point of this issue, and closing does not remove it. Reopen only on evidence that contradicts the measurement — not to re-raise the finding.

PR #451, opened 2026-08-13
fix(fleet): extend gojq parity harness to the spec predicate

### LINKED ISSUE #204: candidates.mjs --jq uses \s and \d, whose classes differ between gh's gojq and the system jq the suite tests with — a U+00A0 in a heading leaks a spec

feigi:
## Summary

Closes #204.

#331 closed the gojq-vs-jq (RE2 vs Oniguruma) parity gap for the dependency scan only. The spec predicate (`candidates.mjs:212`, `^#{2,6}[ \t]+User Stories\s*$`) was still exercised solely under the suite's system jq. Its one remaining engine-divergent position — the trailing `\s*` — is `\s`: Unicode-aware in Oniguruma, ASCII-only in RE2. A `## User Stories` heading padded with a trailing NBSP therefore reads as a spec under the suite and NOT under `gh`, which applies gojq — the exact silent-divergence class #331 fixed for dependencies, left open here. `dropSpecs` is the only line of defence against a leaked to-spec spec, so this gap was silent in both directions.

## Changes

`skills/fleet/scripts/candidates.test.mjs` only — `candidates.mjs` is unmodified, per the ticket's scope split from #383 (this is detection only, not a behaviour decision).

- Added the discriminating assertion to the already-existing gated `"the spec predicate holds under gojq"` test: `isDroppedAsSpec("## User Stories \n\nx\n") === false`. `false` is gojq's actual answer (production's), not a ruling that NBSP should separate — a comment says so and points at #383, which owns that question and may flip it.
- Updated the file-header comment and the local comment above the test to describe what's now covered (previously: "the spec predicate is under test; the ENGINE is not").

## Enumerate the class (covered vs. deliberately left)

The spec predicate's only ASCII-vs-Unicode divergent position, after #377 made the separator (`[ \t]+`) engine-portable, is the trailing `\s*` — one member, now covered.

Deliberately left, and why:
- **Whether NBSP *should* separate** — #383, a behaviour decision this ticket explicitly must not make.
- **`\d` divergence in the dependency scan** (`depnums`) — already covered by #331's gojq test; untouched here, still passing.
- **CI provisioning of gojq** (#337) — out of scope; this test skips on CI exactly as the existing dependency-scan gojq test already does. Known, accepted ceiling, not a defect.
- **Thinning/removing the harness** — refused per `.out-of-scope/gojq-parity-harness.md` (#338, #345); not attempted.

The "must ACCEPT" side is already covered inside the same gojq-gated test — `isDroppedAsSpec("##\nUser Stories\n\nsplit across lines\n") === false` was already there, pinning a real not-a-spec input under gojq specifically, alongside the two "is a spec" cases and the new NBSP discriminator.

## Test plan

- [x] `node --test skills/fleet/scripts/candidates.test.mjs` — 53/53 pass, with `GOJQ_BIN` pointed at a real `gojq 0.12.19` (`go install github.com/itchyny/gojq/cmd/gojq@latest`)
- [x] Mutation-verified: forced the test's `JQ_BIN` to system `jq` (simulating a deleted plumb) — the new assertion goes red (`true !== false`) exactly as it should; reverted before commit
- [x] Verified the NBSP is the ` ` escape, not a pasted literal — `od -An -tx1` on both new lines shows `5c 75 30 30 61 30` (ASCII ` `), no `c2 a0` bytes anywhere in the file
- [x] Ran with `PATH` excluding gojq and `go` entirely — 51 pass, 2 skip (both gated gojq tests), 0 fail, no throw at import
- [x] Existing dependency-scan gojq test (#331) still passes unchanged
- [x] Full fleet scripts suite (`node --test skills/fleet/scripts/*.test.mjs`) — 614/614 pass, no regressions elsewhere

https://claude.ai/code/session_01NL6Lp3fy5pnt7PnQtXVmXV


feigi:
**Review adjudication — 7 findings, 2 applied, 2 recorded on existing tickets, 3 no-change.**

Applied (both comment-only; the `--jq` program is byte-identical):

- `9f6208d` — `candidates.test.mjs:405`. The paragraph claimed #65 "only widened depth/CRLF handling, neither of which touches `\s`". CRLF handling **is** the trailing `\s*`: `b4739c8` (PR #377, branch `fix/65-candidates-spec-heading`) widened `[ \t]*$` back to `\s*$` for exactly that reason. Measured on `## User Stories\r`: `\s*$` → jq `[true]` / gojq `[true]`, `[ \t]*$` → `[false]` / `[false]`, so the CRLF fixture matches through that class alone.
- `593976b` — `candidates.mjs:331`. The `#65` block called the trailing `\s*` harmless and never mentioned the engine split this PR pins, while the dependency-scan comment above it does. Now states the split, the direction it fails in (RE2 matches fewer → `dropSpecs` never fires → the padded spec **ships as a claimable ticket**), and that #383 owns the class question.

Deferred — commented on the existing ticket rather than filing (both `ledger.mjs check` → `tracker-hit`, no duplicate opened):

- **#383** ([comment](https://github.com/feigi/claude-config/issues/383#issuecomment-5282021937)) — dimension `correctness`. The `\d` half of #204's title. New measurement: `Blocked by: #<Arabic-Indic 12>` makes the reduction **exit 5 under jq** (`tonumber` rejects the digits, and `candidates.mjs` reads a non-zero `gh` exit as "the query broke" → exit 2), while **gojq returns `d:[]`** and the blocked ticket reads as unblocked. Out of #204's scope (Out of scope assigns the dependency scan to #331/#340), so not widened here.
- **#337** ([comment](https://github.com/feigi/claude-config/issues/337#issuecomment-5282071375)) — dimension `tests`. There are now two `SKIP_WITHOUT_GOJQ` gates, not one. Measured on this branch: gojq reachable → `tests 614 / pass 614 / fail 0 / skipped 0`; `GOPATH=/nonexistent-probe` → `tests 614 / pass 612 / fail 0 / skipped 2`. `tests` and `fail` are identical either way — `skipped` is the only field that moves.

No change (settled against the tracker, not against a grep of the tree):

- The `#331` pointers at `:11` and `:412` stand. `gh pr view 331` is MERGED, "dependency scan finds refs in the forms tickets are actually published in", and its diff adds `isGojq`/`findGojq`/`GOJQ_BIN`. The review's objection established only that `#331` is uncited elsewhere in the tree, which is not evidence the attribution is false; it does not conflict with `:337-341`, which credits #63/#58 for *owning* the gap rather than for landing the coverage.
- The `#383` pointer at `:441` stands — `gh issue view 383` is OPEN and owns exactly that question.
- Reviewer wording rejected on measurement, twice: "divergence needs a non-ASCII byte" is false (gojq's `\s` is `[\t\n\f\r ]`, so ASCII **VT** diverges — jq `true`, gojq `false`), and "production is the stricter reader" describes a stricter *regex* but a leakier *filter*, in a block whose thesis is that a leaked spec is the expensive direction.

Verification: `node --test skills/fleet/scripts/*.test.mjs` from the worktree at `593976b` with `GOJQ_BIN` set — **tests 614, pass 614, fail 0, skipped 0** (both gojq-gated tests actually executed; a `skipped 0` run is the only one that proves the engine was reached).

PR #452, opened 2026-08-13
fix(fleet): reap.sh prints payload before pruning, refuses on 2 not 1

### LINKED ISSUE #265: reap.sh:102 — a failing `git worktree prune` exits 1 with no payload, after the branches are already deleted

feigi:
## Summary

`git worktree prune` sat as the last command of an AND-OR list after the branch-reap loop. Under `set -eu`, the *last* command in an AND-OR list is not exempt from `-e` — only earlier commands in the list are — so a failing prune aborted the script before the `{applied, reaped[], kept[]}` payload printed, after the branches were already deleted. The caller lost its only record of what had happened, at exit 1 — the code the fleet's script contract reserves for a verdict this script never had.

Triage on #265 closed the open design question in the issue body: approach 2 (print the payload first, prune after) plus a loud refusal on a failing prune.

- Payload now prints unconditionally, before the prune runs.
- A failing prune calls `die "git worktree prune failed"` — exit 2, message names the failed command — instead of leaking prune's own exit status through `-e`.
- Dry-run (`apply=false`) and successful-apply behavior unchanged.

**Exit-1 audit (per Agent Brief's whole-script claim):** walked every command in `reap.sh` for `set -eu` reachability. The only live exit-1 path was the prune line above (now fixed). Two other assignments (`wt=$(...)` at what was line 71, `ignored=$(...)` at what was line 96) are plain command substitutions whose failure *would* reach `-e` if the awk/paste pipeline inside them ever failed — but that requires an internal-tool failure on well-formed input, is pre-existing, outside this ticket's stated interfaces/out-of-scope list, and the resulting exit code isn't guaranteed to be 1 (it'd be whatever awk/paste emits). Declaring rather than fixing here, in the same spirit as the already-tracked #119 (JSON-escaping gap, same script). Everything else in the script's failure paths — the branch-loop's `git cherry`, `git status`, `git worktree remove`, `git branch -D` — is either inside an `if`-condition (fully `-e`-exempt) or already guarded with `cmd || die`/`cmd || { keep; continue; }`, consistent with the rest of the file.

## Test plan

- [x] Added `skills/fleet/scripts/reap.test.mjs` coverage:
  - Failing-prune case: `PATH` shim fails only `git worktree prune`; asserts payload still prints with correct `reaped`/`kept`, exit code 2, stderr names `git worktree prune`.
  - Accept-side control: successful `--apply` run still reaps, still prunes (verified via a genuinely stale worktree registration that must disappear from `git worktree list`), still exits 0.
- [x] Mutation-killed the new failing-prune test against the pre-fix ordering (reverted the tail to the original `[ apply ] && git worktree prune` before the printf) — only that test went red; the accept-side control and all 5 pre-existing tests stayed green, then restored and re-verified green.
- [x] `node --test skills/fleet/scripts/*.test.mjs` — 616/616 pass.

Closes #265


feigi:
## Review applied — deferrals filed

**Applied (11 findings, 4 commits).** The four `critical` findings on `reap.sh:138` were one defect: moving the payload `printf` above the prune made `[ "$apply" = true ] && { … }` the script's last command, so the documented default dry run regressed from exit 0 to a bare exit 1 — the exact failure #265 exists to remove, relocated onto the path every human and `set -e` caller takes. Fixed with an `if`; the comment that vouched for the old form was rewritten, and so was the one above it that still pointed at "the AND-OR list below". A dry-run test now covers the default path (all 8 prior `runReap` call sites passed `--apply`, which is how a 616-green suite shipped this) and was mutation-verified: with the `&&` form restored it is the only test that reds. Spec row 186 now states the shipped exit surface. The `pruneShim` docstring stops naming two causes that measurably do not make `git worktree prune` fail.

`node --test skills/fleet/scripts/*.test.mjs` → tests 617, pass 617, fail 0 before each commit.

**Deferred, out of this ticket's scope:**

| Issue | Finding | Dimension | Label |
|---|---|---|---|
| #460 | `reap.test.mjs` has three near-verbatim PATH-shim builders; one `gitShim(t, body)` collapses them | simplify | `ready-for-agent` |
| #461 | `reap.sh` dies on a failed housekeeping prune while `release-ticket.sh` only warns — one operation, two fleet contracts | simplify | `needs-triage` |

**Not filed, resolved here:** the `suggestion` that a failed prune leaves `"applied":true` unqualified in the payload. Its own remedy said documenting the payload-with-exit-2 shape in the spec row is sufficient today, and that row is now updated. No programmatic consumer of `reap.sh`'s stdout exists in the repo, so the speculative `"pruned"` field stays unbuilt.

**Refuted, not applied:** none — all three `suggestion`-band findings dispatched to a refuter survived. The `types` dimension returned nothing applicable (no static type system in a `.sh`/`.mjs` diff).

Issue #365, opened 2026-08-11
fleet scripts: a misspelled flag name reads as absent and silently takes the default — no unknown-flag sweep anywhere


feigi:
Deferred from PR #360 review (dimension: **correctness**). Finding state: **unverified** — the review's `suggestion` band is budgeted 0 refuters by policy. The measurement below is the reporting specialist's own.

A **misspelled** flag name still reads as absent and silently takes the default, at `skills/fleet/scripts/ci-state.mjs:97` and its siblings — the site #169 names as the priority precisely because the fleet gates PR-green on its verdict.

## Measured

The specialist `sed`-lifted PR #360's exact `arg()` body verbatim from `ci-state.mjs:40-50` and drove it over an argv matrix:

```
["--pr","5","--base","main"]     -> base = "main"
["--pr","5","--basee","main"]    -> base = null   << falls back to the default 'main'
["--pr","5","--base"]            -> DIE: --base needs a value
["--pr","5","--base=main"]       -> DIE: --base needs a space-separated value, not --base=
["--pr","5","--base","--quiet"]  -> DIE: --base needs a value
["--pr","5","--base","  "]       -> DIE: --base needs a value
```

**Every case #169 covers is handled correctly — there is no logic error in the new parser.** The residual is the `--basee` row: an unknown flag produces a real, wrong verdict at exit 0/1 with no refusal. Same consequence #169 describes, reached by a different defect (unknown flag, not absent value).

Worth having because `ci-state.mjs` and `candidates.mjs` are invoked from **markdown read by a model**, where a misspelling is more plausible than in a human-typed command line — `candidates.mjs:66-68` already argues exactly that ("a malformed invocation is more plausible here than the shape of this guard suggests"). Note #173 (closed) recorded the same class in `candidates.mjs`: a doc naming `--label` where the flag is `--require-label`, silently ignored.

## Suggested fix (one candidate, not a ruling)

An unknown-flag sweep after the known flags are read: collect the recognized names, then

```js
process.argv.slice(2).filter(a => a.startsWith("--") && !known.has(a.slice(2)))
```

and die naming the stray.

## Why `needs-triage`

Two undecided items, both material:

1. **Where it lives.** The sweep applies to all five scripts equally, which argues for doing it once — but triage's ruling on #169 explicitly **rejected** a shared module ("fix the four copies in place. No shared module."). A fifth copy of a `known` set per script is the alternative. That is a seam decision. See the shared-module ticket filed alongside this one.
[...]

Issue #362, opened 2026-08-11
ledger.mjs: --file swallows the next flag as its path, silently disabling --require-file and flipping exit 2 to 0


feigi:
Deferred from PR #360 review (dimension: **correctness**). Finding state: **survived** — adversarially verified 2-of-2 not-refuted, both against the immutable snapshot.

`skills/fleet/scripts/ledger.mjs:41-47` is a **fifth site** of the same fail-open defect #169 covers, in a hand-rolled shape the four `arg()` copies do not share: it splices `argv`, so it needs its own fix rather than the drop-in guard the other sites got.

## Mechanism

```js
const file = argv[fileIdx + 1];   // takes the next token, whatever it is
if (!file) die("--file given with no path");
argv.splice(fileIdx, 2);          // consumes that token too
```

`if (!file)` only catches a **truly trailing** `--file`. Given `--file <next-flag>`, `file` becomes the flag's own text (truthy), the guard never fires, and the `splice(fileIdx, 2)` then **eats the next flag's token**, so that flag reads as absent.

## Measured

```
$ node skills/fleet/scripts/ledger.mjs --file ; echo exit=$?
ledger: --file given with no path
exit=2                                    <- the only case that was ever tested

$ node skills/fleet/scripts/ledger.mjs --file /tmp/nope/l.md --require-file check "dup subject" >/dev/null 2>&1 ; echo exit=$?
exit=2                                    <- correct invocation: the guard fires

$ node skills/fleet/scripts/ledger.mjs --file --require-file check "dup subject" >/dev/null 2>&1 ; echo exit=$?
exit=0                                    <- guard silently gone
```

stderr of that last run:

```
ledger: WARNING — ledger file not found: --require-file. Every check will read "safe to file" until it exists.
ledger: not previously filed in this run's ledger — ledger-only answer, tracker unchecked
{"subject":"dup subject",...,"found":false,...,"verdict":"unverified"}
```

So `--require-file` — the flag whose entire job is to turn a missing ledger into a hard failure — is **silently disabled by the flag that precedes it**, and the duplicate-filing check answers "safe to file" at **exit 0** where a correct invocation exits 2. The exit code flip is the harm: a caller gating on it files the duplicate.

This is the identical shape PR #360 **did** fix one file over — `pr-overlap.mjs --a --b 5`, where `a` became the string `"--b"`:

```
$ node skills/fleet/scripts/pr-overlap.mjs --a --b 5 ; echo exit=$?
[...]

PR #462, opened 2026-08-13
fix(fleet): has() refuses --flag=value instead of reading it as absent

### LINKED ISSUE #364: board.mjs / ci-state.mjs: has() is blind to the --flag=value form, so a boolean flag written that way silently reads as absent
Deferred from PR #360 review (dimension: **correctness**). Finding state: **unverified** — the review's `suggestion` band is budgeted 0 refuters by policy, so nothing has adversarially checked this. The measurement below is the reporting specialist's own.

PR #360 taught `arg()` to refuse the `--flag=value` form, because `indexOf` cannot see it:

```js
if (process.argv.some((a) => a.startsWith(`--${n}=`))) die(`--${n} needs a space-separated value, not --${n}=`);
```

Its sibling `has()`, two lines below in the same files, was left with exactly the `indexOf` blindness that guard exists to close:

- `skills/fleet/scripts/board.mjs:35` — `const has = (n) => process.argv.includes(`--${n}`);`
- `skills/fleet/scripts/ci-state.mjs:57` — identical

So `--declare-no-ci=true` still reads as **absent**, silently:

```
$ node -e 'const argv=["--pr","5","--declare-no-ci=true"]; console.log("has(declare-no-ci) =", argv.includes("--declare-no-ci"));'
has(declare-no-ci) = false
```

## Severity: this is fail-CLOSED, not fail-open

`ci-state.mjs:100-104` documents `--declare-no-ci` as "the caller's opt-out, never inferred... so absence never silently reads as pass". Misreading it as absent means a no-CI repo exits **non-zero instead of taking the declared pass** — the safe direction, and not the fail-open class #169 is about. What is still wrong: the caller's explicit opt-out is **dropped with no signal at all**, which is the same "no refusal" complaint #169 makes.

## Suggested fix (one candidate, not a ruling)

```js
const has = (n) => {
  if (process.argv.some((a) => a.startsWith(`--${n}=`))) die(`--${n} is a boolean flag, not --${n}=`);
  return process.argv.includes(`--${n}`);
};
```

in both `board.mjs` and `ci-state.mjs`. `board.mjs`'s other boolean is `--open`.

**`needs-triage` because the approach is genuinely open**: refusing `--declare-no-ci=true` is one answer; *accepting* it (treating `--flag=true`/`--flag=false` as the documented boolean spelling) is another, and that is a CLI surface decision, not a naming one. Pick one before implementing.

Outside #169's scope. #169 is `arg()` reading a flag **given no value**; `has()` is a different function with a different signature and no value to read. Triage's ruling on #169 was "fix the four copies in place. No shared module." — widening to `has()` in-diff would have gone past what was ruled.

feigi:
## Summary

`has()` in `board.mjs`, `ci-state.mjs` and `candidates.mjs` was a bare `argv.includes()` check — the same `indexOf`-blindness `arg()`'s `=` guard exists to close for valued flags. A boolean written `--open=true`, `--quiet=false`, `--declare-no-ci=` or `--allow-fallback=true` silently read as **absent** instead of refusing. For `--declare-no-ci` specifically, that meant the caller's explicit no-CI opt-out was dropped with no signal at all.

Triage ruled: refuse the `--flag=value` form for booleans (not accept it as a documented spelling — that would make the two spellings disagree with `arg()`'s valued-flag refusal inside the same script). Triage also corrected scope: there are **three** copies of `has()`, not the two the ticket named — `candidates.mjs`'s `--allow-fallback` was missed. Fixed all three so this doesn't reproduce the drift #367 (shared-helper extraction, not in this wave) exists to eliminate.

`candidates.mjs`'s `--allow-fallback=true` happened to still exit 2 already, via a downstream `parseArgs` type check added for an unrelated reason (#173) — so the bug was functionally masked there, but `has()` itself was still blind. Fixed anyway, per the ruling and for the same reason: leaving it blind is exactly the copy-drift this ticket is about.

## What changed

- `has()` in all three files now refuses any `--name=value` spelling by name, exit 2, before any query/gh call/listen — wording is boolean-specific (`"--name is a boolean flag, not --name="`), distinct from `arg()`'s `"needs a space-separated value"` (a boolean has no value to take).
- Bare flags (`--open`, `--quiet`, `--declare-no-ci`, `--allow-fallback`) still read as present; absence still reads as absent — unchanged.
- Tests added for all three files, covering `=true`, `=false`, and bare `=` for every boolean flag (`--open`, `--quiet`, `--declare-no-ci`, `--allow-fallback`), plus accept-path controls (bare flag reads present; absence reads absent).
- Two existing `candidates.test.mjs` tests updated: their expected refusal message moved from `parseArgs`'s wording to `has()`'s own, since `has()` now intercepts `--allow-fallback=...` earlier than `parseArgs` ever runs.

## Enumerate-and-declare

Class: boolean flags read via `has()`. Members, from the ticket + the ruling's scope correction:
- `board.mjs`: `--open` — covered
- `ci-state.mjs`: `--quiet`, `--declare-no-ci` — covered
- `candidates.mjs`: `--allow-fallback` — covered (the one the ticket body missed)

Deliberately left out (ticket's own out-of-scope list):
- Accepting `--flag=true` as a documented boolean spelling — ruled against.
- Unknown/misspelled flag names — #365.
- `ledger.mjs`'s `--require-file` (its own spliced-argv parser) — #362.
- `#367`'s shared-helper extraction — not in this wave; fixed against the three copies as instructed.

False-positive check (what this change could wrongly REFUSE): every bare-flag call site was re-verified to still read present (`--open`, `--quiet`, `--declare-no-ci`, `--allow-fallback` bare forms all pinned), and absence still reads absent. No caller in this repo passes any of these flags with a `=` today, so nothing legitimate is newly refused.

Acceptance-criteria-violation check (#265-class regression): the new guard runs synchronously at module top-level, before any gh call, any file write, or (in `board.mjs`'s case) before `listen()` binds — verified directly (`serve --open=true` exits 2 with no port bound, no side effects). It does not touch any exit-code default path — `ci-state.mjs`'s normal not-green-is-nonzero contract and `candidates.mjs`'s exit 1/3 contracts are all unreached by this guard.

## Test plan

- [x] `node --test skills/fleet/scripts/*.test.mjs` — 632/632 green
- [x] Each new/changed assertion mutation-tested: reverted `has()` to the old `argv.includes()` one-liner per file, confirmed exactly the intended tests (and only those) went red, confirmed the accept-path controls stayed green, restored, confirmed green again — done separately for `board.mjs`, `ci-state.mjs`, `candidates.mjs`.
- [x] Manually verified `board.mjs serve --open=true --port 0` dies exit 2 with no listen, and `--open` bare still triggers the `open` attempt (visible via the ENOENT failure message with PATH stripped).

Closes #364


feigi:
## Review adjudication — 17 findings, all six dimensions

Head at review: `6b51b08`. Now `9926a66` (two commits). Full suite **tests 635, pass 635, fail 0**. No source `.mjs` changed by the review pass — tests and comments only.

Every in-scope `suggestion` got one refuter, biased to refuse and required to measure rather than reason. Six dispatched, six reported.

### Applied (5)

| # | dimension | what | verification |
|---|---|---|---|
| survived[1] | tests | `board.test.mjs` + `ci-state.test.mjs` pinned the `=` guard with every case at a **fixed** argv offset | argv-index mutant caught by exactly the new case in each file (board 34/35, ci-state 33/34) |
| unverified[2] *(inverted)* | correctness | the comment **this PR added** claimed `ERR_PARSE_ARGS_INVALID_OPTION_VALUE` was unreachable — it is not | `candidates.mjs --limit 5 --limit` raises exactly that code |
| unverified[7] | types | nothing pinned that `arg("spend-since")` runs before `has("spend-since")` | reorder → `--spend-since is a boolean flag`, 1 red of 9; green under an unrelated mutation |
| unverified[10] | simplify | `--allow-fallback` `=true` standalone + `=false`/`=` loop → one loop | guard deleted → 5 red, same 5 as before the fold |
| unverified[11] | simplify | two `ci-state` loops differing only in flag name → one nested loop | 7-mutant matrix, identical failing sets folded vs unfolded |
| unverified[12] | simplify | `--open` control re-spawned `serve` with argv byte-identical to the `--port 0` test | assertion rides the existing spawn; ~2.03s / 18.6% faster, mutation detection identical |

**On survived[1]:** the review's own note that `ci-state.test.mjs` was already position-robust "because `run()` always prepends `--pr 42`" is backwards, and was measured so. A *fixed* prepend gives a *fixed* offset — it guarantees the absence of position variation. A guard narrowed to `process.argv[4]` left all six `ci-state` cases green. Both files needed the case; `candidates.test.mjs` did not, because it genuinely runs its flag both alone and behind `--require-label nonexistent`.

**On unverified[2]:** the finding asked for the source comment in `candidates.mjs` to be corrected. Refuted — read literally, its clauses are a claim about parseArgs' codes plus past-tense history, both still true. But the refutation surfaced the inverse: the comment **this PR added** to `candidates.test.mjs` asserts "nothing left downstream can raise it", and `arg()`'s `=` guard is `indexOf`-gated, so it fires only when the bare spelling is absent. A repeated string flag whose later occurrence has no value reaches parseArgs. Corrected in the test comment; source left alone.

### Refuted, with evidence (2)

**unverified[9] — "row 4 is now a byte-identical duplicate of row 2, delete it."** Do not delete it. The stderr *is* identical, but the finding tested one mutation. Row 4 is the only test in all 635 that catches a guard narrowed to the **last** argv element: under that mutant, row 4 present → 1 red; row 4 deleted → whole suite green, mutation survives. Row 2's `=` flag is simultaneously first and last, so it structurally cannot catch it. The suggested replacement (bare `--allow-fallback --label …`) is strictly worse — it contains no `=`, so it never exercises `has()`'s guard at all, and it duplicates an existing test. **Row 4 is still doing the "behind" job it is named for, against `has()` instead of `parseArgs`.**

**refuted[1] — "`--declare-no-ci false` fails open."** Upheld as refuted. `--declare-no-ci` is presence-only, documented `[--declare-no-ci]` with no value form; no convention reads `--flag false` as false. The gate opens because the caller passed the flag, which is what passing it means. The probe's own control (`--quiet false` → runs, token ignored) shows the measured property is positional tolerance, not a gate failure. The **residue is real and was not filed anywhere** → now #463.

### Deferred (1 ticket)

**#463** — `board.mjs`/`ci-state.mjs` silently ignore stray positional arguments while `candidates.mjs` refuses them via strict `parseArgs`. `needs-triage`: `board.mjs` legitimately takes `build`/`serve` positionals, so the blanket refusal `candidates.mjs` gets free is wrong there, and #367 owns where the check would live. Distinct from #365, whose `startsWith("--")` sweep cannot catch a bare token.

### Not actioned, with reasons

- **refuted[2] + unverified[1] + unverified[3]** — one claim, that `--spend-since` gets `has()`'s boolean message. Measured false: `arg()` at `board.mjs:401` runs before `has()` at `:403`, so the wrong message is unreachable. The honest residue was unverified[7]'s missing pin — applied above.
- **refuted[3], unverified[6], unverified[13]** — three copies of `has()`/the comment/the `=` predicate with no shared enforcement point. Real, and owned by **#367**, which is open and already ruled to extract. Duplication a tracked ticket owns is not this PR's defect; extracting here would pre-empt it.
- **unverified[4]** — correct positive finding, `candidates.test.mjs` needs nothing. Confirmed.
- **unverified[5]** — claimed `ci-state.test.mjs` was already position-robust. Measured **false**; that is why a case was added there. See survived[1] above.
- **unverified[8]** — n/a, no static type system in these scripts.

Issue #363, opened 2026-08-11
board.mjs: a single child emitting >64 KiB into a piped stderr may swallow die()'s refusal — the console.error half of the #246/#328 family


feigi:
Deferred from PR #360 review (dimension: **correctness**). Finding state: **refuted as filed, narrowed on adjudication** — read the scope note below before working this.

## What was claimed, and what survived

A `critical` finding claimed `board.mjs`'s refusal path is lost whenever its stderr is a pipe — i.e. a no-op in the invocation the fleet actually uses. **That general claim is false.** The refuter ran a byte gradient and the refusal survived intact at **209,865 bytes**, well past the 64 KiB pipe buffer the claim rests on. Do not work the general claim.

What is *not* ruled out, and is what this issue is for: **a single child process emitting more than ~64 KiB into `board.mjs`'s piped stderr**, filling the pipe before `board.mjs` writes its own refusal.

## Why board.mjs is not covered by the existing issues

`board.mjs` is the one script in this family whose `die()` is `console.error`-based, not `writeSync`-based:

```js
// skills/fleet/scripts/board.mjs:16
const die = (m) => { console.error(`${NAME}: ${m}`); process.exit(2); };
```

- **#328** (open) covers `pr-overlap.mjs` / `diff-stats.mjs` / `ci-state.mjs` — the `writeSync` EAGAIN **exit-code inversion** family (2 inverts to 1). `board.mjs` cannot hit that: `console.error` does not throw EAGAIN. Different failure, different fix.
- **#246** (open) is `ledger.mjs read`'s `console.log` + `process.exit(0)` truncating its payload at 64 KiB on a pipe — the *same async-drop mechanism* as this one, on a different script and a different stream. That is the closest precedent and the reason this is plausible at all.
- **#299** (closed, `candidates.mjs`) and **#176** are the earlier rounds of the same two families.

Node makes `stdout`/`stderr` writes **asynchronous when the fd is a pipe** and synchronous for files and TTYs. `process.exit()` does not flush a pending async write. That is exactly #246's mechanism.

## The narrow case to reproduce

`board.mjs` forwards its children's stderr straight through — `tryRun` (`board.mjs:40`) and the `ci-state.mjs` sub-invocation (`board.mjs:62`) both call `execFileSync` with no `stdio` override, so the child inherits the parent's fd 2:

```js
try { return execFileSync(cmd, args, { encoding: "utf8" }); }
```

[...]

PR #464, opened 2026-08-14
fix(fleet): extract arg()/has()/die() into shared arg.mjs

### LINKED ISSUE #328: fleet scripts: die() inverts exit 2 into exit 1 on a full stderr pipe in three more scripts — the defect PR #322 fixed in candidates.mjs only
PR #322 fixed the EAGAIN exit-code inversion in `skills/fleet/scripts/candidates.mjs` only (#299). The byte-identical unguarded `die()` is still live in three sibling fleet scripts:

| file:line | current body |
|---|---|
| `skills/fleet/scripts/pr-overlap.mjs:24-27` | `writeSync(2, \`\n${NAME}: ${msg}\n\`); process.exit(2);` |
| `skills/fleet/scripts/diff-stats.mjs:108-111` | identical |
| `skills/fleet/scripts/ci-state.mjs:26-29` | identical |

All three forward gh's own stderr through `execFileSync` the same way `candidates.mjs` does, so all three hit the same condition #299 names: once ~64 KiB of that forwarded stderr is queued on a pipe, fd 2 is non-blocking, `writeSync` throws EAGAIN, the throw is uncaught, `process.exit(2)` never runs, and the process falls through to Node's **default exit 1**.

## Why `ci-state.mjs` is the sharp one

In the other two, exit 1 is at worst a confusing failure. In `ci-state.mjs` exit 1 is a *meaningful verdict*. Its own header comment states the invariant this defect breaks (`ci-state.mjs:404-405`):

> Exit vocabulary unchanged: 0 only when the gate is satisfied, 1 when it is not, **2 (via `die()`, above) only when the question could not be answered at all.**

An EAGAIN-swallowed exit 2 lands on exit 1, which in that same vocabulary means "gate not satisfied" — CI red / not-green — a real, different, actionable value the fleet's CI Monitor reads. `skills/fleet/commands/review-and-fix.md:23` makes the same split load-bearing: "every failure to *read* is exit 2", against exit 1 for not-green / undeclared no-ci. So this inversion collapses exactly the two states both documents say must never merge: a broken tool reads as a legitimate answer.

## Measurements

Reproduction rate is race- and machine-dependent; the mechanism is not. Independent measurements during the #322 review, on Linux (this repo's CI target), of exit 2 inverting to exit 1:

- 5/10, 6/10, 7/10 (pr-overlap, diff-stats, ci-state respectively)
- a second pass on the same harness: 4/10, 3/10, 0/10, then 6/10, 4/10
- a third harness: 1/40, 0/40, 0/40 — but that harness validated itself with a control at 32/40, so it was reproducing *something*

Treat the rates as machine-dependent and the mechanism as confirmed. The same one-line guard drops all three to 0/10 in the passes that measured it. `candidates.mjs` did not reproduce on darwin at all (PR #322 recorded 0 hits there even at 20 MB bursts), so a local macOS pass is not evidence either way.

## The fix

The minimum viable change is decided — the guard merged in #322, applied verbatim three times:

```js
function die(msg) {
  try {
  } catch {
    // Message may be lost; the exit code below must not be.
  }
  process.exit(2);
}
```

**Optional, implementer's call:** these are now four verbatim copies of the same function. Hoisting one shared `die()` into a module the four scripts import would make this fix and every future one apply once. Only `board.mjs` currently imports a sibling (`./compute-spend.mjs`), so the pattern exists but is not established. Three copies is the smaller diff today; the shared module is the smaller diff on the next occurrence. Either is acceptable — do not block on it.

Out of scope: `skills/fleet/scripts/ledger.mjs:17-20` also has a `die()`, but it uses `console.error`, which does not throw here. That is the separate #176 async-drop family, already fixed elsewhere.

## Tests

None of the three has any coverage for this. `pr-overlap.mjs` has **no test file at all** — it is the one script under `skills/fleet/scripts/` with a CLI and no `*.test.mjs` sibling. Port the pin from `candidates.test.mjs` ("the exit code survives a gh stderr larger than the pipe buffer"), and pair it with the deterministic source pin #322 added next to it rather than relying on the race alone. The #322 review reported the behavioural assertion firing on roughly 13-25% of Linux runs and never on darwin — those figures sat in the review's unverified band, so treat them as indicative, not measured.

## Precedent — this function has hit the partial-fix trap twice before

Three data points on this exact `die()` family, all verifiable from the tree:

1. **#176 / PR #222** (`e7b11e6`). Its own commit body: *"Review of the #176 fix found the one-site patch was a five-site defect."* The `writeSync` form that `candidates.mjs`, `pr-overlap.mjs`, `diff-stats.mjs` and `ci-state.mjs` all carry today landed in that single commit — `git log -S` for the write expression returns `e7b11e6` for all four. Caught in review, fixed everywhere at once.
2. **#169** — "arg() returns undefined for a trailing flag in four more scripts — the defect PR #167 fixed in candidates.mjs only", naming these same three scripts plus `board.mjs`. **Not** caught in review; still open.
3. **This issue** (#299 / PR #322). Caught in review again, deferred deliberately rather than widened in-diff.

Two of three needed a reviewer to notice; one slipped through and #169 is still open. That is the argument for the shared-`die()` option above — the copies staying in lockstep has depended on review catching it every time, which it has not.

---

dimension: `correctness`, `silent-failure`, `tests`, `comments`, `types` — all five independently reported this in the #322 review; it survived adversarial verification in every one, with all twelve refuter votes returning `refuted=false`.

Deferred from PR #322 review — out of #299's scope, which is explicitly the `candidates.mjs` exit contract.

### LINKED ISSUE #367: fleet scripts: arg()/has() is now five copy-pasted copies that have already drifted — extract one shared module? (triage ruled no for #169; re-ruling with post-#360 evidence)

feigi:
## Summary

Overturns #169's "no shared module" ruling per #367's re-ruling: `arg()`, `has()` and `die()` had drifted into 5/3/7 copies across the fleet scripts (post-#360 evidence — `candidates.mjs` alone carried a `Number.isInteger` guard, `board.mjs`'s `arg()` diverged in style, and `die()` was two incompatible shapes). One shared module now owns the contract.

- New `skills/fleet/scripts/arg.mjs` exports `makeDie(name)`, `makeArg(die)`, `makeHas(die)`.
- `board.mjs`, `candidates.mjs`, `ci-state.mjs` import all three; `diff-stats.mjs`, `pr-overlap.mjs` import `makeDie`/`makeArg` (no `has()` there); `ledger.mjs` and `fleet-tick.mjs` import only `makeDie`.
- `ledger.mjs`'s own spliced-`argv` parsing is untouched (#362 owns that — explicitly not a drop-in).
- `fleet-tick.mjs`'s `node:util` `parseArgs` option parsing is untouched — only its `die()` moved.
- Net **-49 lines** across the 9 touched scripts, plus a new 78-line module holding the rationale that used to be repeated at every call site.

### `die()` is the load-bearing half, per the ruling

Seven copies existed in two incompatible shapes: `board.mjs` and `ledger.mjs` used `console.error` + `process.exit()`; the other five used `writeSync`. That split is the direct mechanism behind two open bugs:

- **#328** — `writeSync`'s own `EAGAIN` exit-code inversion, fixed in `candidates.mjs` (#299/PR #322) and `fleet-tick.mjs`, but **not** in `ci-state.mjs`, `diff-stats.mjs`, `pr-overlap.mjs`.
- **#363** — `console.error`'s async-drop-on-a-pipe, live in `board.mjs` and `ledger.mjs`. #363's own Agent Brief names #367 as the fix vehicle: *"Land #367 first and this becomes a deletion."*

The shared `die()` is the `try { writeSync(2, …) } catch {} process.exit(2)` shape — the one already fixed in `candidates.mjs`/`fleet-tick.mjs`, and the one `candidates.test.mjs`'s own pre-existing source-text pin required (a plain `writeSync` shared function would have **regressed** `candidates.mjs`, silently removing protection it already had — verified by running the pin). Consolidating on that shape means `ci-state.mjs`, `diff-stats.mjs` and `pr-overlap.mjs` also pick up #328's protection as a consequence of sharing one correct implementation, not a second one-off fix. **This PR does not close #328 or #363** — #328's own acceptance criteria call for a dedicated darwin-vs-Linux pipe-race test that isn't added here, and #363 is a separate ticket — but both should now be cheap to verify-and-close.

### Observable output change — called out deliberately

`board.mjs`'s and `ledger.mjs`'s refusal format changes from `console.error`'s `NAME: msg\n` to `writeSync`'s `\nNAME: msg\n` (leading newline added, load-bearing for pipe-safety per the other five scripts' existing comments). No test in either file asserted the old format positionally (checked directly — no `^NAME:` anchored positive match existed for either script's `die()` output), so no test pins needed updating for this, but flagging per the ticket's acceptance criteria since it is a real change to what a caller reading raw stderr sees.

### Enumerate-and-declare (per the brief's request)

Verified against the tree at `29a37ac` (current `origin/main`), not trusted from the ticket's own counts:

| helper | copies found | covered |
|---|---|---|
| `arg()` | 5 — `board`, `candidates`, `ci-state`, `diff-stats`, `pr-overlap` | all 5, now shared |
| `has()` | 3 — `board`, `candidates`, `ci-state` | all 3, now shared |
| `die()` | 7 — `board`, `candidates`, `ci-state`, `diff-stats`, `fleet-tick`, `ledger`, `pr-overlap` (2 `console.error`, 5 `writeSync` in 2 sub-shapes) | all 7, now shared and canonicalized |

Deliberately left alone:
- `ledger.mjs`'s spliced-`argv` parser (#362).
- Guard *behaviour*: #364/#366 (already closed independently, verified their current guards carried over unchanged), #365 (open, untouched — unknown-flag sweep is a policy change, not a relocation).
- `util.parseArgs` — not reconsidered, matches the standing ruling.
- #328's own required new pipe-race test, and closing #328/#363 themselves — left for those tickets.

### Accept-path coverage added

`diff-stats.mjs` and `pr-overlap.mjs`'s CLI test files previously asserted only refuse cases for `arg()` — no test ever handed either script a well-formed value and confirmed the script actually ran. Added one `gh`-stubbed end-to-end test to each, asserting exit 0 and the real computed payload, so the shared `arg()` returning `undefined` for a valid input would be caught.

## Test plan

- [x] `node --test skills/fleet/scripts/*.test.mjs` — 640/640 pass (638 baseline + 2 new accept-path tests), zero pins broken
- [x] Manually verified `node skills/fleet/scripts/candidates.mjs --bogus` still refuses correctly post-refactor
- [x] Confirmed no test asserted `board.mjs`/`ledger.mjs`'s old `console.error` stderr format positionally before changing it

Closes #367

https://claude.ai/code/session_01GZwSzihZVAF3iSiyS86GVb


feigi:
Review applied at `9cf2fd7`. `node --test skills/fleet/scripts/*.test.mjs` -> **tests 642 / pass 642 / fail 0** (baseline was 640).

**Applied** — every added test kills its own mutant and only it, with a negative control that stays green:

| what | receipt |
|---|---|
| `arg.test.mjs`: import + wiring + `const NAME` pinned for all 7 consumers | board revert -> fail 1 (that test only); candidates local `makeDie` with the wiring line verbatim -> fail 1; `ci-state` `NAME = "wrongname"` -> fail 1 |
| `arg.test.mjs`: `makeDie`'s leading newline, pinned behaviourally | newline dropped -> fail 1 (that test only) |
| `candidates.test.mjs`: arg.mjs shape pin re-anchored per line under `/m` | revert + trailing-comment decoy -> fail 1 (that pin only); the decoy defeated the old unanchored regex |
| control: import reorder, trailing comment on the NAME line, and a comment inserted inside `makeDie` — all at once | 642/642 green |

Comment corrections, no behaviour change: `arg.mjs`'s header no longer claims a contract fix reaches every caller (`ledger.mjs` splices its own argv, #362); `board.mjs` no longer implies `--port` is guarded on the `build` path (`build --port`, `build --port abc`, `build --open=1` all exit 0, measured); `candidates.mjs` no longer claims two callers when `fleet-tick.mjs:215` is a third, programmatic one.

**Deferred, filed:**

- #467 — collapse the three factories into `makeCli(NAME)`. Behaviour-neutral and measured green, but it re-decides the interface #367's Agent Brief specified and was mutually exclusive with the pins this PR adds.
- #468 — `board.mjs build` never evaluates `argPort()`/`has("open")`. Pre-existing; the prose half was fixed here, the behaviour half has two competing approaches.
- #469 — record of the whole `suggestion` band: 3 applied, 2 refuted with evidence, 2 deferred on scope, 2 folded as duplicates.

The two `important` findings that were refuted 2-0 (`arg.mjs:3` header shape count, `arg.mjs:56` untyped `die` contract) stand refuted and are recorded in #469 so they are not re-raised.

feigi:
Addendum — one change in `9cf2fd7` the commit message does not name: **`arg.mjs:1`'s `#!/usr/bin/env node` shebang is deleted.**

Rationale, stated precisely because two tempting arguments for it are false: `arg.mjs` has **no CLI entry point** — no `import.meta`/`isCLI`/`argv[1]` main detection, it exports only `makeDie`/`makeArg`/`makeHas`, and every internal caller spawns through Node explicitly (`spawnSync(process.execPath, ...)`, `execFileSync("node", ...)`). It is the only pure-library file in the directory carrying one; `compute-board.mjs`, `compute-spend.mjs` and `strip-comments.mjs` all open on a `//` comment.

Two arguments deliberately **not** relied on, both measured false during review:

- *"mode 644, so it cannot be exec'd"* proves too much — `board.mjs` is also mode 644 and also carries a shebang, and it is not a candidate for deletion. The mode axis does not distinguish `arg.mjs`.
- *"the suite stayed green with it removed"* discriminates nothing — deleting the shebang from `candidates.mjs`, a genuine 755 CLI, is **also** 640/640. The suite has no power to detect shebang removal from any file, so a green run is not evidence here.

No pin added: shebangs are never exercised in this repo, so there is nothing meaningful to pin. Full detail in #469.

PR #465, opened 2026-08-14
docs(fleet): finisher gate is the controller's outbox, and tier outcomes accumulate across runs

feigi:
Four commits carried over from the previous `/fleet:run-team` run, committed in a worktree and never pushed. Rebased onto current `main` (the pre-rebase net diff read as large script/test deletions purely because the branch predated #450/#451/#452/#462 merging — nothing here touches a script).

## What lands

- **The finisher gate is the controller's outbox, not the member's last report.** The old text said "relay everything, wait for its final report, then dispatch" — followed exactly, it still dispatched into a live member. The gate is now "anything decided but not received, *including* a withdrawn or reversed ruling"; a ledger-only withdrawal is undelivered. Measured twice in one run, plus a third instance where a member pushed after reporting final.
- **Implementers enumerate what their EDIT newly does**, not only what the code already did wrong — and check the suite can see the mode they changed. From the PR #452 regression: a moved guard became the script's last command and regressed the *default* dry run 0→1, green at 616 tests because all 8 call sites passed `--apply`.
- **Tier outcomes accumulate in `docs/metrics/tier-outcomes.tsv`** instead of being concluded inside one run. The load-bearing finding: `class` and `tier` are perfectly confounded by the dispatch rule, so the comparison the old guard asked for measures class, not tier. Only a same-class comparison across tiers is informative.

## Added in this PR, on top of those four

The accumulate rewrite removed `at least three class=routine PRs` along with the in-run comparison it hung off, leaving the revert trigger with no sample floor. Restored, and bound to the accumulated file rather than to a run.

Two prose pins in `implementer-model-tier.test.mjs` updated **deliberately**, tracking wording these commits changed on purpose — the renamed `**Guard:` heading anchor, `/measure per PR, not per wave/` → `/unit is the PR/`, and `/never read its silence as a pass/` → `/never read the guard's silence as a pass/i`.

Each changed pin was mutation-checked **independently** — mutating its phrase reds that test and only that test, restored text green. One at a time, because the first failing assertion in a slice masks every assertion after it (which is also why the original run showed 2 failures for what was one renamed anchor).

638/638.


feigi:
## Review applied — `c390cad`

Review returned **0 survived / 3 refuted / 8 unverified**. Two of the three refutations were 1-1 splits and were reversed on the votes; a split is not a verdict.

Both reversed findings are defects in this branch author's own text, recorded as found.

### Applied (3)

| Finding | Location | Why |
|---|---|---|
| Stale finisher gate at the trigger site | `run-team/SKILL.md:435`, `:451` | The event-loop row the controller acts on still said "wait for the final report, then dispatch" — the rule this PR supersedes 300 lines later. Replaced with the outbox test at both edges. Measured first-hand this run: a fix-applier sent a final report and then sent two more substantive messages. |
| Tier floor counts PRs while claiming to count runs | `run-team/SKILL.md:261-273` | The minimum was "over the accumulated file, never one run" but counted PRs, and this branch ships four `class=routine` rows all on one `run_date`. Floor now requires **two or more distinct `run_date`s**. |
| `closed_own_ticket` trending `no` had no threshold | same paragraph | "Trending" is not measurable, so whether the guard fired was undefined. Now **two or more `no` rows**. Floor and trigger separated explicitly. |
| Misquote in the metrics note column | `docs/metrics/tier-outcomes.tsv:23` | Primary source (the comment added in `7f1418c1`, before `f666f50b` corrected it) reads "nothing **left** downstream can raise it". One word restored. Survived its refuter. |

The two strings `review-path-default.test.mjs:207` pins — `never dispatch off it while you do` and `never while you still owe it a ruling` — are intact, and that test passes.

### Deferred and filed (6)

| # | Finding | State | Dimension |
|---|---|---|---|
| #472 | `tier-outcomes.tsv` header has no test pinning it against SKILL.md's guard text | out of scope — coverage gap, not a defect in this diff | `tests` |
| #473 | Three unrelated policy changes behind one branch, no ticket | out of scope — the observation *is* this PR; unfixable from inside it | `correctness` |
| #476 | `/unit is the PR/` pin has no polarity guard | **refuted** — the suggested lookbehind passes 4/4 against the very mutation it was written for | `tests` |
| #477 | `tier-outcomes.tsv:21` misquotes the #65 CRLF finding | **refuted** — the phrase is verbatim from commit `9f6208d`'s subject; the proposed replacement appears verbatim nowhere | `comments` |
| #478 | "Measured: three members in one run" unsupported | **refuted** — two of three are base-branch records this PR consolidates; the reviewer's `find -iname "*log*"` searched filenames, not content | `comments` |
| #479 | "Observed twice" doubles a count | **refuted** — the same hunk deleted a second citation, so the region's incident total is unchanged at 2 | `comments` |

All six remain open.

One unverified entry was a positive confirmation rather than a defect (the mandated test command passing in full) and needed no action.

**638/638** from the worktree before commit.

PR #466, opened 2026-08-14
fix(fleet): three-state worktree predicate — absent, unreadable, no-linkage

### LINKED ISSUE #82: worktree-audit.sh:30 reports an unreadable worktree as `MISSING on disk` with `ahead=0 dirty=0` — byte-identical to a genuinely deleted one
Deferred from PR #72 review. Pre-existing; not introduced by that PR.

## Problem

`skills/fleet/scripts/worktree-audit.sh:30` gates on `[ -d "$wt" ]`, which is false for every `stat` failure, not only for a deleted directory. The else-branch at `:45-49` then asserts facts it never measured:

```sh
else
  readable=false
  ahead=0; dirty=0; files=""
  echo "    MISSING on disk: $wt" >&2
fi
```

## Measured — the two states are indistinguishable

Same worktree, holding an uncommitted `precious.txt`:

```
# parent chmod 000 (directory and work intact)
MISSING on disk: .../.worktrees/9-x
{"worktree":".../9-x","branch":"fix/9-x","ahead":0,"dirty":0,"dirtyFiles":[],"readable":false}

# genuinely deleted
MISSING on disk: .../.worktrees/9-x
{"worktree":".../9-x","branch":"fix/9-x","ahead":0,"dirty":0,"dirtyFiles":[],"readable":false}
```

Byte-identical. `dirty:0` is a positive assertion of cleanliness for a worktree that was never looked at, and `MISSING on disk` is a false statement in the first case. Same class as #56 and #74.

## Why it matters beyond a report

This script is used to decide whether a replacement member redoes or destroys work.

## Mitigation already present

`readable:false` is emitted in both, so a consumer keying on that field is safe — which is why this is not urgent. Note the asymmetry though: the *other* failure branch at `:38-44` uses `ahead=null; dirty=null` for exactly this reason. The code already knows `null` is the right answer for "could not look" and does not use it here.

## Open question (why this is needs-triage, not ready-for-agent)

Whether the fix is simply `null` in the else-branch, or whether the audit needs to distinguish three states (present / gone / cannot tell) the way `release-ticket.sh:188` now does after #72. The second is more faithful and more churn.

Verified on git 2.50.1.

### LINKED ISSUE #83: reap.sh and release-ticket.sh take opposite defaults on an unreadable worktree — one says dirty, the other says clean
Deferred from PR #72 review.

## The divergence

Two scripts guard the same thing — "does this worktree hold work I must not delete?" — and answer an unreadable worktree in opposite directions.

`skills/fleet/scripts/reap.sh:53` — unreadable ⇒ **dirty** ⇒ keep:
```sh
if [ -n "$(git -C "$wt" status --porcelain 2>/dev/null || echo dirty)" ]; then
  keep "$b" "dirty worktree $wt"
```

`skills/fleet/scripts/release-ticket.sh:188` after #72 (and after the follow-up in `1e54bac`) — a directory whose absence can be *established* ⇒ **nothing to protect** ⇒ proceed.

## Why this is not simply "reap is right"

`reap.sh`s default is the conservative one and it is cheap there: reaping is a housekeeping sweep, so keeping one branch too many costs nothing. `release-ticket.sh` cannot take that default wholesale — refusing forever is precisely the bug #37 filed. So the two are not obviously reconcilable by adopting one rule.

But they are also not obviously *un*-reconcilable: `1e54bac` distinguishes "gone, established" from "cannot look", and `reap.sh` currently collapses both into dirty. Applying the same three-state predicate would let reap keep its safe answer for "cannot look" while releasing a genuinely absent worktree instead of keeping a dead branch forever.

## Also note

`reap.sh:53` folds a failed `status` into the *value* via `|| echo dirty` rather than testing the exit status. That is safe in direction but it is the same shape `release-ticket.sh:156-158` explicitly rejects for `git cherry`, with a comment explaining why. Worth settling in the same pass.

Whether reap should adopt the established-absence predicate, or whether the two scripts *should* differ because their cost asymmetries differ. That is a design call, not a defect with one obvious fix.

### LINKED ISSUE #128: reap.sh:53 and worktree-audit.sh:36 read a worktree with a deleted .git file as CLEAN at rc 0 — the same parent-repo leak as #74, in two more callers

feigi:
## Summary

Triage split one recurring defect into three issues that land in the same
lines of the same scripts, with a dispatch note that they must be taken as
one set or the "unknown" state ends up with two different shapes in two
scripts. All three close here.

- **#82** — `worktree-audit.sh` collapsed "genuinely deleted" and "cannot
  stat (unreadable parent, dropped mount)" into one `MISSING on disk,
  ahead:0, dirty:0` report — asserting cleanliness that was never measured.
  Now three states: present-and-readable, established absent (real zero
  counts), unknown (null counts, `readable:false`). Also fixed: the
  porcelain worktree path and each dirty file's own path were read as awk's
  `$2`, truncating at the first space.
- **#83** — `reap.sh`'s dirty check folded *any* failed `git status` into the
  literal string `"dirty"` via `|| echo dirty` — including a worktree
  directory deleted by hand, which pinned its `[gone]` branch forever (the
  #37 shape, in the script whose default was supposed to be safe). Split
  into present / established absent (nothing to protect, eligible to reap) /
  cannot tell (keep — fail-closed, same direction the cherry probe already
  takes). Same `$2` path truncation fixed. Reap's worktree-removal path had
  **no** test coverage before this PR; added.
- **#128** — a worktree whose `.git` file is deleted, emptied into a
  directory, or replaced by a dangling symlink doesn't fail `git -C`: git
  walks UP to the enclosing repo and answers about THAT at rc 0. Fixed in
  all three named callers — `worktree-audit.sh`, `reap.sh`,
  `claim-ticket.sh`'s post-install lockfile-mutation check — each
  establishing the `.git` linkage exists before trusting anything read
  through it. Reference shape is release-ticket.sh's own linkage guard
  (`#82`/`#83`/`#128` all cite it by name, not line number, per #129).
  `worktree-audit.sh` additionally has to accept `.git` as a real
  *directory* (not just a file), since — unlike the other two scripts — it
  also audits the **main** checkout, whose `.git` is never a linked
  worktree's pointer file. Caught by a positive-control test before this PR
  ever ran: an early version of the fix refused the main worktree on every
  single run.

No shared helper extracted across scripts (each guard is one line and only
the refusal differs per caller — reap keeps, the audit reports unknown,
claim-ticket refuses); each script's guard cites release-ticket.sh's shape by
name in its own comment instead.

## Enumerated, covered vs. left

- Covered: unreadable-ancestor vs. established-absent (#82, #83); missing
  `.git` file / empty `.git` directory / dangling `.git` symlink (#128, all
  three scripts); the `$2` worktree-path truncation in both `worktree-audit.sh`
  and `reap.sh`; a dirty *file's own name* containing a space in
  `worktree-audit.sh` (named in #82's acceptance criteria — git C-quotes such
  paths itself, so the fix avoids double-quoting them into invalid JSON); the
  main-checkout `.git`-is-a-directory case in `worktree-audit.sh` (a false
  refusal caught by a positive-control test, not named in any ticket body).
- Deliberately left: `reap.sh`'s `awk -v b=…` branch-name interpolation
  (pre-existing, unrelated escape-processing risk, not one of the three named
  defects — issue #86's fix elsewhere was for a *path*, not a branch name);
  `.git` pointing at an *unrelated* worktree's gitdir (#135, explicitly
  out-of-scope in the #128 brief); a shared worktree-identity helper
  (explicitly decided against in the #128 brief); `reap.sh`'s `awk -v b=…`
  worktree lookup never needing the directory-vs-file `.git` distinction
  `worktree-audit.sh` needs, since reap only ever matches a `[gone]` branch's
  linked worktree, never the main checkout.

## Test plan

- [x] `node --test skills/fleet/scripts/*.test.mjs` — 657/657 green
- [x] New `worktree-audit.test.mjs` (none existed before): established-absent,
      unreadable-parent (root-skip guarded), missing/empty-directory/dangling-
      symlink `.git`, spaced repo path, spaced dirty-file name, plus positive
      controls for the healthy clean/dirty/ahead cases
- [x] `reap.test.mjs` gains worktree-removal coverage it had zero of before:
      healthy reap, dirty-keep, deleted-directory reap, dry-run/apply
      agreement, unreadable-parent keep, missing-linkage keep (file/symlink
      shapes), spaced repo path (clean-reap and dirty-keep)
- [x] `claim-ticket.test.mjs` gains a missing-linkage refusal test (shimmed
      `npm` that deletes `.git` mid-install)

Closes #82
Closes #83
Closes #128

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01GZwSzihZVAF3iSiyS86GVb


feigi:
## Review adjudication — 18 findings, all settled

Head after this pass: `b7b894e`. Suite from the worktree: `node --test skills/fleet/scripts/*.test.mjs` → **661 tests / 661 pass / 0 fail** (branch baseline was 657; four tests added, each mutation-checked).

### The headline: this PR had introduced a regression

The two `survived` findings were one defect. The new `dirtyFiles` awk read a staged rename's whole `R  <src> -> <dst>` line as one filename, wrapping the quotes git had already added inside a second pair — **the entire payload became unparseable, every other worktree's entry with it**. The `$2` code it replaced emitted valid JSON for the byte-identical status line.

Measurement while applying the remedy moved it: git C-quotes **any** path holding a space, rename or not, so the reviewer's `index(p, " -> ")` split still produced invalid JSON for `R  "sub/a -> b.txt" -> sub/c.txt`. Gate on the X status byte, and let a quoted source end at its own unescaped closing quote. The suite had no rename case at all, which is why this shipped green.

### Applied — 6

| finding | `file:line` | dimension |
|---|---|---|
| rename destroys the whole payload | `worktree-audit.sh:99` | correctness / silent-failure |
| linkage guard fires on the main checkout with a false reason | `reap.sh:125` | correctness |
| linkage guard missing the `-x` gate both siblings carry | `claim-ticket.sh:117` | silent-failure |
| "an ancestor could not be read" about a path that read fine | `worktree-audit.sh:116` | silent-failure |
| four identical unknown-state blocks collapsed into one helper | `worktree-audit.sh:77` | simplify / types |
| dead `mkdirSync` import, unused `envOverrides` param | `reap.test.mjs:14`, `worktree-audit.test.mjs:73` | simplify |

Each of the first four in-scope suggestions got a refuter told to default to refuted; all three dispatched survived, each verified by running a fixture. Two of them turned out **stronger** than reviewed: `reap.sh:125` is a regression (pre-diff printed the accurate `worktree remove refused`), and `worktree-audit.sh:116` covers three input shapes, not one.

### Deferred — 3, all filed

- **#470** — `reap.sh:101` `wt_present` boolean (simplify). In scope, but it restructures control flow around a guard added in this same pass; the reviewer's verified remedy no longer applies verbatim.
- **#178** (commented) — `reap.sh:26` `gone()` walk ends at `""` instead of `/`. Same defect as that issue, and measurement shows it is now at **three** byte-identical sites, not two.
- **#48** (commented) — `docs/specs/2026-07-23-fleet-plugin-design.md:187` output-shape example. Predates this diff; same file and same class that issue already tracks.

### Refuted — 6, recorded in #471

Including the one `critical` (`worktree-audit.sh:76`), whose headline artifact did not reproduce. #471 carries each refutation with its evidence so a later review does not re-litigate them. Nothing acted on.

The three referenced issues above all remain open; this PR changes nothing in them.

feigi:
## Correction to the adjudication above — two scope reversals

Head is now **`7c6cdb7`**, superseding the `b7b894e` in my previous comment. Suite unchanged: **661/661**.

Two changes in that comment's "Applied" table were reversed on a scope ruling. Both are behaviour-visible, so recording them rather than editing the comment.

### 1. `reap.sh:125` — same defect closed, narrower fix

The false `has no .git linkage` message for a `[gone]` branch checked out in the main checkout is still gone. **How** changed: instead of an explicit main-checkout branch, the existing linkage guard now accepts a `.git` **directory holding `HEAD`** — the shape `worktree-audit.sh` already used. `git worktree remove` then supplies the true reason.

What that gives up: the dry run still predicts `reaped:["feature/merged"]` for a branch `--apply` always keeps. **That asymmetry is pre-existing** — measured identical on both sides of this PR — so it is not something this PR introduces or was ever obliged to close. Building main-checkout detection here would be new logic on a path #82/#83/#128 do not own. Filed as **#474** with the fixture and both outputs.

The test was rewritten, not merely kept: it now pins the `--apply` half and carries an in-place comment saying the dry-run half is deliberately unpinned, because pinning it would cement the wrong behaviour.

### 2. `worktree-audit.sh:77` — the four-way dedup is reverted, not shipped

Applied and then reverted. It restructures the same `if/elif` chain that two of this PR's own fixes touch. Filed as **#475**, with the count corrected to **four** copies — it became four when this PR's not-a-directory split added one, and the review's "three" would have sent whoever takes it looking for a block that no longer matches.

### Corrected totals

**Applied (5):** rename payload regression · main-checkout linkage guard · `claim-ticket.sh` `-x` gate · not-a-directory cause · the two dead-code deletions.

**Deferred (5), all filed:** #470 (`wt_present`, body corrected — its rationale cited the guard now removed) · #474 (dry/apply asymmetry) · #475 (four-way dedup) · #178 and #48, both commented.

**Refuted (6):** #471, unchanged.

All five referenced issues remain open; this PR changes nothing in any of them.

feigi:
## Final state — disregard my previous correction comment

**Head: `5d64e20`.** Its tree is **byte-identical to `b7b894e`** (`git diff b7b894e HEAD` is empty). Suite **661/661**.

My previous comment announced two scope narrowings on `7c6cdb7`. Both were reversed after review; `7c6cdb7` has been reverted rather than force-pushed away, so the history shows the round trip honestly. The first adjudication comment on this PR is the accurate one.

### The two reversals, and why they were wrong

**1. `reap.sh` keeps explicit main-checkout detection.** Narrowing it to a linkage-guard relaxation would have left the dry run predicting `reaped:["feature/merged"]` for a branch `--apply` always keeps. That matters here specifically because `reap.test.mjs` already pins dry/apply agreement for the deleted-directory case, and **that pin belongs to #128 — one of this PR's own three tickets**. Shipping the narrow fix would have left this PR contradicting a property its own suite guards, on a ticket it closes. Detection is in scope, not scope creep.

**2. `worktree-audit.sh` keeps the four-way unknown-state dedup.** It was applied last, after both correctness fixes touching the same `if/elif` chain had landed, green and mutation-checked — which is what the review ruling asked for.

### Final split

**Applied (6):** rename payload regression · main-checkout linkage detection · `claim-ticket.sh` `-x` gate · not-a-directory cause · four-way unknown-state dedup · the two dead-code deletions.

**Deferred (3), filed:** **#470** (`wt_present`, `ready-for-agent` — its body is the original and is accurate; a correction I posted against it has been retracted) · **#178** and **#48**, both commented.

**Records (3):** **#471** six refuted findings · **#474** the dry/apply parity class, with the pre-fix measurement, explicitly *not* a live defect here · **#475** the dedup, recording that the block was **four** copies rather than the three the review counted.

All referenced issues remain open; this PR changes nothing in any of them.

Issue #376, opened 2026-08-11
no-undo-audit: a deleted refs/stash file with an intact reflog prints a confident `stash entries: 0` — the confident-zero class #147 closed, via a path #292 did not cover


feigi:
> *This was generated by AI during triage of the no-undo-audit cluster (#189, #303, #304, #306, #307).*

`no-undo-audit.sh` prints a confident `stash entries (repo-global, not gated): 0` for a repo whose stash reflog still names recoverable stash commits, because a **deleted** `refs/stash` file is indistinguishable from genuine absence under the current discriminant. This is the confident-zero class #147 was filed to eliminate, reached through a path #147 and PR #292 did not cover.

## Measured

`origin/main` @ `5bbd2ca`, git 2.x, darwin:

```
$ git init r && cd r
$ echo a > f; git add f; git commit -m one
$ echo b >> f; git stash
$ echo c >> f; git stash
  stashes created: 2

$ rm -f .git/refs/stash        # the ref file only — the reflog is untouched

  reflog lines : 2             # .git/logs/refs/stash still names both commits
  git stash list : ''  rc=0    # empty, and exits clean
  git show-ref refs/stash: rc=1
```

## Why the guard misses it

The discriminant added by PR #292 is:

```sh
if [ "$stash" = 0 ] && [ "$sr_rc" -ne 1 ]; then
  stash=null   # unknown
```

`-ne 1` is the whole test, so **rc 1 is defined as genuine absence and nothing else**. In the state above `show-ref` exits 1, the audit takes the `else` branch, and prints `0` — with two recoverable stash commits still named in the reflog. `git stash list` exiting 0 means nothing contradicts it.

Contrast with the three states PR #292 *does* catch, all of which leave `refs/stash` resolvable and so land on `-ne 1`:

| state | `show-ref` | reported |
|---|---|---|
| unreadable reflog (`chmod 000`) | rc 0 | `unknown` ✅ |
| unreadable `refs/stash` | rc 0 | `unknown` ✅ |
| ref → missing object | rc 128 | `unknown` ✅ |
| **`refs/stash` deleted, reflog intact** | **rc 1** | **`0`** ❌ |

## Severity

[...]

Issue #307, opened 2026-08-10
no-undo-audit payload: `stash` is number|null, and JS truthiness collapses null and 0 — a future consumer's `if (json.stash)` would erase the distinction #147 introduced


feigi:
**Deferred from PR #292 review** (implements #147). **Dimension: types.** Severity: suggestion. Reported at `skills/fleet/scripts/no-undo-audit.sh:333` (pre-fix numbering; the payload `printf` now sits at `skills/fleet/scripts/no-undo-audit.sh:339`, its governing comment at `:327-331`).

## The finding

The `stash` field is `number | null`, where JSON `null` means "unknown: the list is unreadable or the ref cannot be read". JS truthiness collapses `null` and `0` to the same falsy value, so a future `if (json.stash)` consumer would silently treat "unknown" as "definitely empty" — erasing exactly the distinction this PR introduces.

Failure scenario as filed: a future caller writes `if (result.stash) { warnMaintainer() }` instead of `result.stash !== null && result.stash > 0`. When the stash is unreadable, `stash` is `null`, `Boolean(null) === false`, so the warning fires identically to the genuinely-empty case and the caller never learns the count is unknowable (and may be nonzero).

## Why deferred — no consumer exists to protect

The reporter's own search settles this, and it is why the finding is a hazard note rather than a defect:

- `grep -rn '\.stash\b' --include='*.mjs' --include='*.js'` finds only `skills/fleet/scripts/no-undo-audit.test.mjs`, and every hit is `assert.equal(r.json.stash, <literal>)` under `node:assert/strict` (`===` semantics). No truthy check exists anywhere in the repo.
- `grep -rln 'no-undo-audit'` turns up `docs/specs/2026-07-23-fleet-plugin-design.md`, `skills/fleet/commands/run-merge-bot.md`, `skills/fleet/skills/run-team/SKILL.md` and the test file — all prose instructing a human or LLM reader to check `git stash list` when the count is "nonzero or unknown", never a code path that could coerce.

So this is a hazard designed into the representation for whichever JS consumer parses this payload next, not a live bug. The reporter's own disposition: *"Not worth a schema change now — no consumer exists to protect yet."*

## Partially addressed by #292

The finding's fallback suggestion was *"optionally add a one-line comment at the printf site naming the trap"*. PR #292 did amend that comment, for an adjacent reason — it had described `$stash` as "a digit count", which the `null` made false. It now reads:

```
[...]

Issue #306, opened 2026-08-10
no-undo-audit: a truncated-but-parseable stash reflog reports a too-low count as exact — the remaining ceiling after #147, with no test


feigi:
**Deferred from PR #292 review** (implements #147). **Dimensions: tests and types** — two findings, one theme, collapsed into this issue. Both severity: suggestion. Reported at `skills/fleet/scripts/no-undo-audit.sh:202` (tests) and `skills/fleet/scripts/no-undo-audit.sh:176` (types).

## The finding

A reflog that is merely TRUNCATED — some entries lost, the rest still parses — resolves `refs/stash` and returns a nonempty but undercounted list, so the cross-check sees no disagreement and reports the too-low count as exact. Same shape as the pre-#147 bug, just narrower: `stash: 2` when the real answer is 3, with no `null`/unknown signal distinguishing it from a real count of 2.

No test exercises this. The three (now four) new tests use `chmod 000` or a deleted object — states that fail loudly enough for the cross-check to catch. A truncation fixture would need to corrupt specific reflog *lines* rather than the file's permissions or the object it points at.

## Why deferred

Three reasons, and they agree:

1. **#147's brief puts it out of scope explicitly**, requiring only that it be *named as the remaining ceiling*. PR #292 keeps that sentence in `no-undo-audit.sh` and it is still accurate after the fix (see below).
2. **Both reporting specialists said no action.** tests: *"No action required for this PR; flagging only for completeness."* types: *"None needed now. The author already named this ceiling in-line instead of hiding it."*
3. Neither finding was executed as a probe — both are reads of the script comment plus the test file. The truncation behaviour above was measured independently by a third specialist on a `fx-badlog` fixture (three stashes, oldest reflog line's new-oid overwritten with 40 zeros -> `git stash list` returns 2 of 3 at rc 0), so the underlying claim is real; the *disposition* is what was deferred.

## What changed in #292 that affects this

A related finding claimed the script's stated ceiling was **wrong** — that it named the truncated reflog as the only remaining case while an unreadable `refs/stash` file was a second one, landing on the confident-zero side. PR #292 fixed the unreadable-ref-file case (`show-ref` replaces `rev-parse --verify --quiet` as the discriminator), which **makes the comment's ceiling statement accurate again**: the truncated reflog is now genuinely the only case left open. That finding is therefore resolved, not deferred, and this issue is what remains of the theme.

## Why `needs-triage`

[...]

Issue #437, opened 2026-08-11
run-merge-bot runbook: the `unknown` path names no ref-based recovery — `git show refs/stash` recovers the work in four measured states where both file reads come back empty


feigi:
**Deferred from PR #433 review** (implements #303). **Dimension: correctness.** Severity: important — but see the ruling below, this is an **enhancement, not a defect**.

## The passage as shipped is NOT defective

Stated up front because the finding this came from was **refuted**, and the refutation was upheld on review.

The finding claimed the new `unknown` path "recovers nothing" in the states where `refs/stash` resolves but the reflog does not list it. That is false for the passage as written. In `skills/fleet/commands/run-merge-bot.md:141`, `cat` is not the diagnostic — `ls -l "$c"/refs/stash "$c"/logs/refs/stash`, "where a mode of `----------` on either is the fault", is, and `cat` is explicitly scoped to one cause ("still names the branch when `refs/stash` is the unreadable one"). Reproduced: with `chmod 000 .git/logs/refs/stash`, the runbook's own `ls -l` prints `----------` on the reflog beside a healthy `-rw-r--r--` on the ref — it names the fault. The finding omitted the command that answers its own scenario.

**Nothing in PR #433 needs changing.** This ticket exists only to carry forward a measurement the refuted finding made that is independently true and useful.

## The residual: `git show refs/stash` appears nowhere in the skill

`git show refs/stash` prints the stashed work in `unknown` states where the two file reads the runbook names come back empty. Measured (git 2.50.1, throwaway repo, stash created on branch `feature-303`; audit verdict `unknown` = `stash list` empty at rc 0 AND `show-ref` rc != 1, per `skills/fleet/scripts/no-undo-audit.sh:303`):

```
(a) unreadable reflog — chmod 000 .git/logs/refs/stash:
      stash list rc=0 lines=0; show-ref rc=0            -> unknown
      cat .git/logs/refs/stash -> Permission denied      (the prose's command yields NOTHING)
      git rev-parse refs/stash -> 935484f3c36655ac11f55589c65f29cc3d64ac4b
      git show refs/stash:f    -> `a b`                  (the stashed work, recovered)

(E) git gc first, then chmod 000 on the reflog:
      stash list rc=0 lines=0; show-ref rc=0            -> unknown
      ls -l -> `ls: .git/refs/stash: No such file or directory` + `---------- .git/logs/refs/stash`
      git show refs/stash:f -> `a b`

[...]

PR #480, opened 2026-08-14
fix(fleet): no-undo-audit lets git name the missing stash object itself

### LINKED ISSUE #304: no-undo-audit drops git's own diagnostic on the unknown stash path — the operator gets a three-guess label instead of the fault git named
**Deferred from PR #292 review** (implements #147). **Dimension: silent-failure.** Severity: suggestion. Reported at `skills/fleet/scripts/no-undo-audit.sh:206`.

## The finding

`stash_list=$(git ... stash list 2>/dev/null) || stash_list=""` discards both git's exit status and git's diagnostic, so the operator gets a guess-list label instead of the actual fault git named, and any further unanticipated cause of an empty list is filed under the same label.

Evidence given (fixture: stash object removed):

```
$ (cd fx-corrupt/w && git stash list; echo "rc=$?")
fatal: bad object refs/stash
rc=1
$ no-undo-audit.sh fx-corrupt/w fix/1-thing 2>&1 >/dev/null
    stash entries (repo-global, not gated): unknown — refs/stash resolves but the list came back empty (unreadable or corrupted reflog)
```

The string `bad object` appears nowhere in the audit's stderr — the one line naming which cause it was is dropped on the floor. Every other probe in this script echoes the command it ran (`echo "$ git -C $wt status --porcelain" >&2` at line 163; the merge-tree echo below it); this one echoes neither the command nor the error.

Suggested fix as filed: keep the fallback, keep the stderr — capture git's stderr to a file or variable and append git's own message to the unknown line.

## Why deferred

**Out of PR #292's scope.** #147's brief is "report the stash count as `unknown` rather than `0` when it was not actually measured" — a correctness change to *which value* is reported. Passing git's diagnostic through to the operator is a new diagnostic capability, not part of that. It was deferred rather than applied on that basis, unrefuted either way.

## The suggested fix no longer applies as written — this is why it is `needs-triage`

PR #292 restructured the block (count first, then correct the one ambiguous value). There is no longer a `stash_list` variable to hang a stderr capture off, and no `|| stash_list=""`. The block now reads:

```sh
sr_rc=0
git -C "$wt" show-ref refs/stash >/dev/null 2>&1 || sr_rc=$?
if [ "$stash" = 0 ] && [ "$sr_rc" -ne 1 ]; then
  stash=null
  echo "    stash entries (repo-global, not gated): unknown — the list came back empty but refs/stash is not absent (an unreadable ref or reflog, or a ref pointing at a missing object)" >&2
else
  echo "    stash entries (repo-global, not gated): $stash" >&2
fi
```

Capturing git's stderr here means splitting the pipeline (stderr to a temp file, stdout to `wc`) and deciding what to do when git writes to stderr *and* succeeds. That is a real design choice, not a mechanical edit — hence `needs-triage`.

## Note on motivation, from a sibling review thread

A refuter on a related finding (filed as the `unknown`-message issue) measured that git's diagnostics genuinely differ per state and are genuinely informative:

- object gone: `git stash list` -> `fatal: bad object refs/stash`, rc 1
- unreadable reflog: `git stash list` -> empty, rc 0 (**git says nothing at all** — there is no diagnostic to pass through in this state)
- unreadable ref file: `git stash list` -> empty, rc 0 (same)

So this change would help in exactly one of the three `unknown` states. Worth weighing against the cost before implementing.

feigi:
The `unknown` stash line offers the operator three possible causes. In one of
those states git already knows which one it is — it says `fatal: bad object
refs/stash` — and the counting pipeline's `2>/dev/null` threw that away, so a
precise answer was reported as a three-way guess.

## What changed

Inside the `unknown` branch only, `git stash list` is asked a second time with
its stderr captured, and the result appended to the existing message when
non-empty.

The counting pipeline is **untouched**, deliberately. Capturing stderr there
means splitting stdout from stderr around a `wc`, a temp file on every run, and
a rule for git writing to stderr while succeeding — all paid on every healthy
audit to serve the one path that has already decided something is wrong. The
second call runs only on that path, where it costs nothing.

`|| true` is load-bearing under `set -eu`: git exits 1 in exactly the state this
exists for, so an unguarded substitution would abort the audit before it emits a
payload — turning a report into a refusal.

## The three `unknown` states, measured (git 2.50.1, this box)

| state | `git stash list` rc | stderr | effect |
|---|---|---|---|
| stash object gone | 1 | `fatal: bad object refs/stash` | appended |
| unreadable reflog | 0 | *(empty)* | line unchanged |
| unreadable `refs/stash` | 0 | *(empty)* | line unchanged |

## Tests

- `a corrupted stash ref passes git's own diagnostic through to the operator` —
  new behaviour, watched fail first (`bad object` absent from the audit's
  stderr), and asserts the generic line is still the prefix, so git's words are
  appended rather than substituted.
- `the states git is silent about print the unknown line unchanged` — pins the
  message as a **literal** for both silent states. Every existing `/unknown/`
  match still passes with a separator appended and nothing after it; this one
  does not. Mutation-checked: making the append unconditional fails it with
  `... missing object) — ` vs `... missing object)`.

`node --test skills/fleet/scripts/*.test.mjs` → **667 pass, 0 fail, 0 skipped**.

## Out of scope, as the brief ruled

The JSON payload keeps its settled `stash` contract (#307 owns its shape); the
command itself is not echoed the way the `status`/`merge-tree` probes do; and
which states produce `unknown` is unchanged — so the truncated-reflog ceiling
(#306) and the deleted-ref-with-intact-reflog undercount (#376) are untouched
here and remain open.

Closes #304


feigi:
## Review fixes applied — 8597c4d

12 findings (1 survived / 9 unverified / 2 refuted). Every in-scope `unverified` got its own refuter, biased to refuse and required to verify by running something.

### Applied (4)

|---|---|---|
| separator unpinned (`test.mjs:347`) | tests | `startsWith` + a loose match left the joiner unconstrained — `msg="$msg$diag"` and `msg="$msg: $diag"` both stayed green across all 50. Pinned as the whole line. |
| multi-line `$diag` (`.sh:327`) | silent-failure / types | a **corrupt** (not missing) stash object reaches the same unknown branch and git answers in 7 lines, 10 with a bad `objects/info/alternates` — they landed at column 0 above the next `$ git …` header. Folded with `tr '\n' ' '`. |
| `echo` expands escapes (`.sh:327`) | correctness | dash and macOS `sh` both eat an `echo` operand from a `\c` onward, and an alternates path with a backslash comes back verbatim in that stderr. `printf '%s\n'`. |
| duplicate fixture (`test.mjs:338`) | simplify | the two `a corrupted stash ref …` tests built a byte-identical fixture and the second's assertions were a strict superset. Merged. |

Plus two corrections the review required even though the change behind them was deferred: the comment claiming **both permission cases are silent** was true of `stash list` and false of the `show-ref` call already in hand, and the new `doesNotMatch(/fatal:|warning:/)` is now **narrowed to the reflog case** so #481's fix will not have to edit a passing test. The state table also listed three causes of `unknown` where there are four.

### Deferred

| # | what | why |
|---|---|---|
| #481 | the unknown path discards `show-ref`'s stderr, which git does speak in 2 of the 4 unknown states | out of scope (#304's brief names the unknown branch as the only place that changes); approach genuinely open — the message is one canned line for two causes, naming a null SHA neither ref holds |
| #482 | a corrupt loose object behind a **non-tip** stash entry prints a silently truncated count while git exits 1 and names the fault | line 300 is byte-identical pre/post PR and the brief forbids reshaping it. Not a duplicate of #306 — that is a truncated *reflog*, silent at rc 0; this is a corrupt *object*, loud at rc 1, so detection costs an rc capture rather than a new invariant |
| #437 (comment) | the runbook's `unknown` paragraph does not mention the audit now names one cause itself, and `two of its three causes` is now three of four — `ls -l` cannot see the fourth | not touched by this diff; same paragraph #437 already proposes to edit |
| #306 (comment) | cross-link to #482 | — |
| #483 | 4 suggestion-band findings refuted with evidence, recorded | two of the proposed fixes would have **lost coverage** — one misses the stray-stderr regression, the other pins #481 shut |

### Verification

Seven mutations re-derived against the final tree (`python3` with an asserted match count of 1 — `perl -pe '\Q…\E'` silently applies nothing here, since `$msg`/`$diag` interpolate as perl variables):

| mutation | expected | result |
|---|---|---|
| separator dropped | red | ✖ `a corrupted stash ref reports unknown, not zero, and passes git's diagnostic through` |
| separator changed to `: ` | red | ✖ same test |
| fold removed | red | ✖ `a multi-line diagnostic holding a backslash arrives folded and whole` |
| `printf` → `echo` | red | ✖ same test |
| stray `warning:` in the unknown branch | red | ✖ `the states \`stash list\` is silent about print the unknown line unchanged` |
| `show-ref … >/dev/null 2>&1` → `>/dev/null` (the #481 fix) | **green** | pass 50 |
| `if [ -n "$diag" ]` → `if [ "$diag" != "" ]` | **green** | pass 50 |

`node --test skills/fleet/scripts/*.test.mjs` → **tests 667, pass 667, fail 0, skipped 0**. `shellcheck -x -S warning` clean.

feigi:
## Follow-up — `f964bc8`

Two corrections to the comment above, both on controller ruling after the remaining refuters reported.

### `:242` taken deliberately — read this as scope, not creep

`echo "\$ git -C $wt status --porcelain"` at `no-undo-audit.sh:242` is **pre-existing** and not part of #304. It is taken anyway, with the same one-token fix, because it is a **strictly more reachable** instance of the hazard the stash line had: it carries a caller-supplied path and needs no corrupt repo, no alternates file — a worktree named `back\clue` truncates the header at `back` and the next stderr line lands on top of it. Fixing the theoretical instance and filing the live one is the failure this would otherwise ship.

The suite's pre-existing backslash fixture could not have caught it: it uses `back\slash`, and `\s` is an **unrecognised** escape that passes through unchanged. Only `\c \t \n \b \f \r \v \\ \0` mangle.

**Three same-class sites are still live and are NOT fixed here** — `:247` (`$porcelain`), `:418` (`$conflicts`), `:449` (`$at_risk`). Filed as **#484**. `:449` is the serious one: it carries **commit subjects**, so a member writing `fix the \connection retry` silently truncates its own `at risk:` line — the audit's most safety-critical output, the list of commits a careless resolution would delete. Measured:

```
$ sh -c 'a="abc123 fix the \cthing"; echo "$a" | sed "s/^/    at risk: /"' | od -c
0000000       a t   r i s k :   a b c 1 2 3   f i x   t h e          # truncated
```

### The test merge is reverted

The `simplify` finding at `test.mjs:338` (merge the two `a corrupted stash ref …` tests) **survived refutation** — coverage genuinely preserved, five destructive mutations redden the merged test alone. It was applied in `8597c4d` and is **reverted** in `f964bc8`: the visible triple at `:282`/`:302`/`:321` mirrors the script's rc table one test per state, and collapsing it saved 0.4s at the cost of that mapping, in a file whose whole subject is a state matrix. Recorded in **#483** so it is not re-raised as an unexamined simplification.

### Verification re-derived against `f964bc8`

Eight mutations, each applied with `python3` asserting an exact match count of 1 — the bare anchor `2>&1 >/dev/null` matches **twice** in this file (code and comment), so an unanchored `sed`/`perl` would mutate the comment and measure nothing.

| mutation | expected | result |
|---|---|---|
| separator dropped | red | ✖ `a corrupted stash ref passes git's own diagnostic through to the operator` |
| separator changed to `: ` | red | ✖ same test |
| fold removed | red | ✖ `a multi-line diagnostic holding a backslash arrives folded and whole` |
| `printf`@327 → `echo` | red | ✖ same test |
| `printf`@242 → `echo` | red | ✖ `a worktree path holding a backslash escape reaches the operator whole` |
| stray `warning:` in the unknown branch | red | ✖ `the states \`stash list\` is silent about print the unknown line unchanged` |
| `show-ref … >/dev/null 2>&1` → `>/dev/null` (the #481 fix) | **green** | pass 52 |
| `if [ -n "$diag" ]` → `if [ "$diag" != "" ]` | **green** | pass 52 |

`node --test skills/fleet/scripts/*.test.mjs` → **tests 669, pass 669, fail 0, skipped 0** (667 baseline + 2 new). `shellcheck -x -S warning` clean.

**Deferrals now: #481, #482, #483, #484, plus comments on #437 and #306.**

feigi:
## Class closed in this file — `b3f2615`

Per the controller's final ruling, all remaining `echo` sites carrying caller text are fixed. **No `echo` in `no-undo-audit.sh` now carries caller-controlled data.**

**`$at_risk` is the one a reviewer should look at.** It holds `git log --oneline` output, so an **ordinary commit subject** reaches it — a member writing `fix: the \connection retry` truncated its own `at risk:` line and collided the next. No corrupt repo, no exotic filename, nothing but the message they typed. That line is the audit's most consequential output: the commits a careless resolution deletes.

| site | operand | shape |
|---|---|---|
| `die()` | `$1` | **11** of its messages interpolate `$wt`, and all of them route through this one function — one edit, eleven messages |
| refusal listing | `$porcelain` | git C-quotes here, so this **corrupts rather than truncates**: `echo` collapses git's `\\` to one and silently rewrites the quoted path into a different one |
| conflict list | `$conflicts` | `-z` deliberately turns C-quoting **off**, so the raw backslash truncates outright |
| at-risk list | `$at_risk` | reachable through an ordinary commit message, as above |

**What is left is safe by construction, not by luck**, and the comment now says so: `$stash` is a digit count or the literal `null`, and git forbids a backslash in a refname — `check-ref-format` rejects it, `git branch` refuses to create it — so `$base`, `$branch` and `$fork` cannot carry one. Verified, not assumed.

Worth recording: the pre-existing backslash fixture used `back\slash`, and `\s` is an **unrecognised** escape that passes through unchanged. It exercised the safe member of the class, read as coverage, and was blind to all five defects.

### Verification — twelve mutations, re-derived against `b3f2615`

Each applied with `python3` asserting an exact match count of 1, each killing **exactly one** named test:

| mutation | expected | result |
|---|---|---|
| separator dropped / changed to `: ` | red | ✖ `a corrupted stash ref passes git's own diagnostic through to the operator` |
| fold removed | red | ✖ `a multi-line diagnostic holding a backslash arrives folded and whole` |
| `printf`@`$msg` → `echo` | red | ✖ same test |
| `printf`@`$wt` → `echo` | red | ✖ `a worktree path holding a backslash escape reaches the operator whole` |
| `printf`@`die` → `echo` | red | ✖ `a die message naming an unreachable worktree keeps the path whole` |
| `printf`@`$porcelain` → `echo` | red | ✖ `an uncommitted path holding a backslash escape survives the refusal listing` |
| `printf`@`$conflicts` → `echo` | red | ✖ `a backslash escape in a conflicting path and in an at-risk subject reaches the operator whole` |
| `printf`@`$at_risk` → `echo` | red | ✖ same test |
| stray `warning:` in the unknown branch | red | ✖ `the states \`stash list\` is silent about print the unknown line unchanged` |
| `show-ref … >/dev/null 2>&1` → `>/dev/null` (#481's fix) | **green** | pass 55 |
| `if [ -n "$diag" ]` → `if [ "$diag" != "" ]` | **green** | pass 55 |

`node --test skills/fleet/scripts/*.test.mjs` → **tests 672, pass 672, fail 0, skipped 0** (667 baseline + 5 new pins). `shellcheck -x -S warning` clean.

**#484 stays open, retargeted:** the same class is live in **ten** other fleet scripts — all carry the identical `die() { echo "$NAME: $1" … }`, and `worktree-audit.sh`, `release-ticket.sh`, `reap.sh`, `claim-ticket.sh` and `inflight.sh` echo worktree paths. The issue now states up front that this file is done, so nobody re-audits it.

**Deferrals: #481, #482, #483, #484, plus comments on #437 and #306.**

PR #487, opened 2026-08-14
fix(fleet): restore the ci-state paragraph's two scope words in both fleet commands

### LINKED ISSUE #198: The ci-state paragraph in both fleet commands lost its scope words: no force-push trigger, and 'reads as pending' now hangs off ci-state itself
Deferred from PR #195 review. **Approach still open — see the open question below.**

The #182 dedup rewrote the `ci-state.mjs` paragraph in both command files. Two scope words went with the redundancy.

**1. `review-and-fix.md:19` — the cancelled run now has no cause.**

Old: "A force-push cancels the run under it, but finished jobs keep their conclusions…"
New: "a cancelled run's finished jobs keep their conclusions…"

`grep -c force-push skills/fleet/commands/review-and-fix.md` → **0**. The file names a cancelled run but no longer says what cancels one, so a reader has no cue for when the hazard applies.

**2. `run-merge-bot.md:86` — "reads as `pending`" lost its aggregation scope.**

Old: "…so an absent job reads as `pending` and an inherited one as `pass` **if you aggregate instead of binding to this one run**."
New: "…whatever never ran is missing from the run entirely, which reads as `pending`."

The new sentence's subject is `ci-state.mjs`'s own presence requirement, and `ci-state.mjs` never reports an absent job as `pending` — it pushes the reason `expected jobs absent from the run: …` (`ci-state.mjs:158`) and refuses green. `pending` is what an *aggregating* view like `gh pr checks` shows.

**Open question (why `needs-triage`, not `ready-for-agent`):**

`ci-state.mjs:155` carries the same looseness in its own comment — "An absent job reads as pending and is invisible in a checks summary." The docs are faithfully mirroring the script. So the fix could be either:

- tighten the two doc sentences and leave the script comment, or
- tighten `ci-state.mjs:155` first and let the docs keep mirroring it.

Someone should pick before an agent edits.

**Context:** six adversarial verifiers refuted the stronger version of this finding — that these sentences misattribute the inherited-pass hazard to the job-presence check. They do not: `review-and-fix.md:13` and `run-merge-bot.md:80` both state that head-SHA binding *misses* that case, so crediting the presence check is correct. Only the two scope words above survive as a real defect.

feigi:
Closes #198

The #182 dedup rewrote the `ci-state.mjs` paragraph in both fleet command documents and dropped two scope words that the reference — the comment beside the `missing` computation in `ci-state.mjs` — still carries. That comment is unmodified here; it is the reference, not the defect.

**`review-and-fix.md`** named a cancelled run but never said what cancels one (`grep -c force-push skills/fleet/commands/review-and-fix.md` → `0` before this change), so a reader had no cue for when the hazard applies. Restored: a force-push cancels the run under you, and its finished jobs keep the conclusions they already reached.

**Both documents** hung "reads as `pending`" off `ci-state.mjs`'s own presence requirement. `ci-state.mjs` never reports an absent job as `pending` — it pushes the reason `expected jobs absent from the run: <names>`, and a non-empty `reasons` is verdict `not-green`, exit 1. `pending` is what an aggregating checks summary shows. Both sentences now scope it there and state what `ci-state.mjs` does instead.

The issue named the `pending` defect at `run-merge-bot.md` only; it is present at the sibling sentence in `review-and-fix.md` too, and both are fixed.

## Verification

`skills/fleet/scripts/ci-state-prose.test.mjs` pins both restored clauses in both documents. No pre-existing test asserted these sentences, so nothing needed updating — this is a new pin. Each slice is the single sentence making the claim: both documents discuss `pending`, cancelled runs and force-pushes in neighbouring paragraphs, so a regex over the section stays green with the sentence deleted outright.

Mutation-tested both ways, one clause at a time (a first failing assertion masks the rest of its test):

| Mutation | Want | Got |
|---|---|---|
| drop force-push cause (review-and-fix) | RED | RED |
| unscope `pending` (review-and-fix) | RED | RED |
| unscope `pending` (run-merge-bot) | RED | RED |
| drop ci-state's own outcome (run-merge-bot) | RED | RED |
| rewrite the sentence's opening | RED | RED (anchor fails loudly) |
| reword an unpinned clause *inside* the slice | GREEN | GREEN |
| edit a different paragraph in the same file | GREEN | GREEN |

`node --test skills/fleet/scripts/*.test.mjs` → 674 pass, 0 fail.


feigi:
Review applied at `e109ed9`. 5 findings: **3 applied, 1 deferred, 1 no-op.**

**Applied** — all three land in `ci-state-prose.test.mjs`; `ci-state.mjs` is unmodified (#198 acceptance criterion), read from the test only.

| # | dimension | what |
|---|---|---|
| survived | correctness | force-push clause was pinned for `review-and-fix` only — deleting it from `run-merge-bot` left the suite 2/2 green. Both tests now loop one shared `DOCS` table. |
| U1 | correctness | `sentence()` ran to the next section marker, not the sentence its comment claims — a gutted sentence stayed pinned by a paragraph inserted before the marker. Slice now stops at its paragraph. |
| U2 | correctness | both docs quote `ci-state.mjs`'s reason string (introduced by this PR); renaming it reddened only `ci-state.test.mjs`. The literal is now read out of the source. |

Mutation-verified one clause at a time, **12 red / 4 green**. Each `run-merge-bot`-only mutant reddens naming `run-merge-bot`, with `review-and-fix` intact — i.e. the loop actually reaches the second document, which is the vacuity this PR fixes. `node --test skills/fleet/scripts/*.test.mjs` → **674 / 674 / fail 0**.

**Deferred** — #490 (dimension: tests) — the two-document loop reports only the first document, so a defect shared by both files is undercounted. Diagnostics gap, not a discrimination gap; the regression is still caught. `needs-triage` because the approach is open and five test files slice these same documents, so it may be a repo-wide convention rather than a one-file fix.

**No-op** — U3 (dimension: tests) probed whether a negation planted beside a pinned clause survives. It does, and the test header already states that ceiling verbatim at line 10. Nothing to fix, nothing filed.

One finding was **refuted** during review (it claimed `ci-state.mjs:351` contradicts the new doc sentences; its fix would have edited `ci-state.mjs`, violating the acceptance criterion). Not applied.

Known residual, not chased: with the narrowed slice, a restatement in the **same** paragraph still survives. That is the case the test header already discloses; closing it needs a contiguous-literal-span pin that nobody has measured.

Issue #482, opened 2026-08-14
no-undo-audit: a corrupt loose object behind a non-tip stash entry prints a silently truncated count — git exits 1 and names the fault, the pipeline discards both


feigi:
**Deferred from PR #480 review** (implements #304). **Dimension: silent-failure.** Severity: important. Reported at `skills/fleet/scripts/no-undo-audit.sh:300`.

## The finding

The counting pipeline discards both git's exit code and its stderr:

```sh
stash=$(git -C "$wt" stash list 2>/dev/null | wc -l | tr -d ' ')
```

When a loose object behind a **non-tip** stash entry is corrupt, `git stash list` dies partway through: it prints the entries it managed to read, exits 1, and says so on stderr. The pipeline throws away both, `wc` counts the truncated stdout, and the audit prints a confident number that is simply wrong. The `unknown` branch cannot rescue it either, because that branch is gated on `[ "$stash" = 0 ]` — so the one path where the count is *wrong* is the one path with no diagnostic.

## Measured (git 2.50.1, Apple Git-155, darwin)

Fixture: bare origin + clone, branch `fix/1-thing`, **two** stash entries, then the older (bottom) entry's loose commit object overwritten with `garbage` (`chmod u+w` first — loose objects are mode 444).

```
$ grep -c '' .git/logs/refs/stash          -> 2          # two real entries
$ git stash list 2>/dev/null | grep -c ''  -> 1          # the count the audit takes
$ git stash list 2>&1 >/dev/null           -> rc 1, five lines:
      error: inflate: data stream error (incorrect header check)
      error: unable to unpack 16df154b… header
      error: inflate: data stream error (incorrect header check)
      error: unable to unpack 16df154b… header
      fatal: loose object 16df154b… (stored in .git/objects/16/df154b…) is corrupt
$ git show-ref refs/stash >/dev/null 2>&1  -> rc 0       # the cross-check sees nothing wrong

$ sh skills/fleet/scripts/no-undo-audit.sh <wt> fix/1-thing
    stash entries (repo-global, not gated): 1
    {"…","stash":1,…}                       audit rc 0
```

Two entries exist, one is reported, git named the fault outright at rc 1, and none of it reaches the operator. A member reading `stash entries: 1` cannot tell this from a healthy repo with one stash.

## Why it was not fixed in PR #480

[...]

PR #489, opened 2026-08-14
test(fleet): pin the truncated-reflog stash undercount as a known ceiling

### LINKED ISSUE #306: no-undo-audit: a truncated-but-parseable stash reflog reports a too-low count as exact — the remaining ceiling after #147, with no test

feigi:
## Summary
- Adds a characterization test for `skills/fleet/scripts/no-undo-audit.sh`'s one remaining stash-count ceiling: a stash reflog that is TRUNCATED (some lines lost, the rest still parses) resolves `refs/stash` and returns a shorter-but-nonempty list, so the `show-ref` cross-check sees no disagreement and the audit reports the too-low count as exact. This is the one state in the stash block that had no test — its three siblings (unreadable reflog, unreadable `refs/stash` file, corrupted ref) all fail loudly enough for the cross-check to catch.
- The test pins the known undercount (does not fix it) — the issue's triage ruling is explicit: detection would need real reflog-vs-count complexity not worth building. No detection logic is added to `no-undo-audit.sh`.
- Widens the in-line ceiling comment (the one reading "That gap is a remaining ceiling" above the stash-counting pipeline — cited by phrase, not line number, since this file's line citations are known to drift), which previously named only the truncated-reflog case, to also name #482's sibling ceiling (same silently-too-low symptom, opposite mechanism — loud `rc 1` rather than silent `rc 0`) so the comment stops under-describing the gap. #482 stays out of scope for this PR; it's filed separately and not yet triaged.
- The new test's own citation of that comment is likewise anchored on its text ("a remaining ceiling"), not a line number, for the same reason.

## Enumerate-and-declare
Class: "the stash count is silently wrong" — every member the ticket names, in passing or otherwise:
- **Truncated-but-parseable reflog** (this issue's core case) — COVERED: new characterization test, fixture per the ruling comment (3 stashes, drop oldest reflog line via `tail -n +2` equivalent).
- **Unreadable reflog / unreadable `refs/stash` file / corrupted ref** — already covered by the three sibling `#147` tests; unchanged.
- **Healthy 3-stash reflog counting 3 (an ACCEPT case)** — already covered by the existing "several pre-existing stashes still do not refuse" test; left as-is, not duplicated, since it already proves the audit doesn't over-refuse a healthy multi-stash repo.
- **#482 (corrupt loose object behind a non-tip stash entry — `stash list` prints what it got, exits 1, `fatal: loose object … is corrupt`)** — deliberately LEFT for this PR. Filed separately and still untriaged, precisely so a "not worth the complexity" answer here doesn't gate that one. Named in the widened ceiling comment only so the comment is accurate, not addressed with a test or fix here.
- **#376 (`refs/stash` file deleted but the reflog intact — reports a confident `0`, not an undercount)** — a different symptom from this issue's (a confident *zero* despite recoverable stashes, vs. a too-low but nonzero count), reached through the same `show-ref rc 1 → "genuine absence"` branch of the cross-check. Deliberately LEFT for this PR — filed and open separately, not addressed here, and not folded into the widened ceiling comment since that comment specifically describes the nonzero-undercount ceiling and #376 is a distinct failure mode, not another shape of it.
- **In-file ceiling sentence under-describing the gap** — addressed for #482 (same-shape sibling); #376 (different-shape sibling) is out of scope, per above.

## Mutation test (run against the new test, per issue's verification bar)
- Mutant A (closes the gap): `stash=$((stash + 1))` in the else-branch — new test goes red (`3 !== 2`), confirming it discriminates the pinned value. Reverted, confirmed byte-identical to baseline, re-ran green.
- Mutant B (orthogonal change): relabeled the unrelated `clean` stderr line — new test stays green, confirming it doesn't false-positive on unrelated script changes. Reverted.

## Test plan
- [x] `node --test skills/fleet/scripts/no-undo-audit.test.mjs` — 56/56 green
- [x] `node --test skills/fleet/scripts/*.test.mjs` — 673/673 green (full suite, post-rebase)

Closes #306


feigi:
> *This was generated by AI during review follow-up.*

## Review actions applied — `29b980b`

Full suite green from the worktree before each commit: `node --test skills/fleet/scripts/*.test.mjs` → **tests 673 / pass 673 / fail 0**.

### Applied

| # | Dimension | What | Verification |
|---|---|---|---|
| survived (×3) | `comments`, `correctness`, `silent-failure` | The ceiling paragraph ended *"Filed and ruled on separately"* about **#482**, which has no ruling — OPEN, sole label `needs-triage`, zero comments, `createdAt == updatedAt`. Now *"Filed separately and not yet triaged"*. The PR description carried the identical claim in two places; corrected there too. | #482's state re-checked live before editing. The paragraph's mechanism half was left untouched — it is correct. |
| U5 | `types` | *"That gap is a remaining ceiling, pinned but not closed — closing it was ruled out … — by #306's characterization test"* could be misparsed as the **test** having ruled. Split into two sentences. | Prose only. The anchor phrase `a remaining ceiling` survives on one line, which the new assertion below depends on. |
| U1 | `correctness` | Added `assert.equal(stashLine(r), "    stash entries (repo-global, not gated): 2", …)` — the operator-facing stderr line was unpinned. | Neutering `no-undo-audit.sh:385` to `:` left **all 673 tests green**; with the assertion, exactly **1** goes red and only that one. A sibling-branch mutation reds three *other* tests and leaves this one green. |
| U2 | `silent-failure` | Added the `jsonError` guard before the `r.json.stash` deref. | Appending `JUNK` to the payload `printf`: before, `TypeError: Cannot read properties of null (reading 'stash')`; after, `payload must parse; got Unexpected non-whitespace character after JSON at position 254` with raw stdout dumped. |
| U6 | `types` | The docstring tells the next reader to grep the script for `a remaining ceiling`; nothing enforced it. Now asserted, matching the pattern the spec-row test in this file already uses. | Rewording that sentence in the script reds this test **and nothing else** (673-test suite); rewording an unrelated comment leaves it green. |
| U8 | `simplify` | Test name 157 → 124 chars; dropped the clause U6 now enforces and the escaped quote pair with it. | Still states plainly that it pins a known ceiling rather than desired behaviour, and still names #306 — both halves #306's acceptance criteria require. |

U1 and U2 change test semantics, so each was checked by a dedicated refuter instructed to default to `refuted=true` and to verify by running. Both returned **`refuted: false`** with measurements. Two of U2's supporting claims were corrected in the process: the sibling guard count is **10, not 12**, and those ten all sit in the JSON-escaping cluster where parseability is the subject — every stash-family test omits the guard. So this is a diagnostics improvement, not a consistency one, and the commit message says so. The guard also does not cover an *absent* payload (empty stdout yields `json: null, jsonError: null`) — same residual gap as its ten siblings.

### Deferred / filed

| # | Dimension | Where it went |
|---|---|---|
| U3 | `silent-failure` | **Commented on #482** — its own text scopes it out of #306, and #482 is where the probe belongs. Reproduced independently before posting: 3 stashes, corrupt **non-tip** loose object → audit **exits 0** reporting `"stash": 1` against a true count of 3, operator line prints `1`, no `unknown`. |
| U4 | `types` | **Filed as #494** (`needs-triage`). Pre-existing, diff-untouched: two comments claim a content-corrupt **tip** stash object leaves `stash list` at rc 0. Measured **rc 1** on git 2.50.1, empty stdout, `show-ref` rc 0. The finding named only `no-undo-audit.sh:369`; the identical clause is also at `no-undo-audit.test.mjs:454`, and the issue names both. |
| U9 | `simplify` | **Deferred, recorded as a comment on #494** — not applied. It proposed deleting lines 325-331, the very paragraph carrying the false claim being fixed here, and its own replacement text (*"The measurement detail and the ruling live on #482"*) restates that #482 has been ruled on. Applying it would have deleted the defect and immediately re-introduced it. |

### Rejected

**U7** (`simplify`) — proposed collapsing the two labelled pre-condition probes to two lines. #306's acceptance criteria require exactly that verbosity: *"It asserts the conditions that make the undercount invisible — `stash list` rc 0, `show-ref` rc 0 — so a future reader can see why the cross-check misses it."* The labelled rc assertions **are** that criterion. U7 itself concedes the redundancy may be deliberate documentation. It is. Nothing filed.

`test-coverage` returned zero findings and was confirmed genuinely clean, not unrun.

PR #495, opened 2026-08-15
fix(fleet): dictated refuter prompt names a scratch path, forbids checkout writes

### LINKED ISSUE #258: run-team: the fix-applier's refuter prompt instructs no scratch isolation, so sibling refuters collide on generic filenames in one shared scratchpad

feigi:
## What

The dictated refuter prompt — quoted verbatim in `skills/fleet/skills/run-team/SKILL.md` (fix-applier dispatch block) and `skills/fleet/commands/review-and-fix.md` (step 2) — orders filesystem writes (*"Verify by RUNNING something — compile it, run the test, apply the mutation"*) and named no location. Both copies now append:

- A two-level scratch namespace, `<scratch>/pr<N>/<finding>/`, matching the **Guards** rule already in `run-team/SKILL.md` and `references/isolation.md` — nothing invented, just reaching the dispatch site that rule didn't cover.
- An explicit statement that the checkout and any worktree are never write targets, with an explicit read carve-out (`git show`/`git archive` at a pinned ref stay fine everywhere) — a write ban must not read as a read ban, or a refuter can no longer verify anything.
- A mechanical guard against the 4th measured instance: chain `cd` into the git command (`cd "$D" && git …`, never `cd "$D"; git …`), and assert `git rev-parse --show-toplevel` equals the scratch path before any `git init`/`git commit` in a fixture. That instance's `cd` silently failed and the following `git` commands ran in the maintainer's live checkout, capturing an uncommitted `settings.json` edit into a stray commit — the agent believed it was in scratch and was wrong, so "namespace your path" alone doesn't close it.

A new prose-pin test, `skills/fleet/scripts/refuter-scratch-prose.test.mjs`, pins all three clauses in both documents.

## Enumerate-and-declare (the class this ticket names)

Searched the fleet skills tree for every hand-dispatched grandchild carrying a dictated, verbatim, write-instructing prompt:

- **Refuters** (fix-applier → refuter) — the only match, and this PR's fix. Confirmed as the sole occurrence of `Try to REFUTE this finding` / `Verify by RUNNING` across `skills/fleet/`.
- **Specialists under the hand-dispatch fallback** (reviewer → 6 specialist types, `review-and-fix.md`'s **Specialists** section, which fires whenever the dispatcher lacks the `Workflow` tool — i.e. every fleet subagent) also write to disk (mutation probing). But they're governed by a *different* mechanism already: the dispatcher cuts and hands each specialist its own tree (`snap-ro` + private copies), rather than the specialist self-selecting a generic name from an unlocated instruction. That's not the "instruction names no mechanism" defect this ticket fixes. They also don't `git init`/`git commit` fixtures the way a refuter does, so the mechanical cd/rev-parse guard has little to catch there. Left deliberately narrow — not applying this clause to specialists — but noting the *same* cross-PR collision shape could in principle apply if their scratch-dir naming isn't PR-scoped; that's a candidate follow-up, not silently ignored, and out of this ticket's explicit scope (out-of-scope list excludes the review-pr/specialist path).
- **Probes** ad hoc-spawned by refuters/specialists mid-verification (the `zshtest` and harness-injection-probe instances from this issue's comments) have no dictated prompt anywhere in the docs — nothing to edit. They inherit whatever the dispatching refuter passes down, which is why the refuter's own instruction is the lever that exists.
- Every other `verbatim` dispatch found (`run-team/SKILL.md`'s member prompt block, fix-applier prompt, finisher prompt) is controller→member, not member→grandchild — outside the class the ticket names.

**Ruling: apply narrowly, as above** — refuters only, in both documents.

## Mutation testing (both directions)

- Deleting the clause (reverting both files to their pre-fix content) reddens all 8 new tests. Verified.
- A reflow-only mutation (reordering the two rules, rewrapping the nested `> >` blockquote so a wrap point lands inside a phrase) first **falsely reddened** 4/8 in `run-team/SKILL.md` — the quote gutter (`> `) isn't whitespace, so a literal-space regex can't span a wrap landing inside a pinned phrase. Fixed by stripping the blockquote gutter and rejoining with a single space before matching; re-verified the reflow then stays green while the pre-fix text still reddens.

## Test plan

- [x] `node --test skills/fleet/scripts/*.test.mjs` — 688/688 pass (not `tests 0`)
- [x] New prose pin mutation-tested both ways (see above)

Closes #258


feigi:
Review findings applied in `f5dcf43`.

**Applied (4):**
- **R-A** `important` / correctness — the toplevel guard was unsatisfiable before `git init` (fresh scratch dir → exit 128, `fatal: not a git repository`, no output). Split into two assertions with opposite expected outcomes, in both dictating documents. Pin at `refuter-scratch-prose.test.mjs` widened to cover both halves.
- **R-B / U5** `critical` / tests + comments — the scratch-path test's two unbounded `assert.match` calls let an exception spliced *between* the pinned phrases stay 8/8 green (measured). Collapsed to one contiguous span, matching the fix PR #488 made in the sibling `finisher-pin-race-prose.test.mjs`.
- **U1** `suggestion` / correctness — the header cited four motivating failures; the clause binds only the refuter's own writes, so the fix-applier `unv1`/`unv2`/`unv7`/`unv11` collision is not covered. Says so now instead of claiming coverage.
- **U3** `suggestion` / correctness — the `between()` citation pointed at the function body. Corrected to `review-pr-reads.test.mjs:271-275`, the rationale (re-derived from the file; the review's `:270-273` was one line off).

Plus **R-C**'s optional wording fix, folded into the same commit: "the actual PR #488 failure" → "the actual failure during the PR #488 run", which reads as one thing rather than two.

**Deferred (1):** #496 — `review-pr.js:664-676`, covering **R-D** and **U2** together. The workflow path hands every refuter of a dimension the same one-level `${scratch}/verify-${d.key}/`, and its prompt carries neither the cd-chaining rule nor the toplevel assertion. #258 scopes this path out; its stated rationale ("provisions `scratch` per run") covers the cross-PR collision but not a within-run collision, and has nothing to do with a silently failed `cd`.

**Not applied:** **R-C** and **R-D**'s substantive claims — both refutations accepted. PR #488 *is* the run the failure happened during, and the `review-pr` workflow path *is* out of #258's scope. **U4** was a clean bill (positive control: the `.{0,40}` join bound discriminates both ways); nothing filed.

Mutation-tested both ways in an isolated tree: the exception splice, the pre-fix guard, each toplevel half dropped alone, and the two halves swapped all go **red**; a reflow breaking a line inside a pinned span across the nested `> > ` gutter, and an unrelated rewording beside the clause, both stay **green**. The old two-regex form was re-run against the same splice and confirmed **green** — the defect reproduced before the fix closed it.

`node --test skills/fleet/scripts/*.test.mjs` → `tests 688 / pass 688 / fail 0`.

PR #498, opened 2026-08-16
docs(fleet): scope release-ticket's delete-time recomputation to the live-directory case

### LINKED ISSUE #87: run-team docs: `release-ticket.sh` no longer recomputes all four preconditions at the moment of the delete — SKILL.md:387 and reaping.md:34,43
Deferred from PR #72 review. Filed rather than edited because `SKILL.md` is owned by PR #71 this wave.

## The claims

`skills/fleet/skills/run-team/SKILL.md:387` — release-ticket "recomputes all four preconditions inside the same" invocation as the delete.

`skills/fleet/skills/run-team/references/reaping.md:34` — "Four preconditions, recomputed inside same invocation as delete, same reason reap recomputes".

`skills/fleet/skills/run-team/references/reaping.md:43` — "**`worktree remove` without `--force`** refuses on modified and untracked files ... Its refusal is the dirty check recomputed by git at the moment of the delete."

## Why they are now inaccurate

The first two are about recomputation *inside the invocation*, and remain true for the three branch-ref preconditions. They are wrong about the dirty check on the branch PR #72 added: when the worktree directory is established absent, the dirty check does not run at all.

`:43` is the load-bearing one and it is wrong in a sharper way. `git worktree remove` gates its clean check on the same `stat` the script does, so it is **not an independent second opinion** — on every input where the guard says "gone", git reaches the same conclusion by the same means rather than checking the work:

```
path stattable    -> fatal: contains modified or untracked files, use --force
path unstattable  -> rc 0, entry deregistered, files still on disk
```

That is what made the fail-open fixed in `1e54bac` survive being reasoned about: guard and backstop fail together on the same inputs.

## Desired

State the limitation the script header now states (`release-ticket.sh:15-20` as of `1e54bac`): the delete-time recomputation covers the live-directory case only, and where the path cannot be stat`ed the guard is sole arbiter. Text is settled; the only care needed is sequencing behind PR #71.

Verified on git 2.50.1.

feigi:
Closes #87

## What was wrong

Three documents said `release-ticket.sh` recomputes all four preconditions at the moment of the delete. Two things are wrong with that.

**The dirty check is the conditional one.** It opens inside `[ -n "$wt" ] && [ -d "$wt" ]`, so a claim whose worktree directory is established absent reaches the delete on three preconditions, not four. That is the only path that does: a directory git cannot stat dies on the unknown-existence guard above it, and a non-directory blocks.

**`git worktree remove` without `--force` is no backstop for the case it skips.** It gates its own clean check on the same `stat` the script's `gone` guard does, so it is not an independent second opinion — it reaches the same verdict by the same means. Measured here, git 2.50.1 (Apple Git-155):

| input | result |
|---|---|
| path stattable, untracked file present | rc 128, `contains modified or untracked files, use --force` |
| directory `rm -rf`'d | rc 0, entry deregistered |
| live worktree holding an untracked file, `chmod 000` parent | **rc 0, entry deregistered, file still on disk** |

The third row is the point: git deregistered a worktree that was holding uncommitted work, at rc 0, because it could not stat it either. Guard and backstop fail on the same inputs — which is what let the fail-open fixed in `1e54bac` survive being reasoned about.

## What this does not claim

The correction narrows the claim, it does not withdraw it. The delete-time refusal is still real wherever the directory is there to read, and `branch -d` still recomputes the branch-ref preconditions at the delete. Only on a path that cannot be stat'ed is `gone` the sole arbiter.

## Sites

Corrected — the three #87 names, re-resolved against the tree (its line numbers had drifted: `SKILL.md:387` is now `:991`, `reaping.md:34,43` are now `:35,:44`), plus one neighbour it does not name:

- `skills/fleet/skills/run-team/SKILL.md` — release-claims section
- `skills/fleet/skills/run-team/references/reaping.md` — the four-preconditions lead-in and the `worktree remove` bullet
- `skills/fleet/scripts/release-ticket.sh` — the comment at the dirty check, which said the unqualified thing 545 lines below the header that already said the qualified one

Left alone, deliberately:

- `release-ticket.sh`'s label-last comment — scoped to *refusals*, which only occur on the stattable path, so it is accurate as written
- reap's sibling prose (`reap.sh`'s header, `SKILL.md`'s reap paragraph, `reaping.md`'s reap section) — "recomputed inside this invocation" makes no delete-time claim, and "double-covers dirty check" is about breadth (modified *and* untracked), not independence from the stat; `reap.sh` routes cannot-stat to KEEP before its delete is reachable, so its delete is never sole arbiter of a path it could not stat
- `release-ticket.test.mjs`'s two comments — one describes a worktree that is live and dirty at delete time, the other already states the corrected property

## Test

New `skills/fleet/scripts/reaping-prose.test.mjs`: five presence pins over bounded slices, one per corrected claim, written red first and watched fail for the right reason before the prose was touched. Nothing pinned these sentences before, which is how they rotted.

`node --test skills/fleet/scripts/*.test.mjs` — 694 pass, 0 fail.


feigi:
## Review applied — 1 of 11 findings landed, 6 deferred, 3 needed no action

All 11 findings were in the `suggestion` band, which the review workflow budgets 0 refuters. Each in-scope one got its own refuter, briefed to refute and to verify by running rather than reasoning. Nine of ten checked findings did not survive that.

**Applied — `f5696fe`**

| finding | change |
|---|---|
| `unverified[9]` `simplify` | dropped the duplicate `See references/reaping.md. ` from `SKILL.md:997` — line 1006 already ends the same subsection with it, and 22 of the file's 23 real pointers sit end-of-line. 90 chars → 63. |

**Deferred and filed**

| # | finding | outcome | label |
|---|---|---|---|
| #500 | `[10]` `between()` is pasted identically in 3 test files | out of #87's scope — 2 of 3 copies are in files this ticket has no business editing | `ready-for-agent` |
| #502 | `[3]` rewrite `reaping.md:35`'s "Four preconditions" | refuted — the count introduces a 4-bullet list, and the rewrite is arithmetically false in this document's frame | `wontfix` |
| #504 | `[8]` collapse `release-ticket.sh:571-574` to a pointer | refuted — the header carries none of `rc 0`/`gone`/`this block`; the collapse deletes the *because* and keeps the *therefore* | `wontfix` |
| #505 | `[2]` rewrap `SKILL.md` 991-1001 | refuted *by* `[9]` — the 90-char line is 63 after this commit; no width linter exists and 51 lines already exceed 83 | `wontfix` |
| #506 | `[1]` pin the script header (15-25) | refuted on **scope only** — the header is a fifth site that never made the stale claim. The pin was built and proven to discriminate, so it is open for a maintainer call, not closed | `needs-triage` |
| #507 | `[0]` qualify or carve out `release-ticket.sh:612` | refuted — `:612` is an ordering clause whose subject is a refusal that already happened, and it fails the carve-out list's own admission test | `wontfix` |

**No action, nothing filed** — each verified rather than taken on the finding's word; all three declare `suggested_fix: none needed`:

- `[4]` `tests` — mutation audit passed: 5 targeted reverts each reddened exactly their own test.
- `[5]` `tests` — the trailing-addendum ceiling is disclosed verbatim in the test file's own `THE CEILING` block.
- `[6]` `types` — scope note; the diff declares zero types.
- `[7]` `types` — stands; it and `[3]` were mutually exclusive on the same line and went to one refuter, which ruled for `[7]`.

**Verification:** `node --test skills/fleet/scripts/*.test.mjs` → exit 0, `tests 694, pass 694, fail 0, cancelled 0`, run with an isolated `TMPDIR` after a concurrent-run collision was observed in a sandbox. No `--no-verify`.

PR #501, opened 2026-08-16
fix(fleet): inflight.sh's gh-stderr capture is mktemp'd, not guessable

### LINKED ISSUE #91: inflight.sh: predictable /tmp/.inflight.$$ is symlink-truncatable, and a failed capture reports an empty cause
## Problem 1 — the temp path is predictable and the redirect follows symlinks

`skills/fleet/scripts/inflight.sh:47` and `:74` capture `gh` stderr to a fixed, guessable path in a world-writable sticky directory:

```sh
... 2>/tmp/.inflight.$$
```

`>` follows symlinks, so anything that pre-plants `/tmp/.inflight.<pid>` pointing at a file the running user can write gets that file truncated. PIDs are a small, cheaply-guessable space.

**Measured:**

```
$ printf 'IMPORTANT DATA THAT MUST NOT BE LOST\n' > victim.txt   # 37 bytes
$ ln -sf .../victim.txt "/tmp/.inflight.$$"
$ { echo "some gh error" >&2; } 2>/tmp/.inflight.$$              # :47's exact construct
$ wc -c < victim.txt
14
$ cat victim.txt
some gh error
```

Real data destruction, not a theoretical race. `mktemp` fixes it, and is used **nowhere** in `skills/fleet/scripts/` (grepped: zero hits across the directory). There is also no `trap`, so an interrupt between `:47` and `:67` leaves the file behind.

## Problem 2 — a failed redirect reports an empty cause

`:48` and `:75` read the captured stderr back with `|| true`:

```sh
err=$(cat /tmp/.inflight.$$ 2>/dev/null || true)
```

If the redirect itself failed (unwritable `/tmp`, full filesystem, or the symlink above pointing somewhere unreadable), the script still **fails closed** — measured, exit 2 — but with nothing to act on:

```
inflight: gh issue view 42 failed, so #42's PR links are unknown:
```

Nothing after the colon. `release-ticket.sh:232,244` capture with `2>&1` instead and keep the cause.

## Suggested fix

Both problems are the same three lines of temp-file handling. Replace the fixed path with `mktemp`, add a `trap` to clean it up, and either keep the `|| true` but say "cause unavailable" when `$err` is empty, or switch to the `2>&1` capture `release-ticket.sh` uses.

Deferred from PR #76 review. Pre-existing, unrelated to that PR's mechanism, but verified during it.

feigi:
## Summary
- `inflight.sh`'s two `gh` stderr captures used a fixed, guessable `/tmp/.inflight.$$` path. `>` follows symlinks, so anything on the host that pre-plants that path pointing at a file this user can write gets it silently truncated when stderr lands there (measured in the ticket: a 37-byte victim file left at 14 bytes). Switched to `mktemp` plus a `trap 'rm -f "$errfile"' EXIT`, matching the precedent already in this repo (`no-undo-audit.sh:409`). The trap also covers an interrupt landing between the redirect and the old inline `rm -f`, which previously left the file behind.
- Separately, when the capture itself fails (unwritable `/tmp`, a full filesystem), `$err` used to read back empty — indistinguishable from `gh` genuinely printing nothing — and the operator-facing message trailed off after the colon (`...PR links are unknown: `). Empty now falls back to `"cause unavailable"`.

**Scope check (both halves of the class):**
- Capture sites: ticket named two (`gh issue view`, `gh pr list`), both inside `probe_pr()`. Re-grepped `skills/fleet/scripts/*.sh` for `/tmp/.inflight` and any other fixed-`$$`-style temp path — these were the only two occurrences in the whole directory; no sibling script shares this pattern. `mktemp` is already used correctly elsewhere in this directory (`no-undo-audit.sh`), so no third site to bring into line.
- Left alone, deliberately: `release-ticket.sh:620,632` already capture with `err=$(cmd 2>&1)` (no temp file at all) — not this defect's shape, nothing to change there.

## Test plan
- [x] `./agent-test skills/fleet/scripts/inflight.test.mjs` — 66/66 (64 pre-existing + 2 new), all green
- [x] `node --test skills/fleet/scripts/*.test.mjs` — full repo suite, 691/691 green
- [x] Both new tests mutation-killed: reverting either the `[ -n "$err" ] || err="cause unavailable"` fallback or the `mktemp ... || die` guard turns exactly one test red (verified independently), and a full revert to the old `/tmp/.inflight.$$` scheme reddens both new tests and nothing else
- [x] Negative-path regression left in place: existing suite already pins `taken=false` on a genuinely free ticket and exit 2 with a named cause on a real `gh` failure — both pass unchanged against the refactor

Closes #91


feigi:
Review findings applied in 5fe18ad.

**Applied (11 findings, 4 `survived` + 7 `suggestion`):** the EXIT trap now guards its `rm -f` with `|| echo` so cleanup cannot rewrite the verdict (`survived[0]`, critical); `mktemp` failure records `unknown:["pr"]` instead of aborting the run (`survived[2]`); the duplicated cause-readback is one `gh_cause` helper and the stub's `pr list` branch finally has a failure dial, so both call sites are pinned (`survived[1]`, `survived[3]`); the four intermediate `rm -f "$errfile"` calls are gone, which closes the 0600→0644 re-create window (`unverified` 0/1/7); the `twelve lines up` comment anchor is stable (`unverified` 3); and `issueFailSilent` retires in favour of set-ness on `issueErr` (`unverified` 8).

**Filed, no code change:** #511 — one finding refuted by measurement (its remedy reddens two of the PR's own pins), two no-ops per their own stated remedy.

Tests: `node --test skills/fleet/scripts/*.test.mjs` → 694 pass, 0 fail. Green under dash as well; `shellcheck -x -S warning` clean. Every new pin mutation-checked, with a control mutation that stays green.

PR #509, opened 2026-08-16
docs(fleet): name inflight.sh's exit 2, and brief ./agent-test as <file-or-dir>

### LINKED ISSUE #93: run-team/SKILL.md: the in-flight step says "any hit = taken" and never names inflight.sh's exit 2, which PR #76 makes far more reachable

### LINKED ISSUE #102: run-team docs still describe ./agent-test as file-only, so the directory form PR #75 added is undiscoverable
Deferred from PR #75 review.

PR #75 made `./agent-test <dir>` work. Two places still tell members the opposite shape, and they are the places members actually read:

- `skills/fleet/skills/run-team/SKILL.md:135` — "Brief members with `./agent-test <file>` and nothing else"
- `skills/fleet/skills/run-team/references/isolation.md:7` — same sentence

Neither is *falsified* by #75 — a file argument still works, so both remain literally true — but the PR's own motivation is that agents type a directory anyway and misread the resulting red as a finding against the diff under review. Leaving the docs file-only keeps the capability invisible to exactly the audience that ran into the bug.

Small: mention that a directory works and expands to the test files under it, and that a directory with no test files refuses rather than passing vacuously.

Note `run-team/SKILL.md` was owned by PR #71 in the same wave, which is why #75 did not touch it and why this is an issue rather than an edit.

feigi:
Closes #93
Closes #102

Two corrections to `skills/fleet/skills/run-team/SKILL.md`, taken as one job because two members on that file in one wave conflict at merge. Both tickets' sequencing caveats (#71/#73 "own this file") were stale — those merged long ago, as did #406 — so every citation below was re-resolved against the tree rather than restated from the issue bodies.

## #93 — the in-flight step never named exit 2

The step said only "any hit = taken". In `sh`, both `cmd || handle` and `if cmd; then` treat 1 and 2 alike, so the natural reading files an unanswerable probe under "taken" by accident rather than by rule. The added paragraph carries the four rules `cc60e73` (PR #406) made necessary, each checked against `inflight.sh` itself:

| Rule | Settled at |
|---|---|
| exit 2 from a probe failure prints the same JSON as 0 and 1 plus `unknown[]` | `inflight.sh:14-15`; the single unconditional `printf` at `:703` that every rc reaches |
| `taken` is `false` on that payload — a summary of `hits`, not a free claim | `inflight.sh:548` |
| a hit outranks an unknown, so `hits[]` and `unknown[]` can both be non-empty at exit 1 | `inflight.sh:540-546` |
| four causes print nothing at all, so exit 2 is not assumed to parse | `die()` at `inflight.sh:36`, reached by a bad argument (`:38`, `:40`), not inside a git repository (`:42`), no such issue (`:109`), verdict unwritable (`:705`) |

The last row is what makes the other three safe to state — without it the step would read as "exit 2 always carries a payload", which is false.

The pre-existing sentence "It runs all three probes … so a partial one cannot read as free" is kept unchanged: it was *not* true before #406 (a first-probe failure aborted the run) and `cc60e73` is what made it true.

## #102 — the runner was briefed as file-only

PR #75 made `./agent-test <dir>` work; the two documents members actually read still said `<file>` and nothing else, keeping the capability invisible to exactly the audience #75 was for. Measured against this worktree's own runner before documenting:

- a directory holding one test file ran it — `ℹ tests 1`, exit 0
- an empty directory printed `agent-test: no test files under <dir>` and exited 1

Both halves are stated. The refusal is the load-bearing one: appending nothing leaves argv empty and bare `node --test` then discovers the whole worktree — a green for a suite nobody asked for (`agent-test`'s own comment on that branch).

The same clause is now verbatim in `run-team/SKILL.md` and `references/isolation.md`.

## Class enumerated

**#93 — every document mentioning `inflight.sh`** (`grep -rn --include='*.md' inflight`):

- `run-team/SKILL.md` — **fixed**, the defect.
- `next-ticket/SKILL.md:37,39` — the ticket's claim that it is already correct **verified**: "0 free, 1 taken, 2 unanswerable" plus "Exit 2 is not free … treat it as taken until you know". It reads the exit code and never tells anyone to parse stdout, so the payload rules do not apply to it. **Left, and now pinned** so the two documents cannot drift apart about one script.
- `docs/specs/2026-07-23-fleet-plugin-design.md:181` — the ticket's claim that #45 covers `docs/specs/` **verified**: #45 is CLOSED and the script table now names `unknown[]`, the shared payload shape, and all four payload-less causes correctly. **Left, already correct.**
- `docs/specs/2026-07-22-run-team-agent-fleet-design.md:24` — a superseded-notice giving the exit codes only; accurate, and a historical record. **Left.**

**#102 — every document mentioning `agent-test`:**

- `run-team/SKILL.md:176-182` and `references/isolation.md` — **fixed**, both sites the ticket names.
- Deliberately left, because none claims an argument shape: the "filesystem isolation is not stack isolation" guard in both documents, the reused-worktree note (`run-team/SKILL.md`), `review-and-fix.md`'s specialist test command, `references/reaping.md`'s note that the runner is an ignored file, the two `docs/specs/` mentions, and `.out-of-scope/`.

## What this could get wrong

The failure mode of a correction ticket is a new overclaim, so both directions were checked: exit 2 is **not** described as always carrying a payload, `taken:false` is **not** described as free, and the directory form is **not** described as unconditional — `agent-test` also refuses a directory under `node_modules` and one whose subtree it cannot fully read, neither of which this diff claims either way.

## Tests

Two new prose pins, both red before the edit and green after — `skills/fleet/scripts/inflight-exit2-prose.test.mjs` (4 red → green, plus the `next-ticket` pin that was green from the start, confirming the ticket's claim) and `skills/fleet/scripts/agent-test-dir-prose.test.mjs` (6 red → green). Presence pins over a bounded slice, same ceiling as `refuter-scratch-prose.test.mjs`.

Full suite: **711 pass, 0 fail** — run as `./agent-test skills/fleet/scripts`, the directory form this PR documents.


feigi:
Review record: #512 (closed `wontfix`) — 1 refuted finding narrowed and applied (`bbfd3fd`), 1 unverified attribution confirmed correct. No other findings.

PR #510, opened 2026-08-16
fix(fleet): anchor claim-ticket.test.mjs's five remaining bare pass-count pins

### LINKED ISSUE #127: claim-ticket.test.mjs: five pass-count assertions are unanchored substrings, so they stop discriminating once a fixture grows

feigi:
`assert.match(r.stdout, /pass N/)` is an unanchored substring against node's reporter output — it also matches `pass N0`, `N1`, … `N9`, so the pin silently stops discriminating once a fixture grows past its decade (`/pass 1/` is worst: matches 1, 10-19, 100+).

Five sites in `skills/fleet/scripts/claim-ticket.test.mjs` still had this shape (base line numbers 111, 122, 131, 310, 321 — drifted from the ticket's :101/:112/:121/:136/:147). Each now reads `/(?:ℹ|#) pass N(?!\d)/`: the reporter prefix pins the match to a summary line, the lookahead rejects `pass N0`-`pass N9`.

**Why not `/^(?:ℹ|#) pass N$/m`, the ticket's shape and this file's convention.** Review caught that the line anchors are a regression this PR would have introduced: `apply()` builds the child env as `{...process.env}` minus only `NODE_TEST_CONTEXT`/`NODE_TEST_WORKER_ID`, so `FORCE_COLOR` reaches the `spawnSync`'d runner, which colors its own summary even into a pipe — `\x1b[34mℹ pass 3\x1b[39m` starts with ESC and ends with `[39m`, defeating both `^` and `$`. Measured on this file: `FORCE_COLOR=1` took head to `tests 58 / pass 29 / fail 29` against base's `34 / 24`, a delta of exactly these five tests by name. The `(?!\d)` form keeps every discrimination the anchor buys and survives color (verified against real captured bytes, not typed strings).

The dual `(?:ℹ|#)` prefix is kept rather than the ticket's literal `/^ℹ pass N$/m` suggestion (PR #109's original form): this file already standardized on the dual-prefix pattern at 18 other sites (17 literal `/^(?:ℹ|#) pass N$/m` regexes plus the templated `new RegExp(`` `^(?:ℹ|#) pass ${count}$` ``, "m")` at :245), because node's default reporter is version-dependent — tap (`# pass N`) on older node writing to a pipe, which is every CI run, spec (`ℹ pass N`) on newer. The narrower single-prefix form would false-refuse under an older node in CI. Matched the existing convention for consistency and correctness.

## Scope swept

Grepped every `*.test.mjs` in the repo for `assert.match` against `pass|fail|tests` without a `^…$/m` anchor. Only other survivor: `review-path-default.test.mjs:341`, `assert.match(step3, /tests 0/, …)` — asserts a literal phrase in prose (a skill doc's markdown text, already scoped to one section), not a live reporter count subject to decade growth. Different class, left alone.

## Verification

- Assertions re-checked against real reporter output (spawned via `spawnSync`, not a TTY) on node v26.5.0 — confirmed `ℹ pass N` format; the `#` alternative covers older node in CI per this file's own existing comment.
- Mutation test: changed all 5 assertions' expected counts to wrong values, ran the file — exactly those 5 named tests went red, the other 53 stayed green. Reverted — all 58 pass again.
- Colored runs, node v26.5.0, all piped: `FORCE_COLOR=1` and `=3` both give `tests 58 / pass 34 / fail 24` after the change (from `29 / 29` under the anchored form); uncolored `58 / 58 / 0`. A wrong-count mutant at one of the five goes red both colored and uncolored.
- **Not a CI fix, and the file is not color-clean.** CI runs a non-TTY bash step and nothing in the repo sets `FORCE_COLOR`/`NO_COLOR`, so CI never saw this — it is developer-environment robustness. Nine other assertions in this file remain color-sensitive for unrelated pre-existing reasons; filed as a follow-up. Untested: a real pty, color forced from outside the repo text, and any node old enough to emit the tap `#` prefix.
- Full repo suite: 700/700 pass.

Closes #127


feigi:
Review findings adjudicated. Deferrals filed:

- **#513** — aggregate record for the one refuted finding ("version-dependent" → "version- and TTY-dependent", `comments` @:113, refuted 2-0). Filed `wontfix`, closed, no action wanted. Also lists the residue that *was* applied, so the closed record is not read as "the review changed nothing".
- **#514** — `ready-for-agent`. Nine assertions in `claim-ticket.test.mjs` still fail under a colored reporter for reasons unrelated to pass-count matching. Out of #127's scope; this PR fixed the five it owned plus the color regression its own anchoring introduced.

Applied here: the impossible `pass <n>1<n>` clause dropped (`0c86ac5`), the five pins moved from `/^(?:ℹ|#) pass N$/m` to `/(?:ℹ|#) pass N(?!\d)/` (`e4e857e`), and the PR body's site count corrected from 17 to 18.

PR #515, opened 2026-08-16
fix(fleet): refuse a zero-padded issue number, and cite siblings by construct where the line had rotted

### LINKED ISSUE #121: inflight.sh: a zero-padded issue number emits invalid JSON
Deferred from PR #110 review. Pre-existing — identical at the merge base.

`skills/fleet/scripts/inflight.sh:17` accepts any all-digit argument, and `:240` interpolates it into a JSON *number* slot with `%s`. All-digits is not a JSON number: RFC 8259 forbids leading zeros.

```
$ sh inflight.sh 007
exit=0  {"issue":007,"taken":false,...}   => UNPARSEABLE
$ sh inflight.sh 0     -> parses
$ sh inflight.sh 42    -> parses
```

Low reachability — callers pass numbers straight from `gh`, which never zero-pads — and evidence-only, since the exit code is unaffected. Filed because PR #110 added a comment at `inflight.sh:223-224` asserting `$n` needs no wrapping *because* it is all-digits, which is exactly the reasoning this falsifies. That comment was corrected in #110; the code was left alone to keep that PR decision-invariant.

## Fix

```sh
case "$n" in ''|*[!0-9]*|0?*) die "issue must be a number, got '$n'";; esac
```

**Do not use `n=$((n))`.** Verified octal trap: `$((010))` is 8 — a silently wrong issue number, worse than the bug. `$((08))` errors outright.

Same `"issue":%s` shape with the same guard exists in `claim-ticket.sh` (1 site) and `release-ticket.sh` (3 sites); worth sweeping in one pass.

### LINKED ISSUE #129: fleet scripts: four cross-file line citations in comments point at the wrong line

feigi:
Closes #121
Closes #129

Taken as a set: both edit `skills/fleet/scripts/inflight.sh`, so two members on
that file in one wave would conflict at merge.

## #121 — a zero-padded issue number emits invalid JSON

All-digits is not a JSON number. RFC 8259 forbids a leading zero, so `007`
cleared the `''|*[!0-9]*` guard, reached a `"issue":%s` printf and emitted
`{"issue":007,…}` at exit 0 — unparseable, with an exit code giving the caller
no hint.

**The class, re-derived rather than taken from the ticket.** The ticket names
"`claim-ticket.sh` (1 site) and `release-ticket.sh` (3 sites)". Those are
*emission* counts. The fix site is the **guard**, and there is one guard per
script covering every payload it emits — so four guards cover six emissions:

| guard | emissions it protects |
|---|---|
| `inflight.sh:40` | `:751` |
| `claim-ticket.sh:23` | `:371` |
| `release-ticket.sh:46` | `:250`, `:585`, `:655` |
| `drop-merged-label.sh:28` | `:52`, `:91` — **not named by the ticket** |

`drop-merged-label.sh` is the same defect in a differently-named field
(`"pr":%s`). It was found by sweeping the *shape*, not the field name.

**Search scope, and what the pattern would have missed.** `git grep` over the
working tree for `"issue":%s` finds only `printf`-built JSON in `.sh`. That
pattern alone would have missed JSON built by string concatenation and JSON
built in `.mjs`, so both were swept separately. The concat sweep is what
surfaced `drop-merged-label.sh`'s inner `{"issue":$n,…}`.

**Deliberately left, with the reason:**

- `.github/workflows/rebase-check-refresh.yml`'s `BEHIND_BY` guard — reads
  `gh api` output into `[ -eq ]`, never reaches a JSON slot.
- `drop-merged-label.sh`'s inner `{"issue":$n,…}` — `$n` comes from
  `closingIssuesReferences[].number`, which gh emits unpadded.
- `prove-merge.sh`'s `parentCount`, `worktree-audit.sh`'s `ahead`,
  `claim-ticket.sh`'s `ports` — all derived internally (`git rev-list --count`,
  arithmetic), never caller-supplied.

**Not `n=$((n))`.** Measured under `/bin/sh`, which is what these scripts run:
`$((010))` is **8** and `$((08))` errors outright. Normalising would answer
about a different ticket instead of refusing. (Measured under `zsh` it is 10 —
the interactive shell does not transfer.) The guard also protects
`claim-ticket.sh`'s `$((16000 + issue))` port arithmetic, which a padded number
could otherwise point at another claim's ports.

**What the guard must not refuse.** `gh` never zero-pads, so every legitimate
caller passes a bare number and must still be answered — an over-refusing guard
would strand the whole fleet at exit 2. Both directions are pinned: `007` is
refused at exit 2 with no payload, and an unpadded `7` still reaches a
parseable verdict emitting `"issue": 7`.

## #129 — cross-file citations that had rotted

Every `file:line` in the issue's tables was re-derived against this tree; the
issue's numbers are all stale. Its own history is the argument for dropping
them — three rows had been wrong, fixed, and wrong *again* inside one issue's
lifetime.

Converted **only** the citations measurement showed wrong, each replacement
verified to grep to one construct:

| citing site | was | now |
|---|---|---|
| `inflight.sh` | `release-ticket.sh:173` | its `git ls-remote --heads` lookup |
| `inflight.sh` | `release-ticket.sh:78` | its own `substr($0,10)` awks |
| `inflight.sh` | `release-ticket.sh:115` | its `jstr()` |
| `inflight.test.mjs` | `release-ticket.test.mjs:30` | `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` |
| `inflight.test.mjs` | `release-ticket.sh:78` | its own `substr($0,10)` awks |
| `inflight.test.mjs` | `release-ticket.test.mjs:729` | its hostile-name payload cases |
| `inflight.test.mjs` | `release-ticket.test.mjs:743` | "a control character in the worktree name…" |
| `candidates.test.mjs` | `candidates.mjs:239` | its `dropSpecs` |
| `candidates.test.mjs` | `run-team/SKILL.md:61` | "do NOT pass \`--allow-fallback\`" |

**Six more the issue never listed**, found by resolving *every* cross-file
citation under `skills/fleet/scripts/` rather than by re-reading its tables:
`inflight-exit2-prose.test.mjs` pointing at `inflight.sh:703`, `:548`, `:540`,
`:109` and `:705`; and `implementer-model-tier.test.mjs` pointing at
`ledger.mjs:131` and `compute-board.mjs:168`.

`release-ticket.test.mjs:743` was the sharpest case: it landed on real,
plausible prose about `git worktree add`, so a reader would have read the wrong
paragraph and *believed* it rather than bouncing off an obvious mismatch.

**Left as line numbers, deliberately** — they are CORRECT at this tree, and
correcting a correct citation is a different, larger refactor:
`inflight.sh` → `release-ticket.sh:73`, `no-undo-audit.sh` →
`claim-ticket.sh:26`, `reaping-prose.test.mjs` → `release-ticket.sh:15-25`,
`candidates.test.mjs` → `next-ticket/SKILL.md:15` and `run-team/SKILL.md:60`,
and the `review-pr-reads` / `select-dimensions` / `review-pr-testcmd` ranges.

**Left untouched, not citations at all:** `file.mjs:164` in `ledger.mjs` and
`candidates.mjs:164` in `ledger.test.mjs` are literal fixtures pinning that a
`file:line`-shaped string does not reach `gh` as a search qualifier. Editing
them would have broken the tests.

**This PR demonstrates the rot it fixes.** The #121 commit added six lines to
`release-ticket.test.mjs`, moving the control-character test from `:1388` to
`:1394` — a citation invalidated by an unrelated commit in the same PR.

## Verification

- Full suite: **720 tests, 720 pass, 0 fail**.
- Each new test kills only its own mutant: dropping `|0?*` from each guard in
  turn reddens exactly one test (`fail 1`) in that script's file.
- Restored by `cp` from a pre-mutation copy, never `git checkout -- <path>`,
  which writes the index too and makes the mutation measure nothing.
- Two controls: rewriting the guard to the behaviourally identical `0[0-9]*`
  stays **green** (71 pass), proving the tests pin behaviour and not the literal
  `0?*` text; an over-refusing `[0-9]*` turns the suite **red**, proving the
  accept-half test has teeth.


feigi:
Review findings adjudicated. Head is now `0869d68`; suite **720/720** at that SHA (`node --test skills/fleet/scripts/*.test.mjs`).

**Applied (4 findings, 3 commits)**

- `a237cb0` — all four zero-pad tests hard-coded `007`, so none told the shipped guard `0?*` from a narrower `0??*`: that mutant left all four suites green (59/59, 14/14, 71/71, 64/64) while re-admitting `01`. Each now runs both widths; measured per file with the mutant applied — one failure each, the zero-pad test — restored 208/208. Same commit moves the new `claim-ticket.test.mjs` test off the comment block it orphaned and renames it off the `usage:` prefix no other test in that file carries.
- `dcf0ef9` — the citation swaps left four comment paragraphs unfilled (`# which tests $remote`, `# read the same field:`, `// over a verified-zero baseline`, `// it out. Probe 2 only became a`). Reflowed whole; word sequences unchanged, widest added line 80 chars, nothing cites these files by line.
- `0869d68` — the new comment's claim that "only this direction catches" an over-refusing guard is false: 69 of `inflight.test.mjs`'s 71 tests go red under one. Corrected; the test is kept because rewriting the verdict printf to `{"issue":"%s"` fails that test **alone**, making it the file's only pin on `issue` as a JSON number.

**Filed**

- #516 (`needs-triage`) — the citation sweep is incomplete: 5 line citations remain in files this PR touched, 20 across `skills/fleet`, 23 more elsewhere in the repo. Filed as a record: the finding's remedy (drop the 12 citation hunks) was **rejected** — those hunks are #129's deliverable, and this PR closes #129 as well as #121, which neither refuter could see from the diff alone.
- #517 (`wontfix`, closed) — the refused band: 3 refuted with evidence, 1 survived but declined with its residue applied, 3 resolved to no action. Includes the measurement that rewording all four guard messages to name the leading zero is drop-in at 720/720, for whoever wants an operator-usability ticket.

PR #520, opened 2026-08-16
test(fleet): pin the accept side of #121's numeric guard — a bare 0 must still clear it

### LINKED ISSUE #518: fleet scripts: the accept side of #121's numeric guard is pinned only in an unpushed local commit (a0a893b)
Test coverage produced during PR #515's fix pass, **committed locally but never pushed**, so it is not in `94acb8f` and exists nowhere but one worktree on one machine. Recording the patch here so it survives that worktree.

## What it is

Four tests, one per guarded script, pinning the **accept** side of #121's numeric guard: a bare `0` must still clear it.

The guard shipped in #515 is `case "$n" in ''|*[!0-9]*|0?*)`. The `0?*` arm refuses `007` and `01` while letting `0` through, which is correct and deliberate — **#121's own body lists `sh inflight.sh 0 -> parses` among its PASSING cases**, beside `42`. A bare `0` is a valid RFC 8259 number and `$((0))` is `0`, so neither hazard the guard exists to close applies to it.

**Nothing is broken.** The merged code already behaves as these tests assert; this is unmerged *coverage*, not a pending fix. Verified against the merged scripts: `sh inflight.sh 0` clears the guard and reaches the `gh issue view` probe, `sh inflight.sh 007` dies with `issue must be a number, got '007'`.

## Why it matters

The reject side is pinned (`007` and `01`, added in #515 after a specialist measured that a `0??*` mutant left all four test files green while re-admitting `01`). The accept side is not. Without these, a future widening of `0?*` to `0*` would refuse the ticket's own worked example and **the whole suite would stay green** — measured during the review: `sed 's/|0?\*)/|0*)/'` across all four scripts left 720/720 passing, because nothing covers bare `0`.

So this closes the other boundary of the same guard.

## How it was lost

The fix-applier wrote and committed it after its final report, on `fix/121-zero-padded-issue`, which GitHub deleted on merge. The commit is `a0a893b` and it is **upstream-gone, unpushed**. `reap.sh` correctly refuses to delete that worktree (`KEEP fix/121-zero-padded-issue — unmerged commits`), so it is safe on disk — but only on that disk.

## The patch

Apply to a fresh branch off `main`; the four target files were all touched by #515, so re-derive the anchors rather than trusting the hunk offsets.

```diff
diff --git a/skills/fleet/scripts/claim-ticket.test.mjs b/skills/fleet/scripts/claim-ticket.test.mjs
index 8fb8382..3ba65db 100644
--- a/skills/fleet/scripts/claim-ticket.test.mjs
+++ b/skills/fleet/scripts/claim-ticket.test.mjs
@@ -98,6 +98,15 @@ test("a zero-padded issue is refused", () => {
   }
 });
 
+test("a bare 0 clears the numeric guard", () => {
+  // `0?*`, not `0*`: #121 lists `sh inflight.sh 0 -> parses` among its PASSING
+  // cases, beside `42`. A bare `0` is a valid RFC 8259 number and `$((0))` is
+  // `0`, so neither hazard the guard exists to close applies to it. Widening
+  // the arm would refuse a value the ticket's own worked example shows working.
+  const r = spawnSync("sh", [SCRIPT, "0", "slug", "fix"], { cwd: tmpdir(), encoding: "utf8" });
+  assert.doesNotMatch(r.stderr, /issue must be a number/);
+});
+
 const PASSES = 'import { test } from "node:test";\ntest("ok", () => {});\n';
 // `root.test.mjs` exists so no assertion below can be satisfied by the argv
 // being dropped: bare `node --test` discovers the whole fixture, and every
diff --git a/skills/fleet/scripts/drop-merged-label.test.mjs b/skills/fleet/scripts/drop-merged-label.test.mjs
index 4f2e350..7dc0507 100644
--- a/skills/fleet/scripts/drop-merged-label.test.mjs
+++ b/skills/fleet/scripts/drop-merged-label.test.mjs
@@ -115,6 +115,16 @@ test("usage: rejects a zero-padded PR number", (t) => {
   assert.deepEqual(s.calls(), [], "must refuse before calling gh at all");
 });
 
+test("usage: a bare 0 clears the numeric guard", (t) => {
+  // `0?*`, not `0*`: #121 lists `sh inflight.sh 0 -> parses` among its PASSING
+  // cases, beside `42`. A bare `0` is a valid RFC 8259 number and `$((0))` is
+  // `0`, so neither hazard the guard exists to close applies to it. Widening
+  // the arm would refuse a value the ticket's own worked example shows working.
+  const s = stub(t);
+  const { stderr } = run("0", { env: s.env() });
+  assert.doesNotMatch(stderr, /pr must be a number/);
+});
+
 test("usage: rejects an unknown second argument", (t) => {
   const s = stub(t);
   const { code, stderr } = run(9, { apply: false, env: s.env() });
diff --git a/skills/fleet/scripts/inflight.test.mjs b/skills/fleet/scripts/inflight.test.mjs
index 6b989f2..05f742e 100644
--- a/skills/fleet/scripts/inflight.test.mjs
+++ b/skills/fleet/scripts/inflight.test.mjs
@@ -1588,6 +1588,15 @@ test("accumulate: a zero-padded issue is refused before any probe runs, with no
   }
 });
 
+test("accumulate: a bare 0 clears the numeric guard", (t) => {
+  // `0?*`, not `0*`: #121 lists `sh inflight.sh 0 -> parses` among its PASSING
+  // cases, beside `42`. A bare `0` is a valid RFC 8259 number and `$((0))` is
+  // `0`, so neither hazard the guard exists to close applies to it. Widening
+  // the arm would refuse a value the ticket's own worked example shows working.
+  const r = spawnSync("sh", [SCRIPT, "0"], { encoding: "utf8" });
+  assert.doesNotMatch(r.stderr, /issue must be a number/);
+});
+
 test("accumulate: an unpadded issue number still reaches a parseable verdict", (t) => {
   // The other half of #121's guard, and the half that would strand the fleet if
   // it were wrong: `gh` never zero-pads, so every legitimate caller passes a
diff --git a/skills/fleet/scripts/release-ticket.test.mjs b/skills/fleet/scripts/release-ticket.test.mjs
index 69a8b23..e372b20 100644
--- a/skills/fleet/scripts/release-ticket.test.mjs
+++ b/skills/fleet/scripts/release-ticket.test.mjs
@@ -876,6 +876,16 @@ test("usage errors exit 2", (t) => {
   }
 });
 
+test("a bare 0 clears the numeric guard", (t) => {
+  // `0?*`, not `0*`: #121 lists `sh inflight.sh 0 -> parses` among its PASSING
+  // cases, beside `42`. A bare `0` is a valid RFC 8259 number and `$((0))` is
+  // `0`, so neither hazard the guard exists to close applies to it. Widening
+  // the arm would refuse a value the ticket's own worked example shows working.
+  const r = repo(t);
+  const zero = spawnSync("sh", [SCRIPT, "0", "slug", "fix"], { cwd: r.w, env: r.env(), encoding: "utf8" });
+  assert.doesNotMatch(zero.stderr, /issue must be a number/);
+});
+
 test("a chatty but successful gh does not fake an already-dropped label", (t) => {
   // The label was read with 2>&1 and substring-matched, so gh's own upgrade
   // notice on a SUCCESSFUL call broke the match: the script deleted the worktree
```

feigi:
Closes #518

Recovers test coverage written during PR #515's fix pass that was committed locally after the fix-applier's final report, stranded when #515 merged and GitHub deleted the branch, and preserved as a patch in #518.

## What

Four tests, one per guarded script, pinning the **accept** side of #121's numeric guard: a bare `0` must still clear it.

The shipped guard is `case "$n" in ''|*[!0-9]*|0?*)`. The `0?*` arm refuses `007` and `01` while letting `0` through — correct and deliberate, because **#121's own body lists `sh inflight.sh 0 -> parses` among its PASSING cases**, beside `42`. A bare `0` is a valid RFC 8259 number and `$((0))` is `0`, so neither hazard the guard exists to close applies to it.

**No behaviour changes. Nothing was broken.** The merged code already behaves as these tests assert; this is coverage, not a fix.

## Why it is worth having

#515 pinned the reject side (`007` and `01`), added after a specialist measured that a `0??*` mutant left all four test files green while re-admitting `01`. The accept side had no pin. Without it, widening `0?*` to `0*` would refuse the ticket's own worked example and **the whole suite would stay green** — measured during that review: `sed 's/|0?\*)/|0*)/'` across all four scripts left 720/720 passing.

This closes the other boundary of the same guard.

## Verification

- Suite on top of current `main`: **724/724**.
- Mutation, per file — guard widened to `0*`, restored by `cp` from a pristine copy: `claim-ticket` 59/60, `drop-merged-label` 14/15, `inflight` 71/72, `release-ticket` 64/65. One red each, and it is the new test. Restored: 212/212.
- `range-diff` against the stranded original confirms identical content.


feigi:
Review applied in 300131a — 3 findings, 3 applied, 0 deferred, so no follow-up issues filed.

| # | dim | file:line | outcome |
|---|-----|-----------|---------|
| S0 | correctness | `inflight.test.mjs:1596` | applied — bare spawn replaced with the file's `inflight()` fixture; it was leaking a live `gh issue view 0` and `git ls-remote --heads origin` on every suite run inside a git repo |
| S1 | comments | `inflight.test.mjs:1603` | applied — counts re-derived against the post-S0 tree, not copied from the finding |
| U0 | correctness | `drop-merged-label.test.mjs:118` | applied — survived one refuter; extended to `claim-ticket` / `release-ticket` too, see below |

**S1's counts were re-measured, not copied.** S0 rewrites the body of the very test whose redness the count depends on. With S0 applied and the guard widened to refuse every digit string (`case "$n" in ''|*[!0-9]*|*)`): `tests 72 / pass 2 / fail 70`. Guard restored, verified with `git status --porcelain`. The same comment carried a second claim S0 falsified — that this is the file's only JSON-number assertion and the printf rewrite reds it alone; measured, that rewrite now reds two tests, so the clause was corrected as well.

**U0 went wider than its own suggested fix.** The finding proposed `assert.notDeepEqual(s.calls(), [], …)` for `drop-merged-label` only, arguing a positive assertion in the other two costs more fixture than it is worth. Measured, both premises needed adjusting:

- `s.calls()` only proves gh was *reached* — a refusal inserted after the first gh call leaves it non-empty and the assertion green. `assert.equal(code, 0, …)` costs the same line and reds both the early and the late refusal.
- `claim-ticket` and `release-ticket` need no fixture at all: each already dies at a *named* later precondition, so one `assert.match` on that message (`not inside a git repository`, `no branch fix/0-slug`) is a positive proof of acceptance. `fix/0-slug` additionally shows `0` arrived there as the issue number.

**Mutation evidence, per file, run in the worktree.** The mutant U0 exists for — an early refusal of `0` worded outside the guard's own message — was invisible to all 724 tests before this commit. After it, each of the four reds exactly its own bare-0 test and nothing else. Discrimination checked in the other direction too: widening `0?*` to `0*` reds them, narrowing to `0??*` leaves them green and reds only the zero-padded siblings.

Full suite from the worktree at 300131a: `tests 724 / pass 724 / fail 0`.

PR #521, opened 2026-08-16
fix(fleet): dedupe no-undo-audit.sh atRisk[] on the abbreviated SHA

### LINKED ISSUE #148: no-undo-audit.sh: above ARG_MAX xargs splits the pathspec list and atRisk repeats each commit once per batch
Deferred from PR #131 review. Introduced by that PR, in the fix that replaced the word-split pathspec — the fix is right, this is a rough edge on it.

`skills/fleet/scripts/no-undo-audit.sh:150-152` feeds the conflicting paths to `git log` through `xargs -0`. Above `ARG_MAX` (1048576 on macOS, `getconf ARG_MAX`) xargs runs the child **more than once**, and each invocation reports the commits touching its own batch. A commit touching paths in several batches is then printed once per batch.

Measured at `c17d4b6` — 20 conflicting files, one main commit touching all of them, with `xargs -s 300` forcing the split that a real 1 MiB pathspec list would cause:

```
normal xargs:      conflicts 20  atRisk ['58e874f ONE MAIN COMMIT']
xargs -s 300:      conflicts 20  atRisk ['58e874f ONE MAIN COMMIT', ... x10]
```

Reaching this for real needs roughly 4500 conflicting paths of ordinary length, so it is rare — but `atRisk` is read by a human deciding what a resolution would delete, and 4500 repetitions of one line is not a list anyone reads. The direction is safe (over-reporting, never under-), which is why this is a follow-up and not a blocker.

Pre-PR the same input was worse: the unquoted pathspec blew `ARG_MAX` outright and the `2>/dev/null || true` swallowed the `E2BIG`, so `atRisk` came back empty. This ticket is about polish on top of a real fix.

## Suggested fix

Dedupe on the abbreviated SHA, order preserved, after the xargs:

```sh
| awk '!seen[$1]++'
```

One line, no new process in the common case, and it is a no-op when xargs did not split.

A test can pin it without building a 1 MiB fixture: run the pipeline with a small `xargs -s` in a fixture with several conflicting paths and one shared commit, exactly as measured above. Worth checking whether making the batch size injectable purely for the test is worth it, or whether asserting `atRisk` has no duplicate SHAs on the ordinary fixture is enough.

feigi:
## Summary

Above `ARG_MAX`, `xargs -0` in `no-undo-audit.sh` splits the conflicting-paths pathspec list across more than one `git log` invocation, and each invocation reports every commit touching its OWN batch — so a commit whose changes span more than one batch was named once per batch in `atRisk[]` instead of once. Deferred from PR #131's review as polish on a fix that was otherwise correct (over-reporting, never under-reporting — a real distinct entry was never dropped, which is why this shipped as a follow-up, not a blocker).

Fix: dedupe on the abbreviated SHA (`--oneline`'s first field), order preserved, via `awk '!seen[$1]++'` — run as its **own statement** over the already-captured `$at_risk` string, not appended into the `xargs` pipe. Appended into the pipe, `awk` becomes the pipe's last command, and with no `pipefail` in POSIX `sh` (dash rejects `set -o pipefail` outright) the existing `|| die` guard would then read `awk`'s exit status instead of `git log`'s — silently swallowing a real `git log` failure. That's the exact regression shape #265 shipped (a guard moved to the wrong side of a pipeline, correct path unaffected, error path broken).

**Enumerated and covered:**
- duplicate SHA across xargs batches (the named bug) — deduped
- an empty `atRisk` (no commits at risk) — verified it stays empty through the new awk pass, not turned into a stray blank line
- `git log` failure inside the `xargs` pipe — verified still caught by the existing `|| die`, unaffected by the new dedupe statement (it's outside that pipe)

**Deliberately not addressed:** two genuinely distinct commits abbreviating to the same SHA. Git's abbreviation is computed against the full object set at call time, not per-batch, so two different commits colliding on the same abbreviated prefix within one `git log` call (or across xargs batches, which query the same object set) is not a case this bug or fix touches.

**Test arrangement:** the ticket left open whether to make the xargs batch size injectable in the script for testing, or just assert no duplicate SHAs on an ordinary fixture. Chose neither literally — instead of modifying production code to accept a test-only batch-size override, the test shadows `xargs` on `PATH` with a wrapper that forces `-s 300`, forcing the real ARG_MAX-split code path at 20 fixture files instead of the ~4500 a real split needs. No production code changes for test support, and the test exercises the actual unmodified `xargs -0` invocation.

## Verification

- Confirmed the defect present at `origin/main` before any change (unfixed script gave 4-25 duplicate `atRisk` entries under a forced small-batch split, verified both via a standalone bash spike and via the new test run against `git show origin/main:...`).
- New test confirmed **failing** against the unfixed script (4 duplicate entries), then confirmed **passing** against the fix.
- Full suite: `./agent-test` — 721/721 passing.

Closes #148


feigi:
PR #521 review — apply/defer split, at `beaf394`.

**Applied (one commit, `beaf394`)**

- **S0 / S1 / R0 / U5** (tests + silent-failure, `no-undo-audit.test.mjs:644`) — one defect, one edit. The fixture now carries **two** distinct main commits interleaved through the pathspec list (evens from one, odds from the other), so every batch reports both and the concatenation is `B A B A` rather than `A A B B`; their subjects share a first word; and `withSplitXargs` shadows `git` alongside `xargs` to count invocations carrying a `:(literal)` pathspec — the one call xargs drives — which the test asserts is `> 1`. Measured 4 batches here. Every mutant the review named now goes red: `uniq` → 8 subjects, `!seen[$2]++` → 1, fix deleted → 8, fix deleted + `-s 999999` → the batch-count assertion at `1x`. An awk variable rename stays green.
- **U0 + U2** (correctness + silent-failure, `no-undo-audit.sh:485`/`:477`) — same ordering defect, fixed once, comment-side, per U0's **second** branch. `order preserved` read as git log's order; what survives a split is batch order, i.e. pathspec order. Narrowed to the set, with the reason for not buying the ordering back. U0's first branch was **not** taken: it would have put the `|| die` on a pipeline ending in xargs/git, reintroducing the no-pipefail hazard U7 confirms this PR deliberately avoids.
- **U1** (silent-failure, `:485`) — `|| die` on the dedupe statement, as written. Safe now that `:485` stays a plain `$(…) | awk`.
- **U10** (simplify, `:148`), stale-literal half — assertions read `c.paths.length`, not a repeated `20`. The `n` parameter went with the fixture rewrite.
- **U4** (comments, `:144`) — the `~4500` figure was in the docstring ruling 1 rewrote anyway. Replaced with an environment-qualified `thousands`, since the review could not reproduce `4500` and the number gates no assertion. Strict weakening of a claim, in text this commit already owns.

**Deferred**

- **U3** (silent-failure, `:472`) → #522. Pre-existing guard, context in this diff rather than part of it; wording only, nothing swallowed.

**Declined**

- **U9** (simplify, `:141`) — obsolete, not deferred. It proposed deleting `manyConflictsOneCommit` as a near-duplicate of `bareConflictRepo`; ruling 1 replaced that fixture with a two-commit, two-subset shape `bareConflictRepo` does not have. The premise no longer holds.
- **U6, U7, U8** — clean bills. Nothing filed, nothing changed.

`node --test skills/fleet/scripts/*.test.mjs` from the worktree: **tests 721 / pass 721 / fail 0**.

PR #523, opened 2026-08-16
docs(fleet): revert class=routine to session tier, and probe ticket staleness in phase 0

feigi:
Three controller-tooling changes made during one `/fleet:run-team 2 5` run, under the skill's own **Fix the tooling mid-run** rule. No ticket — these are adhoc, and each was triggered by a defect that run produced.

**Not opened by an implementer, so it has had no review.** Queued deliberately for the next wave.

## 1. The tier guard fired — `class=routine` no longer dispatches at `sonnet`

Both halves met on the accumulated `docs/metrics/tier-outcomes.tsv` once this run's row was appended (`521 148 routine sonnet yes yes`):

- **Floor** — 8 `class=routine` PRs spanning 3 distinct `run_date`s (needs ≥3 PRs, ≥2 dates).
- **Trigger** — 2 rows carry `closed_own_ticket` `no`: **#452**, which moved a guard to a script's end and regressed the default dry run 0→1 — the exact defect its ticket existed to remove — and **#466**, whose new awk emitted invalid JSON for a staged rename, destroying the audit document its three tickets existed to make truthful.

Per the guard's own wording this reverts **`class=routine` only**, never the rule wholesale. Phase 0 still records the class: it selects the correction-ticket discipline, partitions the metrics file, and is what any future control would be drawn from.

### The counter-evidence is recorded beside the revert, not suppressed

Reviewers should weigh this, because it is the part most likely to be wrong:

- On the guard's own metric the split is **2 failures in 7** `routine`/`sonnet` rows against **0 in the single** `routine`/`opus` row (#480). That is **n=1** on the control side.
- Both `no` rows predate the revert — #452 by three days (`run_date` 2026-08-13), #466 by two (2026-08-14); the most recent routine PR (#521) closed its own ticket.
- **Every row so far, at either tier, carries `minted_false_claim` `yes`** — that column discriminates nothing about tier.
- `class` and `tier` were perfectly confounded by the old rule, so these rows establish that **routine tickets sometimes fail to close their own ticket — not that `sonnet` caused it.**

The revert is a pre-committed rule being honoured. It is **not** a measurement, and the prose now says so in those words.

## 2. Phase 0 gains a "Still live?" probe; phase 2 gains a re-derive backstop

Half this run's staged wave was already dead on arrival:

- **#132** — its exact one-line remedy had shipped ten days earlier in `e7b11e6`, whose message references #176 and carries no `Closes #132`. Cost a full claim, worktree and implementer dispatch to discover. Closed unimplemented.
- **#134** — 3 of 4 ACs already satisfied; **AC-2 asked for exit 1 on a case #64 deliberately changed to exit 3**, already pinned at `candidates.test.mjs:941`. Implementing as written would have shipped a test asserting the opposite of shipped behaviour.

Neither fixing commit carried a trailer back, so the tracker showed both as open work. Both probes took seconds. Phase 0 now settles a ticket's central claim against `origin/main` before `decided?`; the implementer prompt carries the backstop for what grep cannot settle.

## 3. `implementer-model-tier.test.mjs` now pins the reverted rule

It anchored on the old binding and failed. **Rewritten, not relaxed** — and it now carries a negative pin in *both* phase 0 and phase 2, because the binding was stated independently in each and restoring one alone is undetectable without a pin on each.

That half-revert is not hypothetical: **it is what this change shipped on its first pass.** Phase 2 was reverted, phase 0 left saying `class=routine` "dispatches at `sonnet`", SKILL.md self-contradictory — and it was only caught because opening this PR meant running a test whose name does not match `*prose*`.

## Verification

- Full suite on this branch: **`tests 720 / pass 720 / fail 0`**.
- 33-pattern rule manifest: 4 misses, **all 4 pre-existing on `origin/main`** (recorded on #48).
- Mutation-verified, `SKILL.md` restored byte-identical between runs:

| mutation | result |
|---|---|
| phase-2 binding restored | **RED** |
| phase-0 binding restored alone (the half-revert) | **RED** |
| unrelated reword in the same slice | GREEN |

## Reviewer notes

- ~~**Behind `main` by 6.** Not rebased…~~ **Superseded — rebased 2026-08-16 and now current.** `cff7330` → `9e8ee3d`, `behind=0`, fully green (`check` and `rebase-check` both success). Done server-side with `gh pr update-branch --rebase`, which sidesteps the local blocker (an unrelated pre-existing `settings.json` modification; stashing is forbidden by this skill).

  Replay proved, not assumed: `git range-diff 94acb8f..cff7330 512f7c71..9e8ee3d` returns `=` for the single commit — pure replay, zero conflict resolution. Cross-checked that nothing the merges brought in was clobbered: `no-undo-audit.sh` and `drop-merged-label.test.mjs` are byte-identical to `main`, #521's `awk '!seen[$1]++'` is present at `no-undo-audit.sh:492`, and the diff against `main` is still exactly the three files below.
- **This PR edits the skill the reviewing controller is itself running.** Findings against it change the reviewer's own instructions on merge.


feigi:
**Review findings applied — `d35a86f`.**

Six edits across the three touched files, from 7 `survived` findings plus 2 the controller's refuters held. `node --test skills/fleet/scripts/*.test.mjs` → **tests 725, pass 725, fail 0**. Every widened pin was mutation-checked: green on the clean tree, red on each mutant it targets, green on a change it should *not* catch.

**Deferred, filed:**

| Issue | Finding | Dimension | Label |
|---|---|---|---|
| #528 | phase 3's tier-guard gate fires "since the last check" — a set nothing persists (`SKILL.md:532-534`) | `correctness` | `needs-triage` |
| #529 | two guard-test pins do not discriminate; `/^\*\*Guard: /` cannot fail at all (`implementer-model-tier.test.mjs:246`, `:237-241`) | `tests` | `needs-triage` |
| #530 | comment says the phase-2 section is ~90 lines; it is 238 (`implementer-model-tier.test.mjs:30`) | `comments` | `ready-for-agent` |

All three are pre-existing text this PR does not touch, deferred on scope — outside the tier-guard revert and the phase-0 staleness probe. Verified against the diff, not assumed.

Also commented on #472 (no test pins the TSV header against SKILL.md's guard text) — this PR is the first observed instance of that drift, and the coverage gap remains open after the header fix.

One finding needed no action: the review's positive control confirming the phase-0 adjacency pin correctly rejects the class-inversion bug. It reports a working test, not a defect.

PR #524, opened 2026-08-17
fix(fleet): file through ledger.mjs check on both paths that create issues

### LINKED ISSUE #151: review-and-fix.md files issues without running ledger.mjs check, so the dedupe guard never sees the path that produced the duplicates
Deferred from PR #150 review.

## The finding

`skills/fleet/commands/review-and-fix.md:10` (step 5, the reviewer's filing step) tells the agent to run its own ad-hoc dedupe:

> `gh issue list --search` first — comment on an existing follow-up, never duplicate.

**`review-and-fix.md` contains zero occurrences of "ledger"** (verified: `grep -c -i ledger skills/fleet/commands/review-and-fix.md` → 0).

PR #150 rebuilt `ledger.mjs check` into a real guard — tracker query, overlap scoring, `--state all`, a warned failure mode, exit 3 on a tracker hit. The reviewer filing step, which files most of a run's issues, still hand-rolls its own search and never calls it.

## Why this is the path that matters

The measured 5x rediscovery of #114 in #145 came through the **reviewer** path, not the controller path — the filed row in PR #150's own test fixture is tagged `(review-pr-108)`. So the guard was improved on a path that was not the one producing duplicates.

Nothing is broken; the blast radius is inverted.

## The fix

Point step 5 at `ledger.mjs check "<subject>"` instead of the bare `gh issue list --search`, and state how to read exit 3 (tracker rows to review) versus exit 1 (already filed this run). `run-team/SKILL.md:608` now documents the exit codes and can be referenced rather than restated.

## Why deferred

PR #150's scope is `ledger.mjs check` itself, and every acceptance criterion on #145 is about that script's behaviour. Changing the fleet's active reviewer workflow is a separate, deliberate edit — and it was the final PR of the run.

Found by the code-reviewer specialist; independently verified during adjudication.

feigi:
`skills/fleet/commands/review-and-fix.md` step 5 hand-rolled `gh issue list --search`, and the file contained zero occurrences of "ledger" — so the reviewer and the fix-applier, the agents that file a run's deferrals, never reached the guard #145 rebuilt. The ledger row for the finding rediscovered five times in one measured run carries `(review-pr-108)`, a reviewer's own source tag, so the guard had been improved on a path that was not the one producing duplicates.

## The two filing sites covered

| Site | Agent | Change |
|---|---|---|
| `review-and-fix.md` step 5 | reviewer, fix-applier | calls `ledger.mjs check "<subject>"`; states how to read every exit code |
| `run-team/SKILL.md` finisher duty 2 | finisher | its `Not filed → file it` branch now files through the same guard |

The finisher is the second site because it reaches `gh issue create` by a different route — it files last, off its own read of what the reviewer left behind, so a deferral it reads as unfiled may already be on the tracker under someone else's wording.

## Reading the guard in both directions

Getting `check` called is half the fix. It exits **0** for `clean` and `unverified` alike, so an instruction reading exit 0 as a green light files over a tracker nobody read; and exit **2** is a usage error where nothing was checked at all, so an instruction reading any non-zero as "duplicate" turns a deferral into a silent drop. Step 5 now says:

- exit **1** already filed this run; exit **3** a clean ledger with tracker rows to read and comment on
- **exit 0 is not automatically "safe to file"** — an unreachable `gh` degrades to a ledger-only answer printing `TRACKER NOT CHECKED`, so the stdout `verdict` separates a tracker that was searched and clean from one nobody read
- exit **2** is a usage error, never a duplicate; and a duplicate is a comment on the existing issue, never a dropped deferral

Full exit-code and `verdict` semantics are cited to run-team's **Run ledger** section rather than restated, per the issue. Only the operational reading is inline, because the agent acting on step 5 is mid-filing and a second file read is a cost it will skip.

## Deliberately left

- **run-team's fix-applier dispatch prompt** and its reviewer blockquote both say "defer and file it" and both route to step 5 by name (`run **steps 2, 3 and 5 only**`) — covered transitively, and duplicating the guard text there is a third copy to drift.
- **run-team's Run ledger section** already mandated `ledger.mjs check <subject>` before every `gh issue create` and already distinguished `clean` from `unverified`. Unchanged.
- **`docs/agents/issue-tracker.md`** documents `gh issue create` as a repo convention, not a fleet run path; `issue-tracker-prose.test.mjs` records that nothing in the tree routes an agent to it.
- **`run-team/SKILL.md`'s `gh issue list --search` for pool counts** is a count query, not a pre-filing dedupe — not in this class.
- **Recording `ledger.mjs filed` from the reviewer.** The controller owns that write today; adding a second writer means concurrent whole-file atomic writes to one ledger, which loses rows. Out of scope for this ticket and a design call, not an oversight.

## Tests

New `skills/fleet/scripts/dedupe-guard-prose.test.mjs`, 7 pins over both slices. Red 6/6 before the fix.

Mutation-verified, each reddening only its own pin:

| Mutation | Result |
|---|---|
| step 5 reverted to the bare `gh issue list --search` | 5 red |
| exit-0/`verdict` warning deleted | 1 red |
| exit-2 clause deleted | 1 red |
| finisher loses the guard | 1 red |
| cited `## Run ledger` section renamed | 1 red |
| run-team remaps exit 1 / drops exit 3 / inverts the exit-0 warning | 1 red each |
| **non-mutation:** unrelated reword inside step 5 | **stays green** |

Full suite `./agent-test skills/fleet/scripts`: **732 pass, 0 fail** (725 before this branch).

The pre-existing step-numbering pin in `review-path-default.test.mjs` reads step 5 as a single line and requires it to still open `File each deferred finding` — step 5 stays one line and keeps its opening, and that test is green.

Closes #151


feigi:
## Review deferrals

Filed from this PR's review: **#531** (`needs-triage`) — the dedupe guard's ledger arm is permanently empty because no filing site runs `ledger.mjs filed`, and the obvious fix races `save()`'s whole-file write. Deferred rather than applied: the gap is verified, but the remedy needs a design call (`impl-151` rejected the naive form as declared item #7, and that reasoning was re-verified against `ledger.mjs:97`).

Applied in 6f2a454, both defects in this PR's own new test file:
- `dedupe-guard-prose.test.mjs` test 5 matched a multi-word phrase against an **un-flattened** slice of hard-wrapped Markdown, so a pure reflow of run-team's Run ledger section reddened the suite — contradicting the file header's own "a reflow of the same words stays green by design". Now `flat()`-wrapped like its sibling.
- Test 6's comment claimed "renumbering in either file alone reddens **this**". True of the suite, false of that test, which reads `RUN_TEAM` only. Comment now names the test that actually holds the step-5 side. No coverage change — measurement showed there is no gap.

Four further findings were checked and dismissed: two refuted at review time, and both `suggestion`-band items (step 5's exit-3 direction; test 3's presence pins) were dispatched to a refuter and came back refuted — the second after showing the proposed tightening reds on a pure clause reorder, i.e. the remedy reintroduced the very defect fixed above.

`node --test skills/fleet/scripts/*.test.mjs` → tests 732, pass 732, fail 0.

feigi:
### Addendum — the un-flattened pin was a class, not a case

Follow-up in e4d1d0d. `survived1` named test 5's slice; the same defect was in **all five** step-5 pins, which match literal phrases containing spaces against an un-flattened slice.

It reads as a no-op because `review-and-fix.md` authors one long line per step, so nothing wraps today. It is not a no-op on the defect: reflowing that file at **width 100** split `never a dropped finding` and reddened test 4, while 80/120/140 stayed green. Width was sampling *which* pin fired, not which were vulnerable.

Every slice in the file is now `flat()`-wrapped. Measured after the change:

| check | result |
|---|---|
| reflow **both** source files @ 60/80/100/120/140 | 7 pass / 0 fail at every width |
| mutate what each pin pins, all 7 | each reds its own test, and only it |

One assertion is **deliberately left raw**: `assert.match(RUN_TEAM, /^## Run ledger$/m, …)`. Flattening would destroy it — the `/m` anchors need real newlines — and a reflow cannot split a heading. No residue issue filed, because there is no residue.

`node --test skills/fleet/scripts/*.test.mjs` → tests 732, pass 732, fail 0.

Issue #143, opened 2026-07-31
the 'tests 0' reading rule is too narrow: a partial tree and an all-skipped run both pass it


feigi:
Deferred from PR #130 review.

PR #130 added a reading rule to `review-pr.js:250-253` and `review-and-fix.md:43`: `'tests 0' is a FAILED run, not a pass`. It closes the zero-match-glob case. Two zero-work runs still walk through it, both measured:

```
partial tree (2 of 11 test files present):  tests 6  pass 6  fail 0   EXIT=0
every test skipped:                         tests 2  pass 0  skipped 2  EXIT=0
```

The first is **realistic and instructed**: the prompt at `:247` tells specialists to work inside their own copy of the snapshot, so a partial copy is the documented workflow, and nothing anchors an expected count.

**Fix (settled):** extend the rule in both places with a floor and a `pass 0` clause — a run reporting fewer tests than the suite has, or zero passes, is unrun. Stating the expected count is worth it even though it drifts: a stale number produces a loud false red, which is the safe direction; its absence produces silent false greens.

**Related, same file:** the prompt test in `review-pr-testcmd.test.mjs` pins only the ruling clause. Deleting any of these individually leaves the suite green:
- `Do not substitute a command of your own.` — the actual anti-substitution instruction
- the whole rationale block
- `report that dimension as unrun and say the command produced no tests.` — the actionable follow-through

So the sentence declaring a zero-test run a failure is guarded, while the instruction saying what to *do* about it is not. Worth pinning the follow-through alongside the rule if this is touched.

Found by the silent-failure and test specialists on PR #130.

feigi:
Additional finding for the same prompt block, from the code-simplifier on PR #130 — fixing this in the same edit.

The surviving rationale in `review-pr.js:250-252` tells specialists not to substitute a command because a bare runner *"picks up a default config that tears down a shared container mid-run for every sibling."* **That is inert in this repo**, and this PR's own `review-and-fix.md:43` edit says so in as many words: `feigi/claude-config` has *"no compose file, no `globalSetup`, no vitest, and no CI."*

So a specialist that checks the reason it was given finds it false here. `review-and-fix.md` argues for carrying the portable rule rather than the local command, which is right — but the prompt states the container story as present-tense fact about the run it is describing, not as a portable rationale.

Fix alongside the `tests 0` floor: state the rule portably (do not substitute; a substituted runner can collide with siblings in repos that have a stack) rather than asserting a teardown that cannot happen here.

[...]

Issue #140, opened 2026-07-31
review-pr.js: snap.path is handed to every specialist without confirming the snapshot exists


feigi:
Deferred from PR #130 review.

`snap.path` comes back from the snapshot agent and is interpolated into every specialist prompt (`workflows/review-pr.js:245`) and every verifier prompt (`:282`) **unchecked**:

```
215  log(`snapshot ${snap.head} at ${snap.path}`);
245  READ ONLY FROM THE SNAPSHOT: ${snap.path} (HEAD ${snap.head}).
282  Verify against the snapshot ${snap.path} by RUNNING something — compile it, run
325    snapshot: snap.path,
```

If `tar -x` failed, or the agent reported an aspirational path, every specialist gets `No such file or directory` on every read and reasons from source instead of measuring.

**That is precisely the defect PR #130 fixed for `testCmd`, one level up — and it affects all six dimensions instead of one.** The schema requires `path` to be a string; nothing requires it to exist.

**Fix (settled):** `existsSync(snap.path)` (or a stat on a known file inside it) between the snapshot phase and the review phase, failing loudly. The snapshot agent is already asked to verify byte-identity of a couple of files (`:186-187`); this makes the caller check rather than trust the report.

Found by the silent-failure specialist on PR #130.

Issue #500, opened 2026-08-16
skills/fleet/scripts: between() is pasted identically into three test files — extract it to a shared prose-pin.mjs


feigi:
Deferred from PR #498 review (ticket #87).

**Dimension:** `simplify`
**Site:** `skills/fleet/scripts/reaping-prose.test.mjs:62`

## The finding

`between()` in the test file PR #498 adds is the **third byte-identical copy** of the same 7-line helper in `skills/fleet/scripts`:

```
skills/fleet/scripts/reaping-prose.test.mjs:62      (added by #498)
skills/fleet/scripts/refuter-scratch-prose.test.mjs:41
skills/fleet/scripts/review-pr-reads.test.mjs:276
```

`diff` of the three extracted bodies is empty in both directions. The repo already has the convention of importing shared test helpers from a sibling `.mjs` rather than re-pasting them — `./strip-comments.mjs` (3 files), `./arg.mjs`, `./board.mjs` (2), `./compute-board.mjs` (2), `./compute-spend.mjs`, `./diff-stats.mjs` (2), `./fleet-tick.mjs`.

## Why deferred rather than applied

**Out of ticket scope.** #87 is "state the delete-time limitation wherever the claim is made" — it corrects three stale prose sites and pins them. Two of the three `between()` copies live in `refuter-scratch-prose.test.mjs` and `review-pr-reads.test.mjs`, files #87 has no business editing. Landing only the third copy's extraction inside #498 would leave the dedup half-done while still touching a file the ticket does not cover, so the whole thing moves here.

## Desired

Extract the helper once to `skills/fleet/scripts/prose-pin.mjs`:

- `export function between(text, from, to, what)`, importing `node:assert/strict` itself
- carry over the both-ends-bounded rationale comment ("an unbounded end lets a later, unrelated occurrence of the same phrase satisfy the assertion with the real clause deleted")
- replace all three inline definitions with `import { between } from "./prose-pin.mjs";`

**Measured by the reviewer** against the #498 head tree: after the extraction, `node --test` over the three files gave `tests 31, pass 31, fail 0`.

Re-verify against `main` at pickup time — #498's own head is not where this will land, and `reaping-prose.test.mjs` may have gained call sites.

feigi:
Census update from the PR #526 review (`simplify` dimension, deferred here rather than filed separately).

**The copy count has grown from three to five** since this was written:

```
skills/fleet/scripts/agent-test-dir-prose.test.mjs:42
[...]

PR #526, opened 2026-08-17
fix(fleet): review-pr reports an unrun or crashed dimension instead of an empty findings list

### LINKED ISSUE #137: review-pr.js: a dimension that never ran the suite is indistinguishable from one that ran it green
Deferred from PR #130 review.

`FINDINGS_SCHEMA` (`workflows/review-pr.js:28-40`) has no field for the test run. A specialist that ran the command and one that never executed anything return **byte-identical shapes** — `findings: []` either way. The script cannot tell them apart, and neither can the controller reading the result.

PR #130 added a reading rule to the specialist prompt (`:250-253`): `'tests 0' is a FAILED run, not a pass: report that dimension as unrun and say the command produced no tests.` **There is nowhere for that report to land.** A specialist that obeys it emits an empty findings list and is indistinguishable from a clean pass.

The comment at `:320` compounds it:

```
320  // `dimensionsRun` names what actually ran: a trimmed fan-out must say so, never
326    dimensionsRun: dimensions.map((d) => d.key),
```

`dimensions` is the list **dispatched**, not the list that ran. The comment asserts what the code does not do.

Verified scope: `review-pr.js` contains no `import` statement and nothing that executes, observes or parses a test run — the rule is entirely an instruction to a model, the same trust level that produced the original defect (grep over the whole file for `child_process|exec|spawn|import |testCmd|tests 0`).

**Fix (settled):** add a required `test_run: {command, tests, pass, fail}` to `FINDINGS_SCHEMA`, and have the caller list dimensions whose `test_run.tests` is 0 or missing alongside `dimensionsRun`.

One thing to check while implementing: whether the workflow runtime hard-rejects a missing required field or is best-effort about it.

Found by the silent-failure and comment specialists on PR #130, independently.

### LINKED ISSUE #138: review-pr.js:267: a reviewer that crashes is reported as a dimension that ran and found nothing
Deferred from PR #130 review.

`workflows/review-pr.js:267`:

```js
(review && review.findings ? review.findings : []).map((f) => () => {
```

That `review &&` guard proves the author expects `agent()` can return falsy. When it does — spend limit, timeout, terminal error — the dimension yields `[]`, which is **identical to a dimension that ran fully and found nothing**, while `dimensionsRun` at `:326` still lists its key. The caller reads it as covered.

**The asymmetry is the tell.** Thirty lines below, the verifier path handles exactly this case correctly and deliberately (`:290-297`):

```js
// Every refuter crashed (spend limit, timeout, terminal error): the
// finding was NOT verified, so it is `unverified`, not `survived`. It is
// still returned — surfaced, never dropped — but a consumer keying on
// "survived" must not read a verification that never ran as one passed.
let verdict;
if (live.length === 0) verdict = "unverified";
```

That reasoning applies verbatim one level up and was not applied there.

**Fix (settled):** mirror the verifier treatment — a falsy `review` yields a dimension marked crashed/unrun rather than an empty findings list, and the caller reports it separately from `dimensionsRun`.

This is the defect class that bit the fleet repeatedly on this run: reviewers declaring dimensions uncovered whose reports existed, and the controller reconciling against relay receipts by hand.

Found by the silent-failure specialist on PR #130.

### LINKED ISSUE #139: review-pr.js:31: scope_searched is documented Required and is not in the required array
Deferred from PR #130 review.

`workflows/review-pr.js:31` omits `scope_searched` from the schema's `required` array:

```js
required: ["dimension", "findings"],
```

while its own description at `:37` says:

> "The exact commands/paths this pass covered. **Required** so a negative claim is bounded: a grep that found nothing looks identical to a grep never run."

A specialist that omits the field validates clean — **reproducing the exact failure mode the description exists to prevent**.

**Fix (settled):** one line, either direction — add `scope_searched` to `required`, or drop the word "Required" from the description. The first is what the description argues for.

Found by the silent-failure specialist on PR #130.

feigi:
Closes #137
Closes #138
Closes #139

One seam, three tickets: in `workflows/review-pr.js`, a dimension that never ran
or that crashed was indistinguishable from one that ran clean. Taken separately
they conflict on one file and the unrun state ends up with two shapes, so this is
one PR giving it one shape.

## What changed

**`FINDINGS_SCHEMA` gains `test_run {command, tests, pass?, fail?}`, required** (#137).
PR #130 added the reading rule — `'tests 0' is a FAILED run, not a pass: report
that dimension as unrun` — with nowhere for that report to land, so a specialist
that obeyed it emitted `findings: []`, byte-identical to a clean pass. The
specialist prompt now names the field and says to report the run **even when it
failed or produced nothing**.

**`scope_searched` joins the `required` array** (#139). Its own description called
it *Required* while the array omitted it, so a specialist that skipped it
validated clean — reproducing the exact failure the description exists to
prevent. This is the direction the description argues for; the alternative (drop
the word "Required") was rejected because the field bounds every negative claim.

**A falsy `agent()` return is classified, not swallowed** (#138). Its `[]` was
read as coverage. The recording is derived from the pipeline's **own result**,
after `pipeline()` returns — one slot per dimension, in order, holding null for a
chain that died, so index recovers which dimension crashed.

It cannot be done inside the verify stage, and the first version of this PR tried:
`pipeline()` short-circuits between stages, running `if (result === null) break`
*before* invoking stage 2, so a reviewer that returned null never reaches that
closure at all and the recording there was dead code. Read out of the harness
bundle rather than inferred — the pre-existing `review && review.findings` guard
proves an author's belief, not a harness contract, and reading it as one is what
made this look handled. Verified by transcribing the bundle's `pipeline()` and
running all four cases (null return, throw, `tests: 0`, clean).

The verify stage keeps #137's half, which does reach it: that reviewer returned an
object, so nothing short-circuits.

**The caller returns `dimensionsUnrun: [{dimension, reason}]`**, alongside an
unchanged `dimensionsRun`. Both tickets specify a sibling field ("alongside",
"separately from"), and it is also the correct shape: subtracting the unrun keys
from `dimensionsRun` would make a crashed dimension indistinguishable from one
the size tier never dispatched — this ticket set's own defect, moved one field
over. The `:711` comment claimed the list named what ran; it now says what the
code does.

`unrunReason` is deliberately **pure** — no `snap`, no `testCmd`. `review-pr.js`
has a top-level `await` and cannot be imported, so a test can only reach this
logic by lifting it out of the source text, the same technique
`review-pr-testcmd.test.mjs` uses for `resolveTestCmd`.

## On #137's open runtime question

Settled by the controller before dispatch and **not re-litigated here**: a schema
forces the subagent to call a StructuredOutput tool and validation happens at the
tool-call layer, so a mismatch is retried rather than passed through. Measured
over `~/.claude/projects/-Users-chris--claude/*/subagents/workflows` and
reproduced independently for this PR: **85** transcripts carry `Output does not
match required schema`, **184** rejection events in all, every one recovered by
retry, and **0** carry `Failed to provide valid structured output`.

The retry is **bounded**, so this is not a zero-risk claim and the code comment
does not make one. Exhaustion returns null from `agent()`, and that null is what
`unrunCrashed` reports as unrun rather than clean — **not** `unrunReason`'s falsy
branch, which `pipeline()`'s short-circuit means a null review never reaches. The
review caught that claim; see the #138 section above.

**Re-counting those figures is a trap, and it caught a reviewer of this PR.** Both
strings are quoted verbatim in the comment under review, so every transcript that
*reads* the file becomes a grep match: a naive re-count counts its own readers and
reports drift that is the observer. One review measured 92/197/6 and called the
figures already stale; excluding the reading session returns **85/184** exactly,
and real exhaustion events — matched with a digit, `after [0-9]+ attempts`, rather
than the `<n>` placeholder that appears nowhere but the comment — number **0**.
The comment now carries the date and the trap. The prior claim on #137 died trying to *measure* it;
members have no `Workflow` tool, so that experiment is structurally unavailable
to an implementer — the transcript corpus is not, which is what makes this
checkable from inside a member.

## The case class — covered vs. deliberately left

Every member of "a dimension's outcome is unreadable from its return":

| # | Case | Ruling |
|---|---|---|
| 1 | Specialist agent returns falsy | **Covered** (#138) — at the caller, from the pipeline's result; the verify stage never sees this case |
| 2 | Specialist returns an object having run no tests (`tests` 0 or absent) | **Covered** (#137) |
| 3 | Specialist omits `scope_searched`, unbounding a negative claim | **Covered** (#139) |
| 4 | `dimensionsRun` names the dispatch, read as coverage | **Covered** — sibling list + corrected comment |
| 5 | Specialist reported `test_run` but never got the field asked for | **Covered** — prompt names it; a schema field nothing asks for comes back absent |
| 6 | Consumers documenting the return shape go stale | **Covered** — `run-team/SKILL.md`, `review-path-default.test.mjs` |
| 7 | Refuter agent returns falsy | **Left** — already correct (`live.length === 0` → `unverified`); #138 cites it as the model to mirror |
| 8 | Snapshot agent returns falsy | **Left** — already guarded with a throw |
| 9 | Dimension trimmed by `selectDimensions`, never dispatched | **Left** — absent from `dimensions`, so absent from both lists; correct by construction |
| 10 | `suggestion` skipping verification by policy | **Left** — already `unverified`, already distinguished |
| 11 | `snap.path` reported but nonexistent | **Left — #140, open sibling ticket** |
| 12 | `tests N` below the suite's size, or all-skipped `pass 0` | **Left — #143, open sibling ticket** |
| 13 | Stage-1 `agent()` *rejecting* rather than returning falsy | **Covered**, by the same derivation. `pipeline` does **not** propagate: `allSettled` maps a rejected slot to null and logs `pipeline[i] failed:`, so a throw is indistinguishable from a null return at the caller and lands in `dimensionsUnrun` too. #527, filed on the old reading, is answered on the ticket rather than auto-closed |

## What this could wrongly refuse

A new required field and a new unrun predicate are both refusal surfaces. A
specialist that legitimately has nothing to report must still come back clean, so
four accept-side cases are pinned as tests: an empty `findings` list with a real
run, a run reporting `fail: 3`, a run that omitted the optional `pass`/`fail`, and
`tests: 0` remaining *expressible* rather than unvalidatable. `pass`/`fail` are
deliberately **not** in `test_run.required`: they are not always separable from a
runner's output, and an unanswerable required field gets answered with a guess.

## The suite could not see this mode before

No test in the repo touched `FINDINGS_SCHEMA` or the falsy-review branch, so a
green run said nothing about either. `review-pr-unrun.test.mjs` is new: **16
tests**, both directions. The doc pin is an exclusion assertion (the superseded
workaround must be **gone**, not merely accompanied) and was verified non-vacuous
by running it against the pre-change text.

**Three of the original pins were measured vacuous against the very bug class they
exist for, and were rewritten to RUN rather than to read source text.** The
order-pin compared the source positions of two substrings, so guarding the push
with `review &&` left both in place and kept the suite green while every crashed
dimension vanished. The two schema pins regexed `required:\s*\[([^\]]*)\]` over
`stripComments` output, which is line-based by design, so an entry commented out
*inside* an array literal — `["dimension", /* "scope_searched", */ …]` — kept its
quoted text visible to the regex while the real array lost it, reproducing #139
and #137 green. Both now evaluate the declaration and assert against the object.

Mutation matrix, all run: **8 mutants each red exactly their own test** — the
`review &&` guard, a lost index→dimension mapping, an empty `unrunCrashed`, either
recording deleted, both comment-disguised schema deletions, and the plain-deletion
control. **4 behaviour-preserving refactors stay green**: renaming the local inside
`unrunReason`, adding `minimum: 0`, reflowing a comment, hoisting the `flatMap`
callback.

`node --test skills/fleet/scripts/*.test.mjs` — **741 pass, 0 fail** (737 before this PR).


feigi:
## Review adjudication — 20 findings, head `af9c608`

`node --test skills/fleet/scripts/*.test.mjs` — **741 pass, 0 fail**.

**Baseline correction:** the review snapshot reported 738 pass / 6 fail in `dedupe-guard-prose.test.mjs`. That file does not exist on this branch, and both refuters that re-cut the tree from `cd0ef5c` independently measured **737 pass / 0 fail**, matching this PR's own original body. The 6 failures came from whatever tree the review actually ran in, not from this commit. So the baseline here is clean and **any** failure would have been ours.

### Applied

| Finding | Severity | What |
|---|---|---|
| **controller ruling on `refuted2`** | critical | #138 was dead code. `pipeline()` short-circuits before stage 2, so the recording never ran for a crashed reviewer. Now derived at the caller from the pipeline's result. Also covers a stage-1 **throw** (#527). |
| `survived0` | critical | Order-pin compared two source positions; `if (review && why)` slipped past it. Replaced with pins that execute the derivation. |
| `survived1` + `survived2` | important | Both schema pins regexed `stripComments` output, so an entry commented out *inside* an array literal stayed visible to the regex while the real array lost it. Both now **evaluate** the declaration — one idiom, both sites. |
| `unverified0` | suggestion | `!run.tests` vs `run.tests === 0` was unpinned; the narrower comparison passed all 737 tests while reading an unrun dimension as clean. **Survived** its refuter. |

Every added or changed test was mutation-checked both ways: **8 mutants each red exactly their own test**, **4 behaviour-preserving refactors stay green**.

### Refuted — checked, not applied

| Finding | Why |
|---|---|
| `unverified3` (log `dimensionsUnrun`) | Refuter transcribed the harness `pipeline()` and ran it: the harness *already* logs `pipeline[i] failed: <msg>` for the dead case, and ships `<diagnostics>`, `<result>` and `<failures>` to the same consumer in one block. Duplicate of data already delivered. |
| `unverified6` (transcript counts stale) | **The refutation corrected me, after I had already applied the finding.** Both grep strings are quoted verbatim in the comment under review, so every transcript that *reads* the file matches: the drift measured was the observer. Excluding the reading session returns **85/184** exactly, and real exhaustion events — matched with a digit rather than the `<n>` placeholder — number **0**, not 15. My first edit asserted exhaustion "HAPPENS"; it does not, and that claim was removed. Kept: the date, and the trap written down where the next re-counter will read it. |
| `unverified11` (move the declaration above the banner) | Premise true, rationale false. The file's *other* `// --- X ---` banner heads a 200-line, 14-statement section, so a `const` under one is normal — and applying the move re-files `dimensionsUnrun` under `// --- Snapshot ---`, across a section boundary from the code that populates it. |

Refuted findings are recorded here rather than filed as tickets: `unverified11`'s refutation shows the change would make things *worse*, and minting a ticket for it invites someone to do the wrong thing later. Flagging the deviation explicitly.

### Deferred and filed

- **#533** (`needs-triage`) — the `new Function` source-lift is pasted into four test files. `unverified10`. Out of scope: doing it properly touches three files these tickets have no business editing, and `select-dimensions.test.mjs`'s three sites may not fit the signature at all.
- **#500** — commented, not duplicated. `unverified8`'s `between()` duplication is already tracked there; the census has grown from three copies to **five**, and #526 adds three call sites (down from four — two disappeared when the schema pins were rewritten to evaluate).
- **#143** — commented. `unverified4`: a run reporting `fail > 0` with an **empty findings list** classifies clean. That is this ticket's widening, and the code comment already points there.
- **#527** — commented. Covered by this PR, and its stated mechanism ("a throw propagates, the workflow dies, five dimensions die with it") is wrong: `allSettled` maps the rejected slot to null and every other slot returns normally. Recommended for closing on merge, not auto-closed — its `needs-triage` design call dissolved rather than got made.

### No action

- `refuted0`, `refuted1`, `refuted3`, `refuted4` — refuted 2-0 at review time.
- `unverified2` — restates `refuted1`, killed 2-0. A suggestion does not resurrect it.
- `unverified5` — no defect and no fix proposed; the PR's edit there is inside a prose comment, and the claim it makes is independently pinned as a live assertion.
- `unverified7` — the stale-snapshot baseline, resolved above.
- `unverified9` — **mooted by `survived2`**. The double `indexOf` with the unreachable `-1` arm was inside the pin that got rewritten; the line no longer exists.

### Comments the fix falsified, corrected in the same pass

`unrunReason` is no longer "the only part of this seam a test can execute"; the exhaustion paragraph credited its falsy branch with catching a null review it never receives; `dimensionsUnrun`'s declaration claimed one writer where there are two; and the PR body's own #138 paragraph and case-table row 13 described the mechanism this fix replaces. All updated.

feigi:
## Adjudication addendum — correcting an off-by-one, and one ruling

Surfaced by `finish-pr-526`, which noticed the adjudication comment above says **20 findings** and assigns **19 labels**, then resolved the count from the review journal rather than assuming a typo: 20 stage-1 findings (2 critical, 6 important, 12 suggestion), and since `suggestion` is budgeted 0 refuters, the buckets are exactly `unverified0..11` + `survived0..2` + `refuted0..4`.

**The missing label is `unverified1`** — "#138 classifies a reviewer that RETURNS falsy but not one that REJECTS". It was omitted because it was folded into the #527 discussion rather than labelled in its own right. Its disposition, for the record:

**Addressed, and #527 is closed.** The harness catches a rejected slot and maps it to null (`pipeline[i] failed: <msg>` → `null`); it does not propagate. So a throw and a null return are indistinguishable at the caller, and `unrunCrashed(reviewed, dimensions)` covers both in one derivation. #527 was filed by the controller on the false premise that a rejection kills the workflow, and has been closed with that correction.

**Separately, `unverified2` is now filed as #535, reversing part of the ruling above.** That ruling dismissed it as "restates `refuted1`, killed 2-0". That conflated two remedies: `refuted1` asked for a **code** change (compare `test_run.command` to the dispatched `testCmd`) and was properly refuted; `unverified2` asks for a **one-word doc** change to a sentence this PR introduces — `is covered` → `ran a suite`. Killing the code fix does not settle the doc wording, and the doc remedy was never adversarially checked.

Filed rather than applied here: the sentence is **imprecise rather than false** (a narrower run is still a specialist running and reporting), and this PR strictly improves the situation by making `test_run.command` a required field, so the command actually run is now recorded where before it was not captured at all. #535 carries the full weighing in both directions for triage.

No finding from this review is unfiled or lost. `ready-to-merge` was gated on `check: success`; the `rebase-check` red is behind-count staleness.

Issue #533, opened 2026-08-17
skills/fleet/scripts: the new Function source-lift is pasted into four test files — extract lift(code, name, signature)


feigi:
Deferred from PR #526 review (tickets #137/#138/#139).

**Dimension:** `simplify`
**Site:** `skills/fleet/scripts/review-pr-unrun.test.mjs:28`

## The finding

`workflows/review-pr.js` cannot be imported — it has a top-level `await` and is a workflow body the harness wraps — so its pure functions are lifted out of the source text with `new Function`. That lift is now pasted into four files:

```
skills/fleet/scripts/review-pr-reads.test.mjs:31      generic lift(name, signature)
skills/fleet/scripts/review-pr-testcmd.test.mjs:39    private copy
skills/fleet/scripts/select-dimensions.test.mjs:27/37/42   three sites
skills/fleet/scripts/review-pr-unrun.test.mjs:28      added by #526
```

`review-pr-reads.test.mjs:31` already carries the generic form — `lift(name, signature)` doing the identical regex-match / assert / eval — so the fourth copy re-implements a helper that lives one file over.

The reviewer demonstrated the shared form: a `skills/fleet/scripts/lift.mjs` exporting `lift(code, name, signature)`, with `review-pr-unrun` and `review-pr-testcmd` rewired to it, ran `tests 21, pass 21, fail 0`.

## Why deferred rather than applied

**Out of ticket scope, and the PR made the copy grow rather than shrink.** #137/#138/#139 are about reporting an unrun dimension. Doing this properly touches `review-pr-reads`, `review-pr-testcmd` and `select-dimensions` — three files these tickets have no business editing.

Note #526 did land the *sibling* change the finding argues for: its lift now handles three functions in one scope rather than one, so the fifth copy is not a straight paste of the fourth. That widens the shape a shared helper has to cover, and is a reason to design it deliberately rather than extract the narrow form.

## Why `needs-triage` rather than `ready-for-agent`

The defect is verified; the **scope is the open question**, and there are two defensible answers:

1. **Narrow** — export `lift(code, name, signature)` and use it from the new file only, leaving the other copies for whoever next touches them. The finding calls this "the minimum defensible version".
2. **Whole class** — rewire all four files. This is what #500 decided for the sibling helper `between()`, so there is precedent for going wide.

[...]

PR #536, opened 2026-08-17
fix(review-pr): refuse a snapshot whose path was never confirmed to exist

### LINKED ISSUE #140: review-pr.js: snap.path is handed to every specialist without confirming the snapshot exists

feigi:
## Summary
`snap.path` came back from the snapshot agent and was interpolated into every specialist prompt and every verifier prompt unchecked, and returned to the caller unchecked. If `git archive HEAD | tar -x` silently failed (bad auth, a dead HEAD), the directory `mkdir -p` already created stood empty, and every specialist got `No such file or directory` on every read and reasoned from source instead of measuring — on all six review dimensions instead of one, per #140.

The Workflow sandbox that runs `workflows/review-pr.js` has no filesystem access of its own ("no filesystem or Node.js API access", documented at `docs/specs/2026-08-06-review-pr-specialist-read-rules-design.md:194` and `skills/fleet/scripts/select-dimensions.test.mjs:14`), so the caller can't `existsSync` the path directly. Instead the snapshot agent now runs a mechanical `test -d && ls -A` check right after extraction and reports it as a **required** `pathVerified` boolean — enforced by the schema so it can't be silently omitted, distinct from the existing byte-identity "Verify it" step, which stays pure unenforced narration. `snapshotMissing(snap)` is the new caller-side gate that reads this field and refuses loudly, consolidating the pre-existing no-tree guard alongside it.

## Cases covered / left
- Missing snapshot object, or missing `path`/`head` — pre-existing behaviour, preserved and now covered by `review-pr-snapshot-path.test.mjs`.
- **New**: `path`/`head` present but `pathVerified: false` (tar/archive silently produced nothing) — refused, naming the path.
- **New**: `pathVerified` omitted entirely (schema bypass or legacy shape) — treated as unverified, not assumed true.
- **Accept case**: a fully verified snapshot (`pathVerified: true`) is not refused — pinned so the guard's false-positive class is tested, not just its false-negative class.
- Left alone: the existing byte-identity "Verify it" narrative step (unrelated to path *existence*, out of scope for #140) and `usableDiff`'s diff-freshness checks (a separate, already-covered concern).

## Test plan
- [x] New `skills/fleet/scripts/review-pr-snapshot-path.test.mjs` — lifts `snapshotMissing` by source text (same technique as sibling `review-pr-*.test.mjs` files, since `review-pr.js` runs a top-level `await` and can't be imported) and exercises all four cases above.
- [x] Updated the two existing schema pins (`review-pr-reads.test.mjs`, `review-pr-testcmd.test.mjs`) that asserted `required: ["path", "head"]` to include `"pathVerified"`.
- [x] `./agent-test skills/fleet` — 752/752 passing.

Closes #140


feigi:
Review findings applied in `0a21ceb`. Suite: `node --test skills/fleet/scripts/*.test.mjs` → **755 tests, 755 pass, 0 fail**, from a self-measured 752 baseline at `f4ab9fc3`.

**Applied — all 9 `survived` findings**

The headline defect: the `SNAPSHOT_NONEMPTY` probe sat *after* the `node_modules` symlink, so the symlink alone satisfied `ls -A` and a totally failed `git archive | tar -x` still reported `pathVerified: true` — in every repo the symlink exists for. Found twice independently (`correctness/critical` and `silent-failure/critical`), reproduced by four refuters. Probe moved above the symlink; the `test -d` conjunct went with it (dead once `ls -A` runs on the dir `mkdir -p` just made). Verified four ways in an isolated fixture: failed archive + `node_modules` now reads `SNAPSHOT_EMPTY`, the three healthy cases still read `SNAPSHOT_NONEMPTY`.

Then the mechanism that *produces* `pathVerified` was pinned, not just the function that consumes it — the shell probe, the sentence binding the field to its output, the ordering of the two, and the call site could each be deleted with the suite green. Also: `pathVerified`'s schema *declaration* (only its `required` membership was pinned), `snapshotMissing` added to the declared-exactly-once guard, and a false `#113/#118` attribution in the test file's header dropped (both were `resolveDimensions` — a hand-written REQUIRED-array check and a disconnected call site, neither a JSON schema).

Every added pin was mutation-verified to red **alone**, and to stay **green** under a reflow of the pinned sentence and a reword of unpinned prose.

**Deferred and filed**

| finding | dimension | issue |
|---|---|---|
| `review-pr.js:617` — is the `node:fs` sandbox restriction real? | `correctness` | #538 `ready-for-agent` |
| `review-pr.js:622` — a dead agent reports as "returned no tree" | `silent-failure` | #539 `needs-triage` |
| `review-pr-{reads,testcmd}.test.mjs` — byte-identical `required` pin | `simplify` | #540 `needs-triage` |
| `review-pr-snapshot-path.test.mjs:62` — merge two tests into a loop | `simplify` | #541 `needs-triage` |
| `review-pr-snapshot-path.test.mjs:31` — single-use `liftSnapshotMissing()` | `simplify` | comment on #533 |
| `snapshotBlock()` duplicates `between()` | `simplify` | comment on #500 |

Two of the four unverified `suggestion`s in scope got a dedicated refuter. `review-pr-reads.test.mjs:378` (add `snapshotMissing` to the declared-exactly-once guard) **survived** and was applied — a duplicate declaration placed *after* the real one measured 22/22 green, because the lift regex takes the first declaration while JS runs the last. `review-pr-snapshot-path.test.mjs:62` was **refuted** on two of its three supporting claims and is filed with the measurements (#541).

The 4 `refuted` findings were not applied and are not filed — the review settled them.

feigi:
**Correction to the comment above:** one of the four `refuted` findings has been reversed and is now applied in `3b1dd72`.

`workflows/review-pr.js:512` — *"`mkdir -p` never wipes and `tar -x` merges, so a reused scratch dir makes `pathVerified: true` certify a stale tree"* — was refuted 2-0 by the review. **The refutation was wrong.** The defect was reproduced operationally: two `review-pr` runs sharing one `scratch` argument left the snapshot holding files from both PRs, and one PR's entire review ran against a tree that was half another PR's branch.

`pathVerified` cannot catch this and never could — a merged tree is non-empty *for real*, so `SNAPSHOT_NONEMPTY` passes it trivially. The directory is not empty, it is full of the wrong repo state, which is exactly what `pathVerified` exists to reject. The symlink reorder in `0a21ceb` addresses a different masking and does nothing here.

Fix: `rm -rf ${scratch}/snapshot` before `mkdir -p`. Sequence is now wipe → extract → probe → symlink.

**Why the wipe and not a per-PR path**, both measured in a fixture:

| | run 1 | run 2 |
|---|---|---|
| no wipe, shared scratch, two PRs | `only-on-A.txt` | `only-on-A.txt only-on-B.txt` ← contaminated |
| with wipe | `only-on-A.txt` | `only-on-B.txt` |
| **per-PR path**, same PR re-reviewed, run 2's archive fails | `only-on-A.txt` | `SNAPSHOT_NONEMPTY` over run 1's files ← **still stale** |
| wipe, same case | | `SNAPSHOT_EMPTY` ← correctly refused |

A per-PR path fixes only the cross-PR case; re-reviewing the same PR is the same path both times. The wipe fixes both, in one line, and leaves `snap.path` and every downstream interpolation untouched. `scratch` is `A.scratch || /tmp/review-pr-${pr}` (`review-pr.js:344`), so it is never empty, and the `/snapshot` suffix means the wipe cannot reach the scratch root.

The root cause is worth naming: every other scratch user is namespaced (`<scratch>/pr<N>/<finding>/` per `run-team/SKILL.md`) while the snapshot alone sat at a bare path — which is why it was the one that merged. That is recorded in the prompt comment.

Pinned as a **sequence**, not a presence check, since two different reorderings each defeat the guard (probe after symlink; wipe missing or after `tar -x`). Folded into the ordering pin from `0a21ceb` rather than added beside it. Both new mutants red that test alone, the six earlier mutants still isolate, and the two green-discrimination checks still pass. Suite: **755/755**.

The other three `refuted` findings stand as refuted, and #538-#541 are unchanged.

PR #537, opened 2026-08-17
docs(fleet): stop pairing the review snapshot with ./agent-test

### LINKED ISSUE #141: run-team/SKILL.md and isolation.md still pair the snapshot with ./agent-test, which review-and-fix.md now contradicts
Deferred from PR #130 review.

PR #130 established that `./agent-test` **cannot exist in a review snapshot** — the snapshot is `git archive HEAD | tar -x`, tracked files only, and `claim-ticket.sh` writes the runner into the worktree and `.git/info/exclude`s it. `review-and-fix.md:43` was corrected accordingly: specialists get `node --test skills/fleet/scripts/*.test.mjs`, and `./agent-test` is kept for the reviewer's own worktree verification.

Three sibling sites still say the opposite and were left unchanged:

- **`skills/fleet/skills/run-team/SKILL.md:515-519`** — *"The snapshot and `./agent-test` solve different problems ... 'I'm on my own copy' is exactly the intuition that skips the runner — **say both, every time**."* For a specialist on a snapshot, saying both hands out a command that is not there.
- **`skills/fleet/skills/run-team/references/isolation.md:11`** — same pairing, same wording.
- **`skills/fleet/scripts/inflight.test.mjs:3`** — `// ./agent-test skills/fleet/scripts/inflight.test.mjs`, a run instruction that fails from a snapshot.

**`SKILL.md:511-514` forbids exactly this**, in its own words:

> "**Specialist tree isolation is `review-and-fix.md`'s job — do not restate it.** It owns the object-store rule, the two-tree split, snapshot provisioning ... Your only duty is to *not* contradict it."

The member-in-worktree path is genuinely different (the runner exists there) and should keep saying `./agent-test`. What needs correcting is the sentence that pairs it with the **snapshot**.

**Fix (settled):** at `SKILL.md:515-519` and `isolation.md:11`, split the two audiences — members in a worktree use `./agent-test`; specialists on a snapshot use the stack-free command `review-and-fix.md` names. Point at `review-and-fix.md` rather than restating it. Fix the `inflight.test.mjs:3` header comment to name a command that works in both places.

**Coordination note:** at the time of filing, another fleet member was editing `run-team/SKILL.md` on a separate ticket. Check for conflicts before claiming.

Found independently by this reviewer and by the comment specialist on PR #130.

feigi:
`review-and-fix.md` established that `./agent-test` cannot exist in a review snapshot: the snapshot is `git archive HEAD | tar -x`, tracked files only, and `claim-ticket.sh` writes the runner into the worktree and `.git/info/exclude`s it. Three sibling sites still said the opposite.

**Measured, not inferred:** a `git archive HEAD | tar -x` snapshot of this HEAD carries no `agent-test`. Both suites whose header comments this changes pass under `node --test <path>` from the worktree root *and* from that snapshot — 35 and 72 tests, 0 fail, in each.

## What changed

- **`run-team/SKILL.md` and `references/isolation.md`** — the "filesystem isolation is not stack isolation" guard paired the snapshot with `./agent-test` and closed with "say both, every time", handing a specialist on a snapshot a command that is not there. Both now name the audience with the command: a member in a worktree uses `./agent-test`; a specialist on a snapshot takes the one `review-and-fix.md` hands out. **Pointed at, not restated** — the bullet directly above reserves specialist tree isolation, snapshot provisioning included, to that document, so naming the command here would trade one contradiction for a duplicate that can drift.
- **`inflight.test.mjs:3`** — header comment said `./agent-test <path>`, a run instruction that fails from a snapshot. Now `node --test <path>`.
- **`fleet-tick.test.mjs:2`** — **the same defect, not named by the ticket.** Found by sweeping every `*.test.mjs` header rather than fixing the one site listed. Seven sibling headers already say `node --test`; these two were the outliers.

## Scope searched

Whole tree (`--exclude-dir=.git --exclude-dir=node_modules`) for `agent-test`, then every file co-occurring with `snapshot`. That leaves five prose sites, and the three not touched here are correct as they stand: `review-and-fix.md:47` (already fixed by #130), and `claim-ticket.sh:65`, whose snapshot mention is about `derive-testcmd.sh` being shared with `review-pr.js`, not about the runner.

## The false-positive half

The member-in-worktree path genuinely has the runner and must keep saying `./agent-test`. An over-broad sweep that stripped it there is the opposite failure, so it is pinned in both directions:

- `agent-test-dir-prose.test.mjs` already pins `` `./agent-test <file-or-dir>` `` in both documents' member-briefing slices — untouched, still green.
- The new `snapshot-runner-audience-prose.test.mjs` pins **both** sides of the split: the guard points a snapshot specialist at `review-and-fix.md`, *and* the guard still scopes `./agent-test` to the worktree audience. All four went red before the prose landed.

## On the new pins

- Not reflow-brittle: `phrase()` matches across any run of whitespace, so `SKILL.md`'s hard-wrapped copy and `isolation.md`'s long lines take the same regexes.
- The isolation.md heading `## Filesystem isolation is not stack isolation` is deliberately left verbatim — it is the **end boundary** of `agent-test-dir-prose.test.mjs`'s slice, and renaming it would have reddened that suite.
- The fifth test is a class guard, not a point fix: no `*.test.mjs` may carry a runner invocation naming a concrete suite path. It stays an escaped regex because as a plain string literal the pattern occurs in the scanning file, and the scan reports itself — observed.

753 tests, 0 fail, after rebase onto `origin/main`.

Closes #141


feigi:
Suggestion band checked and adjudicated. The review returned `survived: 0, refuted: 0, unverified: 8` — that band gets 0 refuters by policy, so all 8 arrived unchecked. Five refuters ran; one took the three findings that targeted the same regex, since they were mutually exclusive and independent refuters would each have approved their own side.

**Applied in c9f19a2** (3 of 8)

- `correctness:50` — **survived.** Both prose pins point at `review-and-fix.md` rather than restate its command, and nothing held the target present: deleting that bullet reddened nothing across all 753 tests. Now pinned, bounded at the generic next-bullet marker — the glob occurs twice in that file, so an unbounded pin would survive the delete.
- `simplify:46` — **survived.** `slices` is a data table; each row carries its label once, so the test name and the doc named in a drift failure cannot disagree. Failure messages byte-identical, `between()` still lazily invoked.
- The regex widening — a **synthesis**, after all three competing remedies were refuted (see #542). `/agent-test\s+(?!\/\/)\S*(?:\/|\.test\.mjs)/`: 9 offender spellings caught, 6 non-offenders green, `offenders: []` across every sibling suite, no self-report.

**Deferred** (5 of 8)

- #500 — `types:36`, the `between()`/`phrase()` duplication. Commented rather than duplicated; the count there is now 8 files for `between()`, 3 for `phrase()`. Out of #141's scope by the finding's own text.
- #542 — the refuted band, with the measurements: `correctness:80` (its widening reds correct prose at 7 of 71 reflow widths), `silent-failure:80` (its pattern silently drops the directory form the old regex already caught), `types:76` (leave-it-alone misreads a false-negative guard as loud), `comments:92` (its suggested wording embeds the literal the guard scans for — applying it verbatim made the file report itself, 5/5 → 4 pass / 1 fail), `simplify:78` (zero lines saved; trades a structural short-circuit for an order-dependent one).

**Tests:** `node --test skills/fleet/scripts/*.test.mjs` from the worktree — **754 tests, 754 pass, 0 fail**. Baseline on `cb0ae83` measured at 753/753/0; the +1 is the new pin. Every applied change was mutation-verified off the post-fix tree: each kills its own mutant, only that test reds, and each stays green on a change it should not catch.

PR #544, opened 2026-08-17
fix(fleet): quote the label into candidates.mjs's search term

### LINKED ISSUE #175: candidates.mjs interpolates the label unquoted — a multi-word label exits 1 against a non-empty queue
Deferred from PR #167 review. Pre-existing on `main`, untouched by #167, same fail-open class as #61.

## What to build

`skills/fleet/scripts/candidates.mjs:69` builds the search term by raw interpolation:

```js
const search = label ? `${EXCLUDE} label:${label}` : EXCLUDE;
```

A label containing a space is silently misparsed by GitHub — the second word becomes a free-text term rather than part of the label. Measured against `feigi/claude-config`, which carries both `good first issue` and `help wanted`:

```
label:good first issue    -> 0 results
label:"good first issue"  -> 5 results
```

So `candidates.mjs --require-label "good first issue"` reports an **empty queue at exit 1** while the queue is not empty. That is exactly the confusion #61 exists to remove: exit 1 means "the query worked and there is no work", and here the query silently did not mean what was asked.

Not reachable from either current caller — both pass the literal `ready-for-agent` — which is why it was deferred rather than fixed in #167.

## Acceptance criteria

- [ ] A label containing spaces is quoted into the search term
- [ ] `--require-label "good first issue"` returns the matching issues rather than exit 1
- [ ] A single-word label (`ready-for-agent`) still behaves identically — pin both in a test

## Notes

Found by the `code-reviewer` specialist during the #167 review; the 0-vs-5 measurement above is its verification against the live repo.

feigi:
`candidates.mjs:166` built its search term by raw interpolation — `label:${label}` — which hands the label to GitHub's **query parser as syntax**, not as a value. An unquoted qualifier value ends at the first space; every word after it becomes a free-text term. A multi-word label therefore queried for something nobody asked for, and its empty result reached the caller as **exit 1** — "the query worked and there is no work" — against a queue that was not empty.

## Measured against `feigi/claude-config`, 2026-08-17

| search term | open issues |
|---|---|
| `label:ready-for-agent` | 72 |
| `label:ready-for-agent candidates` | 12 |
| `label:"ready-for-agent"` | 72 |

Row 2 is the bug: the second word narrows the result as free text. Row 3 is what makes the fix free — quoting a plain value is a no-op.

End-to-end on the **full production term** (`EXCLUDE` + the label both callers pass), quoted and unquoted: **65 and 65**. `candidates.mjs --require-label ready-for-agent` runs live at exit 0 with 64 rows.

The ticket was written from `good first issue` and `help wanted`; the class is *label values GitHub's parser reads as syntax*.

- **Space-containing labels** (`good first issue`, `help wanted`) — both named in the ticket. Covered, pinned end-to-end.
- **A single-word label** (`ready-for-agent`) — AC-3. Covered; measured no-op above, pinned at the wire.
- **`:`-containing labels with no space** (`status:ready`, `area:docs`) — search-significant without a space anywhere. Covered *because* the quoting is unconditional; a `/\s/`-triggered version would miss it, so a test pins it.
- **`"` inside the label** — the failure mode this fix itself newly creates: the `"` closes the quoted term early and the rest becomes free text. Whether GitHub honours `\"` inside a qualifier is unverified, so this **refuses at exit 2** before gh runs rather than sending a query whose empty answer would read as exit 1. Deliberately a refusal, not an escape.
- **The hardcoded `EXCLUDE` terms** — not parameterised, all single-word. Left alone; `drop-merged-label.test.mjs:244` pins that string and stays green.

Reachable rather than theoretical: `docs/agents/triage-labels.md` exists to let a repo remap all five triage label strings.

## Exit-code contract

| code | reached by this change | pinned by |
|---|---|---|
| 0 | multi-word label now returns its rows | new multi-word + single-word tests |
| 1 | the defect — was reached wrongly, no longer | new multi-word test (its RED state was exactly `1 !== 0`) |
| 2 | new `"` refusal | new refusal test, `queriesRun === 0` |
| 3 | unaffected — the change is upstream of the filter | existing #64 tests, still green |

## Test-side change

The stub could not tell the two forms apart: both contain `` label:``, so it served the labeled fixture either way and this bug would have stayed green under any fix. It now models the one split that matters — an unquoted value ends at the first space — opt-in through `$LABEL_EXPECT`, so every pre-existing test keeps exactly its old dispatch.

## Verification

- `./agent-test skills/fleet/scripts` — **759/759 pass**.
- Written test-first: all four new tests RED before the fix, no pre-existing test disturbed.
- Mutation-tested, each half independently: quoting made conditional on `/\s/` kills exactly the single-word and `status:ready` tests (2 fail); deleting the `"` guard kills exactly the refusal test (1 fail).

Closes #175


feigi:
Review applied in 64e27a1. `node --test skills/fleet/scripts/*.test.mjs` — **tests 760, pass 760, fail 0** (baseline on 536cbef was 759).

**Deferrals filed:**
- **#546** (`needs-triage`) — `correctness`: EXCLUDE's five negative labels are still interpolated unquoted. Not reachable today (all five are single-word literals with no caller-supplied path), and the approach is genuinely open — quote them, or document the ceiling. Carries the ripple: two stderr assertions anchored on the literal `-label:in-progress` will red.
- **#547** (`needs-triage`) — the suggestion band, checked: 3 refuted with their measurements, 5 applied, 1 routed to #546.

**Both `survived` findings applied.** The block comment's "the second stays green either way" was half false — the second test's query outcome is identical either way, but its wire-form assertion reds on the raw interpolation too. And the `:` rationale was measurably false: `label:auto:logs` == `label:"auto:logs"` on `renovatebot/renovate`, `label:Team:Core` == quoted on `elastic/kibana`. Space is the character GitHub actually splits on. Unconditional quoting and its test are kept — a mutation to `/\s/`-triggered quoting still reds — but the reason is now uniformity, not a second dangerous character.

**One suggestion turned out to be more than a suggestion.** GitHub *does* honour `\"` inside a qualifier — the fact the shipped comment called unverified. `-label:"zzz\" label:"ready-for-agent"` returns the free-text count, not the quoted one. In the shape `query()` builds, a trailing backslash leaves the value unterminated and the query answers zero rows at HTTP 200 — #175's silent empty, reached *through* the fix. Guard widened to `/["\\]/` with a test that reds when narrowed back.

PR #316, opened 2026-08-10
docs(fleet): name the fields that go undefined, not a scratch field that never existed

feigi:
The `args.dimensions` normalization comment in `skills/fleet/scripts/select-dimensions.test.mjs` read:

    dereferencing to `undefined` prompt/agentType/scratch path

Unhyphenated, `prompt/agentType/scratch` reads as a list of dereferenced fields, implying a `d.scratch` on a dimension object. Grep finds no `d.scratch` / `dim.scratch` anywhere in the repo. `scratch` is a top-level workflow arg (`const scratch = A.scratch || ...` in `workflows/review-pr.js`) that gets combined with `d.key` to build a path — so what goes `undefined` and corrupts that path is `d.key`.

Fix takes the second option the issue offers: match the wording the sibling comment above `resolveDimensions` in `workflows/review-pr.js` already uses for the same defect, listing the three fields `REQUIRED` names and the fan-out dereferences.

```
-// untouched, dereferencing to `undefined` prompt/agentType/scratch path with
-// no throw and no warning.
+// untouched, dereferencing `d.key`, `d.prompt` and `d.agentType` to
+// `undefined` with no throw and no warning.
```

Comment-only, one site. `./agent-test skills/fleet/scripts/select-dimensions.test.mjs` → 27 tests, 27 pass, 0 fail.

### Separate defect found, deliberately not fixed here

Both this comment and its sibling in `workflows/review-pr.js` cite `skills/fleet/commands/review-and-fix.md:49` for the phrase "keys or dimension objects". That phrase is on **line 53**; line 49 is an unrelated bullet about settling the fan-out before applying. Off-by-four in two files — a different defect than the one #276 names, and fixing one site would leave the two out of sync. Worth its own ticket.

Closes #276

Issue #282, opened 2026-08-10
fleet-plugin-design spec: three live bare-numbered citations survived the #117 sweep, two of them into this repo


feigi:
Deferred from PR #272 review. **dimension: `correctness`** (finding `refuted-1`, refuted 1-1, reversed back into scope by the controller because it alleges the same defect class the PR's `critical` finding confirmed — a universal claim the diff added that does not hold).

PR #272 closed `#117` and added a closing clause to the migration row of `docs/specs/2026-07-23-fleet-plugin-design.md`:

> every live citation in this spec now names a quoted fragment and its section, per the convention marker above

That was false as written. **PR #272 narrowed the clause** to `every live citation **into `run-team` SKILL.md** from this spec`, which is true — zero live bare `run-team` pointers remain. The other half of the finding's suggested fix — finishing the sweep on this file — is what is deferred here.

## What remains

Three live, present-tense, bare-line-numbered citations survive in that document. All three currently **resolve**, so this is fragility, not rot — but two of them point into *this* repo, where the numbers rot from here.

| spec line | citation | resolves to | missing |
|---|---|---|---|
| `:91` | `` `${CLAUDE_PLUGIN_ROOT}` (`command-development/SKILL.md:564`) `` | `Plugin commands have access to `${CLAUDE_PLUGIN_ROOT}`, an environment variable that resolves to the plugin's absolute path.` — section `### CLAUDE_PLUGIN_ROOT Variable` (`:562`) | the **section** half; also lacks the `vendored outside this repo so its numbers do not rot from here` parenthetical that its sibling `plugin-structure/SKILL.md:431` received seven lines earlier in this very PR |
| `:98` | `` `plugins/` is gitignored in `claude-config` (`.gitignore:40`…) `` | `plugins/` | the **section** half (`.gitignore`'s `# Auto-managed by Claude Code` header) — and `.gitignore` is in this repo, so the number rots on any edit above line 40 |
| `:108` | `` `.gitignore:49` ignores `skills/*` `` | `skills/*` | same |

Verified at `27ee41ec5b443837c4878ff3ab86d36ca5f1c08a`:

```
$ sed -n '40p;49p' .gitignore
plugins/
skills/*
$ sed -n '564p' ~/.claude/plugins/marketplaces/claude-plugins-official/plugins/plugin-dev/skills/command-development/SKILL.md
[...]

Issue #281, opened 2026-08-10
review-pr.js:415 missing-field error carries no entry index — a multi-entry args.dimensions cannot say which entry failed


feigi:
`resolveDimensions`'s missing-field error carries no entry index — with a multi-entry override you cannot tell which entry failed.

**dimension:** silent-failure (surfaced during adversarial verification of PR #271, not claimed by any finding)
**file:** `workflows/review-pr.js:413-415`

## Observation

    const missing = REQUIRED.filter((f) => !entry || !entry[f]);
    if (missing.length)
      throw new Error(`review-pr: args.dimensions object is missing required field(s): ${missing.join(", ")}`);

The message names the missing *fields* but not **which entry** they are missing from. For a single-entry override that is unambiguous. For

    args.dimensions: ["correctness", {key: "custom", agentType: "x"}, "comments"]

the caller is told `missing required field(s): prompt` with nothing indicating it was entry 1. The `override.map` callback already has the index available as its second parameter, so surfacing it is cheap — e.g. `args.dimensions[1] is missing required field(s): prompt`.

## Provenance and honest scoping

This was **found by a refuter working on a different finding** (`unverified-6`, which claimed the message *mislabels* non-object entries — refuted, filed as #279). The refuter explicitly flagged this as outside what that finding claimed, and disclaimed the reachability half:

> "the message carries no entry **index**, so with a multi-entry override you cannot tell which entry failed. That is a real gap, but it is not what this finding claims, and I did not test whether anyone passes multi-entry overrides."

Filed separately rather than folded into #279 so it does not inherit that finding's refuted framing — the two are independent, and this one was never argued against.

**Reachability is untested.** Nothing in the repo constructs `args.dimensions` programmatically; both dispatch sites pass a fixed arg set with no `dimensions` key, and the only caller-facing mention is `skills/fleet/commands/review-and-fix.md:49` ("Pin a subset explicitly with `args.dimensions`"), which is hand-typed. Whether multi-entry overrides are used in practice — and therefore whether this gap ever bites — was not measured by anyone.

## Note on interaction

If #279 is ever actioned, its rewrite touches the same throw and should carry the index at the same time rather than in a second pass. #277 also touches this line.

Deferred from PR #271 review.

feigi:
Still present, re-measured during the **PR #545** review (dimension: `silent-failure`). Deferred from that PR — comment-only diff, does not touch this code.

[...]

Issue #279, opened 2026-08-10
review-pr.js:413 'object is missing required field(s)' calls a non-object an object — fixing it requires rewriting a #271 test pin


feigi:
The "object is missing required field(s)" message calls a non-object an object — `[null]`, `[123]`, `[["correctness"]]` all get the same string.

**dimension:** silent-failure (`unverified-6` of the PR #271 review) — **note the dimension is itself part of what was refuted, see below**
**file:** `workflows/review-pr.js:413-415`

## The observation

Measured against `1e1aa04c:workflows/review-pr.js` with the repo's own `new Function` lift:

    THROW [null]            -> args.dimensions object is missing required field(s): key, prompt, agentType
    THROW [123]             -> args.dimensions object is missing required field(s): key, prompt, agentType
    THROW [["correctness"]] -> args.dimensions object is missing required field(s): key, prompt, agentType
    THROW [undefined]       -> args.dimensions object is missing required field(s): key, prompt, agentType
    THROW [{}]              -> args.dimensions object is missing required field(s): key, prompt, agentType
    THROW [{key:"x"}]       -> args.dimensions object is missing required field(s): prompt, agentType
    OK    ["correctness"]   -> [{"key":"correctness",...}]

A caller who wrote `[["correctness"]]` — a plausible slip, since the override *is* an array — is told their object lacks three fields.

## ⚠️ Applying this breaks a test added by PR #271 — read before implementing

PR #271 added **"a null or undefined entry stops the run instead of crashing on the field check"**, which pins `resolveDimensions([null], …)` to throw `/missing required field\(s\): key, prompt, agentType/`. The suggested fix below **changes that exact string**. Measured, by applying the fix to a private copy and running the incoming assertion:

    incoming [null] pin: FAIL -> did not match /missing required field\(s\): key, prompt, agentType/
    actual [null] msg now:     "…entry is neither a key string nor a dimension object: null"

So **the pin must be rewritten in the same change.** An implementer applying this alone will turn a green suite red and not know why. The pin itself is load-bearing and mutation-verified — it is the only coverage of the `!entry ||` clause, without which `[null]` crashes with `TypeError: Cannot read properties of null (reading 'key')` — so **rewrite it to the new message, never delete it**.

Suggested fix as originally proposed:

    if (!entry || typeof entry !== "object" || Array.isArray(entry))
      throw new Error(`review-pr: args.dimensions entry is neither a key string nor a dimension object: ${JSON.stringify(entry)}`);
[...]

PR #545, opened 2026-08-17
docs(fleet): anchor the args.dimensions citation on its section, not a line

### LINKED ISSUE #317: review-pr.js:411 cites review-and-fix.md:49 for a phrase that is on line 53
Found by `impl-276` while implementing #276, and re-measured by the controller before filing. **Not** a review deferral — an incidental find, deliberately kept out of #276's diff because that ticket is a one-site comment-wording fix.

## The defect

`workflows/review-pr.js:411` (at `origin/main`):

```js
// `skills/fleet/commands/review-and-fix.md:49` documents `args.dimensions` as
```

The phrase it points at — *"Pin a subset explicitly with `args.dimensions` (keys or dimension objects)"* — is on **line 53**, not 49.

Measured at `origin/main`:

```
$ git show origin/main:skills/fleet/commands/review-and-fix.md | grep -n "keys or dimension objects"
53:**The fan-out scales to the diff — the workflow computes it, you don't.** … Pin a subset explicitly with `args.dimensions` (keys or dimension objects). …

$ git show origin/main:skills/fleet/commands/review-and-fix.md | sed -n '49p'
- **Apply only once the fan-out is settled — every report either in hand or accounted for.** …
```

Line 49 is an unrelated bullet about settling the fan-out before applying. It resolves to real, plausible-looking prose about the same subsystem, which is what makes the citation quietly misleading rather than obviously broken.

## Scope — TWO sites (corrected 2026-08-10 — see the correction comment)

> **The original text of this section was wrong** and claimed one site. It rested on a line-oriented grep that could not match a citation wrapped across two source lines. Corrected below; the reasoning is in the correction comment on this issue.

Two files cite `review-and-fix.md:49` for a phrase that lives at `:53`:

| file | line | note |
|---|---|---|
| `workflows/review-pr.js` | 411 | unwrapped |
| `skills/fleet/scripts/select-dimensions.test.mjs` | 173-174 | **wrapped** across the line break: `review-` / `and-fix.md:49` |

They are copies of one another — fix one without the other and they are out of sync again.

Sweep that finds both (match the tail alone, so the wrap cannot hide a hit):

```
$ git grep -nI "and-fix\.md:[0-9]" origin/main
```

12 hits repo-wide. Scope searched: all tracked files at `origin/main`, no pathspec restriction, cross-checked by stripping comment prefixes and joining lines per file.

**PR #316 is editing the `select-dimensions.test.mjs` comment paragraph** and its review rated that site `important` with both refuters returning `refuted=false`, so it may already be fixed. **Check the tree first** — if `:174` is corrected, this ticket is only `workflows/review-pr.js:411`.

## Not a uniform off-by-four — check each, do not batch-shift

The sibling citation in the same file is **correct**: `workflows/review-pr.js:163` cites `:47`, and `review-and-fix.md:47` is exactly the bullet it describes (*"Give specialists a stack-free test command, not the worktree's runner, and their own scratch dir"*).

So a blanket `+4` sweep would **break** `:163` while fixing `:411`. Resolve each citation against the text it names.

## Also worth a look, but verify before touching

`docs/specs/2026-08-06-review-pr-specialist-read-rules-design.md:47` cites `review-and-fix.md:47` for the **source-of-truth rule**, which is at `:51` (`git show origin/main:skills/fleet/commands/review-and-fix.md | grep -n "source of truth"` → 51). That is a spec, and a spec citation may have been accurate when written — resolve it at the revision it was made, not only at HEAD, before deciding it is stale. It may also be better handled under #282's citation-convention work (quoted fragment + section, so the number stops being load-bearing) than as a number bump.

## Acceptance criteria

- [ ] `workflows/review-pr.js:411` names the correct location for "keys or dimension objects"
- [ ] `skills/fleet/scripts/select-dimensions.test.mjs:173-174` likewise — **unless PR #316 already fixed it**; verify, do not assume either way
- [ ] The two stay in sync with each other
- [ ] `workflows/review-pr.js:163` is left correct (verify, do not assume)
- [ ] Preferably the citation carries a quoted fragment and its section per the convention marker, so the line number is no longer the only anchor

feigi:
`workflows/review-pr.js` cited `skills/fleet/commands/review-and-fix.md:49` for the phrase *"keys or dimension objects"*. That phrase is at **:53**. Line 49 is an unrelated bullet about settling the fan-out before applying — so the citation resolved to real, plausible prose about the same subsystem instead of failing visibly.

## The fix

Drops the line number rather than bumping it, following PR #316's fix to the `select-dimensions.test.mjs` copy of this same comment. The citation now anchors on the section name plus the verbatim quoted fragment:

```js
// The "Specialists" section of `skills/fleet/commands/review-and-fix.md`
// documents `args.dimensions` as accepting "keys or dimension objects" — but
```

Both anchors verified against `origin/main` before writing them: `## Specialists` is at :35, the fragment at :53, the next header (`## Judging findings`) at :57 — so the fragment is inside the named section, and the section header is *already* pinned by `review-path-default.test.mjs:219`, which extracts that exact range through a `section()` helper that hard-asserts both bounds.

## Not a uniform shift

The sibling citation in this same file — `review-and-fix.md:47`, for the stack-free-test-command bullet — **is correct** and is left untouched (re-verified at `origin/main`). A blanket `+4` would have broken it.

## Scope

Swept with the wrap-safe pattern the issue prescribes, `git grep -nI "and-fix\.md:[0-9]" origin/main`, plus a strip-and-join pass over every tracked file (comment prefixes stripped, lines joined) to catch citations wrapped across a line break — the failure mode that produced this issue's first, wrong scope claim. Both passes agree: 11 hits.

| site | disposition |
|---|---|
| `workflows/review-pr.js` (`:49`) | **fixed here** |
| `workflows/review-pr.js` (`:47`) | correct — left |
| `select-dimensions.test.mjs` | already fixed by #316 (`c2c0f1f`); no longer a hit |
| `docs/specs/…-specialist-read-rules-design.md` (`:47` ×3) | **accurate when written** — resolved at `c3c49a4`, where the source-of-truth rule *was* at :47. Historical record, left. |
| `docs/specs/…-cost-and-apply-policy-design.md`, `…-fleet-plugin-design.md` | spec records, out of scope |

The spec citations are left deliberately: per the issue, a spec citation may have been accurate when made, and this one was. Renumbering it would falsify the record rather than fix it. #282's citation-convention work is the right home if they are ever re-anchored.

## Verification

`./agent-test skills/fleet/scripts` → **763 pass, 0 fail** (non-zero count, so not a silent empty run). Eight test files read `workflows/review-pr.js` and all are in that run. Comment-only diff — no code lines touched. No new test added, matching #316's precedent; the structural anchor is already guarded by the existing `section()` pin.

Closes #317


feigi:
**Fix-applier pass** — pushed `b7c09b3`.

**Applied (1/1 survived):** `workflows/review-pr.js:216` cited `review-and-fix.md:47` for the "`git show <sha>:<path>` is the source of truth … Then bound the read" rule, which lives at `:51`; `:47` is the unrelated stack-free-test-command bullet. Re-anchored on `## Specialists` + verbatim quoted fragments, same form as this PR's sibling fix. Anchors re-derived from the tree: heading at `:35`, fragments at `:51`, next heading at `:57`.

Added `skills/fleet/scripts/review-pr-citation-prose.test.mjs` — nothing guarded either citation. Bans line-numbered citations into `review-and-fix.md` from this file, and pins all three anchors as quoted verbatim in `review-pr.js` *and* still inside the section they name. Mutation-verified 4 ways (restore `:47` → 3 red; reword the md fragment → 1 red; delete the comment → 2 red, so the ban is not vacuous; retarget the unrelated `code-reviewer.md:21` → stays green).

**Deferred (1):** `resolveDimensions`' validation throw labels a non-string entry an "object" and carries no entry index. Pre-existing, out of scope for a comment-only PR. **Not filed as new** — already open and covering both halves: #279 (the "object" label, incl. the #271 test pin any fix must rewrite) and #281 (the missing index). Re-measured against `b7c09b3` and commented on both.

Note for the fleet: `ledger.mjs check` returned `verdict: "clean"` with `tracker.ok: true` for this subject — its 3-term query (`resolvedimensions validation neither`) missed both. The step-5 `gh issue list --search` is what caught the duplicates.

`node --test skills/fleet/scripts/*.test.mjs` from the worktree: **767 pass / 0 fail** (baseline measured at `f0027ff`: 763/763; +4 new).

Issue #122, opened 2026-07-31
inflight.sh probe 3: a tab or newline in a worktree path truncates the reported path


feigi:
Deferred from PR #110 review. Pre-existing, and explicitly scoped out of that PR — filing so it is not lost.

`skills/fleet/scripts/inflight.sh:189-193` reformats `git worktree list --porcelain` into `basename<TAB>path` lines, then splits them with `awk -F'\t'`. A tab *inside* the path becomes an extra field separator.

Measured, worktree named `fix-88-ta<TAB>b`:

```
exit=1  {"issue":88,"taken":true,"hits":["local"],"evidence":{...,"worktree":"b"}}
```

The payload **parses** and the verdict is **correct** — this is not a repeat of #90. What breaks is fidelity: the reported path is `b`, which exists nowhere on disk, so a human or tool acting on the evidence is sent to a path that is not the worktree. `jstr` cannot help; the truncation happens two stages upstream, before the payload is built. A newline is eaten earlier still, by `while IFS= read -r p`.

Reachability is low (a tab in a directory name is legal but rare, and fleet worktrees are named by `claim-ticket.sh`) which is why this is triage rather than a fix.

Approach is open: a NUL-separated pipeline (`-z` / `awk -F'\0'`) would handle it, but `--porcelain` itself is newline-delimited, so the real fix may be `--porcelain -z` plus a rewrite of the whole probe-3 filter. Worth deciding whether the fidelity is worth that churn at all.

feigi:
> *This was generated by AI during triage.*

Absorbed into #89 rather than actioned separately, and closed on that basis.

#89 rewrites this exact probe-3 filter — collapsing `awk | while read | awk | paste` into a single `awk` — so both defects live in the same five lines and two agents would collide there.

Measured during triage, git 2.50.1:

- this issue reproduces as filed: worktree named `fix-88-ta<TAB>b` yields `{"…","worktree":"b"}` with the verdict correct
- the single-`awk` replacement **preserves the tab**, because the tab-delimited intermediate is what loses it and the rewrite removes that stage
- `git worktree list --porcelain -z` **is supported**, so the newline case is closable in the same pass

Both are now acceptance criteria on #89. The open question in this body — whether the fidelity is worth the churn — is answered: it is not worth its own churn, and it is nearly free inside #89's.

No `.out-of-scope` entry: this is absorbed work, not a rejected request. Reopen if you would rather it stayed separate.

PR #548, opened 2026-08-17
fix(fleet): a newline in a worktree path no longer makes inflight.sh invent a "free"

### LINKED ISSUE #185: inflight.sh probe 3: a newline in a worktree path yields a wrong "free" — #89's unmet acceptance criterion
Deferred from PR #174 review. PR #174 closes #89 but leaves one of #89's own acceptance criteria unchecked:

> - [ ] A worktree whose path contains a **newline** is likewise reported in full, using the NUL-delimited form of the listing

That criterion arrived when #122 (tab **or newline** truncation) was closed `not_planned` on the basis that it was *"absorbed into #89"* — so if #89 closes with #174, both tickets close and nothing tracks the newline case.

`git worktree list --porcelain` is newline-delimited, so a path containing a newline splits into two records. PR #174 removed the `-F'\t'` join and fixed the **tab** half; the newline half is one stage earlier, in the record splitting itself, and is untouched.

## Measured at `de0ebb1` (PR #174 head), git 2.50.1

Two distinct symptoms, depending on which side of the newline carries the number.

**1. Truncated evidence** — worktree `…/wt/fix-66-a<LF>b`, ticket 66:

```
{"issue":66,"taken":true,...,"evidence":{...,"worktree":".../wt/fix-66-a"}}   exit=1
```

Verdict correct, but the reported path does not exist on disk — the same fidelity loss #122 filed for tabs.

**2. A wrong "free"** — worktree `…/wt/plain<LF>fix-33-slug`, ticket 33:

```
{"issue":33,"taken":false,"hits":[],"evidence":{...,"worktree":""}}           exit=0
```

**This is the answer the script must never invent.** The worktree for #33 is checked out and the probe reports the ticket free — the case that puts two implementers on one ticket. This is strictly worse than #122, whose body explicitly recorded *"the verdict is correct"* for the tab case.

Identical on `d856185` (pre-#174), so this is **pre-existing, not a regression from #174** — which is why it was left out of that PR rather than fixed in it.

## Reachability

Low, and the same order as #122: a newline in a directory name is legal on APFS and ext4 but rare, and fleet worktrees are named by `claim-ticket.sh`. It is the *consequence* that is severe, not the frequency.

## Approach

Settled by #89's criterion and confirmed during #122's triage: `git worktree list --porcelain -z` is supported on the git in use (verified, 2.50.1).

A shell variable cannot hold NUL, so `worktrees=$(git worktree list --porcelain -z)` will not work. The shape that keeps **both** guards #174 established — a lookup whose status is readable, and a filter whose status is readable — is a temp file, for which this script already has precedent at `inflight.sh:47`:

```sh
git worktree list --porcelain -z >"$tmp" || die "git worktree list failed, …"
wt=$(LC_ALL=C awk -v RS='\0' -v n="$n" '…' <"$tmp") || die "could not filter the worktree list for #$n"
```

`RS='\0'` is a gawk/BWK extension, not POSIX — worth confirming against the awk on every target before committing to it.

## Acceptance criteria

- [ ] A worktree whose path contains a newline is reported **in full**, not truncated at the newline
- [ ] A worktree whose path carries the ticket number **after** a newline reads as **taken** (exit 1), never free
- [ ] The three-outcome contract #174 established is preserved: the lookup and the filter each keep a readable status and their own `die` message
- [ ] `inflight.test.mjs` gains a case for each of the two symptoms above, each verified red first

Deferred from PR #174 review.

feigi:
Closes #185

`git worktree list --porcelain` terminates every attribute with a newline, and a worktree path may legally contain one (APFS and ext4 both allow it), so one record split into two. Measured on the pre-fix script, git 2.50.1:

| worktree leaf | ticket | before | after |
|---|---|---|---|
| `fix-66-a<LF>b` | 66 | `evidence.worktree` = `.../fix-66-a`, a path not on disk | full leaf, `taken=true` |
| `plain<LF>fix-33-slug` | 33 | **`taken=false`, `hits=[]`, exit 0** — checkout live | `taken=true`, exit 1 |
| `a<LF>worktree fix-44-slug` | 44 | exit 2, *"the registry read missed entries"* | `taken=true`, exit 1 |
| `fix-88-a<TAB>b` | 88 | already whole (PR #174), pinned by nothing | unchanged, now pinned |
| `fix-99-ordinary` | 55 (free) | exit 0 | unchanged |

Row 2 is the severe one and #89's last unmet criterion: the number lands on the orphaned second record, where the `^worktree ` filter never looks, so a live checkout reads as free — the answer that puts two implementers on one ticket. Row 3 is the same root cause pointed at the *count* rather than the filter, and fails the opposite way: the orphan begins with `worktree `, inflating `listed`, so the probe refused a ticket that was fine.

## `RS='\0'` does not hold — verified, not assumed

#185 flagged it as a gawk/BWK extension worth confirming. It fails on the awk this script actually runs, `/usr/bin/awk`, BWK awk 20200816: it does not ignore the setting, it **stops at the first NUL and reports ONE record for a listing of any length**. Measured `count=1` for all three spellings — `-v RS='\0'`, `-v RS='\000'`, `BEGIN{RS="\0"}`. No awk program here can hold a NUL byte.

Shipping that form would have made `listed=1`, `linked=0` against 4 registry entries, so **every ticket in the repo** exits 2 with *"the listing is incomplete"* — total queue starvation. It fails 15 of the 77 tests in this file, the new accept case among them.

So the split happens before awk sees the stream, in one `tr` pass: NUL becomes the newline awk already splits on, a newline inside a path becomes `\001`. `tr` translates simultaneously from one table, so the two mappings cannot feed each other. **Both awks read the listing unchanged.**

The listing goes through a temp file because NUL cannot survive a command substitution, and because that keeps the lookup's status readable apart from the reader's — a pipeline reports only the last stage's, so a `git` that could not run at all would read as an empty listing, a wrong "free". Both guards #174 established are preserved, each with its own `die` message.

**Ceiling, deliberate:** `jstr` renders `\001` as a space rather than `\n`, so such a path is reported *whole but neutralised*, with `worktreeRewritten: true` telling the caller it is not a literal path. Swapping the byte back would corrupt the opposite case (a path really containing `\001`) for a diagnostic field whose verdict is right either way.

## Effect on the three-outcome contract

Unchanged in shape: 0 free, 1 taken, 2 could not evaluate. `run-team/SKILL.md` reads exit 2 as taken, so every new path below fails closed.

- **Newly reaches 1 (taken)** where it wrongly returned **0 (free)** — the fix's point.
- **Newly reaches 0/1** where it wrongly returned **2** — the inflated-count case above.
- **Newly reaches 2** from one place only: probe 3 now needs `mktemp`, so a broken TMPDIR records `local` unknown instead of answering. That costs the queue nothing — probe 1's unknown alone already forced exit 2 on a ticket with no hit, so a broken TMPDIR starved the queue before this change exactly as after. Both new `add_unknown` sites carry their own message and cannot collapse into 1; no new instance of #328's `die`-inverts-2-into-1 class is introduced.

The EXIT trap is hoisted to the top and now removes both temp files: `sh` keeps a single EXIT trap, so a second `trap … EXIT` in probe 3 would have silently replaced probe 1's and leaked its capture file.

## Tests

77 in this file (was 72), 768 across the repo, all green. The three newline cases were verified **red against the pre-fix script**; the tab and accept cases are green on both sides by design — the tab pins PR #174's unpinned repair (#122), and the accept case guards the direction no reject case can see, a parser that starves the queue by reporting free tickets as taken.

Two fixtures tracked the argument string and had to follow `-z`: the phantom-worktree shim and `registryRaceShim`, both keyed on `"worktree list --porcelain"`. `registryRaceShim`'s `fired` sentinel caught its own breakage, as designed.

## Left out of scope

Three sibling scripts parse the same listing the same way and have the identical defect — `release-ticket.sh:117`, `reap.sh:88`, `worktree-audit.sh:58`. Not touched: #185's criteria name `inflight.sh`, and each has its own suite. Worth a follow-up ticket.


feigi:
## Review applied — `7453685`

`node --test skills/fleet/scripts/*.test.mjs` → **tests 770, pass 770, fail 0** (baseline on this branch before the change: 768/768).

### Headline: this PR regressed the defect it exists to fix

`--porcelain -z` + `tr '\n\000' '\001\n'` rewrites an embedded newline to `\001`, and `\001` is not in the matcher's separator class `([-/]|$)`. Measured on real repos with real linked worktrees:

| leaf | plain porcelain (base) | `-z`, before `7453685` | now |
|---|---|---|---|
| `fix-33<LF>slug` | `taken=true` rc 1 | **`taken=false` rc 0** | `taken=true` rc 1 |
| `fix<LF>33-slug` | `taken=false` rc 0 | `taken=false` rc 0 | `taken=true` rc 1 |
| `fix-33-slug<LF>x` | `taken=true` rc 1 | `taken=true` rc 1 | `taken=true` rc 1 |
| `plain<LF>fix-33-slug` | `taken=false` rc 0 | `taken=true` rc 1 | `taken=true` rc 1 |
| `fix-99-ordinary` (free control) | rc 0 | rc 0 | rc 0 |

Both wrong-"free" rows carried `unknown: []` — the probe *answering*, not refusing, so nothing upstream caught it. Fixed by `gsub(/\001/, "/", b)` in the matcher: the byte counts as a separator for the match only, `p` is untouched, and the evidence is byte-identical under `od -c`.

The suite was green over it because the only newline pin was `plain<LF>fix-33-slug`, where the number sits between real `-` bytes. The two spellings fail in **opposite** directions and neither covers the other, so both are pinned now. Each new test verified red against its own mutant (and *only* that test red), green after, and green under two unrelated probe-3 edits.

### Applied

| finding | action |
|---|---|
| `silent-failure` `inflight.sh:628` (critical) | the `gsub` above, plus the missing `fix-<n><LF>slug` case |
| `comments` `inflight.sh:563` (critical) | the "tab gets the same treatment" clause was false — `\011` has an RFC 8259 short form, jstr escapes it before the `tr` stage, and this PR's own tab test asserts `worktreeRewritten: false`. Inverted. |
| `comments` `inflight.test.mjs:1716` (critical) | "70 of this file's 72 tests" re-measured on the post-fix tree: 77 of 79. Restated as "all but the two cases that expect a refusal anyway", which cannot rot on the next PR. |
| `comments` `inflight.test.mjs:949` (important) | the `\001` payload case's prose asserted a newline "cannot reach `jstr`" and that #185 is "still open" — both falsified by this diff. Rewritten, tab claim included (`comments:953`). |
| `types` `inflight.sh:800` (important) | the printf-guard comment still attributed the EXIT trap to probe 1; this PR hoisted it to file scope. (`correctness:800` and `silent-failure:800` are the same line — one change.) |
| `tests` `inflight.sh:568` (important) | probe 3's `tr` guard had no test; a silent `return 0` there left the suite green. Pinned via the `\000` shim, which addresses that call and no other. |
| `silent-failure` `inflight.sh:545` (suggestion, survived its refuter) | the `>"$wtfile"` redirect is new to that guard, so a temp file the run cannot write to was reported as a git failure. Message now names both operands. |

### Deferred

**#551** (`needs-triage`) — `reap.sh:88`, `release-ticket.sh:117`, `worktree-audit.sh:58` all parse `--porcelain` the same way and all truncate `.../wt/fix-33<LF>slug` to `.../wt/fix-33`, a path not on disk (measured, all three). `release-ticket.sh`'s own incompleteness guard does not fire on it. Its *other* divergences from `inflight.sh` are already tracked under #395. One issue, not three; `needs-triage` because whether the whole `-z` + `tr` shape ports to all three is unsettled.

### Suggestion band, checked (grouped refuters, biased to refuse)

- `silent-failure:544` — **refuted**. `-z` for `worktree list --porcelain` landed in git **2.36.0** (2022-04-18), but this repo's floor is already higher: `no-undo-audit.sh:413` runs `git merge-tree --write-tree --name-only -z` (2.38/2.39), same directory, same CI step. CI's `ubuntu-latest` ships git 2.54.0. Under a simulated pre-2.36 git the suite goes 41/77 red — machine-enforced, not silent. Nothing in the tree declares a git floor for any script (searched; zero hits), so a `-z`-only note would be the only one while a sibling silently demands more.
- `types:568` — **refuted**. The claim was that the three `tr` character classes are held together by prose alone. Dropping `\001` from jstr's replace set reddens 4 tests; dropping it from jrewritten's delete set reddens 1. Drift is loud.
- `simplify:538` — **refuted**. The pipeline variant reproduces the finding's own numbers (76/77) but recovers 11 lines, not 25, and under the *most plausible* trigger it emits "the listing is incomplete" — the exact message `inflight.sh` has a comment forbidding for this case. Rejected as the finding itself offers.
- `simplify:83` — **refuted**. The finding gates itself on a GNU-coreutils measurement no one on this host can take (no coreutils keg, no working container runtime). Simulating an `rm` that rejects an empty operand shows the guards buy more than two expansions: the fallback line degrades to `inflight: could not remove` naming no file.
- `simplify:132` — **refuted**, and it was the dangerous one. Probe 3 commits its branch half at `:466-469`, seventy lines *before* its `mktemp`, so a broken `mktemp` does not blind probe 3 — measured `hits: ["remote-branch","local"]` against a branch-backed worktree, which is the shape `claim-ticket.sh:99` actually creates. Applying the suggested reword would have replaced an accurate comment with a false one.
- `tests:544` — survives, but asks for no action here: pre-existing, unchanged by this PR, already tracked under the issue filed from #406's review.
- The two `refuted` findings from the review itself are not reversed and not filed; one concerns a file outside this PR's tree.

Issue #529, opened 2026-08-17
implementer-model-tier.test.mjs: two guard-test pins do not discriminate — one cannot fail at all


feigi:
Deferred from PR #523 review. Dimension: `tests` (both findings). Band: `suggestion` — 0 refuters by workflow policy; this fix-applier settled them on **scope**, not by refuting them. Both are in the `phase 2's guard is mandatory, runnable, and scoped to one class` test, which PR #523 does not touch.

Sibling record on the same test block, different pin: #476.

## Finding 1 — `assert.match(slice, /^\*\*Guard: /)` can never independently fail

**Location:** `skills/fleet/scripts/implementer-model-tier.test.mjs:246`

> `assert.match(slice, /^\*\*Guard: /, "the guard is no longer stated as an imperative")` can never independently fail — it is logically guaranteed to pass whenever `guard()` successfully returns a slice at all, so any real mutation that would violate it is always caught earlier (and under a different, unrelated message) by `section()`'s own anchor-drift check instead.

Evidence recorded by the reviewer:

- `guard()` is `section(RUN_TEAM, "**Guard: accumulate per PR", "One named member per ticket")`, and `section()` returns `source.slice(at, end)` where `at = source.indexOf(startAnchor)`. By construction `slice[0:9]` is always literally `**Guard: ` whenever the anchor is found at all.
- Deleting the matched span necessarily destroys the `startAnchor` too, so the run reds via a *different* failure (`run-team phase 2 dispatch rule: '**Guard: accumulate per PR' moved — update this test`), never via this assertion's own message.
- Second probe: rewriting the lead to a non-imperative phrasing that avoids every word this assertion targets (`**Guard: accumulate per PR…**` → `**Suggested check: accumulate per PR…**`) again redded only via the two anchor-drift messages (`tests 4 fail 2`). The "no longer stated as an imperative" message appeared in none of ~40 runs, including the one designed to trigger only it.

Candidate remedies (this is the undecided part): drop the assertion as adding no coverage beyond the anchor-drift check, **or** slice a window starting *after* the anchor so the imperative check runs on text that is not itself the anchor.

## Finding 2 — the `compute-spend.mjs` negative is case-sensitive

**Location:** `skills/fleet/scripts/implementer-model-tier.test.mjs:237-241`

[...]

Issue #516, opened 2026-08-16
fleet scripts: 20 file:line citations remain after #129's four — sweep them all, or leave the correct ones?


feigi:
Deferred from PR #515 review. **Filed as a record — the remedy the finding proposed was rejected by ruling; its measurements were kept.**

`dimension: correctness` (specialist finding at `skills/fleet/scripts/inflight.sh:262`), plus a second finding at `dimension: simplify` (`skills/fleet/scripts/candidates.test.mjs:500`) that rests on the same premise.

## What the finding claimed

> 12 of the diff's 21 hunks are `file:line` → prose citation rewrites unrelated to #121, and the sweep is left half-done inside the very files it edits — 5 such citations still stand in three files this PR touched (20 repo-wide).

Its proposed remedy was to drop the 12 citation hunks from PR #515 and land them as their own ticket. The `simplify` twin proposed reverting `candidates.test.mjs` and `implementer-model-tier.test.mjs` out of the diff entirely.

## Ruling: remedy REJECTED, counts KEPT

**The premise is false.** PR #515 closes **two** tickets — its body carries `Closes #121` *and* `Closes #129`, and **#129 is the citation ticket** ("fleet scripts: four cross-file line citations in comments point at the wrong line", whose suggested fix reads *"Re-point the four. Consider dropping the line number and citing the file plus the construct"*). The 12 hunks called "unrelated to #121" are #129's entire deliverable.

Neither refuter could see this: the diff and the review snapshot do not carry the PR body, so both verified the hunk counts — which are **all correct** — against the wrong scope. Applying the remedy would have deleted #129's deliverable while `Closes #129` still fired on merge, closing a ticket that was never fixed.

The residual ("finish the sweep in the touched files") does not survive either. The implementer's stated policy was to fix only citations that **measurement showed wrong** and leave ones correct at this tree, which matches #129's actual ask — dropping the line number is offered there as an option, not a mandate. `inflight.sh:377`'s `release-ticket.sh:73` was explicitly checked and left because it is correct.

## What is worth keeping: the counts

Re-measured at `a237cb0` on `fix/121-zero-padded-issue`, all reproduce:

```
$ grep -rnE '\.(sh|mjs|md):[0-9]+' skills/fleet/scripts/*.mjs skills/fleet/scripts/*.sh | wc -l
20
```

Per file, under `skills/fleet/`:

| file | sites |
|---|---|
| `review-pr-reads.test.mjs` | 4 |
[...]

PR #549, opened 2026-08-17
docs(fleet): drop the rotted line count, and pin the reason that replaced it

### LINKED ISSUE #530: implementer-model-tier.test.mjs:30 says the phase-2 section is ~90 lines; it is 238
Deferred from PR #523 review. Dimension: `comments`. Band: `suggestion` — 0 refuters by workflow policy; this fix-applier settled it on **scope**, not by refuting it.

**Location:** `skills/fleet/scripts/implementer-model-tier.test.mjs:30-32` — the slicing-rationale comment above `section()`.

## Finding (verbatim)

> Comment says 'the phase-2 section is ~90 lines' as the reason paragraph-level (not phase-level) slicing is required.

The comment reads:

```js
// Slice by named anchors and fail loudly when one moves; slice SIZE is what
// does the work. One slice per paragraph, never per phase: the phase-2 section
// is ~90 lines and its member prompt restates enough of this vocabulary to
// satisfy every pin below with both rules deleted.
```

## Measured against the pushed tree (`d35a86f`)

```
$ git show d35a86f:skills/fleet/skills/run-team/SKILL.md | grep -n '^## Phase [23]'
216:## Phase 2 — dispatch implementers
454:## Phase 3 — event loop
```

- `## Phase 2` as a heading block: **216-454 = 238 lines**, not ~90.
- The region this file's `section()` calls actually slice out of phase 2 — `dispatch` (218-287) plus `guard` (287-340) — is **122 lines**.
- `step4` (85-147) is in phase 0, not phase 2.

Neither figure is ~90 under any grouping, and `~` does not stretch to 2.6x.

**Not introduced by this PR, and already wrong before it.** At the merge base `512f7c7` phase 2 spanned 196-400 = **204 lines**, so the figure was already off by ~2.3x; PR #523 grew the section by 34 lines and made a stale number staler. The comment itself is untouched by the diff:

```
$ git diff 512f7c7...9e8ee3d -- skills/fleet/scripts/implementer-model-tier.test.mjs \
    | grep -E '^[+-].*90 lines'
(no output)
```

Outside PR #523's subject (the tier-guard revert and the phase-0 staleness probe). The claim is not load-bearing — no assertion depends on the number — but it is the stated justification for the file's whole slicing strategy, so a reader checking that reasoning finds it does not hold.

## Why `ready-for-agent`

The remedy is a wording fix with both numbers already measured above: state the real figure for the region actually meant, e.g. "the phase-2 section is ~240 lines" or "the dispatch+guard region these anchors slice is ~120 lines". The argument the sentence makes — that a phase-level slice is too coarse — is *strengthened* by the true numbers, so nothing about the strategy is open. Per phase 0's `decided?` test, comment wording is a decided item.

Whoever picks this up: re-derive both line numbers from `origin/main` at that time rather than trusting the ones above — this is a file whose line numbers move, which is why the figure rotted in the first place. Prefer a figure that does not need re-measuring on every edit if one reads naturally.

Deferred from PR #523 review.

feigi:
`implementer-model-tier.test.mjs`'s slicing rationale gave "the phase-2 section is ~90 lines" as the reason the slices are paragraph-tight rather than phase-level.

## Re-derived against `origin/main` (368dc67), not the ticket

```
phase 2 heading block      L216-L454  = 238 lines
region section() slices    L218-L340  = 122 lines   (dispatch 70 + guard 54)
```

The number was **right when written** — phase 2 was 86 lines at `6fc15a2` — and rotted at `9e8ee3d`, the tier-guard revert, which grew the section to 238. Same figures the ticket measured; nothing moved them since.

## The digit is gone, not bumped

Bumping `~90` to `~240` is the obvious fix and the wrong one — it rots on the next edit to a file the fleet rewrites most runs. #317 (`93de7b6`, `02a914b`) and `307819a` both settled on dropping the number and anchoring on a construct, and `307819a` applied that to two citations in **this very file**. Followed here. #530 asked for a figure that does not need re-measuring; the house answer is no figure.

## The reason was false, and is not "strengthened by the true numbers"

#530 supposed the sentence's argument survives with corrected figures. It does not — the mechanism it named never held.

The claim: phase 2's member prompt "restates enough of this vocabulary to satisfy every pin below with both rules deleted." Measured — excise the dispatch and guard paragraphs from a phase-wide slice and the remainder holds **zero** occurrences of `tier`, `sonnet`, `model` or `class=`; **0 of 9** phase-2 pins stay green. Same at `6fc15a2`, where the comment was written: zero. Only the number was ever right.

Replaced with the case that **does** hold, measured on phase 0: delete step 4's `**correction-ticket discipline**` sentence and the pin on that phrase is still satisfied by step 6's "which tickets carry the correction-ticket discipline" — **1 of 6** step-4 pins survives a phase-wide slice. Plus the structural half: `/^\*\*Guard: /` pins the slice's opening characters, so it is expressible only for a slice that starts at the guard paragraph.

## Pinned, because nothing pinned the old one

Nothing in the repo reads this file — the comment rotted twice unnoticed. The new rationale quotes SKILL.md prose, so it gets a pin on step 6's copy of the phrase.

Mutation-checked: reds when step 6 loses the phrase, reds when the `6.` anchor moves, green on an unrelated reword of step 6. Suite 773/773 green.

## Class sweep

Every length/distance claim measured from a moving prose file, under `skills/fleet/scripts/`:

| site | claim | measured | action |
|---|---|---|---|
| `implementer-model-tier.test.mjs:31` | `~90 lines` | 238 / 122 | **fixed** |
| `implementer-model-tier.test.mjs` phase-0 comment | `L224, L235, L303, L304, L314` | all five exact | left — house rule is convert only what measures wrong (`307819a`) |
| `reaping-prose.test.mjs:33,133` | `545 lines below` | exact (header ends `:25`, comment at `:570`) | left — accurate, and `release-ticket.sh` is not my site |
| `review-pr-reads.test.mjs:131` | `900-line file` | hypothetical illustration | not in class |
| `board.mjs:245` | `720 lines an hour` | derived from the 15s interval | not in class |
| `diff-stats.mjs:3` | `8-line docs edit` | illustrative | not in class |

Closes #530


feigi:
## Review findings adjudicated — 4 applied, 4 deferred (`57cce3b`)

Suite: `node --test skills/fleet/scripts/*.test.mjs` → **773/773 pass, 0 fail** (baseline before the fix was also 773/773; no tests added, pins widened).

### Applied

| # | Finding | Band | What changed |
|---|---|---|---|
| `correctness:129` | the new step-6 pin is reflow- and bold-brittle | important, 2/2 survived | pin now `/which tickets carry the\s+\*{0,2}correction-ticket\*{0,2}\s+discipline/` |
| `tests:155` + `tests:97` | both negative pins are direction-dependent | important, 2/2 each | both scans made direction-agnostic, treated as one change |
| `tests:263` (comment half) | the assertion it cites can never fail | suggestion, refuter **could not refute** | the false sentence this PR added has been removed |
| `tests:157` | the revert-note exemption is case-sensitive | suggestion, claim survived / **remedy rejected** | dated-revert pin reordered ahead of the scan; `/i` NOT applied |

**Step-6 pin, five states, running the real test file:**

| state | shipped | ruled form | shipped here |
|---|---|---|---|
| baseline | GREEN | GREEN | GREEN |
| pure reflow (wording byte-identical) | RED | GREEN | GREEN |
| `**correction-ticket discipline**` | RED | GREEN | GREEN |
| `**correction-ticket** discipline` | RED | **RED** | GREEN |
| phrase DROPPED | RED | RED | **RED** |

The ruled form carried `\*{0,2}` only before `correction-ticket` and still redded on the fourth row — the same false-alarm class, one shape left. Both sides carry it now, matching the step-4 sibling. Discrimination preserved.

**Negative pins, both directions:**

| mutation | shipped | fixed |
|---|---|---|
| clean tree | GREEN | GREEN |
| phase 0, forward order | RED | RED |
| phase 0, REVERSED order | **GREEN** | RED |
| phase 2, forward order | RED | RED |
| phase 2, REVERSED order | **GREEN** | RED |

The past-tense revert note stays exempt, so the clean tree stays green.

**On `tests:157`** — the claim reproduced, but its `suggested_fix` (`/i` on the exemption) was measured and **rejected**: it does not clear the red (`/REVERTED on 2026-08-16/` is case-sensitive too, so the downcase just moves which assertion fires) and it *widens* a real blind spot. Applied instead: the dated-revert pin now runs before the scan whose exemption names it, so a reworded note is diagnosed as a reworded note. Verified — downcase → accurate message; both direction mutations → still caught; clean → green.

**Also corrected, not from the findings:** the phase-0 comment's `720/720` was wrong. `d35a86f` introduced the claim and reports **725/725** twice in its own message; `d35a86f^` measures `tests 725 / pass 725 / fail 0` today. Two rotted numbers in one comment block, in a PR whose whole subject is a rotted number in that block.

### Deferred

| Finding | Where it went | Why |
|---|---|---|
| `tests:263` (the assertion itself) | **#529** (comment) | already filed as its Finding 1; commented rather than duplicated |
| `correctness:93` — five `L###` citations | **#516** (comment) | all five measure **exact**; `307819a`'s rule converts only what measures wrong |
| exemption swallows a restoration inside the revert note | **#553** (new, `needs-triage`) | found while applying `:157`; pre-existing, out of #530's scope |
| `comments:20` — historical `720/720` / INVERTED claims | refuted, no ticket | both reproduce from history (`git log -S`); the one real defect (the count) is fixed above |

`comments:166` was refuted by both refuters and is not filed, per ruling.

**#553 in one line:** a restoration written *inside* the exempted revert note — `was REVERTED on 2026-08-16 and is RESTORED on 2026-08-18, so routine members dispatch at \`sonnet\` again` — leaves the suite at **5/5 green**. The exemption drops any hit containing `was REVERTED`, restoration and all.

Note for whoever reads the history: the previous commit message (`e5784d3`) states the same false `/^\*\*Guard: /` claim the comment did. The comment is fixed; the message cannot be.

Issue #241, opened 2026-08-10
docs/fleet-rule-manifest.txt: a hand-synced grep checklist nothing reads, drifted twice — pin it or delete it as a spent restructure gate


feigi:
Deferred from PR #228 review (issue #173).

**dimension:** `code-simplifier` (severity `suggestion`, unverified by the workflow; refuted by one refuter dispatched from the fix-applier, verdict **refuted=false** — the finding holds)

## The finding

`docs/fleet-rule-manifest.txt` is hand-synced against the files it indexes, and nothing reads it. Line 5 carried `` `--label ready-for-agent` mandatory `` — the wrong flag — and PR #228 had to hand-edit it in the same commit as `skills/fleet/skills/run-team/SKILL.md:60`.

## Verified

Measured at `e840bc5`:

- **Nothing consumes it.** `git grep -n "fleet-rule-manifest\|rule-manifest\|ruleManifest" -- '*.mjs' '*.js' '*.ts' '*.sh' '*.json' '*.yml' '*.yaml'` → zero matches. The only reference anywhere is prose at `docs/specs/2026-07-23-fleet-plugin-design.md:482`, a past-tense claim ("verified against `docs/fleet-rule-manifest.txt`, all 33 patterns resolve"), not a check.
- **CI does not cover it.** `.github/workflows/ci.yml` runs `node --check` on `*.mjs`, `jq empty` on `*.json`, `shellcheck` on `*.sh`, a python syntax pass, and `node --test`. Nothing reads a `.txt`.
- **Two mutations both survive.** Reverting line 5 to the wrong flag, and deleting the file outright, each leave the suite unchanged from baseline.
- **It has drifted and been hand-patched twice.** `3db006c` ("docs(fleet): unstale the manifest entry and both design statuses") repaired an orphaned line 6 after `**Light row only.**` was deleted. `e840bc5` is the second instance.
- **PR #228's new test does not cover it.** `candidates.test.mjs` reads `skills/fleet/skills/run-team/SKILL.md` only.

## Why `needs-triage` and not `ready-for-agent`

The defect is verified; **the approach is open**, and the two candidate remedies point in opposite directions:

1. **Pin it** — add a test asserting each of the 33 grep patterns still resolves somewhere under `skills/fleet/`. This is what the manifest was written to enforce, and automating it makes the drift impossible.
[...]

PR #550, opened 2026-08-17
docs(fleet): re-derive the stale figures in the fleet plugin design spec, and repoint two orphaned manifest pins

### LINKED ISSUE #48: docs/specs/2026-07-23-fleet-plugin-design.md: three stale assertions (:208 section list, :225 caller count, :428 'all 33 patterns resolve')

feigi:
Every figure, path and sha below was re-derived from `origin/main` `368dc67`. Nothing is cited from the ticket.

## Item 3 — the manifest. Current miss count is **2**

The thread reports 1, 3 and 4 on different dates. Re-running the ticket's own loop today at three scopes shows why:

```
scope                          total  misses
run-team/SKILL.md only            33       4
skills/fleet/                     33       2
whole tracked tree                33       2
```

Both readers measured correctly; they picked different scopes. Neither of the two current misses is a lost rule, so both are repointed to surviving text — the `3db006c` precedent, not a restoration:

| manifest line | rotted by | repointed to |
|---|---|---|
| `Say the specialist count` | `8b46fdb` deleted `**Say the specialist count in the prompt.**` | `apply the heuristic yourself` — unique in the tree, and sits directly under **Authorize the fan-out explicitly**, whose own pin still resolves there |
| `Per-member scratchpad subdirectory` | `15d50c3` reworded the heading while sharpening the rule ("Per-member is the wrong axis") | `Scratchpad paths need two levels` — resolves in SKILL.md and `references/isolation.md`, and is already pinned in both by `snapshot-runner-audience-prose.test.mjs` |

After both, 33 of 33 resolve — swept against **`origin/main`'s** tree, not this branch's, so the test files added here cannot be what satisfies a pattern.

**The count sentence drops the literal `33`.** The digit is what rotted the last three times and re-rots the moment the manifest grows; "every pattern" needs no re-measuring. It gains the scope and the ref instead, following this file's own convention for pinned measurements.

## Script-surface table — the whole `In`/`Out` class, not the two rows the ticket names

Eight cells were wrong. Three rows were already right and are untouched (`inflight.sh`, `no-undo-audit.sh`, `prove-merge.sh`), as is `candidates.mjs`'s `Out`.

- **Fields that do not exist** — `rid`, `headSha`, `branchHead` (ci-state), `log` (verify-sha), `commits` (worktree-audit), and the `set` subcommand (ledger), which always exits 2.
- **Fields that exist but were omitted** — including the `row` and `check` subcommands, `check` being the entire subject of its own row's `Non-zero when` cell.
- **One wrong type** — worktree-audit's `dirty` is a count and `dirtyFiles` the array, so the old `dirty[]` sent a caller to `.dirty[0]`.

## Reference-table cells — same class, two more phantoms

- `correction-tickets.md`'s cell enumerated three of the file's sections; the file has **five**. Completing the list rots on the sixth, so the cell now states the file's subject, which a new section cannot falsify.
- `member-lifecycle.md` named "the four-cell naming probe". `git log -S` finds it only in the commit that wrote the row — phantom from birth.
- `ci-and-staleness.md` named "behind-count decay (0→2→7→10)". That sequence appears nowhere at `origin/main` but in the cell itself.
- "both callers point at it" is one document, twice: only `run-team` SKILL.md names the reference; `review-and-fix.md` restates the rule inline and never names the file.

`isolation.md` and `reaping.md` were checked item by item and every one names real content.

## Already fixed before this branch, so untouched

The migration-debt row and the `+2` offset note both name their own staleness now ("**as measured at `a03258e^`**", "The offsets are **not uniform** and were never **+2**"). No change needed.

## Tests

Two pins, both derived from a real run rather than a hand-written list — the failure mode the table itself demonstrates. `worktree-audit.test.mjs` follows the idiom `no-undo-audit.test.mjs` established for this same table and adds the other direction, an invented-field check, which is the half that catches `commits`. `ledger.test.mjs` compares the `In` cell against the script's own usage alternation as a set, so a phantom subcommand and a missing one both fail. Both were red before the fix and mutation-checked after.

`./agent-test` → **774 pass, 0 fail.**

## A wrong claim this branch minted and then caught

The spec first cited the precedent repair as `975c3d0`, quoted from the ticket. That sha is **not reachable from `origin/main`** — it is the pre-rebase form of `3db006c`, same date, same subject, byte-identical hunk. `git show` succeeds on it locally and proves nothing; only `git merge-base --is-ancestor` catches it. Fixed in `703c384`.

## Left for triage — not filed as a new issue

**#241 already owns it**, `needs-triage`: *"a hand-synced grep checklist nothing reads, drifted twice — pin it or delete it"*. Commented there rather than filing a duplicate: https://github.com/feigi/claude-config/issues/241#issuecomment-5318340623

Deliberately **not** done here: no checker, no operational definition of "resolve", no target-file column. Picking one inside a correction ticket settles a design question by side effect. `git grep -ln 'fleet-rule-manifest'` still returns this spec alone.

Note: `ledger.mjs check` returned `verdict: tracker-hit` naming only #48 — its query had degraded to `operational definition manifest`. The independent `gh issue list --search` is what surfaced #241.

Closes #48


feigi:
Review findings applied in `1cc1a7c`.

**Applied (3 findings, 2 commits):**

- `800474f` — the `critical` + first `important` were one defect: `worktree-audit.test.mjs`'s missing-field direction matched each payload key against the **whole** table row while its invented-field counterpart three lines below already narrowed to the Out cell. Live consequence: `worktree` was satisfied by the Script cell's `worktree-audit.sh` and `dirty` by the `Non-zero when` prose, so either could be dropped from the type signature with the suite green — and `dirty` is one of the two fields #48 says the row got wrong. Same defect swept in `no-undo-audit.test.mjs`, the sibling the idiom was copied from, where `stash` and `worktree` were masked today (measured before the fix).
- `1cc1a7c` — second `important`: the `ledger.test.mjs` pin derived the "real" subcommand set from `ledger.mjs`'s hand-written usage string, so pin and pinned shared a source. Now read off the dispatch, with **both** hand-typed copies — the spec's In cell and the usage line — measured against it.

**Mutation-verified, both directions, per pin** (spec and `ledger.mjs` restored byte-clean after every mutation, proven by an empty `git diff`):

| Pin | baseline | masked field dropped | control | irrelevant reword |
|---|---|---|---|---|
| `worktree-audit.test.mjs` | green | `dirty` red, `worktree` red | — | prose cell reword: green |
| `no-undo-audit.test.mjs` | green | `stash` red, `worktree` red | `clean` red | prose cell reword: green |
| `ledger.test.mjs` | green | dispatch branch added w/o usage edit: red | `check` from usage line red; `check` from In cell red | prose cell reword: green |

**Deferred (1):** #552 — the `suggestion` on the `| Reference | Carries |` row at spec `:226`. In scope, so it got one refuter biased to refuse, and it was **refuted**: under-listing is that table's uniform convention (4 of 5 rows under-report; `reaping.md` omits 3 sections), the table is not the spec's map of where rules live (`:541` delegates that to `docs/fleet-rule-manifest.txt`), and this PR's rewrite *reduced* the row's gap from two sections to one. Filed `needs-triage` for the record, carrying the open design question rather than a fix.

`node --test skills/fleet/scripts/*.test.mjs` from the worktree: **tests 774 / pass 774 / fail 0** (baseline before these changes was also 774/774).

Issue #118, opened 2026-07-31
Two stale dimension-count comments (review-pr.js:143, diff-stats.mjs:105), and selectDimensions is untestable in place


feigi:
Deferred from PR #107 review. Two wrong comments, one shared root cause.

## The two wrong comments

Both re-derived by lifting the real `computeStats` and `selectDimensions` out of the head blob into a harness:

```
computeStats([])  = {... docsOnly:false, hasSrc:false, hasTests:false, profile:"empty"}
WITH the guard    -> 6 specialists
WITHOUT the guard -> 2 specialists [correctness,comments]   (4 dropped, not 3)
```

**1. `workflows/review-pr.js:143` — "drop three dimensions" is four.**

```
// "nothing to review" would silently drop three dimensions on production code.
```
The comment sits on the empty-profile guard at `:145`, not on the `hasSrc === false` branch. Strip that guard and an empty profile falls through to *both* `hasTests === false` and `hasSrc === false`, dropping tests + types + silent-failure + simplify — **four**.

**2. `skills/fleet/scripts/diff-stats.mjs:105` — the harm it names cannot happen.**

```
// Fail closed. A broken query must not read as "empty diff" — that would
// silently trim a real production PR down to two specialists.
```
It would not trim to two. `computeStats([])` yields `profile: "empty"`, and `review-pr.js:145` widens an empty profile to **all six**. The fail-closed behaviour is still right; the stated consequence is wrong, and it understates by describing a trim where the guard already produces a widen.

Neither is caused by PR #107. Both were *cleared as correct* in that PR’s body without being read in context, which is how they survived — see the correction in the PR body.

## The root cause: nothing can pin these numbers

`selectDimensions` cannot be imported. `workflows/review-pr.js` runs top-level `args` access and `agent()` calls, so requiring it executes the workflow. Consequence: every count claim about it has to be hand-derived, and **both** specialists on the PR #107 fan-out independently had to lift the function into a `new Function(...)` harness to check a number. Hand-derivation is exactly what produced the two wrong comments above.

**Approach open** — hence `needs-triage`:

1. Extract `selectDimensions` (and possibly `DEFAULT_DIMENSIONS`) beside `diff-stats.mjs` as an importable module. One `node --test` assertion over the input matrix would then pin all six reachable sizes — `{2,3,5,6}`, 4 unreachable — and both comments above would have been caught by it.
2. Or leave it in place and just correct the two comments, accepting that the next count claim is unpinned too.

[...]

PR #554, opened 2026-08-17
docs(fleet): record PR #536's tier outcome, and flag it as an accidental sample

feigi:
One row appended to `docs/metrics/tier-outcomes.tsv` for PR #536, ruled during the 2026-08-17 `/fleet:run-team` run. The controller ruled that review but never wrote the row.

```
2026-08-17	536	140	routine	sonnet	no	yes	…
```

## Why `closed_own_ticket=no`

#140 asked for a check that a review snapshot exists. The guard that shipped **cannot fire**: the `SNAPSHOT_NONEMPTY` probe ran *after* the `node_modules` symlink, so a failed `git archive | tar -x` still reported `pathVerified: true` in any repo that has `node_modules` — i.e. exactly the repos the surrounding #142 paragraph exists for. Two dimensions found it independently, 2-0 each, both reproducing against a real repo with an unborn HEAD.

## Why `minted_false_claim=yes`

The new test comment claimed `additionalProperties: false` + `required` "already caught the sibling defect at #113/#118, same schema". #113 was fixed by `resolveDimensions`' own hand-written `REQUIRED` array and #118 was stale test-count comments — neither is a JSON schema, and the two are not one mechanism.

## The tier field is not evidence — read the note

The PR was dispatched at `sonnet` **in error**. That controller session loaded its `SKILL.md` from a checkout 57 commits stale, which still carried the pre-revert phase-2 rule; the revert to session tier had merged the previous day. Under the rule in force this ticket would have run at the session tier.

So this row is an **accidental sample, not the deliberate control** the file's header asks for. The header's warning that `class` and `tier` are perfectly confounded still holds, and this row does not relieve it.

## What it changes for the guard

This is the **second** `closed_own_ticket=no` row (after #452), and the two fall on **distinct `run_date`s**. Both halves of the guard's floor are now satisfiable from the accumulated file, so the floor and trigger are worth re-reading — that is a maintainer call, not something this PR takes.

## Verification

- `node --test skills/fleet/scripts/*.test.mjs` → **782 pass, 0 fail** in a worktree cut from `b5afbf8`.
- All 9 data rows carry 8 tab-separated columns; trailing newline intact; diff is `1 file changed, 1 insertion(+)`.
- Nothing in `skills/` or `workflows/` reads this file — only `run-team/SKILL.md` cites it in prose.

No issue to close: this is the guard's own bookkeeping, not a ticket.


feigi:
## Review findings applied — `3cd40c5`

Suite before commit: `node --test skills/fleet/scripts/*.test.mjs` → **782 tests, 782 pass, 0 fail** (not a zero-test run). No `--no-verify`.

### Applied (1 edit, `skills/fleet/skills/run-team/SKILL.md:234-248`)

`survived0` (critical, `comments`) and `survived1` (important, `correctness`) are one defect reported by two dimensions, with conflicting suggested fixes — freeze the old numbers behind a date anchor vs. replace them with current ones. Applied as one coherent edit doing both: current figures **plus** an as-of anchor, so the next doc-only append leaves the paragraph stale-but-true rather than false.

Every number re-derived from `docs/metrics/tier-outcomes.tsv` rather than copied from the findings:

| claim | was | measured now |
|---|---|---|
| `routine`/`sonnet` rows | 7 | **8** |
| …of which `closed_own_ticket=no` | 2 | **3** (#452, #466, #536) |
| `routine`/`opus` rows / failures | 1 / 0 | **1 / 0** (#480) — unchanged, still n=1 |
| "both `no` rows predate the revert" | true | **false** — #536 (2026-08-17) postdates the 2026-08-16 revert |
| "most recent routine PR (#521) closed its own ticket" | true | **false** — most recent is #536, and it did not |
| all rows `minted_false_claim=yes` | true | **true** (9/9) — kept |

#536's TIER CAVEAT is carried into the paragraph, not dropped: dispatched at `sonnet` in error by a controller session whose SKILL.md came from a checkout 57 commits stale, so it is an accidental sample rather than a designed control and does not relieve the class/tier confound.

**`:226-227` deliberately left byte-identical.** `survived1` also asked for that parenthetical to become "9 PRs / 4 run_dates". Measured: rows dated `<= 2026-08-16` are **8 routine PRs across 3 distinct run_dates with 2 failures** — exactly what it already says. That sentence is a past-tense record of the guard firing, not a live count; applying the suggested fix would have shipped a new false claim.

### Deferred (1)

- `unverified0` (suggestion, `correctness`, `docs/metrics/tier-outcomes.tsv:8`) → **#557**, `needs-triage`. Dispatched a refuter against it (the `suggestion` band gets 0 refuters by policy, so it arrived unchecked); it came back **refuted** by measurement — the header sentence's subject is the dispatch *rule*, which is unchanged; the header's one row-census claim ("exactly one row off the old binding — PR #480") is still literally true with row 33 present; deleting the whole header block leaves the suite at 782/782; and both prior row-appending commits added zero header lines. Full refutation in the issue body.

### Ruled, nothing filed (2)

- `unverified1` (suggestion, `comments`) — states its own answer is "None needed for this PR"; it asserts no defect, only that the row's `Found by 2 dimensions, 2-0 each` and `TIER CAVEAT` clauses record session provenance a tree-only review cannot check. Correct, and not actionable. No issue.
- `unverified2` (suggestion, `comments`) — POSITIVE finding: the row's core technical claim about the `SNAPSHOT_NONEMPTY` probe ordering is corroborated by the tree. Nothing to apply, nothing to file.

### Note for a future reviewer

No test pins these counts, and I did not add one. A pin coupling the paragraph to the TSV would red on every append, which contradicts the guard's own **"That append is the whole duty"** at `SKILL.md:289`. The as-of anchor is what makes a bare append safe instead.

Issue #17, opened 2026-07-30
prove-merge: is_ancestor() collapses git error exit 128 to "false", which on the rebase path makes leg 1 pass


feigi:
Deferred from PR #12 review.

`skills/fleet/scripts/prove-merge.sh:23`:

```sh
is_ancestor() { if git merge-base --is-ancestor "$1" "$2" 2>/dev/null; then echo true; else echo false; fi; }
```

`git merge-base --is-ancestor` exits `1` for "not an ancestor" and `128` for a real error (bad object, unreadable ref, truncated history). This helper maps both to `false`, and `2>/dev/null` throws away the diagnostic that would tell them apart.

The consequence is asymmetric:

- no-rebase path — `head_current=false` makes the gate **fail closed** (`proved=false`). Safe.
- rebase path — `pre_anc=false` makes leg 1 **pass** (`prove-merge.sh:81`). Fail **open**, on the leg the script exists for.

Same asymmetry applies to a *correct* exit 1 from a truncated history: in a shallow clone `merge-base --is-ancestor` can answer "not an ancestor" for a genuine ancestor, which again reads as leg 1 passing.

**Pre-existing** — 9524cba had the same inline if/else on both ancestor checks; #12 only factored it into a helper.

## Not demonstrated end-to-end

Reachability is currently blocked in the common case: the `git cat-file -e` loop at `prove-merge.sh:38-40` validates `$pre`, `$post` and `$merge`, and `git rev-parse --verify` validates `$base`, so the missing-object route to 128 dies at exit 2 first. Confirmed against a `--depth 2` clone:

```
prove-merge: 321e5f0d... is not a commit in this repository   (exit 2)
```

`$first` is the one input never passed through that guard (it is derived from the merge at `prove-merge.sh:63`), but it is only consumed on the no-rebase path, where the collapse fails closed.

So this is a latent fail-open with no demonstrated trigger, filed rather than fixed or dropped.

## Acceptance

- `is_ancestor()` distinguishes exit 1 from any other non-zero: 1 -> `false`, anything else -> `die` with the captured stderr.
- A test pins that a `merge-base` error is an error (exit 2), not a silently passed leg.
- Consider refusing to run in a shallow repository at all (`git rev-parse --is-shallow-repository`), since no ancestor answer from one is trustworthy.

feigi:
Fixed in PR #12 rather than deferred — it was one line once I was in the file.

```sh
is_ancestor() {
[...]

PR #555, opened 2026-08-17
fix(fleet): verify-sha reports a failed merge-base as unanswerable, not "not reachable"

### LINKED ISSUE #266: verify-sha.sh:31 — a merge-base probe that exits 128 is reported as `reachable:false` at exit 1, the same collapse #17 fixed in prove-merge.sh
**dimension:** silent-failure · **severity:** important · Deferred from PR #251 review (ticket #114)

`skills/fleet/scripts/verify-sha.sh:31`

```sh
if git merge-base --is-ancestor "$sha" "origin/$branch" 2>/dev/null; then
  reachable=true; rc=0
else
  reachable=false; rc=1
fi
```

Every non-zero status collapses into `reachable:false`. When `git merge-base --is-ancestor` exits **128** — it never answered — the script reports the definite negative verdict `reachable: false` at exit 1, with git's stderr discarded by the `2>/dev/null`. "The repository could not answer" and "the member's SHA is not on the branch" become indistinguishable — the one distinction this script exists to make.

## Reproduction — real corruption, no fault injection

Bare origin + clone, `main = c1<-c2<-c3`, all pushed. `chmod 000` on the intermediate commit object:

```
$ git merge-base --is-ancestor $A main
error: unable to open loose object 68d1cc3...: Permission denied
error: could not parse commit                              rc=128

$ sh skills/fleet/scripts/verify-sha.sh main $A
    origin/main tip = 7db2c1d...
    aaff678... is NOT reachable on origin/main
{"branch":"main","sha":"aaff678...","reachable":false,"tip":"7db2c1d..."}
VERIFY-SHA rc=1
```

The fetch and both preconditions (`cat-file -e` on the sha, `rev-parse origin/main`) passed — only the walk failed, so none of the four exit-2 guards catch it. Reproduced independently a second time on a separate probe repo.

## Why it matters operationally

The `/fleet:run-team` controller relies on this script's verdict for every SHA it enqueues. A false `reachable:false` is a **stall**, not a data loss: the controller believes a member's pushed work is not on the branch and re-drives or abandons it.

## This exact defect was already found and fixed one file over

**#17** (closed 2026-07-30) was *"prove-merge: `is_ancestor()` collapses git error exit 128 to `false`"* — the same collapse, in the sibling script. `prove-merge.sh:39-44` now carries the guard, with a comment naming the rule:

```sh
# Exit 1 is the answer "no"; anything else is git failing to answer, and a proof
# must never read a failure as a leg it likes.
is_ancestor() {
  if git merge-base --is-ancestor "$1" "$2"; then echo true; else
    [ $? -eq 1 ] || die "git merge-base --is-ancestor $1 $2 failed — cannot prove anything"
    echo false
  fi
}
```

`verify-sha.sh:31` was never given the same treatment. The fix for #17 was applied to the site that motivated it, not to the class — the sweep stopped at one caller.

## Fix — settled

Reuse the sibling guard verbatim (`prove-merge.sh:39-43` already does exactly this):

```sh
if git merge-base --is-ancestor "$sha" "origin/$branch"; then
  reachable=true; rc=0
else
  [ $? -eq 1 ] || die "git merge-base --is-ancestor failed — cannot tell reachable from unanswerable"
  reachable=false; rc=1
fi
```

Drop the `2>/dev/null` so the cause is visible, and add "a probe that could not run" to the script's exit-2 list in `docs/specs/2026-07-23-fleet-plugin-design.md`.

Out of scope for #114 (documentation audit; exit-code changes are explicitly excluded). PR #251 corrects the spec row to state that exit 1 currently carries both cases. **No test covers `verify-sha.sh`.**

feigi:
`git merge-base --is-ancestor` exits 128 when it cannot walk the history — a
corrupt or unreadable object — and `verify-sha.sh` collapsed every non-zero
status into the definite verdict `reachable:false` at exit 1, with git's
diagnosis discarded by `2>/dev/null`. "The repository could not answer" and
"the member's SHA is not on the branch" became indistinguishable, which is the
one distinction this script exists to make. `/fleet:run-team` reads that verdict
for every SHA it enqueues, so a false negative stalls a member whose work really
did land.

This is the same collapse #17 fixed in the sibling `prove-merge.sh`; that sweep
stopped at the site that motivated it. The guard here is the sibling's, in its
words.

## The class, and what is left

Two production `--is-ancestor` call sites exist in the repo:

- `verify-sha.sh:31` — fixed here.
- `prove-merge.sh:41` — already carries the guard from #17. Verified, unchanged.

The file's four other exit-2 guards (arg count, fetch, `rev-parse`, `cat-file`)
all use `|| die`, which fails *closed* — any non-zero already becomes "could not
answer". They collapse nothing into a verdict, so they are deliberately left
alone. `run-team/SKILL.md` names the script but describes no exit codes, so
nothing there is made wrong by this change.

## Verification

Reproduced without fault injection, per the issue: `chmod 000` on an
intermediate loose object, on a SHA that genuinely **is** an ancestor. Before,
`reachable:false` at exit 1 — a false negative on landed work. After, exit 2, no
verdict printed, git's own `error:` lines reaching the operator.

Dropping `2>/dev/null` was measured before relying on it: git is silent and
exits exactly 1 on a plain negative, so the ordinary not-reachable path gains no
noise. A test asserts that directly.

## Tests

`verify-sha.sh` had no test at all. The new suite pins all three modes,
including the two the new guard could wrongly refuse — reachable (exit 0) and
genuinely not reachable (exit 1) — so exit 1 cannot be quietly promoted to exit
2. Both mutations of the guard are caught:

- inverting `-eq 1` → the genuine-negative test fails (over-firing).
- moving the `$?` read below `reachable=false` → the same test fails, since the
  assignment overwrites `$?`. The ordering is load-bearing and pinned.

787/787 green in `skills/fleet/scripts`; `shellcheck -x -S warning` clean.

Spec's `verify-sha.sh` row no longer documents the collapse as intended
behaviour — that case moves to the exit-2 list, in the wording the
`prove-merge.sh` row already uses.

Closes #266


feigi:
## Review findings adjudicated — 3 applied in `a965c56`, 7 deferred

Ten findings came back from the review (0 survived, 3 refuted, 7 unverified). All three `refuted` entries were 1-1 vote splits, so each was ruled individually rather than read as "nothing to do". The five in-scope `suggestion`s — which the review workflow budgets 0 refuters by policy — each got one refuter, biased to refuse and required to verify by running something.

### Applied in `a965c56` — `skills/fleet/scripts/verify-sha.test.mjs`

| dimension | change |
|---|---|
| correctness | quote the git shim's `exec "${REAL_GIT}" "$@"`. Under a git whose path holds a space the shim word-splits, `git fetch` dies first, and the test measures the wrong failure — and it fails loudly only by luck, since that fetch death is itself exit 2 with no JSON, so `code === 2` and `json === null` both still pass. Measured against a git symlinked into a spaced directory: unquoted 4/5, quoted 5/5. |
| simplify | the argc test dropped its bare-origin + clone + commit + push fixture — the script dies at `[ $# -eq 2 ]` thirteen lines before it runs any git. 231ms → 3ms. |
| simplify | the unknown-object test dropped its extra commit+push — a 40-zero sha dies at the `cat-file -e` guard before `merge-base` is reached, so the tip's position is unobservable to every assertion. |

Both trims are mutation-checked against the post-fix tree, since a test that asserts less also passes: deleting the argc guard reddens the argc test **and only it**; deleting `cat-file -e` reddens the unknown-object test **and only it**; each stays green under a change to a die string it does not pin; and the #266 case still kills its own collapse mutant.

`./agent-test skills/fleet/scripts`: **787 tests, 787 pass, 0 fail.**

### Deferred and filed

| # | what | why not here | label |
|---|---|---|---|
| #558 | `run-team/SKILL.md`'s "Verify every reported SHA" guard names no exit-2 branch | pre-existing (four exit-2 paths predate this PR, none documented either) and fail-safe (exit 2 emits no JSON, so a controller cannot read `reachable:true` and cannot enqueue). Three review dimensions reported it independently, which is why it is one ticket and not three. | `ready-for-agent` |
| #559 | `verify-sha.sh:19,22,28` swallow git's stderr behind `2>/dev/null` and assert a false cause list | different class from #266 — message accuracy, not verdict collapse. Both votes reproduced the mechanism; the refutation landed only on the finding's framing. | `ready-for-agent` |
| #560 | the suggestion band, checked — the 2 refuted ones, with the measurements | recorded so they are not re-filed and re-applied | `needs-triage` |

Also commented the measured `evil"branch` → unparseable-JSON-at-exit-0 reproduction on **#119**, which already tracks this script's unescaped splice, rather than opening a duplicate: https://github.com/feigi/claude-config/issues/119#issuecomment-5320039745

PR #556, opened 2026-08-17
fix(fleet): let a failed reachability probe stop prove-merge, not mislabel it

### LINKED ISSUE #267: prove-merge.sh:63 — a probe failure inside the command substitution is reported to the operator as "that merge did not land"
**dimension:** silent-failure · **severity:** suggestion · Deferred from PR #251 review (ticket #114)

`skills/fleet/scripts/prove-merge.sh:63`

When `is_ancestor` dies inside a command substitution, the substitution's exit status is **discarded by `[ ]`**, so `set -e` never fires and the outer `die` prints "NOT REACHABLE — that merge did not land" for a `merge-base` that never answered. A probe failure is reported to the operator as a **disproof**.

## Reproduction — against the real script, twice, independently

Two refuters dispatched on unrelated findings both drove it into existence while probing `prove-merge.sh` with a `BASE_REF` pointing at a **blob** (a real input: `$base` is only `git rev-parse --verify`'d, never commit-checked, so a blob sha passes the gate):

```
$ blob=$(git rev-parse HEAD:f)
$ BASE_REF=$blob sh prove-merge.sh ea29c88 ea29c88 79ba480
$ git merge-base --is-ancestor 79ba480514f4... 78981922613b...
error: object 78981922613b... is a blob, not a commit
fatal: Not a valid commit name 78981922613b...
prove-merge: git merge-base --is-ancestor 79ba480... 789819... failed — cannot prove anything
prove-merge: 79ba480... is not reachable from 789819... — that merge did not land
EXIT=2
```

Both lines print. The first is the truth (the probe could not answer); the second is a **disproof the script never established**, and it is the one phrased as a verdict.

Mechanism: `die` runs inside `$(...)`, so its `exit 2` kills only the **subshell**. The substitution yields the empty string, `[ "" = true ]` is false, and control falls through to the outer `die` at line 63.

Confirmed in isolation too:

```sh
set -eu
probe() { die "probe could not answer"; echo true; }
[ "$(probe)" = true ] || { echo "outer: NOT REACHABLE — that merge did not land" >&2; exit 2; }
```

prints **both** lines and exits 2 — `set -e` does not fire on the discarded substitution status.

## Why it is low severity

The inner cause is printed first, and the exit code is 2 either way, so the exit surface documented in the `prove-merge.sh` spec row stays correct. Only the second message is wrong — but it is the one that reads like a verdict.

## Suggested direction

Assign first, so `set -e` sees the failure — the pattern lines 66-74 already use for `pre_anc`/`post_anc`:

```sh
merge_anc=$(is_ancestor "$merge" "$base")
[ "$merge_anc" = true ] || die "...did not land"
```

## Status

Originally reported at severity `suggestion` (which the `review-pr` workflow budgets 0 refuters), but **since verified against the real script by two independent runs** with the reproduction above — no fault injection, no PATH shim.

Related: **#17** (closed) fixed `is_ancestor()` itself so it dies rather than answering `false` on a git error. That fix works — it is the inner `die` firing above. What #17 did not cover is that the inner `die` cannot terminate the script from inside a command substitution, so the caller keeps going and overwrites a correct diagnosis with a wrong one.

feigi:
## What

`prove-merge.sh` tested its merge-reachability probe inline:

```sh
[ "$(is_ancestor "$merge" "$base")" = true ] || die "… — that merge did not land"
```

`die` inside `$(...)` exits only the subshell, and `[ ]` discards that status, so `set -e` never fired. A `merge-base --is-ancestor` that **errored rather than answered** yielded the empty string, `[ "" = true ]` was false, and control fell through to the outer `die` — printing a disproof the script never established, as the second and more verdict-shaped of two messages.

Now the probe is assigned first, the way every other `is_ancestor` call in the file already does. An assignment is a simple command, so its status is the substitution's and `set -e` sees it.

Reproduced against the real script with no fault injection and no PATH shim, using the ticket's own input (`BASE_REF` naming a blob):

| | before | after |
|---|---|---|
| `prove-merge: … failed — cannot prove anything` | printed | printed |
| `prove-merge: … — that merge did not land` | **printed** | gone |
| exit | 2 | 2 |

## Scope — the class, not the line

Every command substitution in `prove-merge.sh` was enumerated. Line 64 was the **only** member whose status was discarded: the rest either assign first (`pre_full`, `post_full`, `pre_anc`, `post_anc`, `head_current`) or handle their own status (`parent_line=$(…) || die`). Verified in `/bin/sh`, `dash`, `zsh` and `bash` that a bare assignment does propagate the substitution's status under `set -e` while the `[ ]` word-expansion slot does not.

A static test now pins the **class** rather than the instance, so a future `[ "$(…)" ]` in this script fails the suite instead of shipping.

**Deliberately left:** `$base` is still only `git rev-parse --verify`'d, never commit-checked — which is how a blob reaches the probe. Out of scope, and intentionally so: the design spec's `prove-merge.sh` row already documents a non-commit base as landing in the "`merge-base --is-ancestor` that failed rather than answered" exit-2 bucket, and after this fix that bucket is honest — it exits 2 naming the probe failure. Adding the check would only change which message prints, would make that spec row stale, and would remove the ticket's own shim-free reproduction from the suite.

## What this change could wrongly refuse

`set -e` is now live on a line where it was not, so the fix for a false negative could introduce a false positive. Both accept paths are pinned:

- a **legitimate** `false` answer still reaches the outer `die` with `is not reachable from` and exit 2 (existing `ATTACK: a merge commit the caller fabricated` case)
- the **same fixture** as the new failure test, run with a base that can answer, still proves `true` at exit 0 — asserted in the new test, because a suite that only ever fed the probe failing input would not notice it aborting on a good one

The JSON payload and the whole exit surface (0 proved / 1 disproved / 2 unanswerable) are unchanged, so the design spec's script-surface row needs no edit.

## Tests

`784/784` pass (`./agent-test skills/fleet/scripts`), 17/17 in `prove-merge.test.mjs`. Both new tests were watched failing first — the behavioural one produced the ticket's reproduction verbatim, both lines and all.

Closes #267


feigi:
**Review findings applied** — 11 findings, 3 applied, 3 deferred/filed, 5 resolved by the comment narrowing, 1 rejected. Pushed as `254f981`. Suite: `./agent-test skills/fleet/scripts` → **tests 784 / pass 784 / fail 0**.

### Applied

- **The static guard's comment overclaimed** (`tests`, `types`, `correctness`, `silent-failure` — four findings, one defect). It opened *"The class #267 belongs to, rather than its one instance at line 64"*, which a `[`-only regex does not deliver. Narrowed to claim only what it lints. **The regex is deliberately unchanged**: both mutations (`test "$(...)"`, `x=$(...) || true`) were run and each turns the *behavioural* test red while leaving the guard green — the class is closed by the behavioural test, which is fault-agnostic. Widening to `test` would still miss `x=$(...) || true`, so the comment would still overclaim.
- **The exit-2 assertion cannot fail on this bug** (`correctness`, `silent-failure`). `die()` is `exit 2` unconditionally, before the fix and after. Assertion kept — 2 is still right — but its message no longer implies it discriminates; `doesNotMatch /did not land/` above it is and always was the pin.
- **Dropped the blob detour in the fixture** (`simplify`). `${head}^{tree}` clears the same `rev-parse --verify` gate, is never commit-checked either, and drives `merge-base --is-ancestor` to the same exit 128 — so the `writeFileSync`/`git add`/`${head}:f` indirection bought nothing. Mutation-verified: reverting the fix fails the tree fixture on the same assertion at the same line as the blob fixture did, and a `prove-merge.sh` comment spelling the idiom out leaves both tests green.

### Deferred and filed

- **#561** (`needs-triage`) — `is_ancestor`'s `echo true/false` protocol: refactor to an assignment (`simplify`), or validate the literal (`types`). Materially larger than #267's one-line fix, and the two findings are alternatives rather than steps, so the approach is open. Carries the latent `$((` false positive in the guard's regex too, so that observation survives either outcome.
- **#562** (`wontfix`, closed as a record) — dropping the guard's comment filter. Refuted by measurement: the filter is inert on today's script, but without it a comment demonstrating the *correct* idiom trips the guard.

### Rejected

- *"Delete the accept-path half (the second `prove()` call)"* (`simplify`). That half is same-fixture discrimination, added deliberately: the fix makes `set -e` live on a line where it was not, so a suite that only ever fed the probe failing input would not notice it aborting on a good one. Holding every other variable constant is strictly stronger than a different fixture asserting the same pair.

Issue #467, opened 2026-08-14
fleet scripts: collapse arg.mjs's makeDie/makeArg/makeHas into one makeCli(NAME) returning {die, arg, has}?


feigi:
Deferred from PR #464 review (dimension: **simplify**). Finding state: **deferred on scope and precedence — not refuted.** The reviewer applied it and measured it green; nobody disputes it works.

## The proposal

`arg.mjs`'s three factories exist only to thread `die` into `arg`/`has`. One `makeCli(scriptName)` returning `{die, arg, has}` removes the threading and collapses the repeated stanza at all seven call sites.

Call sites become `const { die, arg, has } = makeCli(NAME);` / `const { die, arg } = ...` / `const { die } = ...`.

## Evidence from the review (measured, not proposed)

The reviewer applied the refactor in a private copy: `arg.mjs` became a single `export function makeCli(scriptName)` with `die`/`arg`/`has` as inner functions closing over `die`.

- `node --test skills/fleet/scripts/*.test.mjs` -> `tests 640 / pass 640 / fail 0`, identical to baseline.
- Line count over the eight changed scripts: 2433 -> 2423. `arg.mjs` 78 -> 76.

Name the parameter `scriptName`, not `name`, so `arg()`/`has()`'s own `name` parameter does not shadow it.

## Why it was deferred rather than applied

1. **It re-decides a decision the ticket's own Agent Brief made.** #367's brief specifies `makeArg(die)` — *"`die` must stay injected"*. Re-deciding the exported interface inside the PR that implements it is a second design decision, not a cleanup. That is a triage call, which is why this is `needs-triage` and not `ready-for-agent`.
2. **It was mutually exclusive with two confirmed findings.** It rewrites the exact call-site stanza that PR #464's new regression pins assert on — the finding's own `suggested_fix` says *"Update the call-site pin at candidates.test.mjs:858 to match the new stanza."* A `suggestion` with 0 refuters does not get to invalidate the pins that two adversarially-confirmed `important` findings exist to add.
3. **It is behaviour-neutral by its own claim, so deferring costs nothing** — and it is strictly easier after the pins exist than before.

## What changed under it since

[...]

PR #563, opened 2026-08-17
fix(fleet): ledger's --file stops eating the flag that follows it

### LINKED ISSUE #362: ledger.mjs: --file swallows the next flag as its path, silently disabling --require-file and flipping exit 2 to 0

feigi:
Closes #362

`ledger.mjs` reads `--file`/`--require-file` with its own hand-rolled splice — arg.mjs's shared `arg()` deliberately does not reach it — and took whatever token followed `--file` as the path. `--file --require-file` therefore made the **flag** the path, and `splice(fileIdx, 2)` ate it.

Re-derived against `origin/main` before implementing (the ticket's line numbers had drifted; the defect had not):

```
$ node skills/fleet/scripts/ledger.mjs --file --require-file check "dup subject"
ledger: WARNING — ledger file not found: --require-file. Every check will read "safe to file" until it exists.
{"subject":"dup subject",...,"found":false,...,"verdict":"unverified"}
exit=0        <- correct invocation exits 2
```

A caller gating on that exit code files the duplicate. Now:

```
ledger: --file needs a path
exit=2
```

## The class, and what is left

| Case | Ruling |
|---|---|
| `--file <--flag>` | **Fixed** — the ticket's case. |
| `--file "   "` (whitespace-only) | **Fixed** — same clause; matches arg.mjs's contract for the four sites this reader does not route through. |
| `--file` trailing / `--file ""` | Already died; **unchanged, in its own wording** (see below). |
| `--require-file` | **Nothing to fix.** It takes no value, so its `splice(idx, 1)` can swallow nothing. Its only way to go missing was being eaten by the flag before it — which is the row above. Pinned anyway, since nothing tested it at all. |
| `--file=x` / `--require-file=x` | **Left.** Measured: both already exit 2 loudly (`unknown subcommand '--file=x'`), because ledger's leftover argv becomes the subcommand and the dispatch has an unknown-subcommand guard. Not a fail-open here, unlike at the `arg()` sites. |
| Repeated `--file` | **Left.** `indexOf` takes the first; the second becomes the subcommand and exits 2 loudly. |

## What this change newly REFUSES

A path that legitimately begins with `--` — the same deliberate forfeit arg.mjs documents. A single leading `-`, or a `--` anywhere but the front, is still a path, and that is pinned by a must-accept test asserting the file at such a path is not merely accepted but **read** (exit 1, the `already filed` branch).

The guard is additive and sits above the splice, so the trailing-`--file` case — the one behaviour #362's own Measured block records as already correct — keeps its original `--file given with no path` message. That is pinned too.

Exit vocabulary untouched: `0` clean, `1` already filed this run, `2` usage, `3` tracker hit.

## Mutation-tested

Nothing in `ledger.test.mjs` reached this parser: every test enters through `run()`, which always builds a well-formed `--file <path>` pair. Five CLI tests added (zero new imports, to avoid the same-file import collision that made #361 conflicting). Each clause was mutated and the result observed:

| Mutant | Red |
|---|---|
| drop `startsWith("--")` (the one #362 asks for) | exactly the `--file --require-file` test |
| drop `trim() === ""` | exactly the whitespace test |
| delete the guard | both refuse-tests; all three accept/preserve pins stay green |
| over-broad `includes("--")` | exactly the must-accept test |
| over-broad `startsWith("-")` | exactly the must-accept test |

That last mutant initially stayed **green** and caught a defect in my own pin: it passed `join(dir, "-weird--ledger.md")`, an absolute path beginning with `/`, so the "beginning with `-`" half of its name was never exercised. Fixed in the second commit by passing the value relative to the spawn's cwd.

`arg.mjs`'s header stated this defect as live ("still takes the next flag as the path"); updated to record it as fixed in `ledger.mjs`, and to keep the note that the parser is still ledger.mjs's own so a future change to the shared rule is known to need a second edit.

`./agent-test skills/fleet/scripts` — 792 pass, 0 fail, post-rebase.


feigi:
**Review findings applied — `217c5a3039d90a57d2318ab5a5881a71ddf9e251`.** Gate: `./agent-test skills/fleet/scripts` → `tests 794 / pass 794 / fail 0`. `ledger.mjs`'s exit vocabulary (`0`/`1`/`2`/`3`) re-verified by direct invocation after the change — unchanged, which matters because every member of this fleet run calls it before filing.

**Applied (5)**

- The `=` spelling was unguarded, and it falsified this PR's own claim. `--file=x`/`--require-file=x` exit 2 **only where the token lands in the subcommand slot**; anywhere else neither is refused — the unmatched token falls through to `rest` and `check`'s `rest.join(" ")` folds it into the **subject**. Measured: `check --file=<other-ledger> "<already-filed subject>"` → **exit 0** `verdict: "unverified"` against exit 1 `ALREADY FILED` for the space-separated form, and `--file <missing> check --require-file=true dup` → **exit 0** against exit 2. Two whole-argv guards at `ledger.mjs:55-56`, worded as `arg()`/`has()` already word this spelling; pinned in **both** positions.
- `ledger.mjs:55` "four call sites" → **five**, named. Re-derived three ways (5 `makeArg` importers calling `arg()`; 7 `makeDie` importers − `ledger.mjs` − flagless `fleet-tick.mjs`; #367's own title, *"five copy-pasted copies"*).
- The identical undercount at `pr-overlap.test.mjs:40`, and — **not reported by anyone, a scope extension I made and am flagging** — a third instance at `board.test.mjs:264`. Both from `3a181de`, where **five** scripts carried the clause, so "four" was wrong when written.
- `ledger.test.mjs`'s `cliFixture()` justified its `git` symlink with *"`defaultLedgerPath()` shells out to it"*. No CLI test there reaches a `git` call. Symlink kept, justification replaced with the true one.

**Refuted with evidence (2)** — the `doesNotMatch` at `:726` is **not** a vacuous pin (two mutants make it the only assertion that reds, while exit 2 and the refusal line still pass), and the `fileIdx !== -1` fold is behaviour-identical over 2522 differential cases but measures **longer** (487→489 lines; +14/−0 → +17/−1).

**Rejected on ruling (3)** — dropping the `file &&` term, which protects the one behaviour #362's Measured block records as already correct.

**Deferred (2 findings → 1 ticket)** — #567: `ledger.mjs` now hand-copies **three** of `arg.mjs`'s refusal rules and no test fails when they drift. Triage it with #467.

**Never delivered (1)** — `unverified7`: no text for it ever reached me. Recorded in #568 as unchecked, not cleared.

Full adjudication, mutation receipts and refutations: **#568**.

feigi:
**Post-merge addendum.** This PR merged (`b2e7f90`, head rebased as `8f43b79`) while two of its findings were still being checked — the review payload reached the fix-applier late. Both are now settled, and neither could be applied here:

- **`unverified7` (simplify, `ledger.test.mjs:695`) — SURVIVED a refuter biased to refuse.** `t.after()` plus a fixture-bound `cli` removes **26 lines** from the CLI block (851 → 825), with cleanup-on-failure measured identical to `try/finally` under both an assertion failure and a plain throw — proven with a negative control that leaks all 7 dirs when the hook is removed. Filed as **#569**, `ready-for-agent`.
- **`unverified5` (simplify, `ledger.mjs:57`) — my earlier comment called this refuted; that was wrong and is corrected.** The refutation ran against a paraphrase that claimed brevity. The verbatim finding claims only behaviour-identity and dropping the repeated condition, and both are true and measured. Re-adjudicated as surviving, deferred to **#567** because it rewrites the same lines `isFlagLike` would.

One thing this PR's own diff falsified and I missed on the first pass: **`refuted0`'s suggested fix had a second half** — extend `arg.mjs`'s header, which says #362 landed *"its own copy of the `--`-prefix rule"*. Singular is now wrong: `ledger.mjs:55-56` bring the count to **three** copied refusals. Added to **#567**.

Corrections to the record in #568: `unverified4`'s dimension is `types` at `arg.mjs:9`, not `tests`.

PR #564, opened 2026-08-17
fix(fleet): stop reading a deleted refs/stash as an empty stash

### LINKED ISSUE #376: no-undo-audit: a deleted refs/stash file with an intact reflog prints a confident `stash entries: 0` — the confident-zero class #147 closed, via a path #292 did not cover

feigi:
`show-ref refs/stash` exits **1** for two different worlds: a stash that is genuinely empty, and a `refs/stash` file that was removed while `.git/logs/refs/stash` stayed behind. The guard added by #292 defines rc 1 as genuine absence and nothing else (`-ne 1` was the whole test), so the second world printed a confident `stash entries (repo-global, not gated): 0` over commits still named in the reflog and still reachable.

That `0` is the value the merge bot reads as "nothing to lose", and the runbook's escape hatch only fires on `unknown` or nonzero — so this state routed around the operator check entirely, before an irreversible action.

## The change

rc 1 now buys a second question rather than a verdict: report `0` only when the reflog is gone or empty too.

- **Composes with the existing discriminant, does not replace it.** The three states #292 already catches leave `refs/stash` resolvable and land on rc 0/128, so they never reach the probe. Their line is unchanged, `$diag` included.
- **Lazy.** The probe runs only when `stash=0 && sr_rc=1`. A healthy repo with entries short-circuits on `$stash` and pays nothing.
- **Its own message.** `refs/stash is absent but its reflog is not, and still names entries no ref points at` — reusing the existing line would tell the operator "refs/stash is not absent" about the one state whose whole shape is that it **is**, then send them to `ls -l` on a file that no longer exists.
- **No new exit code.** This extends the existing `stash: null` case from the design spec, exactly as that spec describes it.

## The path comes from git, not from `$wt/.git`

`git -C "$wt" rev-parse --git-path logs/refs/stash`. The stack is repo-global and `$wt` is routinely a **linked** worktree — the fleet's own layout — where `.git` is a *file* and `$wt/.git/logs/refs/stash` reaches nothing. A hand-built path reads "no reflog", takes the accept branch, and prints the same confident `0` in the one layout that matters.

`--git-path` rather than a hand-joined `--git-common-dir` because git owns the per-worktree/common split: `logs/refs` is common, `logs/HEAD` and `logs/refs/bisect` are not, and that rule is git's to change.

## The other half: what this must NOT refuse

Turning every clean repo's honest `0` into `unknown` would be the worse bug — an operator who sees `unknown` on every run stops reading it, and then the three states above go unread too. Measured, git 2.50.1: `pop`, `drop`, `clear` and `update-ref -d` each remove `logs/refs/stash` **outright**. There is no ordinary lifecycle that empties the stack and leaves the reflog behind, which is why the probe can be this blunt.

Six parameterized accept-case tests pin that direction, one per lifecycle plus a zero-byte reflog (`-s`, not `-e` — a zero-byte reflog names nothing recoverable).

## Mutation-tested, both directions

| mutant | killed by |
|---|---|
| path hand-built as `$wt/.git/logs/...` | the linked-worktree test **only** — the plain-repo test stays green, as it must |
| `-s` → `-e` | the zero-byte accept case **only** |
| every rc 1 reports `unknown`, reflog unchecked | all six accept cases + 2 pre-existing |
| the new branch reuses the old `unknown` line | both defect tests |
| runbook count reverted / fourth-cause sentence deleted / recovery action dropped | one prose pin each |

Non-catch controls stay green: a comment reflow in the new block, and bolding or reflowing the new runbook sentence.

**Known unpinned:** the `|| die` on the `rev-parse` probe. It is unreachable — `$wt` has already answered three `rev-parse` calls above — so no fixture produces it; it is there because degrading to an empty path would land on `[ -s "" ]`, false, and print the confident `0` this PR removes. Exit 2 is the honest answer for "the repo went away mid-run", and both the spec and the runbook already list "a probe that could not run" under exit 2.

## Class members left open

Deliberately not covered, each already tracked: a **truncated-but-parseable** reflog (#306's pinned ceiling — resolves the ref, returns a short list, nothing disagrees); a **corrupt loose object** behind a non-tip entry (#482); an unreadable `logs/refs` **directory**, and reflogs disabled repo-wide (`core.logAllRefUpdates=false`) — both defeat the `-s` stat and read as absent. The last two are named in-line as a ceiling rather than fixed: with no reflog to probe there is nothing to distinguish them from an empty stack.

Docs: the runbook and the design spec both enumerated three causes for `unknown`; the fourth has the opposite `ls` signature, so the existing fault tell (`----------`, or `Permission denied`) does not fire for it and an operator following only those two reads a healthy listing and rebases over recoverable commits.

`./agent-test skills/fleet/scripts` — 798/798 green on the rebased head.

Closes #376


feigi:
## Review findings applied — `7ebb721`

`./agent-test skills/fleet/scripts` → **tests 799, pass 799, fail 0**.

**Applied (4 edits, 1 commit):**

- **`survived0` + `survived1`** (`comments` critical / `correctness` important — one defect) — this PR appended a table row directly above an untouched paragraph beginning *"That last row is a fourth CAUSE, not a fourth signature…"*, which describes the **malformed-ref-file** row. Measured: malformed → `show-ref` rc **128** with a healthy `ls -l` mode; deleted → rc **1**, "No such file or directory". Both of the paragraph's claims were false of the row that became last, and the new row is precisely a fourth *signature*, which the sentence denies word for word. Fixed by **naming** the rows (`malformed ref file`, `ref file DELETED`) rather than repositioning, so the next insertion cannot re-break them — this also removed the second, contradictory "last row above" 13 lines later.
- **`unverified2`** (`silent-failure`) — the PR's own new prose stated a ceiling that does not measure. `core.logAllRefUpdates=false` suppresses the HEAD reflog but `git stash push` force-creates `logs/refs/stash` regardless, so such a repo probes normally and reports `unknown` (re-measured here, git 2.50.1). Clause dropped; the unreadable-`logs/refs`-DIRECTORY half is real, re-measured, and stays.
- **`unverified8`** (`simplify`) — `rev-parse --path-format=absolute --git-path` replaces the hand-rolled `case` fix-up. Same spelling this script already uses unconditionally at :211/:213, so no version floor moves. Verdict-identical across 16 measured layouts.
- **`survived2`** (`tests` important) — new test spawns the script from a cwd that is **not** `$wt`, the only way `run-merge-bot.md` ever calls it. Every existing fixture passed `cwd: w`, so this direction had no pin. Pinned as **behaviour**, not mechanism, since `unverified8` deleted the mechanism it was originally written against.

**Mutation-verified** (on a copy, worktree untouched): reverting `--path-format=absolute` reds the new test **and only it** (65/1); re-adding the now-redundant `case` fix-up on top leaves it **green** (66/0) — so it pins that the probe resolves, not how.

**Deferred and filed:** #570 (the unreadable-`logs/refs` ceiling — `correctness` + `silent-failure`, declared in-line, out of #376's scope, not a regression) and #571 (the suggestion band: 5 refuted with measurement, 1 survived but declined, plus the `refuted0` ruling). No action on `unverified4` — the implementation self-disclosed that `|| die` as unpinned rather than claiming coverage, which is the correct disclosure.

Issue #558, opened 2026-08-17
run-team/SKILL.md: the "Verify every reported SHA" guard names no exit-2 branch — all five of verify-sha.sh unanswerable paths are undocumented on the consuming side


feigi:
**Deferred from PR #555 review** (implements #266). **Dimensions: silent-failure, correctness, types** — three specialists reported this independently, in three separate findings. Severity: important (silent-failure), suggestion (correctness, types). Reported at `skills/fleet/skills/run-team/SKILL.md:1167-1170`.

## The finding

`verify-sha.sh` exits 2 when the question could not be answered, and 0/1 when it could. The only caller-facing instruction — the "Verify every reported SHA" guard at `skills/fleet/skills/run-team/SKILL.md:1168-1172` — names exactly one outcome branch:

> Not reachable → flag, do not enqueue, do not return the ticket to the pool until the maintainer rules.

No fleet doc ties an exit code to `verify-sha` at all. `grep -rn "exit 2\|exit code\|unanswerable" skills/fleet/skills/` returns 6 hits, every one of them about `inflight.sh` — `skills/fleet/skills/next-ticket/SKILL.md:37` annotates it inline as `# 0 free, 1 taken, 2 unanswerable`. That per-script inline convention is the house style; `verify-sha` has no equivalent, so the one distinction #266 exists to make has no handling rule on the consuming side.

`grep -rn "verify-sha" .` over the review snapshot confirms `run-team/SKILL.md:1168` is the **only** consumer.

## Why this was deferred rather than fixed in PR #555

Two things the refuting vote got right, and the controller upheld:

1. **Pre-existing, not introduced.** `verify-sha.sh` already had four exit-2 paths before PR #555 — bad argument count (`:14`), failed fetch (`:19-20`), unresolvable `origin/<branch>` (`:22-23`), non-commit sha (`:28-29`). None of them was documented in the guard either, so "no fleet doc mentions verify-sha exit 2" was equally true before the diff. #555 adds a fifth exit-2 path and removes no handling rule.
2. **The current behaviour is fail-safe.** Exit 2 emits **no JSON at all**, plus a loud stderr line. A controller therefore cannot read `reachable:true` and cannot enqueue — strictly at least as conservative as the pre-PR behaviour. Undocumented, not unsafe.

So this is out of #266's stated diff scope and was not a blocker for #555. But three review dimensions converging on it independently is strong evidence the gap is real and worth closing.

## Suggested fix

[...]

Issue #481, opened 2026-08-14
no-undo-audit: the unknown path discards show-ref's stderr, which git does speak in two of the four unknown states


feigi:
**Deferred from PR #480 review** (implements #304). **Dimension: correctness.** Severity: important. Reported at `skills/fleet/scripts/no-undo-audit.sh:325`.

## The finding

PR #480 makes the audit pass git's own diagnostic through on the `unknown` stash path, but only the one `git stash list` speaks in. In two of the four unknown states `stash list` is silent while **`show-ref` — the cross-check the script already runs one line above, under `>/dev/null 2>&1`** — names a fault. That stderr is discarded, so the operator still gets the generic three-cause guess in half the reachable states.

Measured, git 2.50.1 (Apple Git-155), four throwaway fixtures, one stash entry each:

```
state                                        | stash list        | show-ref rc | show-ref stderr
---------------------------------------------+-------------------+-------------+--------------------------------------------------
rm .git/objects/<stash sha>                  | rc1 `fatal: bad   | 128         | fatal: git show-ref: bad ref refs/stash (<real sha>)
  (stash object gone)                        |  object refs/stash|             |
chmod 000 .git/refs/stash                    | rc0, silent       | 128         | fatal: git show-ref: bad ref refs/stash (0000…0000)
  (unreadable ref file)                      |                   |             |
printf 'not-a-sha' > .git/refs/stash         | rc0, silent       | 128         | fatal: git show-ref: bad ref refs/stash (0000…0000)
  (malformed ref file)                       |                   |             |
chmod 000 .git/logs/refs/stash               | rc0, silent       | 0           | (silent)
  (unreadable reflog)                        |                   |             |
```

Only the first is covered by #480. The last is genuinely silent — nothing to pass through. The middle two are the gap.

## Why deferred, and why the approach is open

Two refuters reproduced the same measurements and split 1-1 on what to do:

- **For:** git answered and the audit threw the answer away; the discrimination #304 asks for is sitting in a call the script already makes.
- **Against:** that stderr is **not git naming the fault**. It is the same canned line for two different causes, and in both it is *factually wrong* — the ref does not hold `0000…0000`, and in the `chmod` case the actual fault is `EACCES`, which the message never mentions. Appending it would replace a truthful three-cause disclosure with a confidently false specific. Contrast the covered case, where `fatal: bad object refs/stash` names the real fault and prints the ref's real SHA.

[...]

PR #565, opened 2026-08-17
fix(fleet): let git's own diagnosis reach the operator on verify-sha's three guards

### LINKED ISSUE #559: verify-sha.sh:19,22,28 — three guards swallow git stderr and assert a cause list that is false for the commonest failures (a bad remote URL and a removed remote are byte-identical)
**Deferred from PR #555 review** (implements #266). **Dimension: silent-failure.** Severity: important. Reported at `skills/fleet/scripts/verify-sha.sh:19`, `:22`, `:28`.

## The finding

Three guards in `verify-sha.sh` discard git's stderr behind `2>/dev/null` and replace it with a `die()` message asserting a specific cause list. Two of the three assert causes that are false for the most common real failures.

```sh
19  git fetch --quiet origin "$branch" 2>/dev/null \
20    || die "cannot fetch origin/$branch — branch missing, or no network"
22  tip=$(git rev-parse "origin/$branch" 2>/dev/null) \
23    || die "origin/$branch does not resolve after fetch"
28  git cat-file -e "${sha}^{commit}" 2>/dev/null \
29    || die "$sha is not a commit object in this repository"
```

### Measured — both review votes reproduced this independently

Bare origin + clone fixture, then:

| mutation | verify-sha output | git's own diagnosis, discarded |
|---|---|---|
| `git remote set-url origin /nonexistent/path.git` | `verify-sha: cannot fetch origin/main — branch missing, or no network`, rc=2 | `fatal: '/nonexistent/path.git' does not appear to be a git repository` |
| `git remote remove origin` | **byte-identical** to the row above | `fatal: 'origin' does not appear to be a git repository` / `fatal: Could not read from remote repository` |

Neither failure is "branch missing" and neither is "no network". Two distinct faults collapse into one message naming neither of them, and the operator is sent looking for a missing branch or a network problem that does not exist. Deleting `2>/dev/null` from line 19 in the same fixture surfaces git's real diagnosis.

Line 28 has the same shape with a stronger overclaim: its message asserts `is not a commit object in this repository` for **any** `cat-file` failure — including an unreadable or corrupt object store, where the sha may well be a commit that git simply cannot read.

## Why this was deferred rather than fixed in PR #555

Both review votes agreed the mechanism reproduces. The refuting vote falsified only the finding's rhetorical framing — that this "contradicts the rationale the PR just wrote five lines below at line 34" — not the defect itself. A refutation of framing is not a refutation of the defect, so this is filed rather than dropped.

It is a different class from #266: **message accuracy**, not verdict collapse. #266 is about exit 2 and exit 1 meaning different things; this is about exit 2 arriving with a wrong explanation. Fixing it in #555 would have widened that PR beyond its ticket.

## Suggested fix

Drop `2>/dev/null` and generalise the message, so git's own words reach the operator and the script stops guessing at a cause it does not know:

```sh
git fetch --quiet origin "$branch" || die "cannot fetch origin/$branch"
tip=$(git rev-parse "origin/$branch") || die "origin/$branch does not resolve after fetch"
git cat-file -e "${sha}^{commit}" || die "cannot resolve $sha to a commit in this repository"
```

`verify-sha.test.mjs` asserts `doesNotMatch(stderr, /error:/)` on the ordinary negative path, guarding exactly against leaking git noise into a plain "no". That path is the `merge-base` branch at `:36`, which already runs without `2>/dev/null` and is untouched by these three changes — but add a test per site anyway, since the three above are die paths whose stderr nothing currently pins.

## Related, not duplicate

- #481 — the same stderr-swallow class, but scoped to `no-undo-audit.sh:325`. Does not own this script.
- #266 / #267 — the verdict-collapse class in `verify-sha.sh` / `prove-merge.sh`. Different defect, same file for #266.

feigi:
Closes #559

Three guards in `verify-sha.sh` sent git's stderr to `/dev/null` and replaced it with a `die` message asserting a cause list the script cannot know.

## Measured, on a bare origin + clone

| mutation | before | after |
|---|---|---|
| `remote set-url origin /nonexistent/path.git` | `cannot fetch origin/main — branch missing, or no network` | same guard, plus `fatal: '/nonexistent/path.git' does not appear to be a git repository` |
| `remote remove origin` | **byte-identical to the row above** | now distinguishable — git names `'origin'` |
| `fetch origin nosuchbranch` | same message again | `fatal: couldn't find remote ref nosuchbranch` |

Neither of the first two is a missing branch and neither is a missing network. Two distinct faults collapsed into one message naming neither.

## The change

`2>/dev/null` dropped from the `fetch`, `rev-parse` and `cat-file -e` guards, and each message stopped at what its guard actually observed:

- `cannot fetch origin/$branch` — the cause list is gone; git supplies the cause.
- `origin/$branch does not resolve after fetch` — unchanged, it already asserted only what it saw.
- `cannot resolve $sha to a commit in this repository` — was `$sha is not a commit object in this repository`, which is false for an object that is present but not a commit (git: `expected commit type, but the object dereferences to tree type`) and for an object store it cannot read.

## What this can newly make noisy: nothing

The reason `2>/dev/null` was safe to drop is that all three gits are byte-silent when they succeed. Measured, 0 bytes of stderr each: `fetch --quiet` both up-to-date and with objects to transfer, `rev-parse` on a resolvable ref, `cat-file -e` on a commit. A test pins this exactly rather than sniffing for `/fatal:/`, because this script's stderr is read by the fleet controller and a guard that floods it on healthy runs would be the worse bug.

Dropping the cause list also refuses nothing: the one cause it named correctly — a missing branch — is still reported, now by git and precisely.

## Tests

One per guard, all asserting on **stderr content**, not exit 2 plus empty stdout — those two agree across genuinely different causes, which is how a test in this file once looked like it pinned the `merge-base` probe while actually pinning a failed fetch. The rev-parse guard is reached with real git and no shim by unsetting `remote.origin.fetch`, so the fetch succeeds into `FETCH_HEAD` while the remote-tracking ref stays absent.

Mutation-checked: restoring each redirect reds only that site's test(s); restoring either old message reds only the tests pinning it; dropping `--quiet` reds only the quiet-path test; a reworded comment and a requoted remote stay green.

The existing pin on the `cat-file` message moved to the new wording rather than loosening — it is what proves the exit 2 came from that guard and not from the fetch.

`./agent-test skills/fleet/scripts`: 793 pass, 0 fail.

## Scope

Exit vocabulary unchanged — 0 reachable, 1 not reachable, 2 unanswerable — so the spec row at `docs/specs/2026-07-23-fleet-plugin-design.md:182` still describes what these codes mean and is untouched. This only makes causes visible.

Deliberately left, owned elsewhere: the `merge-base` guard (already fixed by #266 / PR #555, and already unmuted), the consuming side in `run-team/SKILL.md` (#558), the unescaped-JSON `printf` in this file (#119), and `no-undo-audit.sh` (#481).

One adjacent instance is **not** covered by any open ticket: `prove-merge.sh:58` carries the same pattern — `git cat-file -e "${obj}^{commit}" 2>/dev/null || die "$obj is not a commit in this repository"`. Same class, different script, so it is left here rather than widened into.


feigi:
## Review adjudication — 1 applied, 2 refuted, 2 already covered elsewhere

Head at review: `86356657`. Pushed: `c5e72cfd433ec0874efc2c77c3cec906e99e5325`. Gate: `./agent-test skills/fleet/scripts` → **tests 793 / pass 793 / fail 0**.

Nine findings. Every `suggestion`-band one that was in this PR's scope got a refuter biased to refuse.

### Applied (1)

- **`verify-sha.test.mjs:182` — `assert.notEqual(a.stderr, b.stderr, …)` was dead.** SURVIVED refutation. The two assertions above it put `nonexistent` into `a.stderr` and keep it out of `b.stderr`; two strings differing in substring membership cannot be strictly equal (`assert/strict`'s `notEqual` is `notStrictEqual`), so it was a theorem of its neighbours. Unlike a vacuous-*looking* guard that pins where a `die()` sits, this implication is between the assertions themselves, so no edit to `verify-sha.sh` can reach it — 15 mutants against an instrumented replica running each assertion in its own `try`/`catch` produced no row where it reds while the two above stay green, and a 3M-pair fuzz over this stderr's own fragments found none either. Its sentence moved to the assertion that does the work rather than being deleted with the line.

  Mutation-checked after the edit: re-muting the fetch guard reds only this test (at the git's-own-diagnosis assertion); making the fetch `die()` carry a hardcoded remote URL — the collapse-into-one-message class the removed line named — also reds only this test, landing on the assertion that inherited the sentence. Over-broad control stays green.

### Refuted, filed as records (2)

- **#572** — the healthy-run exact-stderr comment is already accurate. The claimed "fourth line over ssh-style transport" was a hand-written banner in a probe using `ext::`, a protocol git blocks by default; git emits nothing there. Measured against this repo's real origin (`git@github.com:…`): fetch stderr is **0 bytes**, and the script produces the same **three** lines the fixture asserts. The suggested rewording would make the comment less true. The `deepEqual` is discriminating — dropping `--quiet` reds it, and only it, with git's own transfer lines.
- **#573** — `verify-sha.test.mjs:206`'s `doesNotMatch(stderr, /cannot fetch/)` **can** fail. Two mutants make it the only red in the suite (8/9 with it, 9/9 without): a plain "DRY the guard message" refactor reusing one `fetch_msg` variable across both `die()` calls, and a `die()` that prints a summary of passed guards. It does not pin where the guard sits — in this fixture the fetch genuinely succeeds (`FETCH_EXIT=0`) — it pins that the fetch guard's string never reaches stderr on a run where the fetch succeeded.

  #573 also records a separate measurement: downgrading the fetch guard from `|| die` to `|| echo … >&2` passes the whole suite 9/9. Nothing pins that a failed fetch is fatal. Not this finding, not plugged by this assertion — recorded so the measurement is not lost.

### Already covered, nothing filed (2 findings, 4 reports)

- **`prove-merge.sh:58` keeps the defect this PR fixes** (survived 2-0, plus the same claim in the suggestion band) — already **#566**, filed independently before this review returned. Out of this PR's scope; not widened here.
- **`docs/specs/2026-07-23-fleet-plugin-design.md:182` is stale** (2-0 refuted on two dimensions, plus two suggestion-band restatements) — the spec row is a prose cause list, never a quotation of `die()` text, and the cause it names still occurs. Settled; spec row untouched.

CI is stale-red from being behind `main` after #563 merged — staleness, not a failure. Not rebased.

Issue #463, opened 2026-08-13
fleet scripts: board.mjs and ci-state.mjs silently ignore stray positional arguments — candidates.mjs refuses them for free via strict parseArgs


feigi:
Deferred from PR #462 review (dimension: **silent-failure**). Finding state: **refuted as framed, residue real** — the review reported this as `--declare-no-ci false` "failing open", which does not hold (see *Why the original framing was rejected* below). What survived the refutation is a narrower, genuine asymmetry, which is not filed anywhere.

## The asymmetry

`candidates.mjs` refuses any stray positional token; `board.mjs` and `ci-state.mjs` ignore it in silence. The difference is not a decision anyone took — it falls out of `candidates.mjs` being the only one of the three that runs `parseArgs`, whose strict mode is on by default.

Measured on `9926a66` (PR #462 head):

```
$ node skills/fleet/scripts/board.mjs build --ledger /nonexistent/x.md junk
   -> exit 0, full board.json on stdout, 'junk' never mentioned

$ node skills/fleet/scripts/candidates.mjs --require-label ready-for-agent junk
   -> exit 2
   candidates: Unexpected argument 'junk'. This command does not take positional
   arguments — accepted: --require-label, --allow-fallback, --limit
```

```
$ grep -c 'parseArgs' skills/fleet/scripts/{board,ci-state,candidates}.mjs
board.mjs:0   ci-state.mjs:0   candidates.mjs:4
```

`ci-state.mjs` behaves like `board.mjs` — measured during the PR #462 review with the `ci-state.test.mjs` fixture rig (git-init'ed repo, `gh` stubbed on PATH): `["--quiet", "false"]` runs normally with `false` ignored, and `["--declare-no-ci", "false"]` exits 0 with the trailing token never mentioned in `reasons`.

## Why this matters here specifically

Both scripts are invoked from **markdown read by a model**, which is the argument `candidates.mjs:66-68` already makes for its own guard ("a malformed invocation is more plausible here than the shape of this guard suggests"). A dropped token in `ci-state.mjs` lands on the fleet's merge gate.

## Why it is NOT #365

#365 is the unknown **flag name** sweep (`--basee` reading as absent). This is a stray **positional value** — a bare token with no leading `--`. #365's suggested fix cannot catch it: that sweep is

```js
process.argv.slice(2).filter(a => a.startsWith("--") && !known.has(a.slice(2)))
```

[...]

PR #575, opened 2026-08-18
fix(fleet): refuse a flag name no script reads, instead of running on defaults

### LINKED ISSUE #365: fleet scripts: a misspelled flag name reads as absent and silently takes the default — no unknown-flag sweep anywhere

feigi:
Closes #365

## The bug

Every flag is read by looking for its own name in `argv`, so a name nothing
reads is never looked for. `ci-state.mjs --pr 5 --basee main` left `base` on
its default and returned a real, wrong verdict at exit 0/1 with no refusal —
the verdict the fleet gates PR-green on. Same fail-open harm as #61/#169,
reached from the other side: not an absent value, an unread flag.

## The fix

`makeSweep(die)` in `arg.mjs`, beside the guards it completes, taking each
script's own set of names — per the brief's ruling on #367, not a fifth copy.

- It refuses `--`-prefixed tokens **only**. Subcommands and values are
  positional and none of its business, so `board.mjs build` / `serve` are
  untouched. `arg()` already refuses a `--`-prefixed value, so there is no
  value to skip over and the sweep never needs to know which names take one.
- `--known=value` is looked up by **name**, so it stays `arg()`/`has()`'s to
  refuse in their specific wording, while `--unknown=value` is caught here.
- Each call site sits **below its own value guards**, the way
  `candidates.mjs` places its `parseArgs` — `diff-stats.mjs --pr --json` still
  reports `--pr needs a value` — and **above the first `gh` call**.
- `board.mjs` and `diff-stats.mjs` call it inside `main()`, never at module
  scope: both are imported by their suites and would otherwise read the
  importer's argv.

## Enumerate and declare

Covered — a misspelling refuses at exit 2, naming the stray, before any query:

| script | mechanism | known set |
|---|---|---|
| `board.mjs` | `sweep()` | `ledger prev port interval open spend-since` |
| `ci-state.mjs` | `sweep()` | `pr base workflow workflow-file declare-no-ci quiet` |
| `diff-stats.mjs` | `sweep()` | `pr` |
| `pr-overlap.mjs` | `sweep()` | `a b` |
| `candidates.mjs` | its own `parseArgs` (#173), **unchanged** | `require-label allow-fallback limit` |

Deliberately left:

- **`candidates.mjs` keeps `parseArgs`.** It accepts no positionals, so it also
  refuses a bare `candidates.mjs ready-for-agent` — which the shared sweep
  cannot, `board.mjs`'s subcommands being exactly that shape. Migrating it
  would be a strict loss of strictness in the script the ticket cites as the
  reason to do this at all. Said out loud in `arg.mjs` so the next reader does
  not "unify" it.
- **`ledger.mjs`** — out of scope, #362.
- **Per-subcommand narrowing in `board.mjs`.** One known set spans both
  subcommands, so `build --port 5` is still accepted-and-ignored. That is the
  pre-existing gap `board.mjs`'s own comment already names; refusing a flag the
  file accepts elsewhere is a different ticket's fix.
- **`--flag=value` (#364) and numeric validation (#366)** — already owned.
- **No fuzzy "did you mean"** — per the ticket's out-of-scope list.

## What this could wrongly REFUSE

The AC's own words: a sweep that refuses a working invocation is worse than
the bug. Every flag literal in every call site (docs, `SKILL.md`s,
`fleet-tick.mjs`, `board.mjs`'s own `ci-state` spawn) and in every suite was
enumerated against these sets before writing them.

Pinned by one test per script driving its **whole** set in one green run —
every other test in these suites passes a subset, so a dropped name refuses a
working invocation with the suite still green. `ci-state`'s reaches
`--workflow-file`, whose `arg()` call sits far below the sweep and which a set
built by reading down to the first `gh` call would miss.

## Tests

821 pass, 0 fail (`node --test skills/fleet/scripts/*.test.mjs`), up from **810**.

The baseline is measured, not remembered: `git archive origin/main` into a clean
extract, 43 test files, `810 tests / 810 pass / 0 fail`. An earlier revision of
this section said "up from 813", which was wrong.

Thirteen mutants, each killed by the intended test and no other: sweep never
refuses; `=`-name split dropped; `argv.slice(2)`→`(3)`; the token echoed as
parsed rather than as typed; the call deleted from each of the four scripts;
a name dropped from each of four known sets; and the `diff-stats` call hoisted
above its `if (!pr)` guard — which reddens the existing `--pr --json` case and
is what pins the below-the-guards ordering.

One finding from that round, fixed in `70b21ad`: the unit probe called `arg()`
before `sweep()`, so its `--base=main` case died in `arg()` either way and held
green with the sweep's `=` split deleted. Sweep-first makes it discriminate.

"Before any `gh` call" is pinned by a stub `gh` that writes a receipt when run,
not by `assert.doesNotMatch` on stderr — a script that queried and stayed quiet
would pass that. The stub also keeps a regressed sweep off this repo's live
GitHub data.

## Review pass (`8ad34e8`)

PR #575's review returned 16 findings across six dimensions. Applied here:

- **Three comments this PR added asserted things the tree does not.** `board.mjs`
  named `serve()`/`gather()` as imported by `board.test.mjs` (neither is —
  `createBoardServer` is); `diff-stats.mjs` named "review-pr's snapshot path" as an
  in-process importer of `computeStats()` (it runs the script as a CLI subprocess);
  `board-cli.test.mjs`'s new accept-test claimed to be "the only one that would go
  red" (measured: all six names redden a pre-existing test too, so it now claims
  what it actually owns — the whole set in one run). Same false inference corrected
  in `ci-state.test.mjs`'s sibling, where only `base` and `workflow` are that test
  alone.
- **Both accept-tests' `doesNotMatch(stderr, /unknown flag/)` are live, and say so.**
  A `simplify` finding called them unreachable. They are not: `tryRun()` forwards a
  tolerated child's stderr and swallows its failure, so a child refusing a flag
  lands on the parent's stderr at **exit 0**. In `board-cli.test.mjs` a `ledger.mjs`
  flag skew makes that assertion the only red in 821. `ci-state.test.mjs`'s copy was
  briefly dropped as subsumed — true of the regex as written, but its tolerated
  children are git's, and git says `unknown option`. Restored widened to
  `/unknown (flag|option)/`, which a bogus flag on `git remote get-url` makes the
  only red while the narrow and deleted forms ship it green at `"behind":null`.
  Measurement and reusable lesson recorded in #595.
- **`arg.mjs`'s `makeSweep` comment gains its two acknowledged exceptions**:
  `board.mjs`'s deliberate above-`cmd` placement, and `ledger.mjs` as deliberately
  unswept — its free-text tail makes a `--` token data, so this sweep would refuse
  working invocations.

Deferred, filed: **#584** (`ledger.mjs`'s unswept tail flips the duplicate-filing
guard to exit 0 — real, unowned since #362 closed without covering it, and the
obvious guard is refuted with evidence), **#586** (the `build|serve` usage line
this placement costs), **#589** (the suggestion band's full record) and **#595** (a `wontfix` record so the next `simplify` pass does not re-flag those assertions). The
single-dash / bare-token residual is measured onto **#463**, which already owns it.


feigi:
## Review adjudication — 16 findings, six dimensions

Applied in `8ad34e8`. Suite: **821 tests / 821 pass / 0 fail** (`node --test skills/fleet/scripts/*.test.mjs`, from the worktree). `origin/main` baseline re-measured from a clean `git archive` extract: **810 / 810 / 0**, 43 test files — the body's "up from 813" was wrong and is corrected.

Every `suggestion` arrives with **0 refuters** by workflow policy, so each in-scope one got its own refuter, biased to refuse and required to verify by running something. Five were dispatched; all five reported.

### Applied (6)

| finding | dim | what |
|---|---|---|
| `survived-1` | comments | `board.mjs:411` named `serve()`/`createBoardServer()`/`gather()` as imported by `board.test.mjs`. Only `createBoardServer` is; `serve` and `gather` are exported and imported nowhere. Now says the module is imported, which is the load-bearing part. |
| `refuted-1` | comments | `diff-stats.mjs:143` named "review-pr's snapshot path" as an in-process importer of `computeStats()`. `workflows/review-pr.js` never imports it — it runs the script as a CLI subprocess with its own argv. Now names the two real importers, `diff-stats.test.mjs:13` and `select-dimensions.test.mjs:5`. Placement inside `main()` unchanged; it is correct and test-enforced. |
| `unverified-6` | simplify | `board-cli.test.mjs`'s accept-test claimed "this is the only one that would go red". Measured false for **six of six** names. Narrowed to the finding's own second branch: test kept, comment corrected. Same false inference fixed in `ci-state.test.mjs`'s sibling, where only `base` and `workflow` really are that test alone. |
| `unverified-8` (ci-state half) | simplify | `ci-state.test.mjs`'s `doesNotMatch(stderr, /unknown flag/)` is subsumed by the exit-0 assertion above it. Dropped. |
| `unverified-10` | simplify | `stubGhBin()` folds `chmodSync` into `writeFileSync`'s `mode`. |
| `refuted-2` residual | types | `arg.mjs`'s below-the-guards contract now names `board.mjs`'s deliberate above-`cmd` placement as its acknowledged exception. |

Plus `arg.mjs`'s `makeSweep` comment now records `ledger.mjs` as deliberately unswept, with the reason and a pointer to #584.

### Deferred and filed (5 findings → 3 issues + 1 comment)

- **#584** — `ledger.mjs`'s unswept free-text tail absorbs a stray `--` token into the check subject, flipping the duplicate-filing guard from exit 1 to **exit 0**. Three findings, one defect (`survived-0` correctness, `refuted-0` silent-failure, `unverified-4` types), reproduced by three independent agents. Real and unowned — #365's out-of-scope list points at #362, which is **closed and never covered unknown flags**. `needs-triage` because the obvious remedy is refuted with measurement: a bare `startsWith("--")` guard would refuse `check "--require-file silently absent when value missing"`, which works today and is the shape of issue titles in this repo. The counter-evidence is in the issue verbatim.
- **#586** — the `build|serve` usage line, now unreachable because the sweep sits above `cmd` (`unverified-2`). Filed rather than fixed: the placement is deliberate and documented at `board.mjs:421-423`, and moving the sweep down would undo the rationale that refuted a sibling finding. Both spellings are exit 2; no script implements `--help` and no doc points at one.
- **#463** (comment, not a new issue) — the single-dash `-basee` and bare `basee` residual (`unverified-0`, `unverified-1`). #463 already owns this class and **predicted exactly this**: "Landing #365 as specified would leave this open." Measured post-#575 and added there, including the single-dash spelling that issue did not yet name. Out of scope by #365's own triage ruling ("the sweep must only ever refuse tokens starting with `--`"), and the naive widening to `-` is measurably unsafe — it would refuse `--spend-since -1`.
- **#589** — the suggestion band's full record: three refutations with their mutant tables, plus the two declined findings and why.

### Refuted by measurement — do not re-raise (3)

- `unverified-3` (append `sweep()` to the `die()/arg()/has()` header): the header's six issue citations map exactly onto those three functions and **none** covers `#365`; sweep's rationale is per-call-site by design because it is placement-sensitive. The remedy would point readers at six irrelevant issues, and was under-scoped to 2 of the 4 importers.
- `unverified-9` (drop the `candidates` row from the STRAYS matrix): the two suites measure **different** properties. `candidates.test.mjs` counts the script's own `$ gh` self-report; `arg.test.mjs`'s row uses a stub `gh` with a receipt file. A one-line *unannounced* `gh` call before the parse guard leaves `candidates.test.mjs` green at 60/60 and reddens only this row. Nine mutants tabulated in #589.
- `unverified-8`, the `board-cli.test.mjs` half: **not** dead. The producer can be a child process — `tryRun` forwards a child's stderr and swallows its failure, so a `ledger.mjs` refusal lands on board's stderr at **exit 0**. Under that mutant the full suite is 821/820/**1**, that assertion alone. Kept, with the reason now in the file so the next simplify pass does not delete it.

### Declined, not refuted (2)

- `unverified-7` (collapse `makeSweep(die)(known)` → `sweep(die, known)`): measured green at 821, saves ~5 lines, breaks `arg.mjs`'s `makeX(die)` symmetry. Not worth it; reversible.
- `unverified-5`: the finding's own `suggested_fix` is "No change recommended". All four `sweep()` arrays verified to match their script's own reads today.

### Blast radius

One overlap: `unverified-6` and `unverified-8` both land inside `board-cli.test.mjs`'s new test. Resolved by taking `unverified-6`'s *keep-and-correct* branch, so one touches the comment block and the other the assertion body — no shared line, no ruling needed.

<sub>Findings applied by an unattended fleet member. Every assertion written into a comment here was re-derived by mutation against the post-fix tree, not copied from the finding text.</sub>

feigi:
## Correction to the adjudication above — `f970ebc`

The comment above says `ci-state.test.mjs`'s `assert.doesNotMatch(r.stderr, /unknown flag/)` was dropped as subsumed. **That was measured right and decided wrong. Reversed in `f970ebc`.**

Subsumed *as written* — every child `ci-state` spawns through `run()` on that accept-test's exit-0 path `die()`s at exit 2, so the status assertion fires first. But the behind-count block **tolerates** its children: `tryRun()` swallows the failure and returns `null` while `execFileSync` has already forwarded the child's stderr. That is the identical mechanism keeping `board-cli.test.mjs`'s copy alive, which the same pass kept for exactly that reason. The two are not structurally different — only the vocabulary is. **The tolerated children there are git's, and git says `unknown option` where the fleet's own scripts say `unknown flag`.**

Mutating `tryRun("git", ["remote", "get-url", "origin"])` to carry a bogus flag:

| assertion | result |
|---|---|
| widened `/unknown (flag\|option)/` | `tests 35 / pass 34 / fail 1` — **only red** |
| narrow `/unknown flag/` (as merged) | `tests 35 / pass 35 / fail 0` |
| deleted (`8ad34e8`) | `tests 35 / pass 35 / fail 0` |

And the mutant is a real fail-open, not a wording nit:

```
$ node ci-state.mjs --pr 575
{"pr":575,...,"behind":null,...}      exit 0
unknown option `bogus-flag'            <- git's refusal, forwarded, unreported
```

`behind` is the count the controller reads for staleness; it degrades to `null` in silence. Restored and widened. Discrimination checked the other way too: a tolerated child failing for an unrelated reason (`get-url nosuchremote`) leaves the suite green at 35/35, so it pins the refusal class and not a noisy child.

Suite at `f970ebc`: **821 / 821 / 0**.

Full measurement, both halves, filed as a record at **#595** (`wontfix`) alongside this repo's existing records for the same class — #573, #562, #568 item 6. #589's table is corrected to match.

<sub>The `board-cli.test.mjs` half of the finding is unchanged: refuted, assertion kept, reason in the file.</sub>

Issue #573, opened 2026-08-17
REFUTED: don't drop verify-sha.test.mjs's `doesNotMatch(/cannot fetch/)` — two mutants make it the only red


feigi:
Deferred from PR #565 review (closes #559). Dimension: `simplify`. Location: `skills/fleet/scripts/verify-sha.test.mjs:206` (reviewed as `:207`, before this PR's own follow-up commit shifted it by one).

Filed as a **record, not as work** — `wontfix`. A future `code-simplifier` pass will re-flag this same assertion; this is the measured answer so nobody re-derives it. Sibling records: #562, and #568 item 6, which refuted the *same class* of claim on `ledger.test.mjs`.

## The finding

CLAIM: `assert.doesNotMatch(stderr, /cannot fetch/, "the fetch passed — this is the guard after it")` cannot fail: `die()` exits 2 immediately, so stderr can never hold both the fetch message and the rev-parse message that the line above already requires.

EVIDENCE: (a) deleting the line leaves `tests 9 / pass 9 / fail 0`; (b) forcing the fetch guard to fire instead (replacing the fetch command with `false`) fails at the *rev-parse* assertion above it, so that line fires first and this one is unreachable. Guard structure `die() { echo "$NAME: $1" >&2; exit 2; }` at `verify-sha.sh:12`.

SUGGESTED_FIX: drop the line and keep its sentence as a comment on the assertion above, or fold it into that assertion's message.

## Refutation — measured, not reasoned

Two mutants of `verify-sha.sh` make this assertion **the only one in the whole suite** that reddens: with it, 8/9; with it deleted, 9/9 green.

**M9 — the mundane one.** A plain "DRY the guard message" refactor:

```sh
fetch_msg="cannot fetch origin/$branch"
git fetch --quiet origin "$branch" || die "$fetch_msg"
tip=$(git rev-parse "origin/$branch") || die "$fetch_msg: origin/$branch does not resolve after fetch"
```

`tests 9 / pass 8 / fail 1`, sole failing site `verify-sha.test.mjs:207`. Delete that line → `pass 9 / fail 0`. Every assertion above it still passes: exit 2, stdout empty, `/origin\/main does not resolve after fetch/` still matches, `/ambiguous argument 'origin\/main'/` still matches. Only this assertion sees that the controller can no longer tell the two guards apart by grepping stderr.

**M5 — `die` printing a summary of passed guards.** Sole failure at the same line:

```
$ git fetch --quiet origin main
[...]

Issue #568, opened 2026-08-17
PR #563 review: the adjudication record — 5 applied, 1 refuted, 3 rejected on ruling, 3 deferred to #567, 1 survived to #569


feigi:
**Superseded in part — read the two CORRECTION blocks below before acting on this ticket.**

Record ticket for PR #563 (issue #362), applied at `217c5a3039d90a57d2318ab5a5881a71ddf9e251`, which **merged** at `b2e7f90` (rebased onto `main` as `8f43b79`) while two findings were still being checked. Gate: `./agent-test skills/fleet/scripts` → `tests 794 / pass 794 / fail 0`. `ledger.mjs`'s exit vocabulary (`0` clean / `1` already-filed / `2` usage / `3` tracker hit) re-verified by direct invocation against the live run ledger after the change — all four unchanged.

Dimension strings are named from each finding's content where the verbatim `dimension` did not reach the fix-applier; those are marked *(inferred)*.

## Applied

1. **`ledger.mjs` — the `=` spelling was unguarded, and it falsified the PR's own claim** (`survived1` types 2-0, `survived2` correctness 2-0, `refuted0` silent-failure 1-1 — one defect reported by three dimensions).

   `impl-362`'s report claimed `--file=x`/`--require-file=x` "already exit 2 loudly". Measured, that holds **only where the token lands in the subcommand slot** — `argv[0]` once the two splices have run. Everywhere else neither flag is refused: `indexOf` matches the exact token only, the unmatched one falls through to `rest`, and `check`'s `rest.join(" ")` folds it into the **subject**.

   | invocation | before | control (space-separated) |
   |---|---|---|
   | `check --file=<other-ledger> "<already-filed subject>"` | **exit 0**, `verdict: "unverified"`, flag folded into `subject` | exit 1 `ALREADY FILED` |
   | `--file <missing> check --require-file=true dup` | **exit 0** | exit 2 |

   So it lost the path *and* corrupted the checked text, and answered "safe to file" for a duplicate — the harm #362 exists to close. Fixed with two whole-argv guards at `ledger.mjs:55-56`, in the wordings `arg()` and `has()` already use for this spelling. Pinned in **both** positions (`ledger.test.mjs:809`, `:836`).

   Mutation receipts, on a scratch copy at the right tree depth (baseline 39/39):

   | mutant | pass | fail | which |
   |---|---|---|---|
   | drop the `--file=` guard | 38 | **1** | the new `--file=` test only |
[...]

Issue #562, opened 2026-08-17
REFUTED: don't drop the comment filter from prove-merge.test.mjs's `[ ]` guard


feigi:
Deferred from PR #556 review (closes #267). Dimension: `simplify`. Location: `skills/fleet/scripts/prove-merge.test.mjs:485-490`.

Filed as a **record, not as work** — `wontfix`. A future `code-simplifier` pass will re-flag this same inert `.filter`; this is the measured answer so nobody re-derives it.

## The finding

CLAIM: the comment-stripping filter never removes anything on the current `prove-merge.sh` — including the very comment #556 added — because the offender regex `/(^|\s)\[\s/` requires `[` at line start or after whitespace, and that comment writes it as `` `[ ]` `` (the `[` is preceded by a backtick).

```js
const offenders = readFileSync(SCRIPT, "utf8")
  .split("\n")
  .map((line, i) => [i + 1, line])
  .filter(([, line]) => !line.trimStart().startsWith("#"))       // <- claimed inert
  .filter(([, line]) => /(^|\s)\[\s/.test(line) && line.includes("$("));
```

SUGGESTED_FIX, two options: (A) drop the comment filter and fold the rest into one pass, or (B) merge both predicates into a single `.filter`.

## Refutation — measured, not reasoned

**The mechanical claim is correct.** Computing both offender sets over the pinned `prove-merge.sh` (125 lines, 46 of them comments): `WITH comment filter: []`, `WITHOUT comment filter: []`, `SETS IDENTICAL: true`. And `prove-merge.sh:65` — `` # and `[ ]` discards that status, so `set -e` never fires and a probe that could `` — fails **both** conjuncts, not just the regex: it has no `$(` on it at all. The `[ "$(...)" ]` spelling lives only in `prove-merge.test.mjs`, which the guard does not read.

**But the direction is wrong. Option A is a regression.** The filter guards a class, not that one comment, and this is the one file whose header comment block explains the `[ "$(...)" ]` idiom in prose (see `prove-merge.sh:38-39`, and the header at lines 7-9 / 19-22 which already uses indented code examples). Appending a realistic future comment in exactly that house style, then running the suite with the comment filter dropped:

```
✖ no probe in prove-merge.sh has its status discarded by `[ ]`
ℹ tests 17 / pass 16 / fail 1
    actual: [ [ 126, '#   [ "$(is_ancestor "$pre" "$base")" ] || die ...   <- WRONG, status discarded' ],
[...]

PR #576, opened 2026-08-18
fix(fleet): name both commands the at-risk guard's status can come from

### LINKED ISSUE #522: no-undo-audit.sh: the at-risk `|| die` blames git log for xargs' own failures
Deferred from PR #521 review. Dimension: **silent-failure** (severity `suggestion`). Out of #148's scope — #148 is the dedupe; this is the pre-existing guard the dedupe sits under, and the line is context in that diff, not part of it.

`skills/fleet/scripts/no-undo-audit.sh:472`

```sh
  at_risk=$(printf %sn "$conflicts" | sed 's/^/:(literal)/' | tr '\n' '\0' \
    | xargs -0 git -C "$wt" log --oneline "$fork".."$base" --) \
    || die "git log failed for the conflicting paths — cannot tell what a resolution would eat"
```

The status this reads is **xargs'**, not git log's. So an xargs-side fault — a single pathspec longer than the size limit is the reachable one — dies saying `git log failed`, sending the reader to git for a fault git never had.

**Nothing is swallowed**, which is why this is wording only and not a defect: the review measured the status mapping on macOS xargs and every child exit code collapses to 1 at the parent, so a real `git log` fatal (128) in any batch still reaches this `|| die`.

```
$ for code in 1 125 126 127 128 129 255; do printf 'a\0' | xargs -0 sh -c "exit $code"; done
xargs_status=1   # for every one
```

Fix: widen the wording to name the pipeline that actually produced the status, e.g.

```sh
    || die "listing commits for the conflicting paths failed (git log or xargs) — cannot tell what a resolution would eat"
```

The sibling `|| die` added on the dedupe statement below it in #521 already distinguishes itself by naming awk, so the two labels stay tellable apart once this one is widened.

Deferred from PR #521 review.

feigi:
`no-undo-audit.sh`'s at-risk `|| die` sits on a four-command pipeline, so the status it reads is `xargs`' — the pipeline's last command — not that of the `git log` xargs drives, and POSIX sh has no `pipefail` to change that. The message named git log alone, so an xargs-side fault died saying `git log failed`, sending the reader to git for a fault git never had.

Widened to name both commands that can produce the status the guard reads. Wording only: the line is a string literal inside `die`'s argument, so no control flow, no exit status and no `set -e` coverage moves.

**Settled against this tree, not inherited from the ticket:**

- An xargs-side fault reaches the guard, and used to be reported as git log's. Reproduced end to end in the suite (new test below); before it, stderr read `no-undo-audit: git log failed for the conflicting paths …` after `xargs: insufficient space for command`.
- A `git log` fatal still reaches the guard — nothing is swallowed. Probed on the pipeline's own shape: a bad revision range gives `fatal: bad revision` and fires the guard; a healthy run does not fire it.
- macOS `xargs` collapses every nonzero child code to 1 (re-measured here for 1, 2, 125, 126, 127, 128, 129, 255; 0 stays 0). GNU xargs maps child statuses to its own set instead — that is documented behaviour I could not settle from this machine (no working container runtime here), so nothing in the diff rests on it. **The new wording asserts nothing about any status mapping** — only about which two commands can be at fault, which holds on both. What is settled on GNU is the part that matters: this PR's CI run on `ubuntu-latest` executed the new test and it passed, so the guard fires there too and prints the widened message.
- The sibling `|| die` on the dedupe statement needs no widening: `awk` is the last command in *its* pipeline, so it names the command whose status it reads. Measured, not assumed.

**Coverage.** Instrumented the guard to record when it fires and ran `origin/main`'s suite: 66 tests, and the guard never fired once — this path had no coverage at all. The new test is the first thing to enter it. It kills its own mutant: reverting the wording reds it and only it (67 tests, 1 fail), and rewording the adjacent dedupe guard leaves the whole suite green.

Full suite: 811 tests, 811 pass.

Closes #522


feigi:
Review follow-ups filed. Nothing applied to this branch — the two in-scope `suggestion`-band findings were both refuted by measurement, and the two out-of-scope ones were already filed by the controller. HEAD unchanged at `d10728bd59e4de4dca795502a94d8a6fffb007a3`; `node --test skills/fleet/scripts/*.test.mjs` → **tests 811 / pass 811 / fail 0**.

| Finding | Dimension | Outcome | Issue |
|---|---|---|---|
| `no-undo-audit.sh:490` locale false-safe (`survived`) | silent-failure | deferred — pre-existing, out of #522's wording-only scope | #582 |
| `no-undo-audit.sh:524` unguarded pipeline stages | silent-failure | deferred — remedy is a design question | #583 |
| `no-undo-audit.test.mjs:188` `withSplitXargs` docstring gap | correctness | **refuted** — the sentence the fix asks for is already at the caller, `:828-829`, added by this same commit | #587 |
| `no-undo-audit.test.mjs:844` "dead `doesNotMatch`" | simplify | **refuted** — four mutants are green without it and red with it, `:844` the sole red (67/67 → 66/67) | #588 |

#587 and #588 are records, not work (`wontfix`), so the next review pass gets the measured answer instead of re-flagging. #588 is the fourth refutation of this claim class here, after #573, #562 and #568 item 6.

PR #577, opened 2026-08-18
fix(fleet): let git's own diagnosis reach the operator on prove-merge's guards

### LINKED ISSUE #566: prove-merge.sh:58 — cat-file swallows git's stderr and asserts "is not a commit in this repository" for causes that are not that
Found by `impl-559` while enumerating the class for #559, in the sibling script. Deliberately **not** widened into by that ticket — out of its scope, and PR #556 had just touched this file. Filed by the fleet controller, verified against `origin/main` before filing.

## The defect

`skills/fleet/scripts/prove-merge.sh:58`:

```sh
for obj in "$pre" "$post" "$merge"; do
  git cat-file -e "${obj}^{commit}" 2>/dev/null || die "$obj is not a commit in this repository"
done
```

Byte-identical in shape to `verify-sha.sh:28`, which **#559 / PR #565 just fixed**. Two problems, same line:

1. `2>/dev/null` discards git's own diagnosis, which is the only thing that distinguishes the causes.
2. The message is a **positive assertion that is false for at least one reachable input**. Measured during #559's work on the sibling: for a tree sha, git says `expected commit type, but the object dereferences to tree type` — the object **is** in the repository. "is not a commit in this repository" is flatly wrong there. It is also wrong for an unreadable object store, where git could not tell.

## Why this is a separate ticket

- **#267 owns `prove-merge.sh`'s verdict-collapse defect** (a probe failure reported as a disproof), fixed in PR #556. This is not that — it is the cause-assertion/stderr-swallow class.
- **#559 owns the same class in `verify-sha.sh`**, three sites, fixed in PR #565.
- This is the one remaining instance of that class in `skills/fleet/`. `impl-559` swept the tree: every other `2>/dev/null` in the fleet is a probe whose non-zero is an *expected answer*, not a `die` asserting a cause.

## Desired behavior

Mirror what PR #565 established for the sibling, rather than inventing a second convention:

- Drop `2>/dev/null` so git's line reaches the operator.
- Narrow the message to what the guard actually observed — `verify-sha.sh` landed on `cannot resolve $sha to a commit in this repository`.
- Exit vocabulary unchanged: this stays exit 2 ("could not answer"), and the design-spec row for `prove-merge.sh` needs no edit, because nothing about *when* exit 2 fires changes.

## The other half — do not skip it

PR #565's measurement is that the healthy path emits **0 bytes** of git stderr, so unmuting adds no noise. **Re-measure that for this script's own invocation** rather than inheriting it, and leave a quiet-path test asserting stderr **exactly** (deepEqual on the trace lines), not a `/fatal:/` sniff. A test asserting only exit 2 + empty stdout cannot tell this guard's failure from a `git fetch` failure upstream of it — that trap was measured in this same file family.

## Deliberately excluded, recorded so it is not re-derived

`docs/specs/2026-07-23-fleet-plugin-design.md:182` still phrases one `verify-sha.sh` exit-2 cause as "the sha is not a commit object in this repository" — the same wording PR #565 removed from the script. `impl-559` argued it is **still accurate as a description of _when_ exit 2 happens**, even though it is no longer the script's wording, and the suite's spec-row consistency test passes. That is a judgment call, not an obvious defect, so it is **out of scope here** and left for a maintainer to rule on. Do not fold it into this ticket without deciding it first.

feigi:
Closes #566

Two guards in `prove-merge.sh` discarded git's own line and then asserted a cause the guard could not know. Both are on the merge bot's proof path (`skills/fleet/commands/run-merge-bot.md:145`).

```diff
-git rev-parse --verify --quiet "$base" >/dev/null || die "$base does not resolve"
+git rev-parse --verify "$base" >/dev/null || die "$base does not resolve"

-  git cat-file -e "${obj}^{commit}" 2>/dev/null || die "$obj is not a commit in this repository"
+  git cat-file -e "${obj}^{commit}" || die "cannot resolve $obj to a commit in this repository"
```

Mirrors what PR #565 established in the sibling `verify-sha.sh`; no second convention invented.

## Claims settled by measurement, not inherited

Throwaway repo, git 2.50.1 (Apple Git-155).

| Claim | Command | Result |
|---|---|---|
| `2>/dev/null` hid the discriminator | `git cat-file -e <tree>^{commit}` | `error: …expected commit type, but the object dereferences to tree type` + `fatal: Not a valid object name` |
| …and an absent sha differs | `git cat-file -e <40 zeros>^{commit}` | `fatal: Not a valid object name` only — no `error:` line |
| `--quiet` hid rev-parse's reason | `git rev-parse --verify --quiet nosuchref` vs without | 0 bytes / exit 1 → 32 bytes / exit 128, `fatal: Needed a single revision` |
| **healthy path emits 0 bytes of git stderr** | full script run, fetch really transferring objects | **896 bytes / 13 lines, every one the script's own trace; 0 from git — identical before and after** |
| `--verify` must stay | `git rev-parse f.txt` | prints the path, **exit 0** — a `$base` naming a file would clear the guard without it |

The healthy-path figure was re-measured for this script's own invocation rather than inherited from PR #565.

## Deliberately unchanged

`git fetch --quiet origin` (the guard above these two) keeps `--quiet`. Measured: against a broken remote it still emits 196 bytes — `fatal: '/nonexistent/path.git' does not appear to be a git repository` — so `--quiet` suppresses progress, not diagnosis, and its `die` reports what it observed. Same call shape PR #565 kept in `verify-sha.sh`. The new quiet-path test now pins this: dropping `--quiet` from the fetch turns that test red.

The design-spec row is untouched — it describes *when* exit 2 fires, which does not change here.

## Tests

Three added; the existing 17 passed against both the muted and unmuted script, so none of them could see the mode this changes.

- a present-but-not-a-commit object is no longer reported as absent, and the absent sha no longer prints the identical line
- a base ref that does not resolve now carries `Needed a single revision` (real git, no shim — `BASE_REF` reaches the guard with the fetch above it passing)
- **healthy run stays quiet**, asserted with `deepEqual` on all 13 trace lines, not a `/fatal:/` sniff

Each failure case also asserts which guard did *not* fire. Every guard here produces exit 2 with empty stdout, so a bare exit-2 assertion cannot tell them apart — the trap already measured in `verify-sha.test.mjs`'s fetch cases.

**Mutation-tested.** Restoring `2>/dev/null`, restoring `--quiet`, restoring the old message, and un-quieting the fetch each turned exactly one test red — the right one. Renaming the loop variable (output byte-identical) left all 20 green, so the tests discriminate rather than merely react to edits.

`node --test skills/fleet/scripts/*.test.mjs` → **tests 813, pass 813, fail 0**.

## Out of scope, and worth a follow-up ticket

#566 states this was the last instance of the class in `skills/fleet/`. **That is not accurate** — five sibling guards still swallow git's reason the same way:

- `no-undo-audit.sh:237` — `rev-parse --verify --quiet "origin/$branch"` → `die "origin/$branch does not resolve — fetch it, or it was never pushed"`. The strongest one left: it swallows *and* asserts an unknowable two-item cause list, the same shape PR #565 deleted from `verify-sha.sh`.
- `no-undo-audit.sh:232`, `worktree-audit.sh:47`, `release-ticket.sh:61`, `reap.sh:42` — all byte-identical to the `$base` guard fixed here.

Left alone deliberately: they are four other scripts, and #566 itself records that #559 declined to widen into this file for exactly that reason.

## Why the rev-parse guard is in this diff at all

`git rev-parse --verify --quiet "$base"` → `git rev-parse --verify "$base"` (line 58) is **not** named in #566's body, which scopes itself to line 58's `cat-file` and asserts "This is the one remaining instance of that class in `skills/fleet/`". It was authorized separately, in a comment on #566 headed `## ADDITION — line 55 is the same class, and this ticket did not name it`: *"Fix both in the same pass — that is the whole point of the class sweep, and leaving one is the fixed-only-the-named-case failure this repo keeps producing."*

Recorded here because reading #566 with `--json title,body` shows the body only, and a reader of this PR alone would otherwise re-derive it as a scope violation. `--quiet` is a different spelling of the same mute, so `grep -n '2>/dev/null'` could never have caught it, and the block comment this diff copies from `verify-sha.sh:18` ("Nothing below silences git") would be false with `--quiet` still attached.

## Applied from the PR review

Three of seven review findings survived a refuter each and are in the last two commits; the other four are settled or filed.

- **Locale pinned.** The shared `ENV` fixed identity and config but not the locale, while two cases here assert git's own diagnostic text across three strings. `rev-parse`'s `Needed a single revision` is `die(_("..."))` at `builtin/rev-parse.c:587` and git ships a live (non-fuzzy) German msgstr at `po/de.po:12193` — compiled with `msgfmt`, it yields `fatal: Benötigte einen einzelnen Commit`, which the regex at test.mjs:530 does not match. CI is `ubuntu-latest` with no `LANG` pinned anywhere in `.github/workflows/`, so today's green rests on the runner image's default. `LANG`/`LC_ALL: "C"` is the pin `inflight.sh` already carries at five sites. (`cat-file`'s `Not a valid object name` and `object-name.c`'s `dereferences to %s type` are unwrapped today — the pin covers all three regardless.)
- **Dead fixture dropped** from both failure cases. `sh -x` traces with and without the merge are command-for-command identical, and every mutation each case exists to catch still reds on the same assertion with the same stderr bytes.
- **Clone routed through `git()`** in the healthy-run case, instead of re-deriving `repo()`'s private `origin.git` layout through a raw `execFileSync`. `git remote get-url origin` returns that same absolute path.

Not applied: the annotated-tag `^{commit}` peel mismatch is real, pre-existing and fails closed — filed separately as **#585**, out of #566's scope. Two further suggestions were refuted by measurement and filed with their refutations.

Re-verified after these changes: `node --test skills/fleet/scripts/*.test.mjs` → **tests 813, pass 813, fail 0**, and each of the three changed cases is still the *only* red on its own mutant (restored `--quiet`, restored `2>/dev/null`, un-quieted fetch), with all three staying green on a mutant none should catch.


feigi:
## Review adjudication — 7 findings, 3 applied, 4 deferred

All seven arrived in the `unverified` band at `severity=suggestion`, which the review workflow budgets **zero** refuters by policy — so none had been checked by anyone. Each in-scope one got a refuter briefed to refute, defaulting to `refuted=true` under uncertainty and required to settle it by running something.

| # | file:line | dim | outcome |
|---|---|---|---|
| `unverified-0` | `prove-merge.sh:58` | correctness | **Not a defect** — the rev-parse `--quiet` drop was authorized by #566's `## ADDITION` comment. Recorded in the PR body; no code change. |
| `unverified-1` | `prove-merge.test.mjs:535` | correctness | **Refuted** → #596 |
| `unverified-2` | `prove-merge.sh:76` | silent-failure | Out of scope, real, fails closed → #585 |
| `unverified-3` | `prove-merge.test.mjs:25` | types | **Survived** → applied (`LANG`/`LC_ALL: "C"`) |
| `unverified-4` | `prove-merge.test.mjs:482` | simplify | **Survived** → applied (dead fixture dropped) |
| `unverified-5` | `prove-merge.test.mjs:521` | simplify | **Refuted** → #597 |
| `unverified-6` | `prove-merge.test.mjs:541` | simplify | **Survived** → applied (clone via `git()`) |

### What the refuters actually tried

- **#596** (`unverified-1`): reproduced the ambiguous-`$base` warning end to end (rc=0, 4 `warning: refname` lines), then restored `--quiet` in a scratch copy — the pre-PR script emits **3 of those 4** from the *untouched* `merge-base --is-ancestor` calls, and fails the same `deepEqual` under the same forced-ambiguity mutant. The fixture cannot reach ambiguity at all (`BASE_REF: undefined` → `origin/main`; `repo()` only creates `main`/`feat`).
- **#597** (`unverified-5`): counted **three** raw `spawnSync` env-override blocks (331, 453, 521), not two; showed `stdout.trim() === ""` is not `json === null` (injected non-JSON stdout throws `SyntaxError` inside `prove()` before any named assertion runs); and found the finding's own evidence does not compile — collapsing the pre-existing block at 453 gives `SyntaxError: Identifier 'code' has already been declared` against line 473.
- `unverified-3`: found `Needed a single revision` is `die(_("..."))` at `builtin/rev-parse.c:587` with a live German msgstr at `po/de.po:12193`; compiled the catalog with `msgfmt` and produced `fatal: Benötigte einen einzelnen Commit`, which the regex does not match.
- `unverified-4`: settled by mutation rather than by reading, per the "deleting setup can make a test pass for a different reason" rule — every mutation each case exists to catch still reds on the same assertion with byte-identical stderr, and `sh -x` traces of both fixtures are command-for-command identical.
- `unverified-6`: confirmed `git remote get-url origin` returns the same absolute path the hardcoded `join(w, "..", "origin.git")` built, and that the fetch still transfers real objects (reachable objects 2 → 4).

### Gate

`node --test skills/fleet/scripts/*.test.mjs` → **tests 813, pass 813, fail 0**, run from the worktree, no `--no-verify`.

Combined mutation matrix on the post-change tree — each changed case is the **only** red on its own mutant, and all three stay green on a mutant none should catch:

| mutant | red |
|---|---|
| `--quiet` restored on `rev-parse --verify` | `a base ref that does not resolve carries git's own cause` (19/20) |
| `2>/dev/null` restored on `cat-file -e` | `an object that is present but is not a commit is not reported as absent` (19/20) |
| fetch un-quieted | `a healthy run stays quiet` (19/20) |
| unrelated die wording (`has no second parent`) | `a non-merge commit as <merge-commit> is a usage error` — none of the three (19/20) |

PR #598, opened 2026-08-18
fix(fleet): let a broken agent sidecar reach stderr instead of booking "other"

### LINKED ISSUE #325: board.mjs:114 — empty catch labelled /* unnamed agent */ swallows genuine meta.json faults, silently booking that agent as "other"
**dimension:** silent-failure
**Deferred from PR #321 review** (which closes #318).

**Pre-existing, and NOT caused by PR #321.** `board.mjs` is not in that PR's diff at all — the diff is `compute-spend.mjs` + `compute-spend.test.mjs` only. Filed here because the review surfaced it, not because the PR introduced it.

## The defect

`skills/fleet/scripts/board.mjs:114`

```js
try { if (existsSync(metaFile)) meta = JSON.parse(readFileSync(metaFile, "utf8")); } catch { /* unnamed agent */ }
```

The comment claims the catch exists for the unnamed-agent case, but `existsSync(metaFile)` **already handles that** — a missing meta file never enters the `try` body. So the catch can only fire on a genuine fault:

- a `meta.json` read truncated mid-write
- `EACCES` / `EISDIR`
- a TOCTOU delete between `existsSync` and `readFileSync`

Each of those leaves `meta = {}`, which flows into `classifyRole({})` → `"other"`, silently reclassifying that agent's **entire spend** with zero bytes on stderr. The spend panel's role percentages — `reviewPct` in particular, the number `compute-spend.mjs`'s own comments say people act on — are then wrong with no signal that anything failed.

The `/* unnamed agent */` comment is independently a defect: it asserts a reason for the catch that the code above it has already made impossible, so a future reader treats a real-fault swallow as intended behaviour.

## Verification status

Two refuters were dispatched and split 1-1, **on different claims**:

- **`refuted=false`** — could not refute the mechanism. Reproduced it: the `existsSync` guard already covers the unnamed-agent case, so the catch is reachable only on real faults. A truncated `meta.json` yields `{}` → `classifyRole` → `"other"` with 0 bytes on stderr, while the intact file yields `"finisher"`.
- **`refuted=true`** — refuted only the **scope** half, i.e. "PR #321 increases exposure to this". It does not; the swallow's consequence is identical pre- and post-diff.

So the underlying defect is **real and unrefuted**; what is correctly refuted is only the claim that this diff makes it worse. Hence: filed standalone, not applied to #321.

## Suggested remedy

Log the fault instead of swallowing it, and fix the comment. The unnamed-agent path is the `existsSync` guard, not the catch — so anything reaching the catch is worth a line on stderr, and the agent should not be silently booked as `other` without one.

*Deferred from PR #321 review*

feigi:
Closes #325

## The defect, re-derived against `origin/main`

The ticket's line number had drifted; the code is `skills/fleet/scripts/board.mjs:200` on `origin/main` (`2653729`), inside `readAgent`:

```js
try { if (existsSync(metaFile)) meta = JSON.parse(readFileSync(metaFile, "utf8")); } catch { /* unnamed agent */ }
```

Each claim the diff restates was settled against the tree, not taken from the body:

- **`existsSync(metaFile)` already covers the unnamed-agent case.** A missing sidecar never enters the `try`; `meta` stays `{}` without throwing.
- **So the catch is reachable only on a real fault** — a sidecar read torn mid-write, `EACCES`/`EISDIR`, or a delete racing the `existsSync`.
- **`meta = {}` flows to `classifyRole({})` → `"other"`.** `gatherSpend` passes `a.meta` to `classifyRole` at `board.mjs:318`; `classifyRole({}) === "other"` is already pinned at `compute-spend.test.mjs:147`.
- **`reviewPct` is the number `compute-spend.mjs`'s own comments say people act on.** Its comments call the review headline "the one number anyone acts on" at `compute-spend.mjs:34` and `:47-48`, and `reviewPct` at `:108` is that headline.

Measured on a two-agent fixture against the unfixed tree — an 800-token reviewer and a 200-token implementer:

| reviewer sidecar | `reviewPct` | roles | stderr |
|---|---|---|---|
| intact | 80 | `reviewer:800 implementer:200` | 0 bytes |
| truncated mid-write | 0 | `other:800 implementer:200` | 0 bytes |

## The fix

Warn once per sidecar path, and keep the `{}` fallback.

Rethrowing was the tempting alternative and is wrong: the fault is in the sidecar, but the **transcript** is still readable, so a throw lands in `gatherSpend`'s per-file catch and drops the agent's real tokens from the totals while counting it as `skipped` — a wrong total in place of a wrong role. That mutant is exercised below.

Warn-once keyed on the full path follows the rule `warnedSkips` already sets two functions down: `serve` rebuilds every ~15s, and a broken sidecar is broken on every tick. Verified — three `gatherSpend` calls over one broken sidecar emit exactly one line.

The misleading `/* unnamed agent */` label is replaced by a comment stating what actually reaches the catch.

`board.mjs` runs as both a one-shot CLI (`build`) and a long-lived server (`serve`), so the new stderr output was checked against both:

- **stdout is untouched.** `build` writes the model with `console.log`; the warning goes to `console.error`. Confirmed a broken sidecar leaves stdout parsing as clean JSON.
- **`serve` cannot be flooded** — warn-once, verified over repeated ticks. The `Set` is keyed on sidecar paths within one session directory, the same bound `warnedSkips` already carries.
- **No pipe-truncation hazard.** `board.mjs`'s only `process.exit(0)` calls are in `serve`'s SIGINT shutdown (`board.mjs:528,530`); nothing exits alongside this write.
- **The file-header invariant holds** — spend telemetry "can only ever populate or omit `spend`, never change a ticket's stage". No field added, no control flow changed.

## Coverage before this PR

Nothing in the suite reached that catch. Proven, not assumed: restoring the empty catch on the full 828-test suite fails **exactly one** test — the one added here.

## Tests

Both added tests were mutation-checked against a scratch copy; each mutant is killed by exactly one test, and the other stays green:

| mutant | result |
|---|---|
| restore the empty `catch { /* unnamed agent */ }` | kills the fault test only |
| `throw e` after warning (drops the agent's spend) | kills the fault test only |
| drop the `existsSync` guard so an absent sidecar warns | kills the silence test only |
| reword the warning, keep the path | **both stay green** — behaviour is pinned, wording is not |

The second test is the false-positive half the fix could otherwise break: a genuinely **absent** `meta.json` — the real unnamed-agent case — must stay silent, or every controller-dispatched agent without a sidecar prints a line every tick.

## Enumerated and deliberately left

`readAgent` is the only sidecar reader in the repo, so there is no sibling caller still swallowing this.

Among `board.mjs`'s other silent catches, none share this ticket's shape — a label naming a cause a guard above has already made unreachable — and each is left, named:

- `mapCi`'s `catch { return "unknown"; }` (`board.mjs:113`) — silent, but no false label, and `unknown` is a documented safe value rather than a confidently wrong one. It is pipeline-stage data, not side-channel telemetry, so its blast radius is different. Separate ticket.
- `newestTranscriptMs`'s `catch { return 0; }` (`board.mjs:173`) — its comment ("a dir we cannot read simply loses") describes exactly what happens.
- the per-file `catch { /* raced away */ }` (`board.mjs:177`) — no `existsSync` precedes it, so a raced-away name really is the expected cause; the label is accurate.
- the per-line `catch { continue; }` (`board.mjs:230`) — a transcript appended to while being read has a torn last line on every tick, so warning here would be pure noise. Over-broad in that a mid-file corrupt line is skipped the same way; separate ticket.
- a sidecar holding valid JSON that is not an object (`null`, `5`) never reaches the catch at all — the read succeeded. `null` already fails loudly through the existing per-file catch; catching a wrong *shape* needs a validity check, not a fault report. Out of scope.

## Verification status of the ticket

The two refuters split 1-1 **on different claims**. The mechanism was not refuted and is reproduced above; what was correctly refuted is only the scope claim that PR #321 increases exposure to it. The defect itself is not contested.

## Test result

`./agent-test .` — 828 tests, 828 pass, 0 fail.


feigi:
Review findings adjudicated. Applied in `e4b07d4`; deferrals filed as #602, #603, #604.

**Applied**
- **Shape guard** (silent-failure + types + correctness, one defect from three dimensions): `JSON.parse` succeeds on `null`/number/string/array, so `meta` became that value and the fault surfaced one frame later at `a.meta.description` — inside the per-file catch meant for unreadable *transcripts*. Measured: tokens left the totals, the agent was counted `skipped`, and stderr named `agent-x.jsonl` for a fault in `agent-x.meta.json` — exactly what this PR's new comment says the `{}` fallback prevents. One-line shape check makes that comment true.
- `unreadable` -> `unusable` in the warn line: a `null` sidecar reads fine, and naming the fault correctly is the point of #325.
- **Warn-once test** (silent-failure + tests + types, one defect from three dimensions): `warnedMeta` was unpinned — both existing tests call `gatherSpend` once, and one call cannot see a dedupe.
- **Test simplification** (simplify): dropped the redundant local + booking assertion from the "stays silent" test. Survived its refuter, which measured three separate silent mutants and confirmed the suite catches all of them without it.

**Deferred**
- #602 — meta faults never reach `board.json`/`board.html`, so `reviewPct` stays silently wrong on the page the operator watches. Defect confirmed by two independent probes; the remedy (new field + renderer) is a larger PR than #325's contract.
- #603 — collapse `warnedMeta`/`warnedNoSpendDir`/`warnedSkips` into one `warnOnce` helper. Would invalidate the warn-once test added here, and overruns a correction ticket's stated diff size.
- #604 — `board.mjs:199-200` "never enters the try" is imprecise. Claim holds; the review's own reword was **refuted** — it drops the scope word and becomes false against `readFileSync` at :242. Remedy left open.

`node --test skills/fleet/scripts/*.test.mjs` -> **tests 830, pass 830, fail 0**. Both new tests mutation-verified: each kills its own mutant and only its own (830/829/1 in both directions), and both stay green under a behaviour-preserving reword of the two message strings.

Issue #583, opened 2026-08-18
no-undo-audit.sh:490 and :524 — pipeline stages ahead of the last one have their status discarded, so a fault under-reports at exit 0


feigi:
Deferred from the PR #576 review (`silent-failure`, severity `suggestion`, so **unrefuted** — the review budgets `suggestion` zero refuters by policy), plus the sibling half of the `important` finding filed as the locale ticket. Pre-existing; PR #576 only reworded one `die` message.

Filed by the fleet controller. Line citations **verified against `origin/main`**.

## The class

In a POSIX shell a pipeline's exit status is its **last** command's. `no-undo-audit.sh` builds two of its answers from multi-stage pipelines, so a fault in any stage **but the last** is invisible: `set -e` cannot see it, and no `|| die` can either. The stage yields empty output, the pipeline exits 0, and the script reports a **smaller** answer with full confidence.

This file's stated standard is that an unanswerable question is exit 2 and never a payload. These two lines are where that standard is unenforceable as written.

## Site 1 — the conflicts pipeline has no guard at all

`no-undo-audit.sh:490`:

```sh
conflicts=$(tr '\n' '\001' <"$mt_out" | tr '\0' '\n' | awk 'NR==1{next} /^$/{exit} {print}')
```

No `|| die` of any kind. A failure in either `tr` truncates or empties `conflicts`, and the audit proceeds to report on whatever survived. (The locale bug filed separately is *one* way to make the first `tr` fail; this ticket is the structural hole that let it pass silently, and it remains after the locale is pinned.)

## Site 2 — the at-risk guard can only answer for its last stage

`no-undo-audit.sh:~524`, the guard PR #576 just reworded:

```sh
at_risk=$(printf %s\n "$conflicts" | sed 's/^/:(literal)/' | tr '\n' '\0' \
  | xargs -0 git -C "$wt" log --oneline "$fork".."$base" --) \
  || die "listing commits for the conflicting paths failed (git log or xargs) — cannot tell what a resolution would eat"
```

PR #576 correctly widened the message to name both commands that can produce the status it reads. The residual is that `printf`, `sed` and `tr` sit **ahead** of `xargs`, and their statuses are discarded — so a fault in one under-reports `at_risk` at exit 0 rather than firing the guard. The new wording is accurate about what it *can* report; it does not and cannot cover the earlier stages.

## Why this is `needs-triage`, not `ready-for-agent`

The defect is not in question. The remedy is, and it is a real design choice:

[...]

PR #599, opened 2026-08-18
fix(fleet): pin the locale, so an invalid-UTF-8 byte cannot forge a safe rebase verdict

### LINKED ISSUE #582: no-undo-audit.sh:490 — an invalid-UTF-8 byte in a conflicting path makes tr exit 1 under the ambient UTF-8 locale, and the audit reports a FALSE SAFE at exit 0
Deferred from the PR #576 review (`silent-failure`, severity `important`). **Survived adversarial refutation 2-0** — both refuters rebuilt the fixture independently and reproduced it end to end against the unmodified script. **Pre-existing; not introduced by #576**, whose diff touches only the at-risk `die` message and the test file.

Filed by the fleet controller. The static claims below are **verified against `origin/main`**; the runtime measurements are the review's.

## The defect

`skills/fleet/scripts/no-undo-audit.sh:490`:

```sh
conflicts=$(tr '\n' '\001' <"$mt_out" | tr '\0' '\n' | awk 'NR==1{next} /^$/{exit} {print}')
```

`tr` is locale-sensitive. Under a UTF-8 locale it **exits 1 on a byte that is not valid UTF-8**. The pipeline's status is `awk`'s, so that failure is invisible to `set -e` and to every guard — and the line carries no `|| die` of its own.

Consequence: a conflicting path containing such a byte **truncates `conflicts`**, and the audit reports a **false safe**.

## Measured

Fixture: branch and main both add `b\377ad.txt` (via `git update-index --add --cacheinfo`, so the filesystem never has to hold the name) **and** `plain.txt`; main's later commit is `MAIN COMMIT AT RISK`. `git merge-tree --write-tree --name-only -z` reports both paths as conflicting (`od -c` shows `b\377ad.txt\0plain.txt\0\0`).

Unmodified script, ambient-style locale:

```
env -i PATH=... LC_ALL=en_US.UTF-8 sh no-undo-audit.sh <fixture>/w fix/1-thing
EXIT=0
stdout: {... "conflicts":["b"],"conflictsRewritten":[false],"atRisk":[],"atRiskRewritten":[]}
stderr: tr: Illegal byte sequence
```

Same fixture, same script, **only the locale changed**:

```
env -i PATH=... LC_ALL=C sh no-undo-audit.sh <fixture>/w fix/1-thing
stdout: {... "conflicts":["b\357ad.txt","plain.txt"],"atRisk":["7cd3bf6 MAIN COMMIT AT RISK"] ...}
```

So under UTF-8 the audit:

- exits **0**, which means *safe*,
- **drops `plain.txt` from `conflicts[]` entirely**,
- names a path `b` that **exists nowhere**,
- positively asserts `conflictsRewritten:[false]` — *"this IS the real path"*,
- and reports **nothing at risk**, on a branch where a main commit really would be eaten.

That is precisely the *"false safe … worse than what this replaced"* outcome the script's own comment at lines 480-489 says must be **exit 2**.

The failing configuration is the one the fleet actually runs under: the operator's ambient environment is `LANG=en_US.UTF-8` with `LC_ALL` unset.

## The fix — one line, measured

`export LC_ALL=C` immediately after `set -eu` (`no-undo-audit.sh:10`).

Verified by the review: with only that line added and `LC_ALL=en_US.UTF-8` still in the caller's environment, the same fixture gives `"conflicts":["b\357ad.txt","plain.txt"]` and `"atRisk":["7cd3bf6 MAIN COMMIT AT RISK"]` — the correct answer restored. Regression check on the patched copy: `node --test skills/fleet/scripts/no-undo-audit.test.mjs` → **tests 67, pass 67, fail 0**.

## The fleet already has this convention — this file is the exception

Verified on `origin/main`: `no-undo-audit.sh` contains **no `LC_ALL` at all**, while `inflight.sh` pins it at five sites, including the byte-identical `tr` pattern:

```
inflight.sh:578  if ! worktrees=$(LC_ALL=C tr '\n\000' '\001\n' <"$wtfile"); then
inflight.sh:367,450,593,648  LC_ALL=C awk ...
```

So this is a missed application of an established in-repo convention, not a new idea. **Sweep the other fleet scripts for the same gap as part of this ticket** rather than fixing one site.

## Reachability — stated plainly, not hidden

APFS refuses such a filename locally (`touch $(printf 'b\377ad.txt')` → `illegal byte sequence`, status 1), so on the fleet's own platform the path can only arrive **inside a fetched tree** — a Linux- or latin-1-authored commit — never from local disk. Rare, and it needs a hostile-ish filename. It is filed at this severity anyway because the failure mode is a *silent false safe on the tool whose entire job is to say whether a rebase would eat commits*, and because the remedy is one line with no measured cost.

Worth noting: the comment at `no-undo-audit.sh:488-489` currently reads *"a restructure bought for a filename shape nobody has produced. Upgrade there if one ever turns up."* One has now been produced. Whoever takes this should decide whether that comment still stands.

## Acceptance criteria

- [ ] A conflicting path with an invalid-UTF-8 byte no longer truncates `conflicts[]` under an ambient UTF-8 locale
- [ ] That case reports the at-risk commit rather than `atRisk: []` at exit 0
- [ ] A test builds the fixture the way the review did — `update-index --add --cacheinfo` plus sparse-checkout — so it needs no filesystem that can hold the name, and runs on macOS and Linux alike
- [ ] The test fails if the locale pin is removed (mutate it and confirm)
- [ ] Other fleet scripts are swept for the same unpinned-locale gap; those covered and those deliberately left are named

## Out of scope

- The **unguarded pipeline stages** in the same file — filed separately; that is about statuses being discarded, this is about the locale. A guard would not have made `tr` succeed.
- Everything else PR #576 owns (#522, the at-risk `die` wording).

feigi:
Closes #582

## The defect, reproduced

Rebuilt the review's fixture independently and reproduced it end to end against the **unmodified** script. `merge-tree` reports both paths (`od -c` → `b\377ad.txt\0plain.txt\0\0`), and then:

```
LC_ALL=en_US.UTF-8   {... "conflicts":["b"],"conflictsRewritten":[false],"atRisk":[]}   EXIT=0
                     stderr: tr: Illegal byte sequence
LC_ALL=C             {... "conflicts":["b\377ad.txt","plain.txt"],"atRisk":["2de1267 MAIN COMMIT AT RISK"]}
```

Only the locale differs. Under UTF-8 the audit drops `plain.txt` from the list entirely, names a path `b` that exists nowhere, positively asserts `conflictsRewritten:[false]` — *"this IS the real path"* — and reports nothing at risk on a branch where a main commit really would be eaten. Exit 0 means safe.

`tr` is locale-sensitive; BSD tr exits 1 on a byte that is not valid UTF-8. In `tr | tr | awk` it is the **first** stage that fails — measured, PIPESTATUS `1 0 0`, with the second `tr` and `awk` both exiting 0 on the short input it hands them — so the real status sits in a non-final slot, where `set -e` never sees it (the pipeline's status is `awk`'s) and the line carries no `|| die`.

After the pin, same fixture, same ambient `LC_ALL=en_US.UTF-8`: `conflicts:["b\377ad.txt","plain.txt"]`, `atRisk:["2de1267 MAIN COMMIT AT RISK"]`, and no `tr` error.

## Enumerate and declare

**Covered — pinned:**

Inventories below are measured per file with comment lines excluded, and each names the tool that is actually exposed. `awk` is **not** one of them: fed the bad byte it is byte-identical under `en_US.UTF-8` and `C`. The four that are: `tr` and `sed` exit 1 (`sed` emitting nothing at all), `grep` silently drops the line holding the byte, and `paste -sd, -` truncates its whole output at the byte while still exiting 0.

| script | tools it runs | what is exposed |
|---|---|---|
| `no-undo-audit.sh` | `tr`, `sed`, `awk`, `paste` | the measured site: `tr \| tr \| awk` over `merge-tree --name-only -z`; also `sed` and `paste` inside jstr/jarr |
| `release-ticket.sh` | `tr`, `sed`, `awk`, `grep` | `grep` over `git worktree list --porcelain`, `git cherry`, the dirty list and gh's labels; `tr` on git's error text; jstr's `sed \| tr` |
| `inflight.sh` | `tr`, `sed`, `awk` (**no `grep`**, deliberately) | `tr` splitting `worktree list --porcelain -z` and flattening gh's error text; jstr's `sed \| tr`. Completes a pin it had already started at five sites |
| `reap.sh` | `tr`, `awk`, `grep`, `paste` (**no `sed`**) | `tr` and `grep` over `git cherry`'s output — commit subjects, which git constrains to no encoding at all; `paste` joining `status --ignored` names |
| `worktree-audit.sh` | `awk`, `paste` (**no `tr`, `sed` or `grep`**) | `paste -sd, -` over `status --porcelain` names. `awk` is immune, so `paste` is the whole reason the pin is load-bearing here — and it truncates silently, at exit 0 |
| `derive-testcmd.sh` | `grep` only | `grep -qE "$testfile_re"` over the changed file paths. Measured: with this regex, a three-name listing whose middle entry is `b\377ad.test.mjs` matches 2 under `en_US.UTF-8` and 3 under `C` |

**Deliberately left, with reason:**

- `prove-merge.sh`, `verify-sha.sh` — no locale-sensitive text tool at all, `printf` only; refs are constrained by `check-ref-format`.
- `drop-merged-label.sh` — its one `grep -qx` reads gh-emitted label text, valid UTF-8 by construction.
- `claim-ticket.sh` — **a real sibling this PR does not close.** Its own execution has no exposed site; the `grep -E` and `sed` over `find` output live inside the `agent-test` runner it *generates* (escaped heredoc), so a pin on this script would not reach them. Same defect class — a test file whose name holds a bad byte drops out of the run silently, another false green — but pinning inside the heredoc changes every member's test runner, which is wider than this ticket authorises. Worth its own ticket.

## What `export LC_ALL=C` newly does

`LC_ALL=C` changes collation and case-folding for every `sort`, `awk`, `grep` and `tr` in whatever script it lands in. Measured across all ten fleet shell scripts: **zero** `sort`, **zero** case-folding (`grep -i`, `tr a-z`, `toupper`/`tolower`, `[:upper:]`/`[:lower:]`), **zero** `[a-z]`/`[A-Z]` collation ranges.

**One POSIX character class, not zero** — this PR originally claimed zero here and in all six script headers, and that was wrong. `no-undo-audit.sh:240` is `back=${back%"${back##*[![:space:]]}"}`, in the worktree-linkage check. The class is genuinely locale-sensitive: measured with the script's own `#!/bin/sh`, a trailing NBSP is stripped under `en_US.UTF-8` and kept under `C`. It is unreachable here — a `gitdir` file always ends `.git` and `$(cat …)` has already eaten the newline, so the last non-space byte is always `t` — so the pin is still safe, but the measurement as stated was not. `no-undo-audit.sh`'s header now says "one, unreachable"; the other five scripts' copies of that sentence say "none" and are accurate as written. The only `\b` occurrences are in sed *replacement* text producing literal `\b` for JSON, which the scripts' own comments already call out. So the two things the pin otherwise changes have nothing in these scripts to act on.

On my own edit: it is an assignment placed after `set -eu`, never the last command, so no script's exit status moves; it reorders nothing and wraps nothing.

## What this could wrongly REFUSE

The false-positive class is the other half, and a suite fed only invalid input pins neither direction. Second test added: a conflicting path of **valid multi-byte UTF-8** (`caf\303\251.txt`) must round-trip byte for byte and stay `conflictsRewritten:[false]`. Every scrub set in the script is `\001-\037` and every byte of a multi-byte UTF-8 sequence is `>= \200`, so the pin cannot reach it — this test is what keeps that true.

## Test construction

The path is spelled as a `printf` **format** and produced by the shell. Node re-encodes every JS string as UTF-8, so a JS `"\xFF"` would arrive as `\303\277` — valid UTF-8, reproducing nothing. Commits go in through a throwaway `GIT_INDEX_FILE` and are pushed straight to origin, so neither the index nor the filesystem ever holds a name APFS refuses; the audit reads only `origin/main` and `origin/<branch>` and compares neither against local HEAD, so no sparse-checkout is needed to keep `status --porcelain` clean. The locale goes in as `LANG`, with `LC_ALL` explicitly unset — see **Review fixes** below for why the first shape did not pin anything.

**Platform ceiling, stated not hidden:** BSD tr (macOS, the fleet's own platform) rejects the byte; GNU tr is byte-oriented and accepts it, so on Linux the unpatched script already answers correctly and the test passes either way. It asserts the correct *answer*, right on both; the macOS run is what kills the mutant.

## Mutation verified

Re-run after the review fixes, on a `git archive` copy under scratch (the worktree was never mutated; restores used `cp`):

| mutant | result |
|---|---|
| delete `export LC_ALL=C` from `no-undo-audit.sh` | RED — the behavioural test **and** its source assertion |
| drop only the `export` keyword (`LC_ALL=C`) | RED — same two. Under the PR's original test env this survived 69/69 |
| delete the pin from `worktree-audit.sh` | RED — that script's source assertion, and nothing else |
| move `reap.sh`'s pin to the bottom of the file | RED — the prologue-placement assertion |
| `export LC_ALL=C LANG=C` (strictly stronger) | **GREEN**, as it must be |
| reflow an unrelated header comment | **GREEN** |
| unrelated change (`at risk:` → `AT RISK:` stderr prefix) | both new tests **green** while three others go red |

Mutations run on copies under a scratch directory; the worktree was never mutated, and restores used `cp`.

## The `:488-489` ceiling comment

The ticket asked whether *"a restructure bought for a filename shape nobody has produced"* still stands. **It does** — the shape it names is a filename holding a literal **newline**, and that one still has not been produced. #582 produced a *different* shape, an invalid-UTF-8 byte, which needs no restructure at all. The sentence is now explicit about which shape it means, because its ambiguity already misled one reader: the ticket itself read it as covering this case.

## Tests

`node --test skills/fleet/scripts/*.test.mjs` → **tests 834, pass 834, fail 0** (826 before; +2 behavioural, +6 source assertions). `no-undo-audit.test.mjs` alone → 69/69, up from the ticket's 67.

## Review fixes

A multi-specialist review of this PR found two defects in it, both now fixed.

**1. The 13-line justifying comment was pasted into five scripts and was false in four.** Its opening sentence claimed "every `tr`, `sed`, `awk` and `grep` below" everywhere; measured, only `release-ticket.sh` runs all four. Each header now carries its own measured inventory (see the table above), the mid-sentence line break the templating left behind is gone, and `no-undo-audit.sh`'s two further errors — the "second-to-last slot" and the "zero POSIX classes" claims — are corrected here and in its header.

**2. The regression test did not pin the fix it exists to protect.** Both new tests ran under `{ ...ENV, LC_ALL: "en_US.UTF-8" }`, which puts `LC_ALL` into the child's environment *before* the script starts. A POSIX shell keeps a variable's export attribute once it is already there, so a mutant dropping just the `export` keyword still reached `tr`, and both tests stayed green on it — while that same mutant under an ambient `LANG=en_US.UTF-8` reproduces the exact `"conflicts":["b"]` / `atRisk: []` false safe this PR exists to fix. Both tests now use `{ ...ENV, LANG: "en_US.UTF-8", LC_ALL: undefined }`, the suite's own idiom and also the real ambient shape.

**Plus the half CI could never run.** GNU `tr`, `sed`, `paste` and `awk` all pass the byte through — measured with coreutils 9.11, gnu-sed 4.10 and gawk 5.4.1 ahead of PATH, the whole of `no-undo-audit.test.mjs` stays green with the pin deleted. `ci.yml` runs ubuntu-latest, so on the only platform CI uses, deleting any of the six pins was invisible. `locale-pin-prose.test.mjs` adds one source assertion per pinned script, pinning presence and prologue placement.

**Coverage asymmetry, stated rather than hidden:** only `no-undo-audit.sh` has a behavioural regression fixture. The other five now have a source assertion, which kills the "redundant export, drop it" cleanup, but not a fixture driving each script's own inputs through a bad byte. That remainder is filed rather than done here.

Two test-fixture lines also came out: `git fetch -q origin` updated no ref the audit reads, and `rm -f "$idx"` insured against a state the next `read-tree` already makes unreachable (verified by comparing `ls-tree -r` output, not by test colour).


feigi:
## Review fixes applied — `e93c6be`

`node --test skills/fleet/scripts/*.test.mjs` → **tests 834, pass 834, fail 0** (828 before; +6 source assertions).

**Applied (3 commits):**

- `a7ff769` — each pinned script's header now names the tools it actually runs. The pasted "every `tr`, `sed`, `awk` and `grep` below" was false in four of five; only `release-ticket.sh` ran all four. Also corrects `no-undo-audit.sh`'s "second-to-last slot" (PIPESTATUS is `1 0 0` — it is the **first** stage) and its "zero POSIX classes" claim (there is one, at `:240`, locale-sensitive but unreachable), and removes the templating's mid-sentence line break from all five.
- `dcd3297` — the two locale tests now run under `{ ...ENV, LANG: "en_US.UTF-8", LC_ALL: undefined }`. The old `LC_ALL: "en_US.UTF-8"` pre-seeded the variable into the child environment, and a POSIX shell keeps the export attribute once it is already there — so a mutant dropping just the `export` keyword stayed green, and **the fix for #582 was unpinned by its own test.** Also drops two dead fixture lines (`git fetch -q origin`, `rm -f "$idx"`) and collapses two `--cacheinfo` calls into one.
- `e93c6be` — `locale-pin-prose.test.mjs`, one source assertion per pinned script. GNU `tr`, `sed`, `paste` and `awk` all pass the byte through, so on `ubuntu-latest` — the only platform `ci.yml` runs — deleting any of the six pins was invisible. Pins presence and prologue placement.

**Mutation-verified** on a `git archive` copy under scratch; the worktree was never mutated.

| mutant | result |
|---|---|
| delete `export LC_ALL=C` (`no-undo-audit.sh`) | RED — behavioural test + source assertion |
| drop only the `export` keyword | RED — same two. **Survived 69/69 under the old test env** |
| delete the pin from `worktree-audit.sh` | RED — that script's assertion, nothing else |
| move `reap.sh`'s pin to the bottom of the file | RED — the placement half |
| `export LC_ALL=C LANG=C` (strictly stronger) | GREEN |
| reflow an unrelated header comment | GREEN |

**Deferred and filed:**

| # | finding | why deferred |
|---|---|---|
| #611 | drop `inflight.sh`'s five per-site `LC_ALL=C` prefixes | ruled a **strict weakening** — they are depth for the moment the global pin is moved or removed |
| #612 | make the `LC_ALL=C` safety precondition an executable check, not prose in six files | well-motivated (its predicted failure already happened — see the POSIX-class correction above), remedy still open |
| #613 | payload carries an invalid UTF-8 byte while asserting `conflictsRewritten:false` | out of #582's scope; a `conflictsRewritten` contract change |
| #614 | behavioural fixtures for the other five pins | the cheap half shipped here; per-script value is uneven and is a judgement call |
| #583 (comment) | residual pipeline-status swallow beyond the invalid-UTF-8 trigger | already owned by an open ticket. Added: the position correction, and that **`pipefail` is not the remedy** — a healthy 200-path run already ends `141 141 0 0`, so it would refuse every large correct run |

#600 (`claim-ticket.sh`'s generated runner) was already filed by `impl-582` and is not re-filed.

The PR body is corrected for the "second-to-last", "zero POSIX classes" and per-script tool-inventory claims, and states the coverage asymmetry rather than leaving it implicit.

feigi:
**Label correction on the deferrals**, per the #239 `/triage` ruling of 2026-08-18 (filing-time bar is "is the defect confirmed", not "is the remedy settled"): **#612, #613 and #614 are now `ready-for-agent`.** All three have a defect confirmed by measurement — the safety precondition shipped provably false; the payload measurably cannot round-trip (`jq` substitutes U+FFFD, `python3` raises `UnicodeDecodeError`); and deleting all five sibling pins leaves 187/187 green. Only the remedy is open in each, which no longer holds a ticket back.

**#611 stays `needs-triage`**, deliberately — it is the one where *nothing is confirmed broken*. The finding asked to delete `inflight.sh`'s five per-site prefixes; that was ruled a strict weakening, so the current code is correct and the only open question is a design preference. That is the narrowed `torn` case, not a confirmed defect.

No code changed. **The pushed SHA is unmoved at `e93c6be`.**

PR #601, opened 2026-08-18
fix(fleet): let gone()'s ancestor walk reach `/`, in all three copies

### LINKED ISSUE #178: release-ticket.sh:158 — `gone()`'s ancestor walk never tests `/`, so an absent path with no surviving ancestor below root reads as "cannot tell"
Deferred from PR #171 review. **Pre-existing** — the walk is byte-identical at base (`d8561853:214`), so #171 did not introduce it. #171 extracted the walk into `gone()` and gave it a **second caller**, which is why it is worth closing now.

## Problem

`skills/fleet/scripts/release-ticket.sh:156-160`:

```sh
gone() {
  look=$1
  while [ ! -e "$look" ] && [ "$look" != "${look%/*}" ]; do look=${look%/*}; done
  [ ! -e "$1" ] && [ -x "$look" ]
}
```

`${p%/*}` on `/foo` yields the **empty string**, not `/`. So a path whose every ancestor below the root is gone falls out of the loop with `look=""`, and `[ -x "" ]` is false — the predicate answers "cannot tell" for a path that is provably absent with a searchable root.

The header at `:142-143` asserts the property the code lacks: "walk up to the nearest ancestor that does exist and require THAT to be searchable". The root is never reached.

## Measured (git 2.50.1, `/bin/sh` and `/bin/dash`)

```
gone /nonexistent-top-level          -> 1 (cannot tell)   look=""    WRONG, / is searchable
gone /nonexistent-top-level/child    -> 1 (cannot tell)   look=""    WRONG
gone /Volumes/nope/repo/.worktrees/9-x -> 0 (gone)        look=/Volumes   correct
gone <repo>/.worktrees/9-x           -> 0 (gone)          correct
```

## Consequence

Both callers land in the safe direction, but both land back on the failure this line of work exists to end:

- **Stray guard (`:184`)** — prints `release it by hand` for a directory that is not there. That is literally the #81 bug, unfixed for this path shape.
- **Dirty check (`:255`)** — `die "cannot tell whether $wt exists"`, the permanent refusal #37/#81 exist to stop producing.

## Reachability

Not reachable through `claim-ticket.sh:26`, which puts worktrees at `<repo>/.worktrees/<issue>-<slug>` under a main checkout that must survive for `git worktree list` to answer at all. Reachable through a **hand-added worktree outside the checkout** whose ancestor chain below `/` was removed (`git worktree add /scratch/wt` then `rm -rf /scratch`) — which is exactly the population the `stray` guard exists to handle.

## Desired

One token, inside the loop:

```sh
while [ ! -e "$look" ] && [ "$look" != "${look%/*}" ]; do look=${look%/*}; look=${look:-/}; done
```

Verified by two independent reviewers across the full input matrix: every current answer is preserved — including `gone ""` staying 1, unsearchable prefixes staying 1, and existing paths staying 1 — while the root-ancestor cases become 0.

Note the rejected variant: `[ -n "$look" ] || look=/` **after** the loop flips `gone ""` to 0. It must go inside the loop.

## Why deferred rather than fixed in #171

It changes the behaviour of `:255`, the gate on an irreversible delete, and no test covers the root-ancestor shape. That deserves its own regression test (both callers) and its own review rather than a drive-by edit on a PR whose whole point was that the hoist changed nothing.

Also fix the `:142-143` comment, which currently claims the property the code will then actually have.

feigi:
Closes #178

`${p%/*}` on `/x` yields the empty string, not `/`. So a path whose every
ancestor below the root is gone fell out of `gone()`'s walk on `""`, and
`[ -x "" ]` answered "cannot tell" — about a path that is provably absent with
a searchable root.

The fix is one token, inside the loop:

```sh
while [ ! -e "$look" ] && [ "$look" != "${look%/*}" ]; do look=${look%/*}; look=${look:-/}; done
```

Landed in **all three copies** — `reap.sh`, `release-ticket.sh`,
`worktree-audit.sh` — which were byte-identical before this change and still
are after it. Fixing one is what makes them disagree.

## What each of the four callers stops doing

| caller | before | after |
|---|---|---|
| `release-ticket.sh` stray guard | `release it by hand` — for a directory that is not there (#81, unfixed for this shape) | names `git worktree prune` |
| `release-ticket.sh` dirty check | `die "cannot tell whether $wt exists"` — the permanent refusal #37/#81 exist to stop | releases |
| `reap.sh` | `keep "cannot tell whether worktree $wt exists"` — forever | reaps |
| `worktree-audit.sh` | `UNREADABLE … an ancestor could not be read` | `MISSING on disk` |

## Inside the loop is load-bearing, and only one test can tell

The ticket names a rejected variant: `[ -n "$look" ] || look=/` placed *after*
the loop. It reads identically and flips `gone ""` to 0, because nothing enters
the loop on an empty `$1`.

Measured, by mutation: that variant passes **all 101 caller-level tests** —
`release-ticket` 67/67, `reap` 19/19, `worktree-audit` 13/13. Every caller
guards on a non-empty path first (`worktree-audit.sh` reads its own off
`git worktree list`, which never emits an empty one), so no caller-level suite
can reach the input that separates the two placements.

That is why `gone-walk.test.mjs` exists: it reads the function out of the three
scripts and runs it directly over the whole input matrix, `gone ""` included.
It also pins that the three copies are still byte-identical.

## Tests

`832/832` pass (`./agent-test skills/fleet/scripts`). Five new cases, one per
caller plus the direct matrix; all five were **red before the fix and green
after**, with the 96 pre-existing tests green in both states — which is the
ticket's "no test covers the root-ancestor shape", confirmed rather than taken
on trust.

Mutation-verified in a scratch copy, worktree untouched:

- fix reverted → exactly the 5 new tests die, nothing else moves
- rejected after-the-loop variant → only `gone-walk` dies, on exactly `gone ""`
- control (`look` renamed to `up`, semantics identical) → all green

The fixture is a rewritten `.git/worktrees/<n>/gitdir`. `git worktree add`
cannot produce this shape, since it has to create the directory it names; in
production it is the hand-added worktree outside the checkout whose ancestor
chain was removed.

## Behaviour deliberately NOT changed

Measured on `/bin/sh` and `/bin/dash`: `gone ""` stays 1, unsearchable prefixes
stay 1, existing paths stay 1, relative paths stay 1, and the walk still
terminates on a path with no slash.

## Also

`release-ticket.sh:265`'s header asserted the property the code lacked ("walk
up to the nearest existing ancestor BELOW `/`") and carried a note that the
walk did not do what "nearest ancestor that exists" promises. Both now describe
what the code does, and the note is replaced by why the restore sits inside the
loop.


feigi:
## Review applied — `2a473a1`

`node --test skills/fleet/scripts/*.test.mjs` -> **832 tests / 832 pass / 0 fail / 0 skipped**.

### Applied (3)

All in `skills/fleet/scripts/gone-walk.test.mjs`. Each is a way this new suite went green without gating anything.

| # | Dimension | What | Mutant killed |
|---|---|---|---|
| 1 | `silent-failure` | Root guard at `:63` skipped the **entire 9-row matrix**, not the one row needing a non-root uid. Now only the `noperm` row drops under uid 0. | Rejected after-the-loop form of this fix, under stubbed uid 0: was **101 / 98 pass / 0 fail / 3 skipped — fully green**; now red on exactly `gone "" must be 1`. |
| 2 | `correctness` | `extract()` lifted the **first** `gone()` while `/bin/sh` runs the **last**, so a shadow definition left both tests green against a predicate the scripts never execute. Now pinned to exactly one match. | Second `gone()` inserted mid-file carrying the rejected form: was **832/832 green**; now `release-ticket.sh must define gone() exactly once, at column 0`. |
| 3 | `silent-failure` | `t.after` is registered before the fixture exists, so a setup failure met a `chmod` with nothing to restore — it threw and took the `rmSync` with it. Now wrapped. | ENOENT in fixture setup: leaked `gone-walk-MvCjDz` before, **0 leaked dirs** after, with the real setup error still reported. |

Negative control for #2: a sibling function with its own column-0 `}` and a near-miss name `gonely()` both stay green, so the new assert discriminates rather than merely fires.

Also names the spelled-out-`SCRIPTS` convention in a comment, per `arg.test.mjs`'s `CONSUMERS`: a globbed set silently *shrinks* when a script drops its copy, which is the regression being pinned. That closes a review finding that argued for deriving the list — deriving it would reintroduce exactly that hole.

### Deferred (5)

Each in-scope `suggestion` got one refuter, biased to refuse. Four were refuted and are filed as records (`wontfix`) so a future specialist pass does not re-derive the answer:

- **#607** — `simplify` — replace `relocate()`'s `.git/worktrees` scan with `git rev-parse --absolute-git-dir`. Refuted: `--absolute-git-dir` resolves to the **main** repo's `.git` on misuse, so the patched helper silently corrupts instead of throwing; dropping the `w` parameter lets a stale 3-arg call `rmSync` the whole fixture repo.
- **#608** — `simplify` + `types` — the duplicated `relocate()` docstring, and the absence of a byte-identity test over the three copies. Refuted: the cited `reap.sh:25` precedent keeps a *full* rationale and adds the pointer; 15/15 copies of the repo's other duplicated test helpers do the same. And unlike `gone()`, each `relocate()` copy has exactly one consumer — its own file — so a shape-destroying drift already fails the owning suite (18/19).
- **#609** — `comments` — the header lacks a `(verified, git X.Y.Z)` tag. Refuted: 80 of 90 git-behaviour comment blocks carry none, and **this PR adds two more uncited instances of the same claim**. The claim itself was verified true against git 2.50.1.
- **#610** — `simplify` — replace `SCRIPTS[i + 1]` with a name-iterating loop. Refuted: the eager `.map(extract)` fires every column-0 guard *before* any comparison; the lazy rewrite hides a structural break behind a drift, and passes **vacuously** on a one-element list.

One was recorded on the ticket that already owns the line, rather than filed twice:

- **#391** (comment) — `silent-failure` — `reap.sh:196` discards git's diagnosis behind `worktree remove refused`. Already owned by #391. What is new: **this diff routes the no-surviving-ancestor population into that line**, where they previously landed on `cannot tell whether worktree $wt exists` at `:106`. Measured both forms; recorded there.

One finding was dropped rather than filed: the argument that `SCRIPTS` should be derived from the repo. That is a deliberate convention here, and the comment added above now says so at the line.

PR #616, opened 2026-08-18
fix(fleet): extract json.sh and sweep six scripts onto it (#119)

### LINKED ISSUE #119: fleet scripts: five more emit unescaped JSON, and jstr is now duplicated

feigi:
Closes #119

Extracts `skills/fleet/scripts/json.sh` and routes every fleet script that
splices a caller-controlled string into a JSON payload through it. No payload
field is added or removed anywhere: this changes escaping, never shape.

## Re-derived against `origin/main` first — the ticket is stale in three places

| ticket says | measured at `aa7bbc5` |
|---|---|
| `jstr` in two files | **three** — `inflight.sh`, `no-undo-audit.sh`, `release-ticket.sh` (the comment is right, the body is not) |
| eight scripts build JSON with `printf '{` | **nine** files, twelve sites |
| `verify-sha.sh` lines `41-42` (comment corrects to `47-48`) | **line 53** |

## The ticket's own prescription is not sufficient, and it is measurable

The ticket says to guard the source with `|| die` "rather than letting `set -e`
decide". Measured — `.` is a POSIX **special builtin**, so failing to open its
operand aborts a non-interactive shell outright and the `||` never runs:

| shell | `\|\| die` fired? | exit |
|---|---|---|
| `/bin/sh` (macOS, bash 3.2) | **no** | **1** |
| `/bin/bash` 3.2, `bash --posix` | **no** | **1** |
| `/bin/dash` | no | 2 |
| Homebrew bash 5.3 | yes | 2 |

Confirmed on the real script, not just a synthetic shell: with the `[ -r ]`
pre-check removed, `inflight.test.mjs` reports `1 !== 2`. Exit 1 from
`inflight.sh` means *taken* — the ticket's own prescription would fabricate a
claim on a free ticket. So the guard is `[ -r ]` **then** `. … || die`, and
each script that has a verdict-carrying exit 1 pins it.

Residual ceiling, stated rather than hidden: a `json.sh` that exists, is
readable, and has a **syntax error** still aborts before either guard. Nothing
in POSIX `sh` catches that; the suite covers it instead.

## Enumerate and declare

**Covered — six scripts, seven sites:**

| site | fields | reachability |
|---|---|---|
| `verify-sha.sh:53` | `$branch`, `$sha`, `$tip` | `$branch` measured reachable; the other two gated, wrapped for uniformity |
| `claim-ticket.sh:371` | `$branch`, `$wt`, `$install`, `$runner` | `<slug>`/`<type>` are unvalidated argv; reachable |
| `worktree-audit.sh:175` | `$wt`, `$short` | both reachable, by different routes — a ref takes `"`, a path takes `"` and `\` |
| `prove-merge.sh:127` | `$second`, `$first`, `$proof_path` | none reachable (rev-parse hex, script literal); uniformity wrap |
| `reap.sh:70,231` | 5 interpolations via `keep()` + the `reaped` accumulator | branch name measured reachable |
| `inflight.sh`, `release-ticket.sh`, `no-undo-audit.sh` | already escaped | lose their inline copies |

- **`drop-merged-label.sh:52,91`** — the tenth site the ticket's table never
  names. Enumerated and excluded **on measurement**: it has no string field.
  `$pr` is `case … ''\|*[!0-9]*\|0?*`, `$apply` is the script's own boolean, and
  `$results`/`$failed` are built from `gh --jq '…[].number'`. Adding `jstr`
  would be escaping numbers.
- **`worktree-audit.sh`'s `dirtyFiles[]`** — git C-quotes those paths itself,
  conditionally, so JSON-escaping on top is a different problem. The script
  already names it in place as "the existing, unaddressed gap … (#119-shaped),
  not one this fix opens".
- **`jrewritten` at the five new sites** — would add `…Rewritten` fields the
  spec table does not carry. Contract change, not an escaping fix.
- **A backslash fixture for `verify-sha.sh`** — written, then deleted:
  `git check-ref-format --branch 'back\slash'` exits **128**. A branch name is
  that script's only string input, so `\` is unreachable there and the test
  pinned fiction. The backslash vector is exercised where it is real —
  `worktree-audit.sh` and `claim-ticket.sh`, whose input is a path.
- **Ticket comment 2's "is this apparatus worth ~160 shell lines?"** — its own
  text says "ruled out of scope on #227".

## The `sed` mask (#119's other half)

`inflight.sh:809` named #119 as the owner of this, so it is fixed here.
A pipeline reports only its **last** stage, so `printf | sed | tr` swallowed a
failed `sed` entirely — empty value, exit 0, `|| die` unfired. POSIX `sh` has
no `pipefail`, so each fallible stage is captured and its status read.

Per ticket comment 3, the test forces **`sed`** specifically, and asserts on the
**output** as well as the status — an assertion reading only the status cannot
tell "the shim fired and jstr caught it" from "the shim never ran".

## What my own edit newly does

Adding a `.` changes what runs first, what `set -e` covers, and what a missing
file does. Each is tested:

1. **A missing `json.sh` is exit 2, never a verdict** — pinned per-script for
   all three whose exit 1 is a verdict, plus a "nothing was mutated" assertion
   on `release-ticket.sh`, `reap.sh` and `claim-ticket.sh` (the guard sits ahead
   of every deletion).
2. **The source must sit below `export LC_ALL=C`** — `locale-pin-prose.test.mjs`
   requires every line above the pin to be comment/blank/shebang/`set -`.
   Green, unmutated.
3. **`shellcheck -x -S warning`** — measured to break without directives
   (SC1090 warning at each call site, SC2148 error in `json.sh`).
   `# shellcheck shell=sh` + `source-path=SCRIPTDIR` + `source=json.sh` fixes
   it, verified from both the repo root and the scripts dir.
4. **`json.sh` never exports `LC_ALL`** — three of its callers do not pin the
   locale, and exporting from a sourced lib would silently re-locale every other
   tool in them. Per-command `LC_ALL=C` prefixes instead.
5. **A pre-existing test broke and was fixed**: the runner-stamp test copies
   `claim-ticket.sh` to a temp dir and already had to carry `derive-testcmd.sh`
   with it. `json.sh` is the second sibling it now needs.

## The false-positive half

Every guard here has a test feeding it input it must **accept**: an ordinary
branch name, path and slug produce output **byte-identical** to what each
script has always emitted (`verify-sha.sh`, `reap.sh`, `worktree-audit.sh`,
`claim-ticket.sh`, `prove-merge.sh`), and `jstr` returns 0 on seven ordinary
strings including a 4 KiB one.

## Mutation testing

Every added test was mutated, confirmed to redden **alone**, and restored with
`cp` from a pre-mutation copy — never `git checkout --`. Twelve mutants, twelve
single-test kills, including `$wt` and `$short` proven to be pinned
*independently* of each other. Plus a discrimination check: a comment reflow in
`json.sh` leaves the suite green.

## Docs

The spec's `reap.sh` contract row documented the unescaped splice as a live
defect. It no longer is, so the sentence is replaced rather than left asserting
something false.

## Verification

- `./agent-test skills/fleet/scripts` — **891 pass, 0 fail** (baseline 869)
- `git ls-files '*.sh' | xargs -r shellcheck -x -S warning` — clean

## Noted, not changed

`ci.yml:175` says `skills/fleet/scripts/` "is 29 flat files". It was already
wrong before this PR (67), and it is incidental colour in a comment about
node's `node_modules` discovery, so it is left alone rather than swept in here.


feigi:
Review fixes applied in `638590e`. `tests 897 pass 897 fail 0` (baseline 891 + 6 new); `shellcheck -x -S warning` over `git ls-files '*.sh'` clean.

**Applied (11 findings):**

- **The causal chain** (`survived[0]`, `[3]`, `[4]`, `[8]`, `unverified[1]`) — the four `jarr`/`jarr_rewritten` sites in `no-undo-audit.sh` and `block()` in `release-ticket.sh` now read their status, and the comment that justified skipping them ("this script has no exit-1 verdict to fabricate") is corrected in both, derived per script. Re-measured at PR head: the unguarded shape gives **status 1, empty stdout, no diagnostic** on a worktree the run had just logged clean; guarded it gives exit 2 with a named cause.
- `survived[1]` json.sh's locale count: six → **five** (the six counts `derive-testcmd.sh`, which does not source json.sh).
- `survived[2]`, `survived[10]` — the pre-existing comments in `inflight.test.mjs` and `no-undo-audit.test.mjs` that tracked jstr's sed masking and the three-copies question as still open under #119.
- `survived[9]` — the design spec's seven `exit 2 only — <causes>` rows.
- `unverified[4]`, `[5]`, `[6]`, `[9]`, `[11]` — see the commit body.
- Plus one defect a refuter turned up while disproving `unverified[2]`: `$sha` in `verify-sha.sh` is a **live** escaping vector (`git cat-file -e "${sha}^{commit}"` resolves a ref name, and `"` is legal in a ref), not the unreachable one its comment claimed. Corrected, with a test — it had none.

**Deferred:** `survived[6]` (worktree-audit `dirtyFiles[]` C-quoting) → **#617**, `ready-for-agent`. Confirmed defect, open contract question on git's conditional C-quoting.

**Refuted with evidence:** six suggestion-band findings, recorded in **#618** (`wontfix`) so they are not re-filed. `refuted[0]` and `refuted[1]` declined on ruling, both noted there.

**Mutation-verified.** Each of the five new guards removed in a scratch copy reds exactly its own test and nothing else; the `prove-merge` and `verify-sha` unwrap mutants likewise. Discrimination checked: restructuring the conflicts guard into two equivalent statements stays 73/73 green, so the tests pin the behaviour rather than the syntax.

PR #619, opened 2026-08-18
docs(fleet): commit the tier-outcome rows and two rule fixes stranded in the working tree

feigi:
Ad-hoc artifacts from two `/fleet:run-team` runs on 2026-08-18 that were written to the working tree and never committed. Opened at the maintainer's request at the end of the second run.

Docs and data only — no code, no test changes. Full suite green at `84039d9`: **916 pass / 0 fail**.

## What is here, and why each one matters uncommitted

**`docs/metrics/tier-outcomes.tsv` — 8 rows.** Four from the morning run (#575, #576, #577, #579), four from the afternoon (#598, #599, #601, #616). Phase 2's tier guard fires on the **accumulated** file across runs, and its floor is "at least three `class=routine` PRs spanning two or more distinct `run_date`s". Rows that live only in a checkout make the guard re-derive that floor from an undercount on every run. This repo has already lost a day to a guard verdict stranded outside `main`.

Of this batch, **#616 and #601 are the file's second and third `minted_false_claim=no` rows**; every other row here is `yes`.

**`skills/fleet/commands/review-and-fix.md` — step 5's filing bar.** Step 5 pointed the *filer* at phase 0's `decided?` test. The `/triage` ruling on **#239** (2026-08-18) superseded that, and the file was never updated to match — so the ruling lived in memory while the document that enacts it disagreed.

Measured cost of leaving the pointer, in one run: two fix-appliers hit it, the controller restated the bar inline in both dispatch prompts, and one member filed correctly *only* by explicitly flagging that it was diverging from the text it had been told to follow.

**`skills/fleet/skills/run-team/SKILL.md` — the mutation-restore rule.** Authored by an **earlier** run, not this one, and left uncommitted. `git checkout -- <file>` reverts the whole file rather than the mutant, so it eats uncommitted edits and the next "clean baseline" is silently the reverted file.

## Deliberately NOT included

`settings.json` is dirty in the same checkout and is **left alone**. It carries an `effortLevel` change and a new permissions block — maintainer configuration, not a fleet artifact, and `run-team/SKILL.md`'s own mid-run tooling-fix scope excludes `settings.json`, permissions and CLAUDE.md by name. It stays uncommitted for the maintainer to handle.

## Review note

No `ready-to-merge` label applied — that is a reviewer's to add, never the controller's. Merge directly or send it through the normal review path, whichever you prefer; it is docs-only and touches no script behaviour.


feigi:
Fix-applier report — pushed `84039d9`, suite `node --test skills/fleet/scripts/*.test.mjs` → **tests 916 / pass 916 / fail 0**.

**Applied (4 defects, one commit):**

- **Header, class claim** (`docs/metrics/tier-outcomes.tsv`) — the header asserted every row is `class=routine`, none `class=correction`, while this branch's own append added three `class=correction` rows (#576, #577, #598). Restated as the claim that still holds and can be rechecked: every row dated before the 2026-08-16 revert is `class=routine` (7 of 7), and the `class=correction` rows are all post-revert.
- **Header, n=1 claim** — "exactly one row off the old binding … That is n=1" now reads as a whole-file count of routine-at-opus rows, of which there are 8. Scoped inline to the rows dispatched while the binding was in force, where it is still exactly one (PR #480).
- **PR-576 row** — "(33 rows prior, all yes)" counted lines, not rows. 9 data rows precede it, all `minted_false_claim=yes`; 33 was the pre-append file's total line count (24 header/comment + 9 rows). Corrected to 9; the "all yes" half was already right.
- **Superlatives + a misquote** — three appended rows each claimed a superlative "of the run" while the file carries no run identifier, and they disagree on the rows' own numbers (#599's 19 findings against #616's 26, both dated 2026-08-18). Superlatives dropped, each row now states its finding total. Separately, the PR-620 row attributed to issue #391's Agent Brief the words "a third filesystem probe here would widen #83"; the Brief says "a third probe here would make that worse", and the quoted string occurs nowhere else in the tree. Replaced with the Brief's own wording.

The header also gains a recount instruction, since an append lands a row without touching it — which is how it went stale in the first place.

**Deferred and filed:** #637 — the pass-1 orphan claiming `<scratch>/pr<N>/mutate/` contradicts the two-level scratch convention. Refuted by measurement (the path is two levels, purpose-named second levels are the existing shape in `review-pr.js`, and the sentence sits in the member briefing rather than the refuter brief); filed as a `wontfix` record with the counter-evidence.

PR #620, opened 2026-08-18
fix(fleet): reap reports what a refused worktree removal actually did, and quotes git's reason

### LINKED ISSUE #391: reap.sh: `worktree remove` refusal is read from the exit code, so a cleared registration reports as "refused" — and git's reason is discarded

feigi:
`reap.sh` derived its `worktree remove` and `branch -D` keep reasons from an exit code and sent git's own diagnosis to `/dev/null`. The worst of them reported a removal that had *already unregistered the worktree* as though nothing had happened.

Closes #391

## What was wrong

A non-zero exit from `git worktree remove` is not proof the removal had no effect. Measured here on git 2.50.1 (Apple Git-155):

| shape | rc | registration after | directory after |
|---|---|---|---|
| locked worktree | 128 | **intact** | present |
| symlink standing in for the directory | 255 | **cleared** | present |
| `chmod 555 .git/worktrees/<id>` | non-zero | **intact** | **deleted** |
| peer session finishes the same removal first | 128 | **cleared** | **deleted** |

Both of the first two produced the same `worktree remove refused`, which reads as a safe no-op while the second has unregistered the worktree and left the branch standing as though deliberately retained. `2>/dev/null` threw away the only thing that says which remedy applies.

The bottom two rows are the review's contribution, and they are why the fix says less than its first draft did: **the registration and the directory are independent facts in both directions.** A cleared registration is no evidence of an orphan; an intact one is no evidence that nothing was removed.

## What changed

- **`git worktree remove` refusal** — re-reads the registry and reports `registration intact` / `registration cleared` / `cannot tell whether the registration survived`, each carrying git's own message with newlines collapsed. It probes the **registration** and claims nothing else, per the brief: that is what was measured, and #83 — this script and `release-ticket.sh` answering an unreadable worktree in opposite directions — is closed, so a third filesystem probe here would reopen that ground. `gone()` is deliberately not reused.
- **`git branch -D` failure** — quotes git's message (`error: cannot delete branch 'x' used by worktree at '…'`) instead of discarding it. This is also what fixes the mislabelling when the `worktree list` lookup dies: the branch still reaches `-D`, still refuses, and now names the worktree holding it rather than reporting a bare `branch delete failed`.

Behaviour is unchanged throughout: still `keep`, still `continue`, **no directory removed under any path**, nothing reaped that was not reaped before, the sweep still finishes the remaining branches, and the dry run untouched.

## Scope — declared, and one half handed back

The brief's acceptance criteria name two sites; a later comment adds the `git worktree list` lookup. This PR originally covered all three and made the lookup fail closed. The review found that half of the third site to be a **behaviour** change, and the brief forecloses it — *"No control-flow change. Still `keep`, still `continue`"*, with triage having ruled *report the state, do not change the behaviour*.

Measured before reverting, rather than conceded: with a shim failing only `git worktree list`, the pre-PR script **reaps** a merged `[gone]` branch that has no worktree (`{"applied":true,"reaped":["feature/merged"],"kept":[]}`), while the fail-closed spelling keeps it. So it is genuinely a control-flow change, and the standing counter-argument does not cover it — `git branch -D` refuses a branch whose worktree is live whether or not `git worktree list` answers, but a branch with no worktree gives it nothing to trip over. Reverted here and filed as **#622** for the maintainer call the brief reserved.

Still deliberately left: `git -C "$wt" status --porcelain 2>/dev/null` and `… --ignored 2>/dev/null`. At both, git's *stdout is the payload* the check reads, so `2>&1` would fold a warning into `$status_out` and make a clean worktree read as **dirty** — a worse, silent failure. Filed as **#625**. Also left: the `%(refname:short)` ambiguity that emits `heads/feat+x` for a branch with a same-named tag (a refname defect, not a discarded-reason one). Filed as **#634**.

## Tests

The ticket says this script has no suite. That has drifted — `reap.test.mjs` exists and covers the worktree path well. The narrower claim held, and I re-measured it: **garbling the `worktree remove` reason, and garbling the `branch delete failed` reason, each left all 24 existing tests green.** Both failure arms were invisible.

Every mutant is killed by its own test, and a behaviour-preserving control stays green for each. Five of the tests I first wrote did *not* discriminate and were fixed:

- the sweep test named branches `locked`/`healthy` — `for-each-ref` sorts by refname, so the healthy one was reaped *before* the refusal and `break` passed as happily as `continue`;
- the dry-run test used only a *locked* worktree, where an attempted removal fails anyway, so a dry run taught to call `git worktree remove` stayed green;
- the registry probe's own failure arm had no test, so spelling it `git worktree list | grep -q` — this ticket's defect inside its own fix — survived the whole file;
- the probe's whole-line anchor had no test; the sibling fixture added for it promptly caught the same unanchored mistake in the suite's own assertion;
- **and the two state phrases, found by the review**: both existing arms use the symlink shape, where the directory happens to survive, so the whole suite stayed green with a hard-coded `— $wt is still on disk` that three measured shapes contradict. The new arm reproduces the peer race and asserts the absence the old wording denied; restoring the clause reds it, and rewording the other arm leaves it green.

The deregistering refusal uses the **symlink** shape rather than `chmod`: reproduces as any user, no permission bits, none of the euid-0 vacuity the `0o000` fixtures here already skip around (#184). The new peer-race arm needs no permission bits either.

`908/908` green on the full suite (`node --test skills/fleet/scripts/*.test.mjs`). No new shellcheck findings (`SC1091` is pre-existing on the unchanged `json.sh` source line).

Routing git's stderr into `keep()` is safe now that #119 landed `json.sh`; before it, that widened a JSON-corruption hole — which is why the brief blocked on it.


feigi:
## Review adjudication — 11 findings, 5 applied, 6 deferred

Pushed at `dc56eb8`. Full suite `node --test skills/fleet/scripts/*.test.mjs` → **908 pass / 0 fail** (baseline 907 + one new arm).

### Applied

| finding | what shipped |
|---|---|
| `survived` ×2 (correctness + silent-failure, `reap.sh:304`) | Both state phrases now claim only the registration the probe read: `registration intact` / `registration cleared`. The disk clause is gone, and so is the comment premise that git always clears the admin entry first — `chmod 555 .git/worktrees/<id>` disproves it. Matching `leaving an orphaned directory` promise dropped from the spec row. New test reproduces the peer race. |
| `refuted[1]` residual (comments, `reap.sh:282`) | `#83 is already open` → closed. The wrong-ticket half is **not** applied: #83's title is verbatim *"reap.sh and release-ticket.sh take opposite defaults on an unreadable worktree"*, so the citation was right. |
| `unverified[0]` (correctness, `reap.sh:152`) | Control-flow half reverted, message half kept — see the PR body and #622. |
| `unverified[4]` first half (simplify, `reap.test.mjs:743`) | `?? process.env.PATH` dropped — unreachable. |
| `unverified[5]` (simplify, `reap.test.mjs:1017`) | Hand-written shim replaced by `failOnlyShim`, counter moved into its `match`. |
| `refuted[0]` residual (comments) | The `all 32 green` count is gone — the note now says which direction each spelling goes. |

Every new or changed test was mutation-checked in **both** directions: the mutant it should catch reds it, and a change it should not catch stays green. The `survived` arm was the sharp one — both pre-existing #391 tests pass with the false disk clause in place, because the symlink shape they use is one where the directory happens to survive.

### Deferred

| # | label | finding | why |
|---|---|---|---|
| [#622](https://github.com/feigi/claude-config/issues/622) | `needs-triage` | `unverified[0]` — a dying `git worktree list` reaps merged `[gone]` branches blind | Fail-closed is a control-flow ruling the brief reserved. Measured both ways. |
| [#625](https://github.com/feigi/claude-config/issues/625) | `needs-triage` | `unverified[1]` — `reap.sh:219,245` still `2>/dev/null` | Out of #391's scope, and `2>&1` is unsafe there: stdout is the payload, so a warning would make a clean worktree read dirty. |
| [#626](https://github.com/feigi/claude-config/issues/626) | `needs-triage` | `unverified[2]` — the three-state outcome has no structured field | Payload-schema change, out of scope; also consistent with the existing convention. |
| [#627](https://github.com/feigi/claude-config/issues/627) | `wontfix` | `unverified[3]` — `2>&1` folding diagnostics into the anchor | **REFUTED.** Half moot after the revert; the surviving half needs a stderr line byte-identical to `worktree <abs path>` — 11 `GIT_TRACE*` channels and 7 corruption shapes all produce zero such lines, and the binary holds no candidate message. The only reachable direction is `cleared`→`intact`, and the remedy would discard diagnostics on the failure path. |
| [#628](https://github.com/feigi/claude-config/issues/628) | `ready-for-agent` | `unverified[4]` second half — migrate 5 inline `PATH:` sites to `withShim` | Survives measurement, but all five are outside this PR's diff. |
| [#629](https://github.com/feigi/claude-config/issues/629) | `wontfix` | `unverified[6]` — a `flat()` helper | **REFUTED** on cost and scope: 3 sites now, not 4; the helper makes the file **+2 lines**; 2 of the 4 lines it touches are outside this PR; and a file-local helper leaves the other 5 repo sites unconsolidated, so the stated benefit is not delivered. |
| [#630](https://github.com/feigi/claude-config/issues/630) | `wontfix` | `refuted[0]` — the `32 green` arithmetic | **REFUTED**: the note describes the *alternative* spelling, and that claim re-measured true here (shipped → `35/34/1`, alternative → `35/35/0`). Only the count drifted; fixed above. |

Each in-scope `suggestion` got one refuter dispatched biased to refuse, verifying by running rather than reasoning; `unverified[1]` and `unverified[2]` were split out on scope and filed without one.

PR #621, opened 2026-08-18
fix(fleet): measure what release-ticket's worktree removal landed, and refuse over an orphaned directory

### LINKED ISSUE #208: release-ticket.sh: `done_wt` records the exit code, not what landed — `worktree remove` can clear the registration and still fail

feigi:
Closes #208

## The defect

`git worktree remove` deletes the registration **before** the directory and does not put it back when that second delete fails. So its exit code answers for neither, and `done_wt=false` — rendered as the headline `HALTED mid-release — nothing landed` — was a positive assertion that is false whenever git cleared the registration on its way to refusing.

Three landing shapes from one call, all measured here on git 2.50.1:

| trigger | rc | registration | directory |
|---|---|---|---|
| dirty worktree | 128 | kept | kept |
| symlink standing in for the directory | 255 | **cleared** | kept |
| unwritable `.git/worktrees` | 255 | cleared | gone |

The third is not named in the ticket and was found while reproducing the first two. All three previously reported `nothing landed`.

## What changed

**`done_wt` becomes `wt_outcome`**, carrying CONTEXT.md's four release-outcome states, taken by a new `release_outcome` *after* the call rather than read off its rc. `done_branch` stays a boolean, because `git branch -d` updates a ref atomically — so the detail line is now honestly uneven rather than evenly shaped over two halves that mean different things. `Indeterminate` asserts neither headline.

Measured on the **refusal only**. A zero rc is different in kind: both deletes completed, so `Released` restates git's own success rather than inferring past a failure — and re-measuring the happy path would let a probe that can answer `Indeterminate` refuse releases that plainly worked.

**A release attempt now refuses over an Orphaned worktree directory.** `stray` is awk over git's registry, so a directory whose registration was already cleared is invisible to it *by construction* — the run after a Deregistered halt deleted the branch, dropped the label and exited 0 with `"released":true,"blockers":[]` over a directory still on disk, and the next `claim-ticket.sh` for that slug then died on it. The probe reconstructs the path from the same `<issue>-<slug>` pairing `stray` matches on and asks the filesystem directly, in the **precondition block** so the dry run predicts it for free. The remedy names a manual removal and never `git worktree prune`: there is no registration left to clear, so prune is an action that cannot work.

No new key in the JSON receipt — the outcome rides in the existing `blockers[]` string. `gone()` is unmodified.

## On the `gone()` tri-valued question

The amendment put `gone()` in scope on the premise that `Indeterminate` needs a probe separating *present* from *could-not-stat*, and nobody else would supply it. **The tree already supplies it.** `gone()`'s own contract prescribes the composition:

> `0` ONLY for established absent; `1` covers present AND cannot-stat, so a caller needing those apart pairs this with its own `[ ! -e ]`, as the dirty check does.

`occupied` + `gone` is that pairing, and the dirty check at `release-ticket.sh` has used it since before this ticket. Making `gone()` itself tri-valued would need a 0/1/2 protocol that breaks its documented condition-context use under `set -e`, would have to land in all three copies pinned byte-identical by `gone-walk.test.mjs`, and would contradict the file's own rule that one question gets one predicate. So `gone()` stays as it is and the brief's original acceptance criterion holds. **`reap.sh` is untouched**, so #391 is unaffected.

## Tests

Seven added, 904/904 green across the fleet suite. Fixtures use the **symlink** trigger, swapped in during the `gh` round trip — the check-then-act window — because the non-directory precondition would otherwise block it before any mutation. The `chmod` shape is named in a comment rather than built (#184).

Both halves of the new refusal are pinned: a clean claim whose worktree sits at exactly the path the probe reconstructs still releases, and the dry run and `--apply` report the orphan word for word identically.

Every added test kills its own mutant, and a semantically-neutral rename leaves all 76 green.

| mutant | killed |
|---|---|
| `Deregistered` arm returns `Unreleased` | the partial-release test, alone |
| remedy names `git worktree prune` | the manual-removal test, alone |
| orphan check gated on `--apply` | the dry/apply parity test, alone |
| `elif ! gone "$orphan"` arm deleted | the cannot-stat test, alone |
| `Indeterminate` headline branch deleted | the neither-headline test, alone |
| orphan block disabled | the end-to-end and manual-removal tests |
| the registration gate dropped | 31 tests, including the false-refusal guard |


feigi:
Review fixes applied in `cbb0e02`. Suite: `node --test skills/fleet/scripts/*.test.mjs` → **905 pass / 0 fail** (baseline 904, +1 new case).

**Applied (4)**
- `release-ticket.sh` `occupied()` doc-comment — the rc-0 measurement was real but attached to the wrong symlink shape. Corrected, and it now states plainly that the rc-0 fast path is sound, so the next reader does not "fix" it on the strength of a wrong comment.
- `release-ticket.test.mjs` — new case `the orphan probe is anchored at the checkout, never at the caller's cwd`. Mutation-verified: `main_wt=$PWD` → that case is the only red (76/77); an equivalent derivation off git's listing (`sed` for `awk`) stays green (77/77).
- `release-ticket.sh:756` — `rc=0; … || rc=$?; if [ "$rc" -ne 0 ]` back to `if ! err=$(…)`, the idiom the sibling `git branch -d` uses eight lines below.
- `release-ticket.test.mjs` — dropped the unread `return c.wt` from the `orphan` fixture helper.

**Deferred (3)** — #631, #632 (both `wontfix`, refuted by measurement, refutations in the bodies), #633 (`ready-for-agent`, out of #208's scope: it refactors two pre-existing shim sites).

**Refuted, not filed here** — the truncated-listing/`Indeterminate` finding and the subdirectory `claim-ticket.sh` probe; the controller measured both and upheld the refutations. Its two residuals are tracked separately as #623 and #624.

PR #635, opened 2026-08-18
fix(fleet): make ledger check's two caps say they capped

### LINKED ISSUE #154: ledger.mjs check truncates tracker hits and near-misses silently, against the no-silent-caps rule its sibling script enforces
Deferred from PR #150 review.

## The finding

Two silent caps in `check`:

1. **`--limit 5`** on the tracker query (`skills/fleet/scripts/ledger.mjs:296`). `gh` returns at most 5 and no total, so 5-of-5 and 5-of-300 are indistinguishable. The overlap re-sort then runs *after* truncation, re-ranking a window `gh` chose.
2. **`.slice(0, 3)`** on near-misses (`:249`). Tied rows are dropped with no count reported — measured, 5 rows scoring 1.00 with the caller shown 3.

## Why this is a rule violation, not a preference

`skills/fleet/scripts/candidates.mjs:78-90`, in the same script family, legislates the opposite and is stricter than a warning — it refuses outright:

> Hitting the limit exactly means the answer may be incomplete, and there is no way to tell from the result itself. Refuse rather than hand back a shortlist that silently omits work — "no silent caps" is the rule this enforces.

## The fix

Request one more than displayed (`--limit 6`, show 5) and say so when the extra row comes back; same for the near-miss cap — report the number withheld rather than the rows. `die()`ing as `candidates.mjs` does is **not** right here: these lists are advisory context for a decision, not the work queue itself, and one hit already forces exit 3.

## Why deferred

Lower severity than the `candidates.mjs` case it echoes: with a 3-term AND query, more than 5 hits is rare (the code-reviewer measured 0 false positives across 4 plausible subjects), and one hit is already enough to escalate to exit 3 — truncation hides *additional* instances of a signal already delivered, not the signal.

feigi:
Closes #154

`ledger.mjs check` truncated both of its advisory lists with nothing in the
output to say so — the rule `candidates.mjs:279` legislates one script over
("no silent caps"), and enforces by refusing outright. Refusing is wrong here,
and the ticket rules it out: these lists are advisory context for a decision,
not the work queue, and a single tracker hit already forces exit 3. So report
the withholding instead.

## What changed

**Near-misses (`.slice(0, 3)`).** Rank, then cut. The payload carries
`nearTotal`, and stderr names the count withheld plus the best score among
them. The ticket's measurement is the case: five rows tied at 1.00 with the
caller shown three — the cut is arbitrary among equals, and a three-row list
was indistinguishable from a complete one. The rows are not printed: that
would give back the length the cap exists to take away.

**Tracker (`--limit 5`).** gh returns no total, so a full page and a truncated
one arrive byte-identical. Ask for one row MORE than is displayed — the probe
row's presence is the only truncation signal available. The whole fetched
window is ranked and *then* cut (the probe row was fetched to be ranked, not
just counted), `tracker.truncated` names the state, and the summary prints
`more than 5` rather than a capped count dressed up as an exact one. What the
extra row cannot repair is that gh chose the window at all — six rows are still
a window — so a second line says the ranking is over that window, not over the
tracker. That is the ticket's third, passing-named finding: "the overlap
re-sort runs *after* truncation, re-ranking a window gh chose."

## Contract

Exit codes and verdicts are untouched. 0/1/2/3 and
`already-filed`/`tracker-hit`/`clean`/`unverified` all still fall exactly where
they did: `already-filed` exits 1 before either list is computed, a truncated
hit list is still non-empty so it is still `tracker-hit`/3, an unread tracker
still degrades to `unverified`/0, and neither new report can move a case
between them. `truncated` is omitted on the `!ok` branches for the same reason
`hits` already is — the search never ran, so there is nothing to claim.

One behaviour is newly reachable and deliberate: the row-shape validation now
sees the probe row too, so a gh returning a malformed sixth row degrades to
`unverified` where before that row never arrived. That is the fail-safe
direction — it can never turn a hit into a clean bill.

## Tests

44 in `ledger.test.mjs` (5 new), 921 across `skills/`, all green. Every new
pin was mutation-tested — broken, confirmed red, restored from a copy — and
then checked against a change it must *not* catch:

| Mutant | Reds |
|---|---|
| `nearTotal = near.length` | the withheld-count pin, alone |
| withheld stderr line deleted | the withheld-count pin, alone |
| `nearTotal` over-counted in the payload | the no-withholding pin, alone |
| `--limit` back to `5` | the probe-row argv pin, alone |
| no display cut on `hits` | the truncation pin, alone |
| `truncated` uses `>=` | the exactly-fills pin, alone |
| slice before sort instead of after | the truncation pin, alone |
| internals renamed, a const moved | nothing — 44 green |

Both false-positive halves are pinned, not just the false-negative ones: a
list that exactly fills its cap must claim no truncation, in the payload and on
stderr. The gh stub cannot emulate `--limit`, so the argv pin is what covers
the real truncation and the fixture pins the reporting — separately, on
purpose.

https://claude.ai/code/session_01Npo59xF5tMhME8aTFpEGWq


feigi:
## Review findings applied — pushed `35c0845`

`node --test skills/fleet/scripts/*.test.mjs` from the worktree: **tests 923 / pass 923 / fail 0** (921 baseline + 2 new tests).

### Applied (4 findings → 1 commit)

- **`correctness` + `tests`, both `survived` 2-0, one defect** — the `highest withheld` score was unpinned. The tied fixture ties every row at 1.00, so highest == lowest and `/highest withheld 1\.00/` held under any index at all: `rankedNear[rankedNear.length - 1]`, `rankedNear[near.length - 1]` and even `rankedNear[0]` each survived the whole suite. A new untied fixture (1.00/0.67/0.50 shown, 0.33/0.25 withheld) asserts the exact printed `highest withheld 0.33`; all three index mutations now red it **and nothing else**.
- **`types`** — exactly one row withheld is the tightest boundary the guard has, and no fixture anywhere produced `nearTotal === 4`. `nearTotal > near.length + 1` went silent there and was green across all 921 tests. Now pinned.
- **`types`** — the comment says the withheld rows are reported as a count and a score, "never the rows"; appending a withheld row to the notice was also green across all 921. Now pinned, reading the notice line itself and matching the `#N` row marker rather than fixture text, so reordering the tied rows cannot red it while the invariant holds.
- **`comments`, `survived` 2-0** — the design spec's script-surface row enumerates `check`'s payload field by field and omitted `nearTotal` and `tracker.truncated`. Row updated; two `suggestion`-band findings said the same thing and are subsumed by it.

Every added assertion was mutation-verified both ways: each new pin reds under the mutation it exists for, and stays green under an unrelated one (`CAPPED at` → `capped at`, which reds only the pre-existing tracker test).

### Deferred and filed (10 findings → 8 issues)

| issue | label | finding | dimension |
|---|---|---|---|
| #638 | `ready-for-agent` | gh catch loses the cause — 500-char clip with no marker, **and** whitespace-only stderr empties it | silent-failure ×2 |
| #639 | `ready-for-agent` | replace the `nearTotal`/`near.length` index arithmetic with one `withheld` slice | simplify |
| #640 | `ready-for-agent` | `1 further near-miss(es)` — the file already has the singular/plural idiom | types |
| #641 | `needs-triage` | a malformed row in gh's 6-row window degrades a hit to `unverified` — **refuted as unreachable** | silent-failure |
| #642 | `needs-triage` | `nearTotal`'s floor is a 0.005 score threshold, not "shares a content word" — **refuted as unreachable** | silent-failure |
| #643 | `needs-triage` | the gh row shape guard typechecks `number` only | silent-failure + types |
| #644 | `needs-triage` | the nested ternary at `:511` — **refuted**, it is the house idiom at 9 sites | simplify |
| #645 | `needs-triage` | `[1,2,3,4,5].map(filler)` spelled twice — **refuted**, extraction costs a line and adds shared state | simplify |

#638 and #643 each merge two findings that land on one expression — two tickets would have put two agents on the same line. Every refuted finding carries its refutation in the issue body.

**#639 and #640 were deferred for ordering, not merit:** both edit the exact lines the new pin covers, and a simplification plus a new pin over the same lines in one pass is the stale-pin hazard. The pin is now in place, so whoever takes them has it first.

Not filed: one `correctness` finding whose refutation was upheld 2-0, and one comment-citation finding the controller measured as correct by the claim it makes.

Issue #348, opened 2026-08-11
ci.yml: the Shellcheck comment enumerates two info-level hits; there are at least three (SC2015 on inflight.sh)


feigi:
`.github/workflows/ci.yml`, the comment above the `Shellcheck` step (~line 146 — re-derive at head).

**Pre-existing; NOT caused by PR #332.** Surfaced by that PR's review because it is adjacent to what the diff touches.

## The finding

The comment claims the only info-level shellcheck hits in the tree are two named false positives:

> the only info-level hits in the tree are false positives (SC2016 on inflight.sh's single-quoted `python3 -c` script … SC1003 on statusline-command.sh's OSC-8 escape)

Running shellcheck at default severity on `inflight.sh` **alone** already shows a third info-level hit the enumeration omits:

```
$ shellcheck skills/fleet/scripts/inflight.sh
In skills/fleet/scripts/inflight.sh line 80:
pr=$(printf '%s' "$pr_json" | NUM="$n" LINKED="$linked" python3 -c '
    ^-- SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.

In skills/fleet/scripts/inflight.sh line 370:
  && wt_j=$(jstr "$wt") && wt_rw=$(jrewritten "$wt") \
                        ^-- SC2015 (info): Note that A && B || C is not if-then-else. C may run when A is true.
```

The comment's next sentence — "Raise this to `style` only after silencing those two at the source" — rests on the count being complete, so a future maintainer raising the severity gate would hit an unlisted third hit. CI runs `-S warning`, so none of these fail today; this is a comment-accuracy defect, not a red build.

## Fix

Re-run shellcheck at default severity across the whole tree (`git ls-files '*.sh' | xargs -r shellcheck -x`), and either enumerate every info-level hit or reword the sentence so it stops claiming an exhaustive list.

## ⚠️ Sequencing

**PR #335 is open against this same workflow file** (node-version pin → version file). Whoever takes this must sequence against #335 — rebase onto it or wait for it to merge — rather than opening a third concurrent PR on `ci.yml`.

`dimension`: `comments`

Deferred from PR #332 review (closes #92). Controller ruling: out of scope for #92, and pre-existing.

feigi:
## A second stale comment in the same file: `ci.yml`'s "29 flat files"

[...]

PR #636, opened 2026-08-18
fix(ci): refuse an empty match instead of a green step that checked nothing

### LINKED ISSUE #161: ci.yml: the four xargs checks are byte-identical in the log whether they checked 23 files or 0
Deferred from PR #157 review.

`.github/workflows/ci.yml` — three steps are `git ls-files '<glob>' | xargs -r <checker>`:

```yaml
run: git ls-files '*.mjs' | xargs -r -n1 node --check
run: git ls-files '*.json' | xargs -r -n1 jq empty
run: git ls-files '*.sh'  | xargs -r shellcheck -x -S warning
run: git ls-files '*.py'  | xargs -r -n1 python3 -m py_compile
```

`xargs -r` exits 0 on empty input, and all four checkers print nothing on success. So **a run that checked 23 files and a run that checked 0 are byte-identical in the log** — confirmed from the real runner log of run 30629390698, where each step emitted zero bytes between `##[endgroup]` and the next `##[group]`.

`pipefail` protects against the left-hand side *failing*. It cannot protect against the left-hand side *succeeding with an empty result*, which is the trigger: one directory rename and a step is vacuously green forever.

PR #157 closed this for the `Tests` step with `shopt -s failglob`. The four `xargs` steps were left, because `failglob` does not apply (the glob is quoted and expanded by `git ls-files`, not the shell) and the fix is a different shape.

## Fix

Per step, make the count visible and an empty match fatal:

```yaml
run: |
  mapfile -t F < <(git ls-files '*.mjs')
  echo "checking ${#F[@]} file(s)"
  [ "${#F[@]}" -gt 0 ] || { echo "::error::no files matched — this check verified nothing"; exit 1; }
  printf '%s\0' "${F[@]}" | xargs -0 -n1 node --check
```

This also picks up the `xargs -0` hygiene item: no tracked path currently contains a space or quote (`git ls-tree -r HEAD --name-only | grep -E "[ '\"]"` → none), and the failure direction is loud rather than silent, so it is hygiene rather than a live bug — but it comes free with the rewrite.

feigi:
Closes #161

## What was wrong

`git ls-files '<glob>' | xargs -r <checker>` exits 0 on an empty match, and every checker in the `check` job prints nothing on success. So **a run that checked 23 files and a run that checked 0 are byte-identical in the log** — confirmed from runner log 30629390698, zero bytes between `##[endgroup]` and the next `##[group]`.

`pipefail` protects against the left-hand side *failing*. It cannot see the left-hand side *succeeding with an empty result*, which is the trigger: one directory rename and a step is vacuously green forever. `shopt -s failglob` — PR #157's fix for the `Tests` step — does not apply either: the glob is quoted and expanded by `git ls-files`, not by the shell.

## The fix

`.github/scripts/check-tracked.sh '<glob>' <xargs args and command...>` prints the count and exits 1 on zero. The four `xargs` steps route through it:

```yaml
run: .github/scripts/check-tracked.sh '*.mjs' -n1 node --check
```

Everything after the glob goes to xargs verbatim, which is why `-n1` stays at the call site — shellcheck wants all its files in one invocation, the other three want one file each. NUL-delimited end to end, which picks up the `xargs -0` hygiene item the ticket noted: no tracked path holds a space or a quote today, and the day one does the failure is not silent.

## The count is four steps, and a fifth member the ticket did not list

The ticket's title says four and its prose says "three" above a four-line block. At `origin/main` there are **four** `xargs` steps (ci.yml:96, 154, 162, 167) — the prose was wrong.

The `Syntax-check .js scripts` step is the same defect in a different shape: `for f in $(git ls-files '*.js')` iterates zero times on an empty match and exits 0. Its per-file `case` dispatch is what keeps it off the shared script, so it carries the count and the refusal inline. Covering the four named steps and leaving it would have shipped this ticket with its own defect still in the same file.

Deliberately left, with reasons:

- **`Tests` step** — already guarded by `shopt -s failglob` (PR #157).
- **`rebase-check` job** — not this shape; it compares SHAs and errors rather than iterating a list that can empty.
- **`rebase-check-refresh.yml`** — same class, already handled, and correctly the *other* way: `if [ -z "$PRS" ]` announces the no-op and exits 0, because zero open PRs is a legitimate everyday state. It also already carries `BEHIND>0 && RERAN==0` and `FAILED==TOTAL` accounting guards. Making emptiness fatal there would be a false positive.

Only a glob that legitimately empties. All five match files today — mjs 57, js 2, json 2, sh 14 (13 + the new script, which the `Shellcheck` step therefore lints), py 7 — so the refusal cannot fire on this tree. The ticket flagged `*.py` specifically as a maybe-empty glob: it matches 7 files under `skills/caveman-compress/scripts/`. If that skill is ever deleted, CI goes red asking for a one-line edit here instead of going quietly green — which is the intended trade, and the script says so.

## Tests

`skills/fleet/scripts/ci-vacuous-green.test.mjs`, 6 cases, both halves:

- **refuse** — empty match is exit 1, the message names the glob, and the checker never runs
- **accept** — a populated tree still runs the checker on every file, prints the count, exits 0, and the glob still selects
- a failing checker still fails the step (the guard does not swallow the real check's status)
- a path holding a space reaches the checker as one argument
- source assertions that ci.yml actually routes through the script, and that `xargs -r` is gone

Mutation-tested — six mutants, each killing exactly one test and a different one: delete the refusal → refuse case only; `xargs -0` → `xargs` → space case only; append `|| true` → status case only; drop the count line → accept case only; revert the `*.py` step to `xargs -r` → routing pin only; delete the `.js` refusal → `.js` pin only. Two controls stay green: reflowing ci.yml's comments and re-indenting the `.js` block, and renaming the script's internal array.

Suite: 922 pass, 0 fail.

**Coverage this does not have:** the workflow YAML is not executed by any test. The helper script is (four real cases against throwaway git repos), and the source assertions pin that ci.yml calls it. The `.js` step's inline guard is pinned by source assertion only — its behaviour was verified by hand, extracting the step body from the YAML and running it under `bash -eo pipefail` (GitHub's `shell: bash`): populated tree → `checking 2 file(s) matching *.js`, exit 0; empty tree → `::error::`, exit 1.

https://claude.ai/code/session_01Npo59xF5tMhME8aTFpEGWq


feigi:
Review findings applied and deferred (fix pass on `9d8f48c` → `a67a6db`).

**Applied** (one commit, `a67a6db`):

- `survived[0]` (critical, tests) — the `.js` step's assertion pinned only the `::error::` message, so deleting just the `exit 1` at `ci.yml:138` left all six tests green. It now requires the `exit 1` adjacent to the message.
- `survived[1]` (important, tests) — a routed `run:` line is now matched unflattened and anchored at both ends, so ` || true` appended to it goes red; `continue-on-error: true` is refused too.
- `refuted[0]` + `unverified[0]` (one edit) — `check-tracked.sh` refuses a call with fewer than two arguments. Handed no checker it exited 0 after `xargs` fell back to its default `echo`; handed nothing at all it died on bash's `unbound variable` with no `::error::` line naming the script. Exit 1, the code the empty-match branch already uses. Two behavioural tests cover it.
- `unverified[2]` (suggestion, comments) — the header comment's runner-log citation is now marked as an observation from outside this repo that nothing in-tree can settle. The evidence is kept, not dropped.

Each new pin was mutation-verified against the exact edit it must catch (delete `exit 1` at `ci.yml:138`; append ` || true` to a routed `run:`; delete the arity guard), and against a change it must **not** catch (re-indent the routed line; reword the guard's message). `node --test skills/fleet/scripts/*.test.mjs`: **tests 925 / pass 925 / fail 0**.

**Deferred** (1):

- `unverified[1]` (suggestion, comments) — `ci.yml:186-187` quotes `claim-ticket.sh` as saying "this **filter**, not node …"; the file says "this **walk**, not node …". Out of #161's scope (pre-existing, untouched by the diff, belongs to the Tests step's vendored-tree argument). Recorded on #348, which already owns two rotted `ci.yml` comments in that same block: https://github.com/feigi/claude-config/issues/348#issuecomment-5333054352

`refuted[1]` (the "29 flat files" count) was already recorded on #348 by an earlier run; left alone rather than duplicated.

Issue #574, opened 2026-08-17
verify-sha.test.mjs: nothing pins that a FAILED FETCH is fatal — downgrading the guard to a warning passes the whole suite


feigi:
Surfaced by a refuter during PR #565's review, as a by-product of refuting a *different* finding. Not part of any finding, so it is filed here rather than left buried in the `wontfix` record at #573. Deduped: `ledger.mjs check` → `tracker-hit` on #573 only, which is the refuted-record ticket that carries the measurement, not this work.

## Measured

`skills/fleet/scripts/verify-sha.sh:19` is:

```sh
git fetch --quiet origin "$branch" || die "cannot fetch origin/$branch"
```

**Downgrading that `|| die` to `|| echo … >&2` passes the entire suite, 9/9.** Nothing in `verify-sha.test.mjs` asserts that a failed fetch is **fatal** rather than advisory.

## Why it matters

`verify-sha.sh` is the fleet controller's gate for every SHA it enqueues, and its exit vocabulary is load-bearing: `0` reachable, `1` not reachable, `2` the question could not be answered. If a failed fetch stopped being fatal, the script would continue on **stale refs** and answer `reachable: false` — a confident negative verdict computed against a branch it never updated. That is the precise collapse #266 and #559 both exist to prevent, reachable through the one guard nobody pinned.

The `die` is correct today. The gap is that **no test would notice if it stopped being.**

## Desired behavior

A test that fails when the fetch guard is not fatal. It should assert the full failure signature, not just one half:

- exit **2** (not 1, not 0),
- **no JSON on stdout** — a question that was not answered emits no verdict,
- git's own diagnosis reaches stderr.

**Assert on stderr content, not only exit code and empty stdout.** Those two agree across genuinely different causes, and this same file already has a measured near-miss of exactly that shape: a test that looked like it pinned the `merge-base` probe was actually pinning a `git fetch` failure, because a broken shim killed the fetch first and *that* is also exit 2 with no JSON.

## Fixture note

The existing suite reaches a failing fetch by pointing the remote at a nonexistent path — see the tests added by PR #565, which measured that a bad remote URL and a removed remote produce byte-identical output. Reuse that shape rather than inventing a shim; PR #565's own work established that git is silent and exits exactly 1 on a plain negative, so the healthy path stays quiet.

**A test you add must kill its own mutant:** downgrade the `die` to a warning, confirm this test and only this test reds, restore — then a change it should *not* catch, staying green.

## Out of scope

[...]

PR #646, opened 2026-08-18
fix(refresh): fall back to the older candidate runs when a rerun POST is rejected

### LINKED ISSUE #159: rebase-check-refresh.yml: a rejected rerun never falls back to the older candidate runs it already fetched
Deferred from PR #157 review.

`.github/workflows/rebase-check-refresh.yml` — the job-resolution loop walks `$RUN_IDS`, takes the **first** run containing a `rebase-check` job and `break`s. The rerun POST then happens once, outside that loop.

If that POST is rejected because the parent run is still in progress, the older candidates in `$RUN_IDS` — precisely the *completed* runs that would accept a rerun — are never tried.

`per_page=5` was chosen specifically to have those fallbacks in hand (see the comment about cancelled stubs); the code then discards them. The file's own comment presents this as an inherent timing race. It is not inherent.

## Fix

Move the rerun POST inside the `for RUN in $RUN_IDS` loop; `break` only on a 2xx. Keep the 401/403 fatal branch and the in-progress/queued skip branch (added in #157) where they are, and make sure `REJECTED` increments once per PR, not once per candidate.

## Why deferred, not applied in #157

It restructures the loop rather than adding a guard, and it interacts with the counters. #157 kept its refresh-workflow changes to guards and classification so this behavioural change lands reviewable on its own.

feigi:
Closes #159

## The defect

`.github/workflows/rebase-check-refresh.yml`'s job-resolution loop walked `$RUN_IDS`, took the **first** run holding a `rebase-check` job, and `break`ed. The rerun POST then fired once, *outside* that loop.

So a POST rejected because the parent run is still in progress never reached the older candidates — precisely the *completed* runs that would have accepted it. `per_page=5` was chosen to have those fallbacks in hand; the code fetched them and threw them away, and the file's own comment presented the result as an inherent timing race. It was not.

## The change

The candidate walk moves into `.github/scripts/rerun-rebase-check.sh`, which POSTs per candidate and returns only on a 2xx. The 401/403 fatal branch and the in-progress/queued skip branch keep their order and their bodies — GitHub answers an in-progress rerun with a 403, so testing that first is what stops a routine skip being diagnosed as a broken token.

### The counter

The ticket's hard clause: `rejected` must count PRs, not candidates. There were two `REJECTED=$((REJECTED + 1))` sites, both inside what is now a retry loop. Both become plain log lines, and the classification comes back as an **exit status** — `0` re-run, `3` rejected, `4` no rebase-check job, `1` fatal — which the caller maps to a counter in a `case`. The script runs once per PR and exits once, so a PR that tried three candidates cannot increment anything twice. `case`'s `*` arm catches every status it does not name (127 for a missing script, 2 for a bash error), because `|| STATUS=$?` disarms `set -e` for the whole call and an unnamed status would otherwise be a silent no-op.

## Why a script, and how it is tested

Nothing in this repo executes a workflow. Left inline, the only available pin is an assertion over the `run:` block's source text — and #161's review measured that such assertions survive deleting the behaviour they name. Extracted, the loop runs for real: `skills/fleet/scripts/rerun-rebase-check.test.mjs` lifts the step's shell out of the YAML, runs it under `bash` with a stub `gh` on `PATH`, from the repo root — so the `rerun-rebase-check.sh` it invokes is the shipped script, not a copy. The lift asserts non-emptiness and that the step still routes through the script, so an indentation change upstream cannot hand every test an empty string.

Nine cases: the fallback (a rejected candidate is followed by a POST to the next; a 403 saying "in progress" is a skip); the counter (three rejected candidates summarise `rejected: 1`, and so does the foot-of-step warning that reads it a second time, asserted in the two-PR case where it is reachable); and what must **not** happen (an accepted first candidate POSTs exactly once, a token failure exits on the first candidate rather than retrying five times per PR, a PR with no `rebase-check` job POSTs nothing, an up-to-date PR touches no run at all, and an unnamed exit status fails the step).

Every one was mutation-checked. Restoring the original defect (`break` after the first rejection) reddens the four fallback/counter tests and leaves the five must-not tests green; counting per candidate reddens exactly the two counter tests; reordering the 403 branches reddens exactly the ordering tests; dropping the `case`'s `*` arm reddens exactly one test. Renaming an internal variable stays fully green.

## Measured, not assumed

The all-or-nothing guard (`BEHIND > 0 && RERAN == 0` → exit 1) exits *before* the skipped-PR warning at the foot of the step, so a run where every behind PR was rejected never prints that warning. Pre-existing, unchanged here — noted because it is where the counter assertion had to move to be reachable.

## Not covered

- A run whose jobs could not be listed still classifies the PR as `no-job` rather than `failed` if no other candidate has one. Pre-existing misclassification, unrelated to the rerun path, left alone.
- `printf … | grep -q` under `pipefail` can in principle read as false on a SIGPIPE. Lifted verbatim; it cannot fire for `gh` error strings, which are far under a pipe buffer.
- Tests: 941/941 pass. `shellcheck -x -S warning` clean over 15 `.sh`; `node --check` clean over 59 `.mjs`.


feigi:
### PR #646 review — applied / deferred

**Applied** in `0e5c38d` (`node --test skills/fleet/scripts/*.test.mjs` → **943 pass, 0 fail**, up from 941):

- **`survived[0]` (critical, correctness)** — the `refresh` job had no `actions/checkout`, so `.github/scripts/rerun-rebase-check.sh` did not exist on the runner: 127 on the FIRST behind PR, `*` arm exits 1, nothing re-run. Added `actions/checkout@v5` (matching `ci.yml:36`/`:78`) and pinned it as text ahead of the `run:` block, since `runStep()` always runs from a checked-out tree and structurally cannot observe the omission.
- **`survived[1]` (critical, tests)** — *"an unexpected exit status … instead of vanishing"* passed with the `*` arm's own `exit 1` deleted, because its single-PR fixture left `RERAN=0` and the downstream all-or-nothing guard produced the same status and matching text (the #574/#565 shape). Rebuilt on two PRs — one re-run normally so that guard is silent, one whose jobs payload is unparseable so jq aborts the script with an unrouted 5 — asserting the step wrote **no summary at all**.
- **`survived[2]` (important, tests)** — added the missing test for the jobs-listing-failure → next-candidate fallback (`rerun-rebase-check.sh:48-52`).

Every pin mutation-verified in a scratch copy, never the worktree: removing the checkout reds **only** the deployment pin (bumping to `@v4` stays green); removing the `*` arm's `exit 1` reds **only** the unroutable-status test, while removing the **downstream** guard's `exit 1` leaves that test **green** — the discrimination that was missing; turning the listing branch's `continue` into `exit 4` reds **only** the new fallback test.

**Deferred and filed:**

| Finding | dimension | Issue | Why |
| --- | --- | --- | --- |
| `unverified[0]` rerun into an older run vs `ci.yml` `cancel-in-progress` | correctness | #648 `needs-triage` | GitHub-side semantics, not executable locally — a real open question, not a guess |
| `unverified[1]` malformed jobs payload → unrouted exit 5 read as a step defect | correctness | #649 `ready-for-agent` | defect confirmed by measurement, remedy open; out of #159's scope (same axis as the `no-job`/`failed` classification left open) |
| `unverified[2]` + `[3]` exit-code table, SIGPIPE note | comments | #650 `wontfix` | **both refuted by measurement** — recorded so neither is re-applied |

Note for #649: the new `*`-arm test reaches an unrouted status *via* exit 5, so routing 5 reds it. It then needs a different unrouted status, not deleting.

PR #647, opened 2026-08-18
fix(review-pr): a run that did no work is unrun, not a pass (#143)

### LINKED ISSUE #143: the 'tests 0' reading rule is too narrow: a partial tree and an all-skipped run both pass it

feigi:
Closes #143

The `tests 0` reading rule read as true and was not: it closes the zero-match-glob
case and nothing else, so three other runs that did no work walked through it.

## What this covers, and what it deliberately leaves

| case | where it is caught | why there |
|---|---|---|
| `tests 0` / absent count | `unrunReason` (already) | unchanged |
| every test skipped — `tests 2 pass 0` | `unrunReason` **+ prose** | the counts say it |
| failing suite, reviewer filed no findings | `unrunReason` | the counts say it |
| **partial tree** — `tests 6 pass 6` of a larger suite | **prose only** | see below |
| container-teardown rationale asserted as local fact | prompt rewritten portably | |
| `Do not substitute a command of your own.` unpinned | pinned | |

**The partial-tree floor is deliberately not in the classifier.** `unrunReason` is
kept pure — `review-pr-unrun.test.mjs` lifts it out of the source text with
`new Function`, so a free variable would throw there — and the suite's size is not
a fact it holds. The agent that ran the command is the only reader positioned to
compare, so that case is a reading rule in the specialist prompt and in
`review-and-fix.md`, and both statements are now pinned.

The ticket's fourth "related" item — the `report that dimension as unrun` follow-through
being unpinned — was already closed by `review-pr-unrun.test.mjs`'s
"the specialist prompt names test_run as where a zero-test run gets reported".
Verified, not re-added.

## Two shapes the ticket proposes that would have been wrong

**A bare `pass 0` clause over-refuses.** A suite where every test FAILED also
reports zero passes, and it is the run that most needs reporting — refusing it is
the over-refusal #137 removed. The clause is `run.pass === 0 && !run.fail`, and
`===` rather than `!run.pass` because `pass` is optional and an absent count is not
a zero one. Both guarded by accept-side pins; mutating to the ticket's literal
shape turns one red.

**`fail > 0` alone would too**, which is why the third case ships as a conjunction
with an empty findings list, exactly as #143's third comment specifies.

## What this newly refuses

The conjunction moves a currently-clean case to unrun, as the ticket says it must.
Blast radius worth knowing: when a PR's suite is red, **every** dimension that files
no findings is now reported unrun, not just one. That is the intended direction — a
red suite nobody wrote up is not a reviewed PR — but it is louder than the single
case the ticket describes.

## Verification

Full suite `./agent-test skills/fleet/scripts/`: **937 tests, 937 pass, 0 fail, 0 skipped**.

Every added pin was mutation-tested — each mutant killed by its own assertion and no
other, then a benign equivalent change confirmed green. Notably `run.pass === 0 && !run.fail`
mutated to the ticket's bare `pass === 0` is killed by the all-failed accept-side pin,
and `run.fail > 0 && !review.findings?.length` mutated to bare `run.fail > 0` is killed
by the pre-existing accept-side pin.

Settled against this tree before restating: `derive-testcmd.sh . HEAD` -> `node --test`;
`review-pr.js:833` carries "your own copy of the snapshot"; `review-and-fix.md` says
"no compose file, no `globalSetup`, and no vitest" and separately records that CI now
exists — #143's comment quotes it as also saying "no CI", which is stale, so this PR
does not restate that. Whether `additionalProperties: false` drops or rejects an
undeclared field is stated both ways in this directory and is not settled here, so no
comment added by this branch asserts either.


feigi:
**Review adjudicated — 16 findings, 1 applied, 5 deferred and filed, 10 closed without an issue.**

Applied here (`0794d73`): the all-skipped clause in `unrunReason` was only ever exercised with `findings: []`, so ANDing `!review.findings?.length` into it survived all 937 tests. The test now loops both arms of `findings`. Mutation-verified on this branch: that mutation reds this test **and only** this test, while dropping the findings conjunct from the sibling `fail > 0` clause leaves it green and reds the two `fail > 0` pins instead.

`node --test skills/fleet/scripts/*.test.mjs` -> **tests 937 / pass 937 / fail 0** at `0794d73`.

**Deferred and filed**

| Issue | Label | Finding |
| --- | --- | --- |
| #651 | `ready-for-agent` | `unrunReason` refuses only exactly-zero passes: `{tests:937, pass:1, fail:0}` classifies clean. Defect confirmed by direct measurement; remedy is an open design choice, and the body carries the four-case measurement plus the measured `todo` hazard that rules out a strict `pass + fail === tests` equality. |
| #652 | `wontfix` | Record of the three in-scope `suggestion`-band findings, each given one refuter and each refuted by measurement — including two whose proposed remedies would have shipped a false statement or a silent hole. Filed so none is rediscovered and re-applied. |
| #653 | `ready-for-agent` | The declared-exactly-once guard does not cover `unrunReason`/`unrunEntries`/`unrunCrashed`, so a duplicate-after-real declaration is invisible to every test that lifts them. |
| #654 | `ready-for-agent` | The cost-and-apply-policy spec cites `review-pr.js:284-286` for the `tests 0` convention; that range is the `BOUND EVERY READ` paragraph. Claim true, citation stale. |
| #655 | `needs-triage` | The specialist-prompt slice preamble is open-coded 3x in `review-pr-testcmd.test.mjs`. Whether to extract locally or retire all 12 hand-rolled copies is the open question. |

**Not applied, not filed:** seven `important` findings whose refutations were upheld (including both prompt-channel claims, the absent-`fail` claim, and the `run-team/SKILL.md:817` echo — that block already orders the agent to read `review-and-fix.md`, which this PR does widen), plus three `suggestion`s that are moot: one asks for a pin on a rule the upheld refutation says should not be there, one's own remedy is "none needed", and one restates #651 more weakly.

PR #658, opened 2026-08-19
test(fleet): make release-ticket.test.mjs's permission fixtures survive their own failure, and refuse to run vacuously under euid 0

### LINKED ISSUE #184: release-ticket.test.mjs: the three 0o000 permission fixtures leak on failure, and go vacuous under euid 0

feigi:
Closes #184

Six fixtures in `release-ticket.test.mjs` reach their condition through `chmod`. Both weaknesses the issue names are confirmed live at `origin/main` **by mutation**, not by reading.

## Six sites, not five

The body names three, the #330 comment adds two more plus a `0o400` leg. There is a sixth the issue names nowhere: **`an unsearchable worktree is not reported as having no .git`** chmods `0o644` rather than `0o000`, so it does not match a `0o000` grep — but root ignores a missing search bit just the same. Covered.

| # | Fixture | Mode |
|---|---|---|
| 1 | `a stray worktree the script may not stat keeps the hand-release remedy` | dir `0o000` |
| 2 | `a worktree directory the script may not stat is unknown, never a release` | dir `0o000` |
| 3 | `an unsearchable worktree is not reported as having no .git` | dir `0o644` **(unnamed by the issue)** |
| 4 | `a worktree the script may not look at is unknown, never a release` | dir `0o000` |
| 5 | `an unreadable worktree registry is unknown, never a release` | dir `0o000` **and** `0o400` |
| 6 | `an entry git cannot read INSIDE is unknown too…` | **file** `0o000` |

## 1. The restore is not failure-safe — fixed in cleanup, not in `release()`

Reproduced by making the script emit a malformed payload: `release()` throws on `JSON.parse`, the restore is skipped, `.worktrees` is left `d---------`, and **the temp dir leaks**. `rmSync` throws `ENOTEMPTY` (measured) and node drops that after-hook error on the floor, so the leak happens with *nothing reported at all* — worse than the issue describes.

**I did not apply the fix the issue prescribes**, because I measured it to be harmful. Guarding the `JSON.parse` inside `release()` swallows a parse failure into `json: null`, and nine `assert.equal(json, null)` cases read `null` as *"refused before any mutation"*:

| Script mutation: `die()` also prints garbage to stdout | Result |
|---|---|
| `origin/main` today | ✖ red — `SyntaxError`, caught |
| with the issue's prescribed `JSON.parse` guard | ✔ **green — regression missed** |

That is a second masking layer, against the issue's own point that a masked failure is worse than the original. Instead the fix restores the search bits in `repo()`'s `t.after` before deleting. Same fix-once property — every permission fixture in the file, including any added later — while `release()` is left completely untouched, so its ~50 call sites keep their current error semantics and the real `SyntaxError` still propagates. Re-measured: **leak gone, real error preserved, no vacuity introduced.**

## 2. `chmod` denies nothing under euid 0

One-line guard on each of the six. Named once (`EUID0`) rather than inlined, deliberately diverging from the sibling suites' fifteen `process.getuid?.() === 0` guards: a guard that fires unconditionally turns all six into skips with nothing failing, and a single named predicate is the only thing a test can pin. `geteuid`, not `getuid`, because the effective uid is what the kernel checks against.

## Verification

The euid-0 half is unreachable on this machine and in CI (`ubuntu-latest`, no `container:`), so **a green suite proves nothing about it and I did not exercise it directly** — no root, no container was used. What is pinned instead is the half a green suite *can* speak to: that the guard is not firing here, and that every mode the six fixtures use really denies.

Mutation battery, full suite each:

| Mutation | Expected | Result |
|---|---|---|
| `EUID0 = true` (guard fires unconditionally) | new test red | ✖ 1 fail + **6 loud skips** — killed |
| reword the skip reason (negative control) | green | ✔ 78 pass — discriminates, not merely fragile |
| `gone "$stray"` → `[ ! -e "$stray" ]` | fixture 1 alone red | ✖ fixture 1 only — **#171 discrimination preserved** |
| registry guard `&&` → `\|\|` | fixture 5 alone red | ✖ fixture 5 only — **#330's `0o400` leg preserved** |
| new test's file leg `0o000` → `0o644` | new test red | ✖ killed |

Each guard sits on a test whose *entire* subject is the permission fixture — including fixture 5, whose two loop rows are both permission modes. None gates a permission-independent assertion.

`./agent-test skills/fleet/scripts` → **949 tests, 949 pass, 0 fail, 0 skipped.**


feigi:
Review deferrals filed: **#660** — `inflight.test.mjs` carries the same two defects #184 fixed here (seven denial-dependent chmod fixtures with inline-only restores, zero euid-0 guards, three `rm -rf` temp-root teardowns that throw on an unrestored 0o000 directory). Out of #184's scope, so not widened into this PR. Dimension: `tests`. Label: `ready-for-agent`.

Issue #165, opened 2026-07-31
rebase-check-refresh.yml: four remaining comments describe behaviour the code does not have


feigi:
Deferred from PR #157 review. The high-severity comment defects were fixed in #157; these four are the remainder, each measured against the tree.

All in `.github/workflows/rebase-check-refresh.yml`:

**1. The "dependent jobs" rationale is inert.** The header says `POST /actions/jobs/{id}/rerun` "re-runs a job AND ITS DEPENDENT JOBS", in load-bearing capitals, as justification for the `behind_by > 0` gate. The API fact is true, but `ci.yml` declares **no `needs:`** anywhere, so `rebase-check` has zero dependents. A reader reasonably concludes `check` gets re-run too. The gate is justified without the clause — re-running flips `rebase-check` itself back to in-progress, which is enough.

**2. The concurrency-supersession note predicts a benign skip where the code can produce a spurious red.** It says a replacement run "will see those jobs as running and skip them". The replacement counts them `REJECTED` — and if *every* behind PR lands there, the `BEHIND>0 && RERAN==0` guard converts the whole run to `exit 1`. Under rapid main pushes that is a red workflow, not a skip.

**3. "every later push re-resolves the same stub"** reads as a push to the PR, which would change `headRefOid` and produce a fresh run. It means every later push **to main**. One word.

**4. "Ordering is created-descending in practice; GitHub does not document a guarantee, hence the walk rather than an index."** The walk still consumes `$RUN_IDS` in returned order and takes the first match, so it inherits exactly the ordering dependency it disclaims. It buys stub-skipping, not order-independence — say that instead.

Also minor, same file: the `if !` comment says a bare assignment would fail "silently, with no annotation". gh's stderr still reaches the log; only the `::error::` annotation is absent. "Silently" overstates it.

Each correction is stated; mechanical.

Issue #160, opened 2026-07-31
rebase-check-refresh.yml: a jobs-API failure is reported as 'no rebase-check job in any candidate run'


feigi:
Deferred from PR #157 review.

`.github/workflows/rebase-check-refresh.yml` — when `gh api .../runs/$RUN/jobs` fails, the loop prints an error and `continue`s **without incrementing `FAILED`**. If every candidate run's jobs query fails, `JOB` stays empty and control reaches the `NO_JOB` branch, whose message asserts the job does not exist.

That points a debugger straight at `ci.yml`'s load-bearing job name when the real cause was an API error — and because `FAILED` was never incremented, neither the `FAILED == TOTAL` guard nor the `FAILED > 0` warning can see it.

The same file guards exactly this confusion ~20 lines earlier for the *runs* query: *"This is an API failure, not a missing run."* The jobs query omits it.

## Fix

```bash
JOBS_ERRS=0
# on query failure: JOBS_ERRS=$((JOBS_ERRS + 1)); continue
if [ -z "$JOB" ]; then
  if [ "$JOBS_ERRS" -gt 0 ]; then
    echo "::error::#$PR ($SHA): could not list jobs for any candidate run ($JOBS_ERRS API error(s)) — an API failure, not a missing job."
    FAILED=$((FAILED + 1))
  else
    NO_JOB=$((NO_JOB + 1)); echo "#$PR ($SHA): no rebase-check job in any candidate run — skipping."
  fi
  continue
fi
```

Approach settled; mechanical.

PR #663, opened 2026-08-19
fix(ci): close the three accounting holes in rebase-check-refresh

### LINKED ISSUE #162: rebase-check-refresh.yml: three accounting holes that exit green having evaluated nothing
Deferred from PR #157 review.

`.github/workflows/rebase-check-refresh.yml` — three related accounting holes where the workflow reports success having evaluated nothing.

**1. An empty `gh pr list` result is indistinguishable from "no open PRs".** `gh pr list` *exiting non-zero* is caught. Exiting **zero with an empty projection** is not — the step prints "No open PRs targeting <branch>" and exits 0. Triggers: a `gh` upgrade changing `--json` field names or `--jq` output shape, or a degraded API returning `[]`. This is the most-frequently-taken path in the workflow, so there is no baseline from which to notice the day it starts lying.

Fix — cross-check the count independently:
```bash
N=$(gh pr list --base "$GITHUB_REF_NAME" --state open --limit 100 --json number --jq 'length')
if [ -z "$PRS" ] && [ "$N" -gt 0 ]; then
  echo "::error::gh reports $N open PR(s) but the --jq projection produced no rows — field names or output shape changed."
  exit 1
fi
```

**2. All-malformed rows exit green.** The malformed-row guard `continue`s past `UPTODATE`, `BEHIND` and `FAILED`. If every row is malformed: `TOTAL=N`, everything else 0 → the `BEHIND>0 && RERAN==0` guard is false, the `FAILED==TOTAL` guard is false → **exit 0**, with a summary reading all zeros. One assertion closes this and every future variant:

```bash
if [ $((UPTODATE + BEHIND + FAILED)) -ne "$TOTAL" ]; then
  echo "::error::accounting mismatch: $TOTAL listed but $((UPTODATE+BEHIND+FAILED)) evaluated — rows were dropped."
  exit 1
fi
```

**3. The 100-PR cap warning is off by one and non-fatal.** `--limit 100` with `TOTAL -ge 100` warns at exactly 100 when nothing was truncated, and above 100 the number missed is unknowable. Low practical risk at this repo's size; if kept, the message should state the limit rather than assert truncation.

Approach stated for all three; mechanical.

feigi:
`.github/workflows/rebase-check-refresh.yml` had three ways to exit green having evaluated nothing. All three are closed; each guard was mutation-tested both ways — the mutation it exists to catch goes red, and a change it should not catch stays green.

## 1. An empty `gh pr list` result

The step now asks for `--json` and projects the rows itself, so `length` and the rows come out of the **same** response. The empty-list exit reads the count; a row count that disagrees with it refuses before any of the list is acted on.

Not the ticket's two-call shape: a second `gh pr list --jq length` disagrees with the first whenever a PR opens or closes between the two calls, and would report that race as a dropped row. Projecting locally also *deletes* the failure mode the cross-check was asked to detect rather than detecting it — a `gh` upgrade changing `--jq` semantics cannot drop rows from a projection gh no longer performs.

Still not covered, and the comment says so rather than implying otherwise: a degraded API answering `[]` is indistinguishable from a quiet week from inside this step.

## 2. All-malformed rows exiting green

One assertion over the counters, not a counter for malformed rows — a counter only covers the `continue` that exists today.

It sums the **terminal** dispositions, not the ticket's `UPTODATE + BEHIND + FAILED`. `BEHIND` is intermediate: a PR is counted behind and then again as re-run, skipped or failed, so that formula reads 4 for three rows when one is behind and then hits an API error — a warn today, not a failure. Measured: the literal formula turns two legitimate runs red, one of them a pre-existing #160 test.

## 3. The 100-PR cap warning — kept, reworded

Truncation is the one member of this class nothing else in the step can see: `TOTAL`, the count gh returned and the accounting assertion all describe the rows that came back, never the PRs past them. Dropping the warning would leave it no surface at all, so it stays. It now states the limit instead of asserting PRs were missed, and the two literals that had to agree (`--limit 100`, `-ge 100`) are one `PR_LIMIT`.

Measured: 0 open PRs on this repo today, so the cap is unreachable in practice — but "unreachable" is not "reported".

## What this changes for a run that used to pass

Three new `exit 1` paths on a previously-green step. The accept side is pinned, since a guard at the foot of a file is exactly how the last fix in this family regressed the default run: 100 up-to-date PRs, a three-disposition mix, and the existing #160 cases all still exit 0.

Every counter and early-`continue` in the step is now accounted for. Terminal: `UPTODATE`, `RERAN`, `NO_RUN`, `NO_JOB`, `REJECTED`, `FAILED` — the six the assertion sums, covering the malformed-row `continue`, both API-failure `continue`s, the non-numeric `behind_by` arm, the up-to-date arm, the no-usable-run arm and all four script-status arms. Not summed, deliberately: `BEHIND` (intermediate, double-counts), `SKIPPED` (derived from three terminals), `TOTAL`/`N` (the totals being partitioned).

Tests: 958 pass in `skills/fleet/scripts/`, 0 fail. Shellcheck clean on the lifted step block.

#165 targets this file and is not staged; none of its items are touched here.

Closes #162


feigi:
Review findings applied in `5348ab5`.

**Applied (4):**
- **F1** `correctness`/important — `jq 'length'` is not an array check: it answers 0 for `null`, the key count for an object, the character count for a string, so the guard's message was unreachable on exactly those three and they died at the unguarded `PRS=` assignment with no annotation. Type check + `if !` on the projection.
- **F2** `correctness` — jq on empty stdin prints nothing at exit 0, so `if !` cannot see it and `N` is left empty; both readers are `[ ]` integer comparisons that abort with `integer expected` and read FALSE, leaving this PR's own headline cross-check inoperative on that input. Verified under `bash` that F1's type check does **not** close this, hence the separate numeric `case` guard.
- **F3** `tests`/important — the empty-list fast exit (`[ "$N" -eq 0 ]`) had zero coverage; `if false; then` left the suite green.
- **F4** `tests`/important — the parse-failure guard had zero coverage. Assertions derived from the post-fix tree, not from the review text.

**Deferred (1):** #665 — `correctness`/suggestion, the accounting assertion failing the run on a single malformed row. **Refuted** by a dispatched refuter on three measurements: the proposed `MALFORMED`-counter variant reopens the all-malformed hole #162 was filed for (exit 1 → exit 0), turns #162's own acceptance test red (`958/957/1`), and the finding's premise reads #162's bug description as an acceptance criterion. Filed as `needs-triage` with the refutation rather than dropped; recommendation is `wontfix`.

**No action (2):** F5 (message overclaim) is subsumed by F1's fix. F7 was a positive finding — the four accounting tests already discriminate in both directions.

**Suite:** `node --test skills/fleet/scripts/*.test.mjs` → `tests 962 · pass 962 · fail 0` (baseline 958 + 4 new).

**Mutation matrix** — each new test kills its own mutant and only its own; benign change stays green; baselines re-measured first and last:

| mutant | tests | pass | fail | red |
|---|---|---|---|---|
| M0 baseline (first) | 104 | 104 | 0 | — |
| `if [ "$N" -eq 0 ]` → `if false` | 104 | 103 | 1 | empty-list fast exit |
| type check → bare `jq 'length'` | 104 | 103 | 1 | not-a-JSON-array |
| numeric guard on `N` deleted | 104 | 103 | 1 | empty-not-malformed |
| projection guard → bare assignment | 104 | 103 | 1 | unprojectable array |
| M5 benign (comment wording) | 104 | 104 | 0 | — |
| M0 baseline (last) | 104 | 104 | 0 | — |

(Matrix scoped to the three test files that read the workflow. Restores were byte copies from a pristine snapshot, never `git checkout --`.)

PR #664, opened 2026-08-19
test(fleet): pin the four unguarded invariants of the default review path

### LINKED ISSUE #215: review-path-default.test.mjs: four invariants of the default review path are still unpinned
Deferred from PR #214 review. `skills/fleet/scripts/review-path-default.test.mjs` now pins the load-bearing invariants of the controller-run review path (default-vs-preference, the fix-applier prompt, the relay/reconcile split, the concurrency cap, the `unverified` rule, the workflow error path, and the cross-file step numbering). Four claims the same PR introduced or renamed are still not pinned — each was mutation-checked and every current test stayed green.

**1. The return shape** — `skills/fleet/skills/run-team/SKILL.md`, the `It returns {pr, head, snapshot, dimensionsRun, survived, refuted, unverified}` sentence.
Mutation: replace the whole sentence with "It returns its findings." → all tests pass. The controller then has no documented field list, and `workflows/review-pr.js`'s actual `return` can drift from it in either direction with nothing to notice. A pin should assert the seven field names against the prose *and*, ideally, against `review-pr.js`'s return statement — that cross-check is what would catch real drift rather than a reworded sentence.

**2. `fix-pr-<pr#>` in the member-naming rule** — `SKILL.md`, the "Names follow the unit of work" paragraph.
Mutation: drop `fix-pr-<pr#>` from that list, leaving only `impl-`/`review-pr-`/`merge-bot-` → all tests pass. The name is the mechanism that carries the `Agent` tool, so a controller reading only that list names the member something else and loses delegation with no error.

**3. `fix-pr-<M>` in the Report table example** — `SKILL.md`, the running-table code block.
Mutation: revert both rows to `review-pr-<M>` → all tests pass. This one has a machine consumer: `skills/fleet/scripts/compute-board.mjs` parses ledger rows for the per-PR member, so the documented example and the parser's regex have to agree. PR #214 fixed the regex to accept both names; nothing keeps the example honest.

**4. The `(fallback path)` qualifiers on two red flags** — `SKILL.md`, the "Tell the reviewer to ping its specialists" and "I relayed it, so the reviewer has it" entries.
Mutation: strip both qualifiers → all tests pass, and the relay red flags read as universal again. That is the same defect the Phase 3 relay assertion exists to prevent, one section further down: on the default path `agent()` returns into the script and no relay ever occurs, so an unqualified relay rule is an obligation that can never be discharged.

## Why deferred

All four are `suggestion`-band coverage gaps, not defects in shipped behaviour — the prose is currently correct, only unguarded. PR #214's review applied every critical and important finding; these were split out rather than grown into the same diff.

## Approach

Settled: add assertions to the existing `review-path-default.test.mjs`, reusing its `section()` helper. Two constraints from that file's own history, both learned the hard way in the #214 review:

- Slice as small as the claim. A regex over a whole section is satisfied by incidental prose elsewhere in it — the reason `fixApplierPrompt()` slices the blockquote rather than the section.
- `section()` takes the **first** `indexOf` hit, so anchor within an already-sliced parent when the anchor text is not unique file-wide.

Mutation-verify each new assertion both ways: the mutation fails it, and a benign reword does not.

Deferred from PR #214 review.

feigi:
Deferred from PR #214's review. `review-path-default.test.mjs` pinned the load-bearing claims of the controller-run review path; four the same PR introduced or renamed went in unguarded.

**The ticket is partly stale — every item was re-derived against the tree, not taken from the issue.**

| # | Ticket says | Measured | Built |
|---|---|---|---|
| 1 | return shape has **seven** fields | **eight** — `dimensionsUnrun` was added after the ticket was filed; `SKILL.md:681` and `workflows/review-pr.js:950` both carry it | yes, cross-checked against the script |
| 2 | `fix-pr-<pr#>` in the "Names follow the unit of work" rule | genuinely unpinned. The existing `:90` assertion slices `### Reviewers` (SKILL.md 645-1015); the naming rule is at **:34**, inside `## Rules that fail silently` — a different site | yes |
| 3 | `fix-pr-<M>` in the Report example | still reads that way (SKILL.md:1382-1383); `compute-board.mjs:31` still parses both names, and its `fix-pr-` branch had **no test at all** | yes, plus the parser cross-check |
| 4 | two `(fallback path)` qualifiers | 2 in SKILL.md, 0 in the test file — unpinned | yes |

## What each pin does

**1 — return shape, cross-checked not transcribed.** Reads `review-pr.js`'s own top-level `return` and compares it to the prose field list, so drift fails in *either* direction: a field added to the script and not the doc, or a name dropped from the doc and not the script. A transcribed list would just be one more copy to drift. Compared as **sets** — field order carries no instruction, and a pin that fires on a reorder is one people learn to edit around.

**2 — the naming rule.** Sliced to the naming *paragraph*. The Reviewers lead-in also says `fix-pr-<pr#>` (pinned separately at the site that dispatches it), and that hit is what any wider match resolves against, leaving this list free to lose the name while green.

**3 — the per-PR member name, in the example and through the real parser.** Literal match on the Report table, plus `doesNotMatch(/review-pr-<M>/)` to catch the half-revert the positive match walks through. Then the doc's own ledger row is fed through `compute-board.mjs`'s actual `parseRow`, so the widened regex and the documented example can only drift together.

**4 — the two relay red flags.** Sliced per *bullet*. On the default path `agent()` returns into the script, so no relay occurs and no specialist can be pinged; unqualified, both read as obligations that can never be discharged.

## Slice boundaries, measured

Both narrow slices were verified load-bearing rather than assumed:

- Revert **both** Report rows to `review-pr-<M>` → a file-wide `/fix-pr-<M>/` **still matches**, satisfied by the refill section's `fix-pr-<M>-b` at SKILL.md:1329. File-wide would be green with the example fully wrong.
- Strip **only** the ping bullet's qualifier → a Red-flags-list-wide `/\(fallback path\)/` **still fires** on the relay bullet's.

## Mutation matrix

Run in an isolated copy under the scratchpad; the worktree was never mutated and every restore came from a pristine copy, never `git checkout --`. Baseline is the first **and** last row.

| row | tests | pass | fail | red | mutation |
|---|---|---|---|---|---|
| R0-first | 15 | 15 | 0 | — | baseline |
| M1a | 15 | 14 | 1 | P1 | drop `dimensionsUnrun` from the prose sentence |
| M1b | 15 | 14 | 1 | P1 | add an `elapsed` field to `review-pr.js`'s return (drift the *other* way) |
| M1-benign | 15 | 15 | 0 | — | reword the sentence's tail, field list untouched |
| M2a | 15 | 14 | 1 | P2 | drop `fix-pr-<pr#>` from the naming list |
| M2-benign | 15 | 15 | 0 | — | reword the paragraph incl. the ticket's own lead phrase, token kept |
| M3a | 15 | 14 | 1 | P3 | revert **both** Report rows to `review-pr-<M>` |
| M3b | 15 | 14 | 1 | P3 | revert **one** Report row (half-revert) |
| M3c | 15 | 14 | 1 | P3 | narrow `compute-board.mjs`'s `reviewer` regex back to `review-pr` only |
| M3d | 15 | 14 | 1 | P3 | revert the ledger example's member to `review-pr-346` |
| M3-benign | 15 | 15 | 0 | — | renumber the ledger example 346 → 350, name shape kept |
| M4a | 15 | 14 | 1 | P4 | strip **both** `(fallback path)` qualifiers |
| M4b | 15 | 14 | 1 | P4 | strip **one** (the ping red flag only) |
| M4-benign | 15 | 15 | 0 | — | reword the ping red flag's consequence, qualifier kept |
| R0-last | 15 | 15 | 0 | — | baseline |

Every mutation kills exactly its own pin and nothing else — zero cross-hits across 15 rows. Every benign reword stays green, so each pin is discriminating, not merely failable.

## Tests

`./agent-test skills/fleet/scripts/` → **tests 958 / pass 958 / fail 0 / skipped 0**.
`review-path-default.test.mjs` alone → **tests 15 / pass 15 / fail 0** (was 11).

Closes #215


feigi:
## Review applied — `cea64c1`

Three findings, **all three applied**, zero review deferrals. Both `suggestion`-band findings (which the workflow budgets 0 refuters) were checked by a dedicated refuter first; both survived, and one of the two prescribed **remedies** was refuted and replaced with a measured one.

| # | dim | finding | verdict | shipped |
|---|---|---|---|---|
| 1 | comments | `:482` "the one claim in this file with a machine on the other side of it" is false as of its own diff — the same commit imports `parseRow` and adds a second machine-backed claim 12 lines below | survived | reworded to name what it is pinned against |
| 2 | correctness | `:579` the `(fallback path)` pins match a literal space against a **raw** slice of hard-wrapped prose, so a pure reflow reds them — the defect `:207` already fixed with `flat` | survived, remedy sound | `flat` applied to both slices, slicing left raw |
| 3 | correctness | `:546` `doesNotMatch(/review-pr-<M>/)` bans the fallback member name anywhere in `## Report`, reding on a legitimate additive row | survived, **remedy refuted** | a **floor**, not the prescribed `=== 2` |

### Finding 3 — why not the prescribed fix

`assert.equal(count, 2)` trades one false-positive class for another: it reds on a **fourth default-path row**, which a table already enumerating three outcome shapes plainly invites. The ban also had a **false negative** neither the finding nor the assertion's own comment mentioned — a row **deleted outright** left it green, the same walk-through its comment claimed to close. A floor (`>= 2`) reds on the half-revert, the full revert **and** the deletion, and greens on every legitimate state. It subsumes the `assert.match` above it, which is gone.

### Finding 2 — widened past the two pins named

Every pin this PR adds was audited for the class, not just the two named. **Only `/\(fallback path\)/` had it.** The rest are `section()` anchors (which fail loudly and self-describingly by design), inside fenced code blocks, or structurally un-flattenable (`/^ {2}(\w+)[,:]/gm` needs its newlines). Each was confirmed by injecting a newline into its own pinned phrase.

### Matrix — 24 rows, 0 mismatches

impl-215's 15 rows re-run against this tree (they measure something different now), plus 9 rows for the states these edits change. Baseline is the first and last row; every restore came from a pristine copy, never `git checkout --`.

| row | tests | pass | fail | mutation |
|---|---|---|---|---|
| R0-first | 15 | 15 | 0 | baseline |
| M1a | 15 | 14 | 1 | drop `dimensionsUnrun` from the prose sentence |
| M1b | 15 | 14 | 1 | add an `elapsed` field to `review-pr.js`'s return |
| M1-benign | 15 | 15 | 0 | reword the sentence tail, field list untouched |
| M2a | 15 | 14 | 1 | drop `fix-pr-<pr#>` from the naming list |
| M2-benign | 15 | 15 | 0 | reword the naming paragraph, token kept |
| M3a | 15 | 14 | 1 | revert BOTH Report rows |
| M3b | 15 | 14 | 1 | revert ONE Report row (half-revert) |
| M3c | 15 | 14 | 1 | narrow `compute-board.mjs`'s `reviewer` regex |
| M3d | 15 | 14 | 1 | revert the ledger example to `review-pr-346` |
| M3-benign | 15 | 15 | 0 | renumber the ledger example 346 → 350 |
| M4a | 15 | 14 | 1 | strip BOTH `(fallback path)` |
| M4b | 15 | 14 | 1 | strip ONE (ping only) |
| M4-benign | 15 | 15 | 0 | reword the ping red flag's consequence |
| N1-reflow60 | 15 | 15 | 0 | reflow Red flags @ 60, words identical — **was 14/1** |
| N2-reflow55 | 15 | 15 | 0 | reflow @ 55 — **was 14/1** |
| N3-reflow80 | 15 | 15 | 0 | reflow @ 80 |
| N4-wrapping | 15 | 15 | 0 | newline inside `(fallback path)` — **was 14/1** |
| N5-addrow | 15 | 15 | 0 | + a CORRECT additive fallback row — **was 14/1** |
| N6-prose | 15 | 15 | 0 | + prose naming `review-pr-<M>`, table untouched — **was 14/1** |
| N7-4th-default | 15 | 15 | 0 | + a fourth DEFAULT-path row — `=== 2` would red this |
| N8-delrow | 15 | 14 | 1 | one row DELETED outright — **was 15/15, a false negative** |
| N9-benign | 15 | 15 | 0 | benign reword of the Report lead-in |
| R0-last | 15 | 15 | 0 | baseline |

Every "was" column is my own measurement against `e5c38fb`, not the reviewer's.

**Width sweep, finding 2.** Unpatched reds at 55 and 60 only; 61-64 and 65-120 green. Patched: green 45-120. At 40 the patched tree reds on `run-team ping red flag: '…' moved — update this test`, which is `section()`'s designed maintenance signal, not a false claim about the doc.

### Tests

`node --test skills/fleet/scripts/*.test.mjs` → **tests 958 / pass 958 / fail 0 / skipped 0**.

### Deferred

- **#667** — `ready-for-agent`. The return-shape pin misses a field added on a line it **shares** with another (`dimensionsUnrun, elapsed: 1,` → 15/15 silent; the same field on its own line → 14/1 red). Confirmed defect, remedy open. Out of this PR's scope: fixing it means changing what the pin parses, not what it claims.

Not filed, reported instead: `/fix-pr-<pr#>/` at `:525` reds with a **false** message if the doc ever breaks inside the backticked token at the hyphen. Confirmed reachable by hand, `flat` cannot fix it (`fix- pr-` still fails), and no markdown wrapper breaks at hyphens inside code spans. No remedy available that is not worse than the exposure.

PR #666, opened 2026-08-19
fix(review-pr): state the model-tier rule the table actually follows

### LINKED ISSUE #221: review-pr.js:81's stated model-tier rule contradicts its own dimension table — silent-failure should be downgraded by it, and isn't
Found reviewing PR #217. The code is right; the one-line rule above it is wrong, and it is the sentence someone will cite later.

## The contradiction

`workflows/review-pr.js:81`:

> `// The rule: downgrade only dimensions whose findings face refuters.`

The table below it downgrades three dimensions to `model: "sonnet"` — `tests`, `comments`, `types` — and leaves `correctness`, `silent-failure` and `simplify` with no `model` key.

**Refuters attach to severity, never to a dimension.** `verifiersBySeverity` (`review-pr.js:189-193`) is `{critical: verifiers, important: verifiers, suggestion: 0}` with `verifiers = 2` (`:161`). Nothing anywhere keys a refuter budget off a dimension. So `silent-failure` findings at critical/important each get 2 refuters exactly like `tests` findings do — and by the stated rule `silent-failure` should therefore be downgraded. It is not.

`correctness` is in the same position, though it has an independent reason to be omitted (below).

## What the actual reasons are

They are already written in the file, just not in the sentence that claims to be the rule:

- **`correctness` and `simplify`** — pr-review-toolkit pins `code-reviewer` and `code-simplifier` to `model: opus` in frontmatter (`:73-79`). Omitting `model` is the only durable lever that keeps the vendor pin, since the frontmatter is vendored third-party and an edit is clobbered on the next plugin update.
- **`correctness` and `silent-failure`** — their misses are silent and permanent. That is the same pair, for the same stated reason, that `SIZE_TIER_DIMS` keeps (`:202-207`).
- **`simplify`** — genuinely faces 0 refuters, but *via severity*: its prompt forces `severity 'suggestion'`, and suggestions are budgeted 0. Its cost is addressed by the size tier instead.

So the omissions have two distinct causes and neither is "this dimension has no refuters".

## Why it is worth a ticket

A false rule in a comment is load-bearing here: the next person tuning cost reads `:81`, applies it consistently, and downgrades `silent-failure` — the dimension whose whole justification is that its misses are silent and permanent, and which the size tier deliberately protects for that reason. The false-negative argument in `:82-83` is the real rule and it argues the opposite.

## Agent Brief

**Decided.** Comment-only change to `workflows/review-pr.js`. No behaviour change, no test change required beyond whatever pins the comment.

Rewrite `:81-84` so the stated rule matches the table. It needs to carry: refuters are per-severity not per-dimension; the downgraded set is the three whose misses are recoverable; the three omissions split into the vendored-`opus`-pin reason and the silent-permanent-miss reason, with `simplify` refuter-free via severity.

**Out of scope — do not change which dimensions carry `model`.** Whether `silent-failure` *should* be downgraded is a separate question needing evidence, not a comment fix. If the rewrite makes that look tempting, file it instead.

**Correction ticket** (`class=correction` under run-team phase 0 step 4): every factual claim restated in the new comment needs a settling command run against the tree first, and the diff must match the stated size — added prose is where minted claims enter. See `skills/fleet/skills/run-team/references/correction-tickets.md`.

**Collides with #220**, which also rewrites `review-pr.js` comment blocks (~70 lines of triple-stated rationale). Sequence these — do not run both in one wave.

feigi:
Closes #221

`DEFAULT_DIMENSIONS`'s stated rule — "downgrade only dimensions whose findings face refuters" — contradicted its own table. Under it, `silent-failure` should be downgraded; it is not, and the size tier deliberately protects it for the opposite reason.

## Why the old rule was false

Refuters attach to severity, never to a dimension. `verifiersFor` takes a severity and nothing else, and `verifiersBySeverity` is keyed `critical`/`important`/`suggestion`. So `silent-failure` findings at critical/important draw the same refuter budget `tests` findings do — "faces refuters" does not separate the six.

## What it says now

Downgrade only where a **miss is recoverable**. The three omissions have two causes, not one:

- vendored `model: opus` frontmatter pin on `code-reviewer` (`correctness`) and `code-simplifier` (`simplify`) — already named in the paragraph above, not restated;
- silent-and-permanent miss for `correctness` and `silent-failure` — the same pair, for the same reason, `SIZE_TIER_DIMS` keeps.

`simplify` does draw 0 refuters, but **via severity**: its prompt directs every finding to `suggestion`, which is budgeted 0. Its cost is paid by the size tier.

The rule now also tells a reader adding a seventh dimension what to ask: is a miss recoverable, and does its agent carry a frontmatter pin worth keeping?

## Settling commands behind each restated clause

Every claim was derived by evaluating the constructs with comments stripped, so the new comment cannot satisfy its own assertions:

| Clause | Settled value |
|---|---|
| "these six" | `DEFAULT_DIMENSIONS` = 6 entries |
| refuter budget keyed on severity | `const verifiersFor = (sev) => verifiersBySeverity[sev] ?? verifiers;` — one param, mentions no dimension key |
| `silent-failure` draws the same refuters as `tests` | `verifiersBySeverity` = `{critical:2, important:2, suggestion:0}`, dimension-independent |
| the downgraded three | `model: "sonnet"` set = `tests`, `comments`, `types` |
| the omitted three | no-`model` set = `correctness`, `silent-failure`, `simplify` |
| "the same pair `SIZE_TIER_DIMS` keeps" | `SIZE_TIER_DIMS` = `{correctness, silent-failure}`, size 2 |
| vendored `opus` pin | all 6 pr-review-toolkit agent frontmatters enumerated: `code-reviewer` and `code-simplifier` are `opus`, the other four `inherit` |
| `simplify` prompt directs `suggestion` | prompt contains `severity 'suggestion'`; budget 0 |

"directs", not "forces": the findings schema enum still permits `critical`/`important`, so the prompt cannot enforce it.

## Second commit — the same false rule, one file over

`select-dimensions.test.mjs` restated the deleted sentence verbatim and named its test "only the refuter-backed dimensions carry a model downgrade". Fixing one copy and leaving the other is this repo's named second-copy defect, and the test file is the copy someone cites as the pin.

Assertions are byte-identical; only the comment and test name change. The edit is deliberately **line-neutral** (6 out, 6 in) because this file is cited by line from `review-pr-reads.test.mjs` (`:211-216`) and `review-pr-snapshot-path.test.mjs` (`:251-256`) — both regions verified unchanged after the edit.

## Out of scope, honored

No dimension gains or loses a `model` key. Nothing filed about whether `silent-failure` *should* be downgraded, because the corrected rule answers it: its misses are not recoverable, so it correctly stays at the session model.

## Verification

- `./agent-test skills/fleet/scripts/` — **tests 968, pass 968, fail 0**
- Mutant: adding `model: "sonnet"` to `silent-failure` turns "only the recoverable-miss dimensions carry a model downgrade" red. Restored from a pristine copy by absolute path.
- Benign reword of the new comment: stays green.


feigi:
Review applied at `675cf131830280a016364161ddb7ebcbb24002ad`.

**Applied (3)**

1. `docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md` — the spec still stated **"Rule: downgrade only dimensions whose findings face refuters"** as governing at `:142`, with the table repeating it at `:159-162`. Added an addendum in the spec's own post-implementation pattern and updated the header's change count. The table is left intact as the dated record of what was decided at `8a84402` — rewriting a past-tense record into a present claim is itself how this repo has minted false claims.
2. `workflows/review-pr.js:401` — `SIZE_TIER_DIMS`'s comment still argued the split in the retired words, while the PR's new block at `:143` cross-references it as holding "the same reason". Rewritten to the recoverable-miss rule. Counts re-derived by lifting the real `selectDimensions`, not carried over: **6 dimensions, 2 kept, 4 dropped** (`tests`, `comments`, `types`, `simplify`), **2 of those 4 carved back in** below (`comments` on a docs file, `tests` on a test file), and `simplify` is the **only** one of the 4 carrying no `model` override.
3. `skills/fleet/scripts/select-dimensions.test.mjs` — new test `the refuter budget is keyed on severity alone, never on a dimension`. Lifts `verifiersBySeverity`/`verifiersFor` from source with `A` and `verifiers` supplied as `new Function` parameters, and pins the claim the corrected rule rests on.

**Mutation matrix** (`node --test skills/fleet/scripts/select-dimensions.test.mjs`; `workflows/review-pr.js` restored from a pristine scratch copy between rows, verified byte-identical):

| # | mutation | result |
|---|---|---|
| 1 | baseline | 28 pass / 0 fail |
| 2 | `(sev, dim) => dim === "silent-failure" ? 0 : …` | 27 / **1** — only the new test |
| 3 | `(sev) => sev === "critical" ? verifiers : 0` | 27 / **1** — only the new test |
| 4 | `(...a) => a[1] === "tests" ? 0 : …` (arity 0) | 27 / **1** — only the new test |
| 5 | benign: param rename `sev` → `severity` | 28 / 0 — green |
| 6 | baseline restored | 28 pass / 0 fail |

**Tests:** `node --test skills/fleet/scripts/*.test.mjs` → `tests 969, pass 969, fail 0` (baseline before the change: 968).

**Deferred and filed (4)**

- #668 `needs-triage` — the `DEFAULT_DIMENSIONS` model-tier invariant is prose plus a hand-enumerated test. Out of scope: a design question, not a comment fix. Body records that the existing guard is stronger than the finding assumed.
- #669 `ready-for-agent` — `selectDimensions`' two `: all` fallbacks (`:444`, `:482`) are unreachable; one production call site, `correctness` survives every filter (2592 stats combos exercised, 0 empty results). Pre-existing and untouched by this diff. Remedy open.
- #670 `needs-triage` — **refuted**: the `tests` recoverability clause vs the "vacuous pin is a recurring defect" line. Different scope of action (drop vs downgrade), and `git log --grep=vacuous -i` → 40 commits is evidence *for* the clause.
- #671 `needs-triage` — **refuted**: the test comment restating the corrected rationale. It is a ~5:1 compression carrying the only `#221` provenance in the tree; both original false copies were born in the same commit `c8fc9d4`, so duplication did not cause the drift.

Not filed: the `refuted[0]` finding on the same spec line, which framed the drift as something this PR caused. The spec was already wrong; the real half is applied item 1 above.

PR #673, opened 2026-08-19
fix(fleet): restore the behind-count qualifier and inline the mechanism in review-and-fix step 6

### LINKED ISSUE #196: review-and-fix.md step 6 dropped the behind-count qualifier and its mechanism, diverging from run-team/SKILL.md:327 and board.mjs:54
Deferred from PR #195 review.

Two related residues of the #182 dedup pass, both in `skills/fleet/commands/review-and-fix.md:11` (step 6).

**1. Bare `staleness` where the twin says `behind-count staleness`.**

- `review-and-fix.md:11` — "a `skipped` heavy job is staleness and fine"
- `skills/fleet/skills/run-team/SKILL.md:327` — "a `skipped` heavy job is behind-count staleness and fine"
- `skills/fleet/scripts/board.mjs:54` — "ci-state's verdict already excludes behind-count staleness"

The controller-side twin and the script comment both name the cause; the reviewer-side sentence no longer does. Cross-file drift between exactly these two files has bitten before.

**2. The mechanism behind "which a behind PR never reaches" is gone.**

Step 6 previously read: "**not** on `ci-state --quiet` exit 0: a behind PR — the normal wave case — never reaches full green (its `rebase-check` is red and heavy jobs `skipped` from the behind-count)". It now reads "**not** on `ci-state --quiet` exit 0, which a behind PR — the normal wave case — never reaches". A reader must chain three separated sentences (:25 board shape, :27 behind-count, ci.yml) to reconstruct why.

Two specialists reported this independently.

**Why deferred:** severity `suggestion` — neither costs a wrong action, unlike the step 4 loss fixed in c940227. Prose precision, not a gate defect.

**Fix:** restore `behind-count` at :11 to match SKILL.md:327, and re-inline the six-word mechanism (`its currency gate red and heavy jobs skipped`).

feigi:
Closes #196

Two residues of the #182 dedup pass in `skills/fleet/commands/review-and-fix.md` step 6. One line of prose changed, plus a pin so the drift cannot recur silently.

## 1. The qualifier (cross-file drift, confirmed)

The reviewer-side clause read `a `skipped` heavy job is staleness and fine`; its twin in `run-team/SKILL.md` reads `... is behind-count staleness and fine`. Re-derived against `origin/main` by construct, not by line — the ticket's line numbers had rotted (it cites `board.mjs:54`; the comment is at `board.mjs:108`):

```
$ git show origin/main:skills/fleet/commands/review-and-fix.md | grep -o 'heavy job is [a-z-]* staleness\|heavy job is staleness'
heavy job is staleness
$ git show origin/main:skills/fleet/skills/run-team/SKILL.md   | grep -o 'heavy job is [a-z-]* staleness\|heavy job is staleness'
heavy job is behind-count staleness
```

Bare "staleness" names no cause, so the sentence no longer says *which* staleness is benign. `skipped` off a non-zero behind-count is; a heavy job skipped for any other reason is an unverified suite, and this clause is what stands between that and a `ready-to-merge` label.

## 2. The mechanism — genuinely absent, not merely reworded

The step gave the **consequence** (`an exit-0 gate strands every behind PR unlabelled`) where the ticket wanted the **mechanism**. Those are different propositions: the consequence says what a wrong gate does, not why exit 0 is unreachable. So the mechanism was absent, and I restored it — in the file's own idiom, copied from step 4's existing wording rather than minted:

> ... never reaches: a `rebase-check` red or heavy jobs `skipped` off the behind-count hold it short of full green, so an exit-0 gate strands every behind PR unlabelled.

Settled against the tree, not the issue body:

- `references/ci-and-staleness.md:44-46` — "`rebase-check` fails fast when behind, gates slow jobs, so rest show *stale green* ... then flip to `rebase-check: FAILURE` with `integration`/`mutation` **skipped**".
- `.github/workflows/ci.yml:59-62` — `rebase-check` exits 1 when `MERGE_BASE != BASE_SHA`.

Kept **disjunctive** ("or"), because both twins state it that way (`SKILL.md`: "a `rebase-check` red, or heavy jobs `skipped` off a non-zero behind-count"; step 4: "a `rebase-check` red or heavy jobs `skipped` **solely from a non-zero behind-count**"). That also keeps it true under this repo's own two-job `ci.yml`, where `check` carries no `needs:`/`if:` and so is never the skipped half.

## 3. The pin

Nothing in the suite compared the two documents, so the drift was invisible to CI — which is why this is the second time it has been filed. `skills/fleet/scripts/staleness-qualifier-prose.test.mjs` pins the shared clause in both files.

**Both halves verified.** Dropping `behind-count` from `review-and-fix.md` reds test 1 and leaves test 2 green; dropping it from `SKILL.md` reds test 2 and leaves test 1 green — so the two pins are independent, not one accidentally covering both. Reflow-safe via `phrase()`: each of the clause's six inter-word gaps broken individually, and all six at once, still match, and both containing paragraphs rewrapped at 60-400 cols stay green. The measured ceiling is recorded in the header.

## Sweep for other divergences

Compared every mirrored CI construct across the two files (`diff-validating`, `heavy job`, `currency gate`, `behind-count`, `no-ci`/`--declare-no-ci`, `--quiet`, `tests 0`, `--no-verify`, `rebase to`). `review-and-fix.md` uses "staleness" exactly twice — a pointer clause and the divergent one — so item 1 is the only site where one file names the behind-count cause and its twin drops it.

One wording difference found and **deliberately left**: `SKILL.md`'s finisher-duty site says `--quiet` "drops `jobs`", while `review-and-fix.md` says it "drops the per-job `jobs`/`missing`". `ci-state.mjs:459` is `if (!quiet) Object.assign(payload, { jobs, missing })`, so `review-and-fix.md` is the complete one and `SKILL.md` is merely incomplete, not wrong. Out of this ticket's scope and in the file it does not name for edits.

## Tests

`./agent-test skills/fleet/scripts` — **975 tests, 975 pass, 0 fail** (973 baseline at `342e2fe` + the 2 new pins).


feigi:
Review fixes applied at `39b87a8`. Deferrals filed: **#676** (`needs-triage`).

**Applied**

- **`survived` · correctness · `:51`** — the two pins matched the clause anywhere in the file, so the qualifier could be gutted in *both* documents and the suite still pass off a stray copy of the old wording elsewhere. Each document is now cut to the paragraph carrying the rule before matching (anchors `6. Diff-check green` and `**The fix-applier pushes and exits`), the way `ci-state-prose.test.mjs` already does, and a missing anchor asserts rather than silently widening the slice back to the whole file. The anchors go through `phrase()` too, so a rewrap that the clause survives does not break the slice.
- **`unverified[0]` · `:42` (survived refutation)** — the mechanism clause `#196` also asked back, dropped by the same `283d985` dedup pass, shipped unguarded. Now pinned over the same bounded step-6 slice, single-document: `run-team/SKILL.md`'s copy of that gate carries the consequence alone, so there is no twin to compare.

**Deferred**

- **`unverified[1]` · `:54` (refuted)** — the assertion message's "its twin document still names one" is false when both documents drift together, but the two-sided failure prints both `✖` lines and the correct BOTH-documents remedy in the same output, so a reader is not misdirected. Prose imprecision in a diagnostic → #676.

**Mutation matrix** — full suite `node --test skills/fleet/scripts/*.test.mjs` = **976 / 976 / 0** first and last. Mutation rows are file-scoped (`staleness-qualifier-prose.test.mjs`, 3 tests):

| row | tests | pass | fail |
|---|---|---|---|
| baseline | 3 | 3 | 0 |
| qualifier gutted, `review-and-fix.md` | 3 | 2 | **1** |
| qualifier gutted, `SKILL.md` | 3 | 2 | **1** |
| qualifier gutted, both | 3 | 1 | **2** |
| **qualifier gutted in both + decoy line appended to each** | 3 | 1 | **2** |
| mechanism reverted to pre-PR wording | 3 | 2 | **1** |
| **mechanism reverted + decoy line appended** | 3 | 2 | **1** |
| slice anchor moved, `review-and-fix.md` | 3 | 1 | **2** |
| slice anchor moved, `SKILL.md` | 3 | 2 | **1** |
| benign: unrelated sentence added beside the rule | 3 | 3 | 0 |
| benign: 6 inter-word gaps broken singly + all six at once, both docs (14 runs) | 3 | 3 | 0 |
| benign: both paragraphs rewrapped at 60·72·80·100·140·200·300·400 cols (16 runs) | 3 | 3 | 0 |
| baseline (final) | 3 | 3 | 0 |

The two decoy rows are the defect this fixes: before the bound, gutting the live clause and appending one `<!-- historical note: … -->` line carrying the old wording put the suite back to **3 / 3 / 0**.

One measurement note, no change made (per ruling): the header comment's "Python `textwrap` at width 40 hyphen-breaks it" is correct, and width **72** hyphen-breaks it too — the behaviour tracks `textwrap`'s default `break_on_hyphens`, not that one width. The 60-400 col sweep above was therefore run with `break_on_hyphens=False`, which is what a Markdown wrapper does.

Issue #661, opened 2026-08-19
release-ticket.test.mjs: the euid-0 comment block says six EUID0 fixtures; PR #658 made it seven


feigi:
Found by `finisher-pr-658` during its pre-label spot-check, and verified by the controller. **Comment prose only — no code defect.** PR #658 is green, its teardown fix is mutation-verified, and this was filed rather than blocking the merge.

## The defect

PR #658's fix commit `7fdf9f2` added a **seventh** `if (EUID0) return t.skip(NO_DENIAL);` — at `release-ticket.test.mjs:1849`, inside the new `repo()'s teardown deletes a root a fixture left unsearchable (#184)` test, which reaches its condition via `chmodSync(join(r.w, ".git"), 0o000)` and so is a denial-dependent fixture of exactly the class the comment describes.

The same commit edited the comment block and left the count at six.

Measured at `7fdf9f2`:

```
$ grep -n 'if (EUID0) return t.skip' skills/fleet/scripts/release-ticket.test.mjs
495 1391 1628 1671 1747 1781 1849      <- seven
```

Stale wording, all in the same file:

| line | text |
|---|---|
| `:54` | "Named once rather than inlined **six** times" |
| `:56` | "turns all **six** fixtures into skips" |
| `:1814` | "the **six** fixtures above" |
| `:1815` | "would turn all **six** into skips" |
| `:1816` | "indistinguishable, in the summary, from **six** tests that passed" |
| `:1829` | "four of the **six** fixtures rest on it" — needs its own recount, not a blind `six`→`seven` |

Under `EUID0 = true`, **seven** tests now skip, not six. The PR body's mutation table likewise records "6 loud skips", measured at `261c92f` before the seventh existed.

## Also stale in the PR body, not the code

PR #658's body still says the sibling suites carry "**fifteen**" `process.getuid?.() === 0` guards. The shipped comment was corrected to **fourteen** during the review (fifteen was the repo-wide count, and the fifteenth occurrence is `release-ticket.test.mjs:55` — the comment quoting the string, not a guard; this file's real guards use `geteuid`). The body was not updated with it. Nothing reads a merged PR body, so this is recorded for provenance rather than as work.

## Why it is worth a ticket rather than a shrug

[...]

Issue #659, opened 2026-08-19
run-team/SKILL.md: the liveness probe's "prints nothing at exit 0" rationale describes the search variant the same paragraph forbids


feigi:
Deferred from the PR #657 finisher's spot-check. **Prose accuracy only — no recipe misbehaves, and the rule the clause argues for is correct.** Filed because this paragraph exists specifically to teach against unsettled claims, so an imprecise justification inside it is worth more than it would be anywhere else.

## The clause

`skills/fleet/skills/run-team/SKILL.md`, the phase-0 `Still live?` block, argues that dropping the `origin/main` argument from `git log -S` means **"a checkout behind the remote prints nothing at exit 0"**, which is indistinguishable from "the change never landed".

## Measured

Run against this repo by `finisher-pr-657`, using the recipe's own worked example (#206, `skills/fleet/skills/next-ticket/SKILL.md`):

| search string | ref | result |
|---|---|---|
| **old** wording (what the shipped recipe prescribes) | `0dc39ef^` | names `4fd2f73` at exit 0 — **not empty** |
| **old** wording | `4fd2f73^` | empty at exit 0 |
| **new** wording (what the shipped recipe forbids) | `0dc39ef^` | empty at exit 0 |

So the stated symptom — empty output — is a true instantiation only of the **new**-wording search, which the very same paragraph tells you not to use. With the prescribed old-wording search, the realistic failure on a behind checkout is the *other* one the paragraph already documents: it names a refactor commit (`4fd2f73`), and `git merge-base --is-ancestor` returns **exit 0 for both** that and the real fix (`0dc39ef`), so the ancestry gate does not discriminate them.

## Why it is not a defect in the rule

Pinning to `origin/main` is load-bearing either way — unpinned you get a wrong commit or nothing, never reliably the right one. The remedy the clause argues for is right; only the failure mode it names is misattributed.

## Suggested shape

State which search each symptom attaches to, rather than presenting "prints nothing" as the consequence of omitting `origin/main`. Roughly: unpinned, a behind checkout yields a **wrong commit** for the prescribed old-string search and **nothing** for the forbidden new-string one — and the ancestry check clears the wrong commit, so it cannot be the backstop.

## Out of scope

- The `--reverse` hazard itself — already documented in the shipped text and verified (`--reverse` names `4fd2f73`; `--follow --reverse` returns empty at exit 0 on git 2.50.1, and `--follow` with two pathspecs exits 128).
[...]

PR #681, opened 2026-08-19
chore(fleet): record wave 3's four tier-outcome rows

feigi:
Controller-authored chore PR closing out wave 3 of the 2026-08-19 `/fleet:run-team 2 5` run. **Closes no issue.** Data only — four appended rows, no deletions.

Per the standing ruling that controller chore PRs go through `review-pr` like any other, this will be reviewed rather than self-labelled. That ruling has earned itself twice: the two chore PRs reviewed this session carried **six** false claims between them, all mine, all caught before merge.

## What it carries

One row per implementer PR ruled in wave 3:

| row | PR | ticket |
|---|---|---|
| 1 | #673 | #196 — review-and-fix staleness qualifier |
| 2 | #674 | #191 — no-undo-audit test hygiene |
| 3 | #678 | #659 — liveness-probe rationale |
| 4 | #679 | #661 — euid-0 fixture count |

`docs/metrics/tier-outcomes.tsv` is the phase-2 tier guard's only input and is read **across** runs. Rows left in a working tree make the next run derive its floor from an undercount — which has happened, and cost a day of stale guard readings.

## File state after this PR — recounted, not carried forward

**29 data rows**, spanning six `run_date`s (`2026-08-13` … `2026-08-19`).

| class / tier | `closed_own_ticket` |
|---|---|
| `routine` / opus | 13 yes, 0 no |
| `routine` / sonnet | 5 yes, 3 no |
| `correction` / opus | 8 yes, 0 no |

`minted_false_claim`: 18 yes, 11 no.

**None of that is a tier verdict, and the rows do not claim to be one.** Every opus row is post-revert, so tier is confounded with time, and the fleet's own rules changed materially during the period — several of this run's own PRs edited them. Recount from the file before citing any of these figures; a doc-only append lands a row without touching prose that quotes it.

## Verification

- `git diff --stat` → **1 file changed, 4 insertions(+)**, zero deletions. Confirmed no removed lines.
- TSV shape preserved: one physical line per row, tab-separated, no wrapped rows.
- Data only — no script, no doc, no test, no behaviour change.
- The maintainer's uncommitted `settings.json` was deliberately left unstaged and untouched.

## Note on the notes

The `note` column on these four rows is unusually long, because wave 3 produced several results worth recording in full — a correction that caught itself three levels deep (#678), an implementer that walked into the self-inclusion trap and backed out (#679), and two cases where a controller ruling was corrected by measurement. The column has no length convention and is read by humans, not parsed, so this is deliberate rather than sloppy. Worth flagging in case a reviewer reads the length as accidental.


feigi:
Fix-applier report — 1 applied, 0 deferred, 0 issues filed.

**Applied** — row `678`, column 7 `minted_false_claim`: `no` → `yes`, plus the one note clause that stated it as `no`. Column 7 only; `closed_own_ticket` untouched, and the rest of the note's past-tense narrative left as written.

The header rule (`:26-27`) has no survived-to-merge qualifier, and the parent-commit probe — re-run here against blob SHAs rather than a working tree — says the diff added it:

| commit | blob | `grep -c 'goes empty only'` |
|---|---|---|
| `e23291e^` | `f7884df` | 0 — absent |
| `e23291e` (implementer) | `ecee2bd` | **1 — introduced** |
| `4338765` (fix-applier) | `74867a8` | 0 — corrected |
| `origin/main` | `74867a8` | 0 |

Precedent checked rather than assumed: row `664` is coded `yes` although its offending comment never reached merge either — `git grep 'machine on the other side' origin/main` hits only this file's own 664 row, never the source.

**Not filed** — the suggestion that row `679` carries the same defect. It does not. Same probe on `and 0o644 drop the search bit` in `release-ticket.test.mjs`: the clause was introduced by `ea18929`, which `git merge-base --is-ancestor ea18929 9074e49^` confirms predates the implementer's commit. Inherited, which the header rule exempts explicitly. `679 … no` is correct as recorded.

**Shape:** 58 lines before and after, 7 tabs on every data line before and after, trailing newline intact, `git diff --numstat` = `1 1`.

**Tests:** `node --test skills/fleet/scripts/*.test.mjs` from the worktree — `tests 978 / pass 978 / fail 0 / skipped 0`. No test covers this data file, so the count is unchanged.

Pushed `508d5aa`.

Issue #603, opened 2026-08-18
board.mjs: three hand-rolled warn-once idioms collapse into one warnOnce(key, msg) helper


feigi:
Deferred from PR #598 review (issue #325). Dimension: **simplify** (`unverified[6]`, severity `suggestion`).

`skills/fleet/scripts/board.mjs` — verified against `e4b07d4`:

- `warnedMeta` — declared :215, gate at :233-234
- `warnedNoSpendDir` — declared :281, gate at :309-310
- `warnedSkips` — declared :288, gate at :348-349

Three module-level names and three `if (!seen) { seen.add(); console.error(...) }` blocks doing one thing. They collapse into one `warnOnce(key, msg)` helper over a single Set, with byte-identical stderr and model output.

## Why this was deferred, not applied

The controller ruled it out of PR #598 explicitly. It is a reasonable simplification, but:

1. It would invalidate the warn-once test that PR #598 adds in the same commit, which is written against `warnedMeta` directly (`a broken sidecar warns ONCE across ticks, not once per tick`, `board.test.mjs`).
2. #325 is a **correction ticket**, whose diff must match its stated size. A refactor of two untouched call sites is not that.

Neither is an argument against the change — only against landing it inside #325.

## Note for whoever picks this up

`warnedNoSpendDir` is a bare `let` boolean, not a Set — it has no key. Folding it into a keyed helper needs a sentinel key (or leaving it alone). Decide that before starting; it is the only part of the collapse that is not mechanical.

The suite is `node --test skills/fleet/scripts/*.test.mjs` (830 tests at `e4b07d4`). `tests 0` is a failed run, not a pass.

feigi:
The PR #912 review (ticket #606) re-derived this finding **independently** — its `code-simplifier` dimension raised it without knowing this issue exists, from a fresh grep of `board.mjs`. Recording that here rather than filing a duplicate, and adding what the review measured.

**`warnedLines` is now a third call site.** #912 adds it at `board.mjs:247`, so the count is `warnedMeta` (:245) + `warnedLines` (:247) + `warnedSkips` (:330) — three declarations and three structurally identical `has`/`add`/`console.error` blocks. #912's implementer was explicitly instructed **not** to introduce a `warnOnce` helper, so the two changes would not collide; that instruction is why the third site landed hand-rolled.

[...]

Issue #606, opened 2026-08-18
board.mjs:244: the per-line catch is position-blind — a mid-file corrupt transcript line drops a turn's whole cache spend as silently as a torn last line


feigi:
Deferred from PR #598, which named it and left it: "*the per-line `catch { continue; }` (`board.mjs:230`) — a transcript appended to while being read has a torn last line on every tick, so warning here would be pure noise. Over-broad in that a mid-file corrupt line is skipped the same way; separate ticket.*"

Filing it so the promise survives the merge that closes #325. (Line is `board.mjs:244` on the PR #598 head `e4b07d4`.)

## The defect

`skills/fleet/scripts/board.mjs:242-244`, inside `readAgent`'s single-pass transcript loop:

```js
for (const line of readFileSync(file, "utf8").split("\n")) {
    if (!line) continue;
    let j; try { j = JSON.parse(line); } catch { continue; }
```

The comment above it (`board.mjs:182-184`) justifies the skip with exactly one cause:

> Malformed lines are skipped rather than fatal: a transcript being appended to WHILE we read it will have a torn last line, every tick.

That reason is correct and is a real argument against throwing. But the catch is **position-blind**: it applies to every line in the split, so a corrupt line in the **middle** of a completed transcript is dropped on the identical path, in the identical silence. The stated cause explains the tail only; the code covers the whole file.

This is the same shape as #325 one layer up in the same function — a catch whose justification a narrower guard has already fully covered, silently absorbing a wider class of fault than its comment claims. #598 fixed the sidecar half; this is the line half.

## What a dropped mid-file line costs

Not nothing — the loop bills spend off these lines (`board.mjs:246-268`):

- A turn's `cache_creation_input_tokens` / `cache_read_input_tokens` are billed **once, on the first line seen for a given `message.id`** (`board.mjs:253-263`). Lose that first line and the entire turn's cache spend vanishes, `maxCtx` with it — every later line of the same turn finds the `turnById` entry missing and opens a *new* turn, so the loss is silent and the turn count is also wrong.
- `output_tokens` is a streaming snapshot taken as a running `Math.max` (`board.mjs:265`). Lose a later line and the max may settle below the turn's true final value.

Either way the spend panel under-reports with 0 bytes on stderr — the silent-telemetry-skew class #598 was filed against.

## Remedy is open, and the naive fix is wrong

[...]

Issue #605, opened 2026-08-18
board.mjs:113: mapCi swallows an UNPARSEABLE ci-state payload into the same "unknown" as no-run-yet — the red-ci flag is suppressed for that PR in silence


feigi:
Deferred from PR #598, which named it explicitly and left it: "*`mapCi`'s `catch { return "unknown"; }` (`board.mjs:113`) — silent, but no false label, and `unknown` is a documented safe value rather than a confidently wrong one. It is pipeline-stage data, not side-channel telemetry, so its blast radius is different. Separate ticket.*"

Filing it so the promise survives the merge that closes #325.

## The defect

`skills/fleet/scripts/board.mjs:110-113`:

```js
export function mapCi(ciJson) {
  if (!ciJson) return "unknown";
  let d;
  try { d = JSON.parse(ciJson); } catch { return "unknown"; }
```

The catch collapses an **unparseable** ci-state payload into the same `"unknown"` that a still-running run and a not-yet-created run produce. Three different states, one indistinguishable output, nothing on stderr.

The silent path is reachable, and it is reachable precisely where the loud path was deliberately declined. `runCiState` (`board.mjs:97`) prints to stderr **only when stdout is empty** — that is by design, and its comment at `board.mjs:90-96` says why:

> ci-state.mjs exits 0 for green, 1 for not-green, 2 for a hard failure — and on exit 1 it has ALREADY printed its verdict JSON to stdout before exiting. So a thrown non-zero exit whose stdout is non-empty is a real verdict (feed it to mapCi); only an empty stdout (the exit-2 die() path) is a genuine read failure.

So **non-empty but unparseable** stdout — a truncated pipe write, a warning line interleaved ahead of the JSON, a crash after a partial write — is handed straight to `mapCi` and swallowed there. `board.mjs:404`:

```js
ci[p.number] = out === null ? (prevCi.get(p.number) ?? "unknown") : mapCi(out);
```

`out` is non-null, so the previous tick's value is not even reused as a fallback — a PR that was correctly `red` last tick reverts to `unknown` on this one.

## What it costs

The same comment states the consequence of a `mapCi` that cannot see red, for the adjacent failure mode it was written about:

> every non-green PR reads as "unknown" and the red-ci flag, the top of the attention strip, never fires.

Scope is **per PR**, not board-wide: one malformed payload suppresses the flag for that PR only. That is the difference in blast radius #598 cited, and it is why this is a separate, lower-severity ticket rather than part of that PR.

[...]

Issue #602, opened 2026-08-18
board: a meta-sidecar fault never reaches the browser — reviewPct stays silently wrong on the page the operator watches


feigi:
Deferred from PR #598 review (issue #325). Dimensions: **silent-failure** (`refuted[0]`, 1-1 split) and **types** (`unverified[4]`, same subject).

`skills/fleet/scripts/board.mjs` (gatherSpend / readAgent) and `skills/fleet/scripts/board.html`.

## Why this was deferred, not applied

The controller ruled the **defect stands** but the **remedy is out of scope for #325**, whose contract is "a real fault stops being swallowed". Adding a fault field to `board.json` plus a renderer in `board.html` is a different, larger PR. PR #598 applied the in-scope half (the catch, plus a shape guard for a sidecar whose JSON parses to a non-object, in `e4b07d4`).

## The defect, measured

Two refuters reproduced it independently, with their own fixtures, against the review snapshot.

Refuter 1's probe (two agents, reviewer 4000 cache-write + implementer 1000):
```
INTACT: reviewPct 80 | skipped 0 | roles [["reviewer",4000],["implementer",1000]] | stderr 0
TORN  : reviewPct 0  | skipped 0 | roles [["other",4000],["implementer",1000]]    | stderr 1
board.json keys, byte-identical in both:
  totals,roles,top,reviewPct,tools,attributedPct,skipped,since
```

Refuter 2's independent probe (reviewer `review-pr-598` 4000 + `impl-598` 1000):
```
INTACT reviewPct 80 skipped 0 roles [["reviewer",4000],["other",1000]] stderr 0
TORN   reviewPct 0  skipped 0 roles [["other",5000]]                   stderr 1
keys byte-identical: totals,roles,top,reviewPct,tools,attributedPct,skipped,since
(node --test skills/fleet/scripts/board.test.mjs -> 37 pass, 0 fail)
```

So the JSON the browser consumes carries **no fault field**; `skipped` — the only fault key — stays `0`. `board.html` has exactly two fault paths, `sp.error` and `sp.skipped`, and neither fires. The page renders `review ${sp.reviewPct.toFixed(0)}%` unconditionally, so a corrupt sidecar shows as "review 0%", indistinguishable from a genuine zero, every tick, forever.

## The two things the split vote turned on, both settled

[...]

PR #682, opened 2026-08-19
fix(board): the guard skips the read, not the try — say so

### LINKED ISSUE #604: board.mjs:199-200: "the UNNAMED-AGENT case never enters the try" is false — the try block IS entered (review's own reword is worse)
Deferred from PR #598 review (issue #325). Dimension: **comments** (`unverified[3]`, severity `suggestion`). Filed with its refutation, per the review protocol.

`skills/fleet/scripts/board.mjs:199-200` (at `e4b07d4`):

> The meta read below is guarded by existsSync, so the UNNAMED-AGENT case never
> enters the try — it simply leaves `meta` at {}.

## The claim HOLDS

The `if (existsSync(metaFile))` guard sits **inside** the `try`, so the try block is always entered; only the throwable branch is skipped. A refuter settled it by running the exact shape rather than reading it:

```
UNNAMED-AGENT (no sidecar): tryBlockEntered=1 conditionEvaluated=1 throwableRan=0 catchReached=0 meta={}
named (sidecar present):    tryBlockEntered=1 conditionEvaluated=1 throwableRan=1 catchReached=0 meta={"name":"reviewer"}
```

The wording asserts block entry; the true statement is about the catch never being reached. It matters here because #325 is a **correction ticket** — a PR whose subject is a comment that misdescribed the code should not ship another one.

## The review's suggested reword was REFUTED — do not apply it

Suggested: *"...so the UNNAMED-AGENT case never reaches code that can throw — it simply leaves `meta` at {}"*.

That is **false, and more damaging than the original**. It drops the scope word "the try", so the claim becomes one about the whole function — and the UNNAMED-AGENT case does reach code that can throw, one line later: `readFileSync(file, "utf8")` at `board.mjs:242`, outside any try. The refuter reproduced it:

```
A) UNNAMED-AGENT, transcript present:  ...past the try, meta = {} -> ok, entries = 1
B) UNNAMED-AGENT, transcript vanishes: ...past the try, meta = {} -> THREW Error ENOENT from readFileSync(file)
```

Run B is the readdir/read race the caller already documents at `board.mjs:322-325`. So the reword trades one imprecision for a claim the next line of the same function contradicts — in a file whose whole ticket is unguarded throws in `readAgent`.

## Remedy is open

The refuter proposed keeping the scope and fixing the boundary, in the same character budget: *"never runs the guarded read"* — true, and it preserves what the original was trying to say. Not applied, and not verified as final wording; pick it or better.

Also note the reword churns more than it looks: at the file's 80-col comment width the paragraph reflows across roughly 7 lines for a ~13-character change.

feigi:
Closes #604.

`skills/fleet/scripts/board.mjs`'s `readAgent` comment claimed the UNNAMED-AGENT case "never enters the try". The `if (existsSync(metaFile))` guard sits **inside** the `try`, so the try is always entered — only the read inside the guard is skipped.

## What was measured

Exact shape of the meta block (origin/main `board.mjs:216-237`), instrumented:

```
UNNAMED-AGENT (no sidecar): tryEntered=1 condEvaluated=1 guardedReadRan=0 catchReached=0 meta={}
named (sidecar present):    tryEntered=1 condEvaluated=1 guardedReadRan=1 catchReached=0 meta={"name":"reviewer"}
```

Confirmed against the real function too, by mutation: replacing the guard with `if (true)` fails exactly one test — `board.test.mjs`'s "a genuinely absent meta.json — the real unnamed agent — stays silent" — because the guarded read then runs and throws ENOENT. That is the same fact the new wording asserts, measured on shipped code rather than a replica.

## The review's suggested reword is NOT applied

Suggested: *"never reaches code that can throw"*. That drops the scope word and is false — the UNNAMED-AGENT case reaches throwable code one line past the try, at `readFileSync(file, "utf8")`, which is outside any `try` inside `readAgent`. Same probe, transcript removed between the guard and that read:

```
UNNAMED + transcript gone: guardedReadRan=0 catchReached=0 meta={} -> readFileSync(file) threw ENOENT
```

That is the readdir/read race the caller already documents at `board.mjs:322-325`. The wording shipped here — *"never runs the guarded read"* — keeps the scope to the meta read, so it says nothing about the transcript read and cannot be read as a whole-function safety claim.

## Diff size

One phrase, `enters the try` -> `runs the guarded read` (14 characters). The paragraph reflows across 8 of its 9 lines at the file's 80-col comment width; word-for-word it is otherwise identical, verified by normalising both revisions and asserting `was.replace("enters the try", "runs the guarded read") === now`. No sentence added.

## Scope

Checked the other try/catch-scope claims in `readAgent` rather than fixing one and leaving its twin — the `gatherSpend` per-file catch claims at `board.mjs:209-211` and `:222-225` are both accurate (`readAgent` is called at `:331` inside the `try` at `:327`, and `a.meta.description` is read at `:341`, also inside it). No sibling correction needed. Not widened into #603/#605/#606.

## Tests

`./agent-test skills/fleet/scripts` — 978 pass, 0 fail. No new test: `board.test.mjs` already covers the unnamed-agent mode (proved by the mutation above), and the precedent for comment-only corrections in this repo (`113fbe6`) ships no prose pin. No test pins the rewritten strings — `grep -rn -F` for `enters the try`, `UNNAMED-AGENT`, `guarded by existsSync`, and `The meta read below` across the repo hits only `board.mjs` itself.


feigi:
## Review adjudication — 3 findings, 0 applied, HEAD unchanged at `ad72751d`

All three findings came back at severity `suggestion`, which the review workflow budgets **0 refuters** by policy — none of them had been checked when they arrived. Each was checked here. Nothing survived into the diff, so this PR is unchanged and its test gate re-ran clean: `node --test skills/fleet/scripts/*.test.mjs` → **978 pass / 0 fail**.

### 1. `board.mjs:199` — reword "never runs the guarded read" to "never reaches the catch" — **REFUTED, not applied**

Dimension `correctness`. The only in-scope finding, so it got a refuter biased to refuse. The refuter agrees "never reaches the catch" is **true** — 13-case instrumented matrix over the pinned head, plus 163,897 race iterations with a child process churning the sidecar, and the falsifying cell (`existsSync → false` **and** `readAgent`'s catch reached) was **0** throughout. `fs.existsSync` did not throw on any of 25 hostile inputs (EACCES parent, ELOOP, ENOTDIR, `NAME_MAX`/`PATH_MAX` overflow, embedded NUL, Symbol, throwing Proxy), and `metaFile` is always a string anyway.

But true is not the bar here, because the shipped clause is also true — the finding concedes that. So there is no `correctness` defect to fix, and the tiebreak went the other way on robustness:

```
existsSync stock            ->  guardedReadRan=0  CATCH=0
    "never runs the guarded read": TRUE     "never reaches the catch": TRUE
existsSync MUTATED to throw ->  guardedReadRan=0  CATCH=1  ["Error:EIO: i/o error, stat"]
    "never runs the guarded read": TRUE     "never reaches the catch": FALSE
```

The shipped wording is true by **local control flow** — the `if` alone guarantees it, and it survives the mutation. The replacement is true only while `fs.existsSync` stays total, a non-local Node invariant no reader of `board.mjs` can check. Node v26.5.0's own `existsSync` dispatches through a vfs handler *outside* its `try`; the refuter could not reach that path from userland on this build, so it is latent rather than a live hazard — but it is enough that the replacement rents its truth from Node while the shipped clause owns it.

The refuter also settled the question #604 left open: the vanishing-transcript race lands in **`gatherSpend`'s** per-file catch, not `readAgent`'s. So "never reaches the catch" is genuinely narrower than the previously-refuted "never reaches code that can throw" and does not inherit its defect — the claim was fairly raised, it just loses on other grounds.

The finding's one fair point is the circularity: the next sentence does open *"Everything that does reach the catch…"*, so a catch-antecedent would read better. Weighed against this being the **third** reword of this one sentence (`"never enters the try"` was false, `"never reaches code that can throw"` was refuted), churn wins for a `suggestion`-severity emphasis preference. Not filed — no defect is confirmed, so there is nothing for a ticket to say.

### 2. `board.mjs:244` — position-blind per-line catch — **deferred, already tracked as #606**

Dimension `silent-failure`. Pre-existing; this comment-only PR does not touch it. Already open and `ready-for-agent` as **#606**, whose body has the mechanism right. The review's probe output is new evidence that ticket did not carry, so it went there as a comment rather than a second issue: feigi/claude-config#606 (comment). Re-run against this PR's head rather than the review's snapshot; numbers reproduced exactly:

```
CLEAN     cacheWrite= 600 skipped= 0 stderr= []
MID-TORN  cacheWrite= 400 skipped= 0 stderr= []
TAIL-TORN cacheWrite= 300 skipped= 0 stderr= []
```

Mid-file corruption loses the middle turn's entire 200 with `skipped` still 0 and stderr empty — indistinguishable from the documented-benign tail tear.

### 3. `board.mjs:235` — warning understates the degradation — **deferred, filed as #686**

Dimension `silent-failure`. Pre-existing and out of this PR's scope. `ledger.mjs check` returned exit 3, but the two tracker rows it named (#149 merge-bot force-push, #39 cockpit concurrency) both scored 0 and are unrelated — an artifact of the query terms, not a real hit. Searched the tracker directly instead: #602 (fault never reaches the browser), #603 (warn-once idioms) and #605 (`mapCi`) are adjacent but distinct, so this is filed clean as **#686**, `ready-for-agent` with the remedy marked open. Defect confirmed by probe against this head:

```
INTACT label= "review-pr-1" role= reviewer stderr= []
TRUNC  label= "r"           role= other    stderr= ["board: agent-r.meta.json unusable, classifying agent as \"other\": Unterminated string in JSON at position 42"]
```

The `{}` fallback degrades the panel label as well as the role (`board.mjs:341`), and the warning mentions only the role.

PR #683, opened 2026-08-19
fix(rebase-check-refresh): correct five comments that describe behaviour the step does not have

### LINKED ISSUE #165: rebase-check-refresh.yml: four remaining comments describe behaviour the code does not have

feigi:
Closes #165

Comment corrections in `.github/workflows/rebase-check-refresh.yml`, plus one test that pins the claim item 2 mints. The non-comment body of the `run:` block is byte-identical to `origin/main` — both scalars parsed and compared with comment lines stripped.

### The five, each re-derived against `origin/main`

1. **The dependent-jobs rationale was inert.** `grep -n 'needs:'` over both workflows finds no `needs:` anywhere, so `rebase-check` has no dependents and the load-bearing capitals told a reader `check` gets re-run too. Clause dropped; the gate is justified by the rerun taking `rebase-check` itself back to in-progress.
2. **The `concurrency:` note predicted a benign skip.** The replacement run's POSTs are refused while the superseded run's already-POSTed reruns hold their parent runs in progress; `rerun-rebase-check.sh` exits 3 and the PR lands in `rejected`. If every behind PR lands there, `BEHIND > 0 && RERAN == 0` fails the step — red, not a skip. Now stated, and measured by the new test.
3. **"every later push"** read as a push to the PR, which changes `headRefOid` and resolves different runs. It is a push to main; the script already qualifies its own sibling claim that way.
4. **The walk disclaimed an ordering dependency it has.** It consumes `$RUN_IDS` in returned order and takes the first candidate whose rerun is accepted — exactly the dependency an index has. What it buys is skipping candidates that hold no `rebase-check` job, not order-independence.
5. **"silently, with no annotation" overstated a `set -e` abort.** `$( )` captures stdout only, so gh's stderr still comes out; what a bare assignment loses is the `::error::` naming which call failed. The file's own `— see gh error above.` messages assume the same thing.

### Class enumeration

Item 5's claim appears at **both** `if !` comments, not only the one the issue quotes — the second says "would abort here silently" while back-referencing the first. Both corrected; fixing one would have left the ticket open.

Deliberately left: `silently` at the `grep -c`, fd-3 and `*`-arm comments — those claim a guard is disabled or a no-op is unobserved, not that a failure is unlogged, and the issue names none of them. Also left: `block the merge it was meant to unblock` in the header, which sits oddly beside "main is NOT a protected branch" two paragraphs above it — pre-existing, outside this ticket's five.

### Test

`skills/fleet/scripts/rerun-rebase-check.test.mjs` — every candidate refusing with the in-progress 403 leaves `rejected: 1`, `RERAN` at 0, and exit 1. Written and run **before** the comment was rewritten: had it come back green-and-skipping, item 2's replacement wording would have been the false claim. Verified live by flipping the guard's `-gt 0` to `-gt 99` (red), then restoring.

The accept side is already covered by *"a 403 that says the parent run is in progress is a skip, not a token failure"* — two in-progress refusals then an acceptance exits 0 with `1 re-run`, so the guard is not made to fire on a supersession that recovers.

`./agent-test skills/fleet/scripts/rerun-rebase-check.test.mjs` — 24/24. `ci-state.test.mjs` + `ledger.test.mjs` (the other files naming this workflow) — 81/81.


feigi:
Fix pass on review findings — pushed `53a0337`.

**Applied (2)**
- `survived[0]` *(important / correctness, `.github/workflows/rebase-check-refresh.yml:28`)* — the `concurrency:` note claimed unconditionally that a superseded refresh leaves its PRs in `rejected`. `rerun-rebase-check.sh` exits 0 on its first accepted POST, so a supersession leaves at most **one** candidate's parent run in flight per PR; the replacement run falls back to that PR's remaining candidates and counts it `re-run`. Note now qualifies `rejected` as the exhausted-fallback case (a page with a single usable run, or enough successive supersessions to put every candidate in flight).
- `unverified[1]` *(suggestion / correctness, `skills/fleet/scripts/rerun-rebase-check.test.mjs:274`)* — **survived** its refuter (`refuted: false`). The added test's title and comment attributed its all-three-refused fixture to one supersession, a state one supersession cannot produce. Renamed to what its assertions measure and the supersession causality dropped from the comment.

**Deferred (1)** — #685, `needs-triage`
- `unverified[2]` *(suggestion / correctness, walk comment above the `per_page=5` fetch)* — **refuted**. The clause the finding says is missing ("takes the first candidate whose rerun is *accepted*") sits in the same sentence it quotes, and a mutation killing the #159 fallback falsifies exactly that clause while leaving the stub-skipping one true. No defect confirmed.

**Verification**
- `node --test skills/fleet/scripts/*.test.mjs` → `tests 979 / pass 979 / fail 0` (baseline before the edits: identical).
- Renamed test mutation-checked in a scratch copy, never the worktree: mutating the all-or-nothing `::error::` wording reds **only** it (978/979); mutating the `::warning::` skipped-PR wording leaves it green and reds a different test instead.

Not touched here: the header self-contradiction at `:2-4` vs `:11-12`, already filed as #684.

PR #692, opened 2026-08-20
test(fleet): execute die()'s EAGAIN guard — deterministically, and per script

### LINKED ISSUE #328: fleet scripts: die() inverts exit 2 into exit 1 on a full stderr pipe in three more scripts — the defect PR #322 fixed in candidates.mjs only

feigi:
Closes #328

## What this is

#328's **code** half landed with #367 / PR #464, which folded all seven `die()` copies onto `arg.mjs`'s `makeDie()` and carried the `try`/`catch` with it. Reconfirmed at `origin/main` (`0008b6b`): `ci-state.mjs:28`, `diff-stats.mjs:108` and `pr-overlap.mjs:26` all read `const die = makeDie(NAME);`, and `makeDie()` is guarded.

The **test** half never landed, and the ticket's last comment reduces scope to exactly that. No script source is touched here.

## The gap

Nothing in the repo ever *executed* that catch.

- `candidates.test.mjs`'s source pin is deterministic, but it is a text pin and its regex stops at `try { writeSync(2,`. It says nothing about the catch body, nor about `process.exit(2)` still running below it. **Measured: `catch { process.exit(1); }` leaves it green** — the whole defect back, under a green suite.
- The behavioural pin beside it races the pipe: 3/10–7/10 on Linux, and on darwin EAGAIN never fires at all.

## What was added — `skills/fleet/scripts/arg.test.mjs`, where `die()` now lives

1. **Deterministic, every platform.** A probe closes fd 2, so the same `writeSync` throws `EBADF` rather than racing for `EAGAIN`. Different errno, same and only promise `die()` makes about it: the message may be lost, the exit code may not.
2. **The three scripts #328 names, end to end** — `ci-state`, `diff-stats`, `pr-overlap` — behind a `gh` that floods 200 KB of forwarded stderr (`execFileSync` with no `stdio`, so Node re-emits it through the async `process.stderr`; the flood has to come from `gh` or fd 2 is never the fd under pressure). Each row also runs an ordinary 64 B case, pinning what the guard must **not** change: the refusal still lands, the code is still 2. Anchored on the script's `NAME` prefix only, so rephrasing a refusal is not this test's business.

## Mutation results (darwin)

| Mutant | New deterministic test | `candidates.test.mjs` source pin | New flood rows |
|---|---|---|---|
| `try`/`catch` reverted | **RED** | **RED** | green (EAGAIN never fires on darwin) |
| `catch { process.exit(1); }` | **RED** | green | green |
| control: comment line added inside `makeDie` | green | green | green |
| control: `diff-stats` refusal reworded | green | green | green |

So test 1 is the assertion that discriminates on this machine, and it catches a mutant the existing source pin cannot. Test 2 is a CI (Linux) gate whose local green proves nothing — stated in its own comment so no one reads it as more than that.

`./agent-test skills/fleet/scripts`: **988 pass, 0 fail** (984 before).


feigi:
## Review adjudication — 4 findings, 2 applied, 2 no-op, 1 filed

Pushed `4a88d6a6b8c4228ff77b30b6ed2a5dd5124871ea` (`ae12dd5` + a one-line citation fix: `board.mjs:77` → `76-79`, since the log and the `return null` it names are on 78). `node --test skills/fleet/scripts/*.test.mjs` → **tests 989, pass 989, fail 0** (baseline 988 + the one row added).

| finding | band | outcome |
|---|---|---|
| `arg.test.mjs:168` — flood rows don't discriminate on darwin | refuted (1-1) | **no-op.** Refutation stands. |
| `arg.test.mjs:148` — fleet-tick excluded from `GH_FLOOD` but has the same forwarded-stderr path | suggestion | **applied**, row added + comment corrected |
| `arg.test.mjs:113` — the EBADF test discriminates | suggestion | **no-op**, it is a confirmation, not a defect |
| `arg.test.mjs:183` — "3/10 to 7/10" unsourced | suggestion | **applied**, cited to #328 and corrected |

### `arg.test.mjs:168` — no-op

The headline claim ("a contributor on a Mac gets a fully green suite regardless of whether the fix is present") is false as stated. Re-measured here with `makeDie()`'s try/catch removed: `node --test skills/fleet/scripts/arg.test.mjs` → **16 tests, 15 pass, 1 fail** — the EBADF test at `:113` catches the mutant locally, in the default invocation. Only a run pattern-filtered down to the flood rows is green. The narrow true version ("those rows don't discriminate on darwin") is already stated verbatim in the file's own comment. No defect, nothing filed.

### `arg.test.mjs:148` — applied, row added

One refuter, dispatched biased-to-refuse, returned **`refuted=false`** on every measurable claim. `fleet-tick.mjs:177` `execFileSync`s gh with no `stdio`, so it forwards gh's stderr like the other three; with `--implementers 1 --reviewers 1 --merge-bots 1 --pool 1` it reaches gh and refuses through `die()` at **65,613 B forwarded, exit 2**, and the message text (`fleet-tick: gh pr list failed: …`) proves which `die()` fired, so it is not a right-code/wrong-cause false positive. It has no `process.env` read, no git/cwd/`GH_TOKEN` dependency, and nothing fallible before the gh call. Positive control: a gh stub exiting 0 makes the same argv exit 0.

Chose **adding the row** over rewording the exclusion alone. The alternative is a comment defending fleet-tick's absence on argv verbosity, which is not a reason a later reader should respect, while the table's own rule is the forwards-then-refuses shape. The row makes the comment true by construction. Cost measured at ~3.1 s, same as its siblings.

The other three exclusions were checked too and are sound, each for a different reason, now stated individually instead of as one blurred clause: `candidates` carries this row itself; `ledger` **captures** gh's stderr (`stdio: ["ignore","pipe","pipe"]`, `ledger.mjs:441`); `board` **logs** a failed gh and returns `null` (`tryRun`, `board.mjs:77`), so it has no exit 2 to invert.

**Mutant discipline for the new row, honestly:** on darwin it cannot discriminate the fix — measured, with the guard removed all **four** flood rows still pass (4/4), so the comment now says 4/4 and the row is a CI gate, not a local one, exactly like its siblings. It is not vacuous here either: bypassing fleet-tick's gh-failure `die()` with `process.exit(0)` fails it on the ordinary path.

### `arg.test.mjs:183` — applied; and no, the range did not reconcile

The comment read "On Linux #299/#328 measured the inversion at 3/10 to 7/10 unfixed and 0/20 fixed" — no source, so it reads as measured here. **Both halves are wrong against #328's own body**, which records:

- 5/10, 6/10, 7/10 (pr-overlap, diff-stats, ci-state)
- a second pass at 4/10, 3/10, **0/10**, then 6/10, 4/10
- a third harness at 1/40, 0/40, 0/40, against a self-validating 32/40 control
- fixed: "the same one-line guard drops all three to **0/10** in the passes that measured it"

So the unfixed span is **0/10 to 7/10**, not 3/10 to 7/10 — the 3/10 floor silently drops a script that measured 0/10 in the same pass its siblings measured 4/10 and 3/10, plus the whole third harness. And **0/20 is not #328's number at all**: it is `candidates.test.mjs`'s, for `candidates.mjs` under `JQ_OVERRIDE` (#299) — a different script under a different flood mechanism. That is also where the sibling 7/15 comes from, which is why the two never lined up.

Rewritten to cite #328, quote the passes, state the real span, keep the `candidates` series explicitly separate, and carry #328's own reading — the rate is machine-dependent, only the mechanism is confirmed, so a single green Linux run is not evidence the guard is present.

The finding's own suggested remedy (have the author or CI attach fresh Linux counts) was not available: unattended, on darwin, where the race does not fire. Citing the existing source fixes the actual defect, which was attribution.

### Filed

- **#696** — `arg.mjs:17` says `fleet-tick.mjs` "reads no flags at all"; it parses six through `node:util`'s `parseArgs` with its own refusals, and `fleet-tick.mjs:115-118` says so. Out of scope here (this PR is test-only; #328's code half already merged), `ready-for-agent`, dimension `comments`. Surfaced by this PR putting a measured four-flag fleet-tick invocation next to that header.

Deduped via `ledger.mjs check` before filing: `found: false`, highest near-miss 0.22, no tracker hit on the subject.

Issue #484, opened 2026-08-14
fleet scripts: echo expands backslash escapes in caller text — ten scripts still affected (no-undo-audit.sh closed by #480)


feigi:
> **RETARGETED after PR #480.** `skills/fleet/scripts/no-undo-audit.sh` is **now clean — do not re-audit it.** All five of its sites are fixed and pinned there. What remains is the same class in the **other ten** fleet shell scripts, enumerated below.

**Originally deferred from PR #480 review** (implements #304). **Dimension: correctness.**

## The mechanism (measured, not inferred)

These scripts are `#!/bin/sh`. Both dash — CI's `/bin/sh` on `ubuntu-latest` — and macOS `sh` expand backslash escapes in an `echo` **operand**. `\c` truncates the line and swallows its newline, so the following stderr line collides with it:

```
$ sh -c 'a="abc123 fix the \cthing"; echo "$a" | sed "s/^/    at risk: /"' | od -c
0000000       a t   r i s k :   a b c 1 2 3   f i x   t h e          # truncated, newline gone
```

Recognised: `\c \t \n \b \f \r \v \\ \0`. **Unknown escapes like `\s` pass straight through** — which is the trap: a fixture using `back\slash` exercises the *safe* member of the class, reads as coverage, and catches nothing. `no-undo-audit.test.mjs` had exactly such a fixture and was blind to five live defects.

Remedy, proven five times in PR #480: `printf '%s\n' "…"`.

## What PR #480 already fixed — do not redo

| site | operand | why it mattered |
|---|---|---|
| the stash diagnostic | `$msg` | git's own stderr, reachable via a bad `objects/info/alternates` |
| the `git status` step header | `$wt` | caller-supplied path; a worktree named `back\clue` is enough |
| `die()` | `$1` | 11 of its messages interpolate `$wt`; one edit covered all 11 |
| the refusal listing | `$porcelain` | git C-quotes here, so `echo` collapses `\\` to `\` and **rewrites the quoted path into a different one** |
| the conflict list | `$conflicts` | `-z` turns C-quoting **off**, so the raw backslash truncates outright |
| the at-risk list | `$at_risk` | `git log --oneline` output — an **ordinary commit subject** is enough |

Each is pinned by a test that reddens when that one `printf` is reverted to `echo`, and by nothing else.

[...]

PR #693, opened 2026-08-20
fix(fleet): port PR #384's three registry cross-check fixes into release-ticket.sh

### LINKED ISSUE #395: release-ticket.sh: the worktree-registry cross-check carries all three defects PR #384 fixed in inflight.sh
`release-ticket.sh:96-124` is a near-verbatim copy of the worktree-registry cross-check in `inflight.sh`. PR #384 fixed three defects in the `inflight.sh` copy. All three are still live in this one, and its tests have the same blind spots that let them ship green.

Measured against `8f3cd20`.

## 1. A stray DIRECTORY in `.git/worktrees` makes every release die

`release-ticket.sh:108-110`:

```sh
for entry in "$wtroot"/*; do
  [ -d "$entry" ] || continue
  registered=$((registered + 1))
done
```

Measured (git 2.50.1, Apple Git-155): `mkdir .git/worktrees/stray-dir` leaves `git worktree list --porcelain` reporting 2 worktrees while this loop counts 3, so the check at `:123` fires. One `mkdir` under the registry turns **every release in that repo** into a hard refusal until a human notices.

`release-ticket.test.mjs:992` covers a stray **file** (`stray-note`), which the `-d` test skips anyway — which is exactly why the directory case shipped green. Same gap `inflight.test.mjs` had.

**Fix, as applied in PR #384** — a registry entry is a directory holding a `gitdir` file, and the `-x` test has to come first so an entry chmod'd 000 (which git *does* drop, measured: 2 listed, then 1) is still counted rather than skipped as a stray:

```sh
if [ -x "$entry" ] && [ ! -f "$entry/gitdir" ]; then continue; fi
```

## 2. `grep -c … || true` swallows a counter that could not run

`release-ticket.sh:122`:

```sh
listed=$(printf '%s\n' "$wt_list" | grep -c '^worktree ' || true)
```

`grep -c` exits 1 on zero matches, which is legitimate, so the `|| true` is load-bearing — and it absorbs a `grep` that could not run *at all* just as happily. The count is then the empty string, `$((listed - 1))` is `-1`, and the die at `:124` blames `git worktree list` for a count no listing can produce.

This fails **closed** (exit 2, no verdict), so it is not the wrong-"free" class — it is a wrong *attribution*, sending the operator after git when the fault was a fork failure. Note `release-ticket.sh` already removed exactly this shape elsewhere for the same reason.

**Fix, as applied in PR #384** — one awk, whose program contains no `exit`, so every non-zero status is a real failure:

```sh
listed=$(printf '%s\n' "$wt_list" | LC_ALL=C awk '/^worktree /{c++} END{print c+0}') ||
  die "could not count the worktrees git listed for #$issue"
```

## 3. The mismatch message is direction-blind

`release-ticket.sh:123-124` reports "the listing is incomplete" for **both** directions. When git lists *more* worktrees than the registry count — a sibling agent's `git worktree add` landing between the two reads, routine in a parallel fleet — the on-disk count is the stale read, and "the listing is incomplete" is the exact opposite of what happened. It sends the operator hunting a permissions fault that does not exist.

**Fix, as applied in PR #384** — split on the observed direction, and name it:

```sh
linked=$((listed - 1))
if [ "$linked" -lt "$registered" ]; then
  die "git listed $linked worktrees for $registered registry entries in $wtroot — the listing is incomplete, so no absence it reports can be trusted"
elif [ "$linked" -gt "$registered" ]; then
  die "git listed $linked worktrees but only $registered registry entries were counted in $wtroot — the registry read missed entries git can see, so no absence it reports can be trusted"
fi
```

## Tests

PR #384 added four cases to `inflight.test.mjs` for these, each mutation-verified to kill only its own mutant. Mirror them into `release-ticket.test.mjs`:

- a stray **directory** in the registry does not block the release
- a registry entry chmod'd 000 still refuses (pins the `-x`-before-`gitdir` ordering)
- git listing MORE than the registry produces the *other* message (shim `git` so `worktree list --porcelain` emits one extra `worktree ` line — git derives its listing from the registry, so the interleaving cannot be raced deterministically)
- a counter that could not run refuses rather than reporting `-1` (the existing `awkFailWhenProgramHas` shim, keyed on `{c++}`)

## Why not fixed in PR #384

Different file, and #384's ticket (#95) is scoped to `inflight.sh`'s probe 3. Editing `release-ticket.sh` there would have widened the diff past its ticket.

See also #367, which tracks whether these two copies should share code at all. This ticket is the defect; #367 is the duplication question. Fixing this one by hand is correct regardless of how #367 is decided.

*Deferred from PR #384 review. Dimensions: `correctness`, `silent-failure`, `simplify`.*

feigi:
`release-ticket.sh`'s worktree-registry cross-check is a near-verbatim copy of `inflight.sh`'s. PR #384 fixed three defects in that copy; all three were still live here, and this file's tests had the same blind spots that let them ship green.

Re-derived against `origin/main` (`0008b6b`), not the ticket's line numbers — they had drifted. Measured on git 2.50.1 (Apple Git-155), darwin, euid 501.

## Measured state table

`linked = listed - 1`. Every row measured directly against real repos, all three candidate discriminators:

| on-disk state | listed | linked | bare `-d` | emptiness-keyed | `gitdir`-keyed |
|---|---|---|---|---|---|
| healthy | 2 | 1 | 1 ✅ | 1 ✅ | 1 ✅ |
| stray empty `mkdir` | 2 | 1 | **2 ✗ refuses** | 1 ✅ | 1 ✅ |
| `gitdir` file deleted | 1 | 0 | 1 ✅ refuse | 1 ✅ refuse | **0 ✗ wrong "free"** |
| entry dir `chmod 000` | 1 | 0 | 1 refuse | 1 refuse | 1 refuse |
| `gitdir` file `chmod 000` | 1 | 0 | 1 refuse | 1 refuse | 1 refuse |

## 1. A stray DIRECTORY refused every release in the repo

The count was a bare `[ -d "$entry" ] || continue`, so one `mkdir` under `.git/worktrees` made it 2 registered against 1 linked and the check fired — turning every release of every ticket in that repo into a hard refusal until a human noticed. The existing test covered a stray *file*, which the `-d` skipped anyway; that is why the directory case shipped green.

**The ticket body's prescribed remedy is not the one applied here, deliberately.** It quotes

```sh
if [ -x "$entry" ] && [ ! -f "$entry/gitdir" ]; then continue; fi
```

which is PR #384's *first* commit. #384's own second commit (`62f0f65`, "the stray-entry skip keyed on gitdir, which freed a corrupt entry") replaced it, and `inflight.sh:554` on `origin/main` carries the replacement. Reproduced independently here: git **drops** an entry whose `gitdir` was deleted, so keying the skip on that file waves a corrupt entry through as "not git's", the count agrees at 0, and the claim releases while its checkout may still be on disk. That is the wrong "free" this check exists to rule out, reintroduced one layer in. Keyed on **emptiness** instead — a stray `mkdir` is empty; even a corrupt entry still holds git's own files (`commondir`, `HEAD`, `index`, `logs`, `refs`).

## 2. `grep -c … || true` swallowed a counter that could not run

The `|| true` is load-bearing (`grep -c` exits 1 on zero matches, legitimate) and absorbed a grep that could not run at all just as happily. The count was then the empty string, `$((listed - 1))` was `-1` (measured), and the die blamed `git worktree list` for a count no listing can produce. Replaced with one `awk` whose program contains no `exit`, so every non-zero status is a real failure.

## 3. The mismatch message was direction-blind

"the listing is incomplete" was reported for both directions. More listed than registered is the opposite fault — a sibling agent's `git worktree add` landing between the two reads, routine in a parallel fleet — and that message sends the operator hunting a permissions fault that is not there. Split on the observed direction; the FEWER wording is unchanged, so the existing pin at `release-ticket.test.mjs` still holds.

## Class enumeration

Class: *defects PR #384 fixed in `inflight.sh`'s worktree-registry cross-check that are still live in `release-ticket.sh`'s copy.* Verified by diffing the two implementations rather than trusting the ticket's inventory.

**Covered (3 of 3 named, plus one the ticket got wrong):** all three defects above, with defect 1 fixed using the discriminator `inflight.sh` actually carries rather than the one the ticket body quotes.

**Deliberately left — the recount.** `inflight.sh:650` (`[ "$linked" -eq "$registered" ] || count_registry || return 1`) is a fourth divergence, and `inflight.sh:511-516` names it alongside the other three as open against this copy. It is **not** a class member: it landed in `b4ee3801` ("inflight.sh accumulates probe results instead of dying on first failure"), which `git merge-base --is-ancestor` confirms is **not** in PR #384. It also presupposes `inflight.sh`'s accumulate-and-continue model, which this script (fatal `die`) does not have. Worth its own ticket; out of scope here.

**Out of scope by collision:** `die()` at `:29` and every other `echo` in this file are untouched — #484 is queued to change them. The diff is confined to two hunks in the cross-check.

**Not applicable:** `release-ticket.sh` uses plain `--porcelain` where `inflight.sh` uses `--porcelain -z` + `tr`. That divergence is issue #185's lineage, not #384's.

This is a guard on a refusal path, so the failure mode that matters is refusing a healthy repo. Net effect: the change **narrows** refusal (a stray directory no longer refuses) and **widens** it in exactly one direction (a counter that could not run now refuses loudly instead of reporting `-1`). A new acceptance test drives a genuinely healthy repo with **two** live worktrees, so the arithmetic is exercised at `registered=2/linked=2` rather than the trivial 1-against-1.

No test anywhere asserted on the old wording except the FEWER-direction pin, which is byte-identical under the split.

## Tests

Six added, each mutation-verified to kill only its own mutant, plus a negative control.

| mutation | result |
|---|---|
| drop the skip line (bare `-d`) | 84/1 — kills only *a stray DIRECTORY…* |
| drop `-x`, emptiness only | 84/1 — kills only *a registry entry git cannot even open…* |
| `gitdir`-keyed skip (ticket's remedy) | 84/1 — kills only *a registry entry whose gitdir is GONE…* |
| revert counter to `grep -c \|\| true` | 84/1 — kills only *a worktree COUNT that could not run…* |
| collapse the direction split | 84/1 — kills only *git listing MORE than the registry…* |
| miscount registry (refuse-everything) | 17/68 — kills the acceptance case among 68 |
| **control:** swap the `&&` operands | **85/0 — green, as it must be** |

That control is a finding: the ticket and `inflight.sh`'s comment both frame `-x` as an **ordering** requirement, but `&&` is commutative in truth value, so both orders behave identically. What is actually load-bearing is that the `-x` test is **present** — dropping it makes a `chmod 000` entry read as empty, the count return 0 to match `linked=0`, and the claim release. The mutation table above separates the two: the order-swap is green, the removal is red.

**euid-0 caveat:** run at euid 501, and the `chmod` fixtures genuinely discriminated — 0 skipped in every run. The `gitdir`-deleted pin uses `rm`, not `chmod`, so it discriminates as root too.

`./agent-test skills/fleet/scripts`: **990 tests, 990 pass, 0 fail, 0 skipped** (79 → 85 in `release-ticket.test.mjs`).

Closes #395

https://claude.ai/code/session_01Q7MJFUiGzFvjFPmpeGZPgf


feigi:
## Review applied — 13 findings adjudicated

Pushed `8933be3` (two commits on `5ef7791`). Gate: `node --test skills/fleet/scripts/*.test.mjs` → **991 pass / 0 fail / 0 skipped** (baseline 990; +1 new fixture). euid **501**, so the `chmod` fixtures really deny — `skipped 0` confirms no EUID0 guard fired.

### Applied (5)

| finding | what changed |
|---|---|
| `SURVIVED[0]` + `SURVIVED[2]` + `UNVERIFIED[0]` — one defect, found by three dimensions | `release-ticket.sh:178` now reads `ls`'s **status**, not only its output |
| `SURVIVED[1]` — this diff falsified `inflight.sh`'s comment | `inflight.sh:511-518` now names the recount as the one item still open (#694), plus the new divergence (#697) |
| `UNVERIFIED[3]` — the `-x`-ordering overclaim | dissolved: the `-x` conjunct is subsumed by the status check and is gone with its rationale |
| `UNVERIFIED[1]` — "would land here and nowhere else" | reworded to what the case actually adds; refuted=false, measured |
| `UNVERIFIED[5]` — subprocess `rm` in the test file | `rmSync`, the file's own convention; refuted=false, throw-on-missing verified preserved |

### The headline: this branch introduced a data-loss regression

The `2>/dev/null` added to `ls -A` made *"I could not read this entry"* indistinguishable from *"this entry is empty"*. An entry at **0111** is searchable — so it passes the `-x` guard the comment leaned on — but unreadable: `ls -A` fails EACCES, prints nothing, and `-z` is satisfied **by the failure**. Git drops the same entry, so the counts agree and the crosscheck never fires.

Re-derived here at euid 501 rather than taken from the review, with an ordinary `git worktree move` (directory name kept, so `stray` would have matched had git listed it):

```
pre-fix guard : code=0 released=true blockers=[] branchGone=true wipStillOnDisk=true
                gh: ["issue view 9 …","issue edit 9 --remove-label in-progress"]
fixed guard   : code=2 released=null branchGone=false gh: []
```

**Correction to the review's framing:** with the checkout at its *canonical* path the orphan probe still catches it and blocks at exit 1 — refused, but naming an orphan instead of the unreadable entry. The wrong "free" needs the checkout moved off that path. Both halves are the same skip; the new fixture pins the cheap half and the comment records the expensive one.

**Fixture mutation evidence** (`an entry that is searchable but UNREADABLE is unknown too, not an empty stray (#395)`, guarded by `EUID0`/`NO_DENIAL` like its 0000 sibling):

- revert the guard to the shipped output-only form → **85 pass / 1 fail**, and the one failure is this fixture (`actual: 1, expected: 2`). It is the *only* case of the 86 that catches it — the existing suite was blind.
- the `-r`-flavoured alternative remedy (`[ -r ] && [ -x ] && [ -z … ]`) → **86 pass / 0 fail**. It pins the invariant, not the syntax.

### Deferred and filed (4) · commented (1) · dropped (1)

- **#697** `ready-for-agent` — `inflight.sh:554` carries the identical skip. Same mechanism, other script, out of #395's scope.
- **#698** `ready-for-agent` — `REFUTED[0]`, 1-1 split, **ruled partly refuted**. The awk half is genuinely refuted: the new `|| die` sits on the same pipeline 15 lines above, so any standing "awk cannot run" fault hits the canary and exits 2 (the refuter re-ran the probe with a realistically broken awk and got exactly that). The half neither vote disputed survives: `:178`'s unguarded `git worktree list` exits with git's own code (measured 3), outside the 0/1/2 vocabulary. Pre-existing context lines.
- **#699** `ready-for-agent` — `UNVERIFIED[2]`, the negative tally's second route (empty-but-successful listing → `git listed -1 worktrees`). Shim-only reachable; refusal direction correct.
- **#700** `needs-triage` — `UNVERIFIED[4]`, redundant `LC_ALL=C`. Torn on purpose: the line is byte-identical with `inflight.sh:633` and holding them identical is the premise of this port.
- **#694** — `SURVIVED[3]` and `REFUTED[2]` are the same subject; the measurement went there as a [comment](https://github.com/feigi/claude-config/issues/694#issuecomment-5355244615). `REFUTED[2]` **ruled refuted as a blocker**: measured, the pre-PR `-eq` form and the post-PR direction split refuse identically on the same simulated race, so this diff changes the message and not the rate. Not widened into this PR — inflight.sh's recount returns a status into an accumulate-and-continue probe while this script dies fatally, so the port is a design decision.
- `REFUTED[1]` — **dropped**, 2-0. Both refuters showed the "shape" the comment names (a `|| true` swallowing a command that could not run) really was removed at both other `grep -c` sites; the finding counted surviving `grep -c` tokens rather than the anti-pattern.

PR #695, opened 2026-08-20
fix(fleet): name both fields ci-state.mjs's header drops, and pin it

### LINKED ISSUE #690: ci-state.mjs:32-33: the header says --quiet drops "the raw job list", naming only jobs where the assignment drops jobs AND missing
Surfaced by `impl-677` during #677's enumerate-and-declare sweep, declared out of scope, and confirmed by the controller against `origin/main` at `c05dd22`.

## The defect

`skills/fleet/scripts/ci-state.mjs` describes what `--quiet` drops in two places, and the header is incomplete.

`:32-33` — the file's own orientation comment, the first thing a reader meets:

```
// --quiet suppresses the diagnostic stream (command echoes, per-job/per-field
// lines) and drops the raw job list from the payload.
```

`:455-456`, at the assignment that actually does it:

```
// On the quiet hot path drop `jobs` and `missing` too — `reasons` already
// states every failing/absent job, so they are pure duplication [...]
const payload = { pr, branch, prHead, runId, attempt, runHeadSha, status, conclusion, behind, verdict, reasons };
if (!quiet) Object.assign(payload, { jobs, missing });
```

Ground truth is `:459`: `--quiet` drops **both** `jobs` and `missing`. The header names only "the raw job list".

## Why `missing` is the field that matters

This is the same defect just fixed one file over in #677 (PR #689), and it matters for the same reason. **`missing` is what distinguishes "the expected job is absent" from "the expected job failed."** A reader who trusts the header believes a `--quiet` payload still carries `missing`, and a finisher gating a label on it would read absence as a pass — the failure mode the fleet's "absence never reads as pass" rule exists to prevent.

It is **incomplete, not false** — which is why #677's implementer correctly declined to widen its one-sentence diff to reach it.

## Acceptance

- [ ] `:32-33` names both fields, matching `:455-456` and `:459`
- [ ] The wording is derived from the payload assignment at implementation time, not copied from this issue — the field list can change
- [ ] Consider adding this site to `skills/fleet/scripts/quiet-payload-prose.test.mjs`, added by PR #689 for exactly this class of drift

## The pin already exists and would cover this cheaply

PR #689 ships `quiet-payload-prose.test.mjs`, which **derives the field list from `ci-state.mjs`'s own assignment** rather than restating it, and pins three prose sites (`run-team/SKILL.md` ×2, `review-and-fix.md`). Adding `:32-33` as a fourth source is the intended extension point — a field added at `:459` then reddens every prose site instead of silently un-completing them.

Note its bounds are load-bearing and were mutation-proved in PR #689: an unbounded file-wide variant stayed GREEN when the live clause was gutted, because the file's other correct statement served as its false green. Follow the bounded pattern; do not add an unbounded match.

## Out of scope

- `run-team/SKILL.md:986` — that is #677, merged as PR #689.
- `docs/specs/2026-07-23-fleet-plugin-design.md:178` — already names both fields correctly; `impl-677` left it unpinned with a measured reason (its sibling spec pins run the script against `gh` fixtures, which is not cheap).
- Whether `--quiet` should drop those fields at all. This is about describing what it does.

## Provenance

`impl-677` sweep (declined in scope, offered as a follow-up candidate), controller re-measured against `origin/main`. Filed by the controller because an implementer's follow-up promise has no owner in the fleet and `Closes #677` would have buried it at merge.

`ledger.mjs check` → exit 3 / `verdict: tracker-hit`, but the rows it named (#334, #415, #483) all scored **0** and none is related — the third instance of #388 in this run. Two targeted tracker searches surfaced only #594 (a different `ci-state.mjs` defect, about behind-count probe reporting) and closed #48. Not a duplicate.

feigi:
`ci-state.mjs`'s orientation comment said `--quiet` "drops the raw job list from
the payload", naming only `jobs`, where the assignment it describes drops both.
Incomplete, not false — but `missing` is the field separating "the expected job
is absent" from "the expected job failed", so a reader who trusts the header
believes a `--quiet` payload still carries it.

**AC-2, derived not copied.** The wording comes from reading the assignment in
this tree at implementation time — `if (!quiet) Object.assign(payload, { jobs, missing });`
— not from the issue text.

## The one real decision: route 1, backtick the header

#689's clause anchor is ``/`--quiet`[^.]*?(drops[^.]*)/``, which needs a
backticked `` `--quiet` `` in the same sentence as `drops`. The header wrote it
bare, so #690's proposed SITES row would have tripped `dropsClause`'s assert
instead of pinning the site. Reproduced independently before choosing, with a
positive control — the same probe returned real clauses for two sites the suite
already pins green, then `null` for this header.

Took **route 1** (backtick the header) over route 2 (widen the matcher):

- Route 2 changes a matcher **three other sites depend on** and would require
  re-proving #689's whole mutation battery. Route 1 touches one line.
- The backtick is not a contortion to satisfy a test: `ci-state.mjs:120` already
  writes `` `--base --quiet` `` backticked, and the tree spells the flag
  backticked in all 18 other places it names it.

## Enumerate and declare — statements in this repo of what `--quiet` drops

Swept the whole tree, not just the paths the issue named.

| # | Site | State | This PR |
|---|---|---|---|
| 1 | `ci-state.mjs` header | named `jobs` only | **fixed + pinned** |
| 2 | `ci-state.mjs` assignment comment | already names both | left (ground truth) |
| 3 | `run-team/SKILL.md` CI-Monitor | complete, pinned (#689) | left |
| 4 | `run-team/SKILL.md` finisher-duty | complete, pinned (#689) | left |
| 5 | `review-and-fix.md` step 6 | complete, pinned (#689) | left |
| 6 | `docs/specs/2026-07-23-fleet-plugin-design.md` script-surface row | complete, **unpinned** | **deliberately left** |

Site 6 is correct today and stays unpinned for the measured reason already in the
test's CEILING: its sibling spec rows (`worktree-audit`, `no-undo-audit`) pin by
running the script against fixtures, which for `ci-state.mjs` means standing up
`gh` fixtures. I confirmed those two spec pins cover their own scripts and not
this one, so the row really is unpinned — not pinned-by-a-neighbour.

No seventh site: `ci-state.mjs:26/:104/:120/:394` and `SKILL.md:571/:983` mention
`--quiet` but state exit-code or stderr behaviour, not payload drops.

## What this can wrongly refuse, and what the edit newly does

- **False red the SITES row could cause:** a legitimate future edit that keeps
  the header accurate but drops the backticks, or re-words `drops` to `omits`,
  now reddens. It reddens as *"slice anchor moved — re-anchor this test"*, which
  points at the fix, not as a silent pass. Measured.
- **What backticking newly breaks:** nothing greps the bare form. `"raw job list"`
  and `"suppresses the diagnostic stream"` each had exactly one hit in the tree
  (the line itself). Only two files read `ci-state.mjs` as text —
  `quiet-payload-prose.test.mjs` (the `Object.assign` line) and
  `ci-state-prose.test.mjs` (the `reasons.push` template); neither touches the header.
- **A false claim this PR had to avoid minting:** the test's CEILING said the
  pinned paragraphs "deliberately do not discuss" the stderr stream. The header
  states it in the same sentence, so adding the site would have falsified that
  line. Corrected rather than left standing.

## Also folded

`ci-state.mjs` was being read twice — via `read()` for a SITES row and directly
via `readFileSync(join(import.meta.dirname, ...))` for FIELDS. Now one
`read(CI_STATE_PATH)`.

## Mutation evidence

Run on an isolated mirror (four files copied into a bare directory tree; verified
no `.git` anywhere under it, so not a live worktree), committed before mutating.

| Mutation | Result |
|---|---|
| M1 gut the live clause to the pre-#690 wording | **RED — the new site and only it** (4 pass / 1 fail) |
| M2 keep both fields, un-backtick `--quiet` | **RED**, and specifically `slice anchor ... moved` |
| M3 reword the paragraph *around* the clause | green — non-catch control |
| M3b rewrap the whole header comment at 55 cols | green — reflow-safe |
| M4 decoy naming both fields *inside* the bound, real statement gutted | **RED** — the axis that defeated #689 does not reopen |
| M5 add a third field to the assignment | **all four sites RED**, derivation test green |
| M6 reshape the assignment so FIELDS cannot be derived | derivation guard RED — never a vacuous green |

M4 is the one worth naming: #689 shipped defeatable because a decoy inside the
paragraph satisfied an unanchored first-match. Here the clause anchors at the
slice start, so the real statement is always the match — checked rather than assumed.

## Tests

`./agent-test skills/fleet/scripts` — **985 pass / 0 fail**, from a measured
baseline of **984** on this branch at `0008b6b` (not 988; #692's 4 tests are
unmerged). +1 is the new site row. No `--no-verify`.

Closes #690


feigi:
Review findings applied in `2952b1c`.

**Applied (5).** The root defect reached by four dimensions — `SURVIVED[0]`, `SURVIVED[1]`, `UNVERIFIED[0]`, `UNVERIFIED[4]` — was one thing: the fourth pin site was given a document-shaped slice, and both of its bounds were wrong in source. Fixed structurally: the source site now anchors on `const quiet = has("quiet");` and slices the `//` block above it, so both bounds are code and no prose edit can move either. Also applied: `${SITES.length}` in place of the hard-coded "four" (`UNVERIFIED[5]`), and the rewrap of the orphaned comment line this diff created (`UNVERIFIED[1]`).

**Deferred and filed (2).**

- #701 — `ready-for-agent` — `quiet-payload-prose.test.mjs:1` says the two SKILL.md statements are "345 lines apart"; measured 353. Out of scope: line 1 is untouched by this diff.
- #702 — `needs-triage` — `ci-state.mjs:455` restates the drop-fact the header now owns, unpinned. Pre-existing, correct today; scope call.

**Deferred without an issue (1).** `UNVERIFIED[2]` (the `?? []` fallback lets four sites pass vacuously) — a refuter reproduced every fact and then measured the remedy: `assert.ok(FIELDS.length)` turns one precise red into five, four of which only point back at the first. The dedicated guard is already the single accurate red, and the vacuous greens only occur inside an already-failing run. Recorded in-tree at the `FIELDS` declaration instead of the backlog.

Mutation matrix, pre-fix (`81b24c9`) vs post-fix (`2952b1c`), `node --test skills/fleet/scripts/quiet-payload-prose.test.mjs`:

| axis | mutation | 81b24c9 | 2952b1c |
| --- | --- | --- | --- |
| — | pristine | 5/5 green | 5/5 green |
| bound-end | header statement gutted | red | red |
| decoy-inside-old-bound | gutted + decoy in code after `const quiet = …` | **5/5 GREEN** | **red** |
| decoy-inside-new-bound | gutted + decoy inside the `//` block | green | green (by design — twin markdown mutation is also green) |
| decoy-outside | gutted + decoy past the blank line | red | red |
| reflow | header rewrapped at 36 cols, content identical | **red (spurious "anchor moved")** | **5/5 GREEN** |
| anchor / wrong-fault | PROBE B: reword "suppresses" + drop `missing` | red, **"slice anchor moved"** | red, **names `missing`** |
| bound-start | anchor declaration renamed | n/a | red, "slice anchor moved" |
| spelling | header `--quiet` un-backticked | red, "anchor moved" | red, "no `--quiet` sentence saying what it drops" |
| field-list | a field added to the payload assignment | 4 sites red | 4 sites red |
| markdown | SKILL.md finisher clause gutted | 1 red, correct site | 1 red, correct site |

Gate: `node --test skills/fleet/scripts/*.test.mjs` → `tests 985 / pass 985 / fail 0`, matching baseline. No `--no-verify`.

Issue #388, opened 2026-08-11
ledger.mjs check: the near-miss rows carry the signal the verdict throws away — clean alongside a 0.45 near-miss, tracker-hit on two score-0 issues


feigi:
Measured across **five** invocations in a single `/fleet:run-team` run (2026-08-11), by three different agents. The verdict field was wrong in **both** directions, and in every case the `near` rows in the same payload pointed at the right issue.

## The pattern

| # | subject | verdict | reality |
|---|---|---|---|
| 1 | fix-applier filing a citation-drift deferral | `clean` | direct `gh issue list --search "citation drift"` surfaced **#129, #317, #48** — all that class. Nearly filed a duplicate of the open aggregator. |
| 2 | fix-applier filing the missing-KB-file debt | `tracker-hit` | matched #340, #356, #280, #349, #268 — five refuted-record issues sharing `wontfix` vocabulary, **none** about the subject. Honouring it would have dropped a real deferral. |
| 3 | controller filing the reap/detached-worktree gap | `clean` | correct, but the nearest row was #265 at 0.30 — the genuinely adjacent issue. |
| 4 | fix-applier recording an NBSP divergence | `clean` | its own `near` rows listed **#204 twice** (0.22, 0.20), one reading verbatim `nbsp/gojq dependency-scan divergence — comment on existing issue, PR331`. #204 was exactly the right issue. |
| 5 | controller, filing **this** issue | `tracker-hit` | named **#381 (score 0)** and **#351 (score 0)** — both unrelated — while its `near` rows named **#231** at 0.45, which is the genuinely adjacent one. |

Row 5 is this issue's own dedupe check. It is self-demonstrating.

## The mechanism, and why the fix is small

The tracker query is an ANDed reduction of three terms drawn from the subject. Three ANDed terms over issue titles is simultaneously **too narrow** (rows 1 and 4: the right issue uses different words, so zero hits, reported as `clean`) and **too coarse** (rows 2 and 5: three generic tokens match anything, including issues the scorer itself rates **0**).

Meanwhile the ledger-side `near` rows — which already exist, are already computed, and are already printed — named the correct issue in rows 1 (via #129's sibling), 4 and 5. The signal is being computed and then discarded by a verdict that only consults the tracker query.

## Direction (approach open — hence `needs-triage`)

[...]

Issue #283, opened 2026-08-10
zsh silently rewrites an unquoted `git show $ref:<path>` for most path prefixes — every citation check in this repo is exposed


feigi:
Measured while adjudicating a review finding on PR #272. Three parties had three different accounts of this and **all three were wrong**; the real behaviour is worse than any of them.

## Measured, zsh 5.9

`$SHA` holds `/a/b/c`; the command is `print -r -- $SHA:<path>` with **no quotes**:

| unquoted expression | result | verdict |
|---|---|---|
| `$SHA:skills/x` | `zsh: bad substitution` | loud failure |
| `$SHA:hooks/x` | `/a/books/x` | **silent, wrong** |
| `$SHA:tests/x` | `cests/x` | **silent, wrong** |
| `$SHA:agents/x` | `/a/b/cgents/x` | **silent, wrong** |
| `$SHA:runme/x` | `/a/b/cunme/x` | **silent, wrong** |
| `$SHA:lib/x` | `/a/b/cib/x` | **silent, wrong** |
| `$SHA:etc/x` | `tc/x` | **silent, wrong** |
| `$SHA:queue/x` | `/a/b/cueue/x` | **silent, wrong** |
| `$SHA:docs/x` | `/a/b/c:docs/x` | correct |

**Cause:** zsh applies *history modifiers* to a `${var}:x` suffix. `a A c e g h l p q r s t u x` are modifiers. `:h` takes the head, `:t` the tail, `:e` the extension, `:r` the root, and so on — the modifier consumes the first character of the path and the remainder is concatenated onto the transformed value. Only `s` (substitute) errors, because it demands a delimiter it never finds.

So `git show $SHA:hooks/statusline.js` does not fail. It runs `git show` against a **different, silently-constructed ref**.

## Why this matters here specifically

`docs/` — by luck — is one of the safe prefixes. **`skills/` is the loud one and `hooks/`, `tests/`, `agents/`, `lib/` are all silent.** This repo's own tree is `skills/`, `hooks/`, `docs/`, `workflows/`, `.github/`. Anyone verifying a citation with an unquoted variable ref gets:

- a loud error for `skills/…` (fine — you notice)
- **the wrong file's content, no error, exit 0, for `hooks/…` and `tests/…`**

A citation-verification sweep that hits the silent case concludes "the quoted fragment is not in that file" and re-anchors a citation that was correct. That is a fix that introduces the defect it was looking for.

## Three wrong accounts, for the record

[...]

PR #703, opened 2026-08-20
fix(fleet): head-equality settles the reflog, and finisher duty 2 covers applies

feigi:
Three finisher-rule defects **this run produced**, each measured more than once. Mid-run tooling fix per `run-team/SKILL.md`'s own rule — scoped to files this run invoked, no speculative polish.

## 1. The two-cause halt block over-fires on a settled reflog

`finish-pr-688`, `finish-pr-689` and `finish-pr-693` each spent effort adjudicating a `rebase` reflog entry that head-equality had already answered. `merge-bot-7-2` named the mechanism: the divergence gate is `worktree HEAD == PR headRefOid`, so on a clean tree at the pin, a rebase entry at or behind it is that head's **provenance** — the implementer's own pre-push replay — not divergence from it.

The rebase bullet already said "head still differs from your pin", so a careful reader wouldn't halt; but three careful readers still paid the cost. The short-circuit now leads the block instead of being implied by one clause inside it.

## 2. Duty 2 tested deferrals only, never applies

A fix-applier's "applied" is a claim like any other, and **nothing else checks it** — the review ran against a snapshot cut before those edits existed, and the finisher is the last agent before the label.

`finish-pr-692` and `finish-pr-693` both extended duty 2 this way *unprompted*, and `finish-pr-693` found **five** applies where the controller's hand-written list named three. A controller's list of what to verify is itself a claim.

## 3. "Filed as an issue" implies a tracker number; the real test is a home that outlives merge

The failure mode recorded from #598/#599 is a promise living only in a PR body, where `Closes #N` buries it at merge — that is a property of *where* it is written, not of deferrals. `finish-pr-689` proposed the correction; `fix-pr-695` then used it correctly, homing a finding in a committed `FIELDS` comment after its refuter **measured the suggested remedy worse than the defect** (one precise red becomes five, four pointing at the first).

## Pin re-anchored, and mutation-verified

`dedupe-guard-prose.test.mjs`'s `finisherDeferrals()` sliced on duty 2's heading, which change 3 renames — the documented "a prose pin's slice boundary is a heading you must not rename" trap, hit deliberately here because the old heading is exactly what the correction supersedes. So the anchor moves with it:

| mutation | result |
|---|---|
| strip `ledger.mjs check` from duty 2 | **995/1 — that pin and only it** |
| reword prose *outside* duty 2 (non-catch control) | **996/996 green** |

The other `ledger.mjs check` occurrence elsewhere in the file survived the first mutation, which confirms the slice is still correctly bounded.

Suite: **996 pass / 0 fail**, unchanged from baseline. Committed before mutating; restored with `git show HEAD:<path> > <path>`, never `git checkout --`.

## Not included

- `settings.json` is dirty in the checkout and **untouched** — out of bounds for tooling fixes, and it predates this session.
- `ledger.mjs check`'s exit-3-on-score-0-rows (measured 5× this run) is #388, already open, and is a script fix rather than a doc one.
- The zsh `$ref:$path` `bad substitution` trap, which I tripped while verifying, is #283, already open.


feigi:
Deliberately left unlabelled and open at drain: the maintainer ruled this goes through the review workflow **next run** rather than being chore-labelled, because it edits the rules the fleet itself runs on and this run produced three consecutive prose pins that their own authors had mutation-tested and that were still defeatable. Self-verification is the thing that kept failing.

CI green at behind=0 when filed; expect staleness once #704 lands.

feigi:
## Review findings applied — `e406c6d`

**Applied (4 survived findings → 1 prose fix + 2 pins).** The three test findings were three framings of two rules, so two pins were written, not four.

| Finding | Action |
|---|---|
| `survived[0]` critical/correctness — head-equality compared `PR headRefOid`, not the dispatch pin | **Applied.** `run-team/SKILL.md` halt block now reads `worktree HEAD == the SHA you were dispatched against`, with `headRefOid` demoted to a second, separate read (pushed vs unpushed). |
| `survived[1]` + `survived[3]`(b) critical/important — head-equality rule unpinned | **Applied** as one pin in `finisher-pin-race-prose.test.mjs`, derived from the post-fix tree (not copied from the finding — `survived[0]` rewrote the sentence they quoted). |
| `survived[2]` + `survived[3]`(a) critical/important — duty-2 claimed-APPLY rule unpinned | **Applied** as one pin in `dedupe-guard-prose.test.mjs`. |

Suite: `node --test skills/fleet/scripts/*.test.mjs` → **998 tests, 998 pass, 0 fail** (52 files; baseline 996 + the 2 new pins).

Each pin was mutation-verified — killed by its own mutant, and green under a change it should not catch:

- **pin a / M1** operand reverted to `worktree HEAD == PR headRefOid` → 1 fail of 998, only this pin.
- **pin a / M2** head predicate spliced in front of the live-editor dirty read → 1 fail of 998, only this pin; the pre-existing loose `git status --porcelain.*dirty` assertion stayed green, which is the discrimination it was added for.
- **pin a / C1** block re-wrapped at a different width + an unpinned sentence reworded → 998/998 green.
- **pin b / M3** the `git diff origin/main...HEAD` sentence deleted, claim sentence kept → 1 fail of 998, only this pin.
- **pin b / C2** the clause between the two anchors reworded and reflowed → 998/998 green.

**Deferred (1 issue filed):** #708 — `REFUTED` : duty 2's `git diff origin/main...HEAD` needs no `BASE_REF` override. A refuter measured the remedy as a byte-identical no-op (nothing in the repo sets `BASE_REF`), 199/199 PRs based on `main`, and `claim-ticket.sh:110` cutting every fleet branch from `origin/main` by construction. Filed `needs-triage` so it is not re-raised and re-applied.

**No action, not filed:** the other two `suggestion`-band findings each report a pre-existing, in-tree-documented property rather than a defect this PR introduces — the presence-pin ceiling (already disclosed verbatim in `dedupe-guard-prose.test.mjs`'s own header, and re-measured here as C2) and the unverifiable run anecdotes (same category as the file's existing `Measured:` claims, contradicted by nothing in the tree).

The three `refuted` findings stand refuted and were not applied.

PR #707, opened 2026-08-20
fix(fleet): report caller text with printf, not echo, across nine fleet scripts

### LINKED ISSUE #484: fleet scripts: echo expands backslash escapes in caller text — ten scripts still affected (no-undo-audit.sh closed by #480)

feigi:
Closes #484

`echo` expands backslash escapes in its operand in dash (CI's `/bin/sh`) and macOS `sh`. `\c` truncates the line *and* swallows its newline, so the next stderr line collides with it. Measured on this branch before the fix — three consecutive `claim-ticket.sh` lines arriving as one:

```
    would: git worktree add .worktrees/7-back    would: (cd .worktrees/7-back    would: write .worktrees/7-back
```

Remedy is the one PR #480 proved five times in `no-undo-audit.sh`: `printf '%s\n' "…"`.

## Enumerate and declare

**Covered — the reachable class, 9 files.**

*The nine `die()` definitions* the ticket enumerates. All nine now match `no-undo-audit.sh`'s already-fixed form. Exit codes untouched: `derive-testcmd.sh` refuses at **1**, the other eight at **2** — an asymmetry the ticket does not mention, now asserted per script.

*Path-carrying sites the ticket names:* `worktree-audit.sh` `unknown()` / `MISSING on disk` / the status line; `release-ticket.sh` ×3; `reap.sh` would-remove; `claim-ticket.sh` ×3; `inflight.sh` worktrees line.

*Siblings the ticket does **not** name*, found by enumerating the class rather than the list. Four are shared helpers, so one edit each covers every caller:

| site | callers | what reaches it |
|---|---|---|
| `reap.sh` `keep()` | 11 | `$wt` (×4), `$ignored` (filenames), git stderr (×2) |
| `inflight.sh` `add_unknown()` | ~20 | `$refsdir`, `$bad`, `$wtroot`, `$wtfile`, git/gh stderr |
| `release-ticket.sh` `block()` | 14 | `$stray`, `$orphan`, `$wtroot`, `$wt` |
| `release-ticket.sh` `halt()` | 3 | `$wt`, `$branch`, git stderr |

Plus `claim-ticket.sh`'s other three `$wt`/`$runner` lines, `inflight.sh`'s EXIT trap (`mktemp` paths), and the `agent-test` runner `claim-ticket.sh` generates — its `$arg` is a user-supplied path.

**One correction to the ticket's own reasoning.** It rules refnames safe by construction, since `git check-ref-format` rejects a backslash. True for refs read back from git — but not in `claim-ticket.sh`, where `slug=$2` is unvalidated (only `$issue` is checked numeric, as the script's own comment at :382 says). `$branch` there is a *proposed* name git has not accepted yet, so it is swept.

**Deliberately left — measured inert, not skipped.**

- Refname / SHA / issue-number / count operands (~45 sites). The ticket's ruling holds everywhere except the `claim-ticket.sh` case above.
- Literal-only operands (`$install` is one of five literals, `$testcmd` one of two).
- **Function stdout return values** — `prove-merge.sh`'s `echo true`/`false`, `release-ticket.sh:496-505`'s `echo Indeterminate`/`Released`/…, `derive-testcmd.sh`'s `npm test --`. These are payload, not operator text; all literal. Converting them buys nothing and puts a `printf` at a function's last command for no reason.
- **`statusline-command.sh` — confirmed inert, as the ticket asked, rather than swept in.** It is `#!/usr/bin/env bash` (bash `echo` does not expand escapes without `xpg_echo`) and every site is a literal `|| echo "0"` fallback feeding `awk`, not operator output.

**Scope cut:** none. The ticket sanctioned a narrower PR ("one PR per script, or just the ten `die()` lines first"); the full reachable class turned out to be ~30 lines, so taking the cut would have left siblings broken for no saving.

## What this change could newly get wrong

`printf` treats its first argument as a format string — a hazard `echo` never had. Each risk was measured, not assumed:

- **`%` in a message.** `printf '%s\n' "$msg"` passes `100% done %s %d` through verbatim; `printf "$msg\n"` emits `100 0one  0`. The message is always an *argument*. Pinned in both directions, and mutation-proven below.
- **Exit status.** `printf` and `echo` both return 0 as a function's last command, and on a closed fd both abort identically — the feared divergence does not reproduce. `die()` ends in `exit` anyway, and every script's code is asserted.
- **`inflight.sh`'s EXIT trap** documents that its `||` reporter must return 0, or a failing cleanup would exit 1 and the contract reads 1 as "taken". `printf` returns 0 exactly as `echo` did (measured); the comment naming `echo` is updated.

## Verification

- Full suite **1012 pass / 0 fail**.
- CI's exact gate `shellcheck -x -S warning` passes. Over the tracked `*.sh` set, **shellcheck output is identical before and after — same codes, same files, same count** (10: SC2015 ×8, SC2016 ×1, SC1003 ×1). Nothing added or removed on `inflight.sh` or any swept script, so the Shellcheck comment in `ci.yml` needs no adjustment for this change.
- **Mutation-checked, 8 mutants.** Seven reverted one `printf` each: every one reddened **exactly one** test, the right one (14 pass / 1 fail). The eighth regressed the fix to the format-string form — it reddened **only** the `%` test, all nine `\c` tests staying green, which is what proves the acceptance direction earns its place. A comment-only control stayed 15/15 green. Scripts restored byte-exact.

Fixtures use `\c` throughout, never `\s`: unknown escapes pass through unharmed and read as coverage while catching nothing — the trap that left `no-undo-audit.test.mjs` blind to five live defects.

https://claude.ai/code/session_01TUuYukPqzpXEB9Qb4RTaPo


feigi:
## Review fixes: what the new guards actually catch

Applied at `5a6e8da`. Every guard added or changed here was mutation-verified against the full suite (`node --test skills/fleet/scripts/*.test.mjs`, 1015 tests) in a `git archive` extraction — one mutation per run, restored from a pristine copy between runs.

| # | mutation | result |
|---|---|---|
| m1 | `verify-sha.sh:33` printf → `echo` | 1014/1015 — RED: `verify-sha.sh: die() reports a \c-carrying message verbatim`, that one only |
| m2 | `verify-sha.sh` `die()` → `printf "$NAME: $1\n"` | 1014/1015 — RED: `every printf … literal format string`, that one only |
| m3 | `inflight.sh:703` printf → `echo` | 1014/1015 — RED: `probe 3: a backslash in the worktree path…`, that one only |
| m4 | `release-ticket.sh` dry-run printf → `echo` | 1014/1015 — RED: `the dry run's plan names a backslashed worktree path verbatim`, that one only |
| m5 | `die "simulated unrelated failure"` injected into `reap.sh` | reap KEEP test fails as `AssertionError: stderr: reap: simulated unrelated failure` — the status assertion, not the escaping headline |
| c1 | trace arg `"$branch"` → `"${branch}"` | 1015/1015 green (control) |
| c2 | die format `'%s: %s\n'` → `"%s: %s\n"` (double-quoted **literal**) | 1015/1015 green (control) |
| c3 | `inflight.sh` printf arg → `"${wt}"` | 1015/1015 green (control) |
| c4 | `release-ticket.sh` printf arg → `"${wt}"` | 1015/1015 green (control) |

**m3 and m4 are the two mutants that passed all 1012 tests before this change.** Reverting either of those lines to `echo` was invisible to every suite in the tree.

**m2 is the mutant the `\c` fixtures cannot see.** `printf "$NAME: $1\n"` still fixes the backslash bug — POSIX printf expands `\c` only under `%b`, never in a format it was handed — so all nine `\c` fixtures stay green over it while a `%`-carrying message is silently consumed. The new static test is the only thing that reds it.

**c2 is the control that matters most for anyone editing this later.** The invariant is *interpolation*, not quote style: a double-quoted **literal** format is fine and stays green. `inflight.sh`'s EXIT trap needs exactly that, because the trap body is already single-quoted. A rule mandating single quotes would break it.

## Coverage figures

`every printf in the fleet scripts is handed a literal format string` scans `skills/fleet/scripts/*.sh` and rejects any `printf` whose format token interpolates: **119 printf calls checked, 0 flagged.** It carries floor assertions on both the file count and the call count, so a glob matching nothing cannot pass vacuously.

The four tracked `.sh` files outside that glob — `.github/scripts/check-tracked.sh`, `.github/scripts/rerun-rebase-check.sh`, `hooks/cost-reset-on-clear.sh`, `statusline-command.sh` — were scanned with the same rule: **39 printf calls, 0 flagged.**

`shellcheck`'s SC2059 is the same check, but it reports at `info` and CI runs `shellcheck -x -S warning`, which filters it out before it prints. That is why the invariant is pinned in the suite rather than delegated to the lint step.

## Also measured

`verify-sha.sh:33` was the last site in these scripts where raw argv reached stderr before any git command accepted it. The other candidates were checked rather than assumed: `release-ticket.sh:803,836` interpolate a `$branch` built from unvalidated argv but are gated on `has_branch`, set only by a `rev-parse --verify` hit; `no-undo-audit.sh:440` is gated on its own `rev-parse --verify` at `:212`.

`./agent-test skills/fleet/scripts` → tests 1015, pass 1015, fail 0, skipped 0.

PR #719, opened 2026-08-20
simplify(fleet): bind ledger.test.mjs's CLI fixture to t.after(), drop 7 try/finally blocks

### LINKED ISSUE #569: ledger.test.mjs: the seven CLI tests hand-roll try/finally cleanup — t.after() plus a fixture-bound cli removes 26 lines
Deferred from PR #563 review (issue #362), finding `unverified7` — severity **suggestion**, dimension **simplify**, `skills/fleet/scripts/ledger.test.mjs:695`.

**State: SURVIVED a refuter biased to refuse.** Not applied only because PR #563 merged (`b2e7f90`) while the finding was still being checked — the fix-applier never held its text until after the push. Nothing about it is unresolved; the approach is decided and measured, hence `ready-for-agent`.

## The finding (verbatim claim)

> The five new CLI tests each repeat an identical `try { ... } finally { rmSync(dir, ...) }` and thread `(env, cwd)` into a free-standing `cli` helper; `t.after()` plus a fixture-bound `cli` removes ~20 lines of boilerplate and both threaded parameters with no loss of cleanup.

## Suggested fix

```js
function cliFixture(t) {
  /* …same setup… */
  t.after(() => rmSync(dir, { recursive: true, force: true }));
  return { dir, cli: (...args) => spawnSync(process.execPath, [SCRIPT, ...args], { encoding: "utf8", env, cwd: dir }) };
}
```

then `test("…", (t) => { const { cli } = cliFixture(t); … })` — no try/finally, no `env`/`cwd` threaded at the call sites. The free-standing `const cli = (args, env, cwd) => …` goes away.

## Verification (refuter, biased to refuse — could not break it)

Measured against **`217c5a3`** (the post-review tree, **seven** CLI tests, not the five the finding was written against). Baseline `tests 39 / pass 39 / fail 0`; rewritten copy likewise `39 / 39 / 0`.

**Cleanup on failure — the whole point of the `try/finally` being replaced.** `t.after()` and `finally` diverge exactly when the body throws, so that was measured directly, each variant in its own isolated `TMPDIR`:

| variant | dirs before | dirs after | result |
|---|---|---|---|
| rewrite, plain `Error` thrown | 0 | **0** | 7 fail, no leak |
| rewrite, assertion failure | 0 | **0** | 7 fail, no leak |
| original `try/finally`, plain throw (control) | 0 | **0** | 7 fail, no leak |
| original `try/finally`, assertion failure (control) | 0 | **0** | 7 fail, no leak |
| **negative control** — `t.after` line deleted | 0 | **7** | leak detected |

The negative control is what makes the zero rows evidence rather than a broken counter.

**Hook timing.** Instrumenting `cliFixture` to count live `ledger-cli-*` dirs printed `LIVE-AT-SETUP 1` seven times — one fixture alive at a time, never accumulating. A standalone probe confirmed the `SETUP A / BODY A / AFTER A / SETUP B …` interleave on the pass path, the assertion-failure path and the uncaught-throw path. `t.after()` is per-test here, not file-end.

**Size.** 851 → 825 lines, **26 removed** (the finding claimed ~20 and understated it, because two more tests using the same pattern landed after it was written). Against base `0e76840`: `+167` → `+141`. The finding's own boilerplate grep yields 21 on `217c5a3` and 0 after.

**Collateral: none.** `ledger.test.mjs` is referenced only by itself — no source-lift or prose test reads it, and neither globbing test (`no-undo-audit`, `arg.test.mjs`) scans it. All eight `cli(` call sites already pass `cwd: dir`, so binding `cwd: dir` inside the fixture loses no capability.

One hazard found and dismissed: a throw between `mkdtempSync` and the `t.after` registration would leak — but the original leaks identically, since its `mkdtempSync` also runs inside `cliFixture()` before the `try` opens. Not a differentiator.

## Two things to preserve while doing it

1. **`cliFixture()`'s comment must stay true of the rewritten helper.** PR #563 replaced a false justification there (finding `survived3`: it claimed the `git` symlink was needed because *"`defaultLedgerPath()` shells out to it"*, which no test in that block reaches). The replacement explains the symlink as deliberate convention plus a guard for the next test added. It sits inside the function body and transplants verbatim — do not let the signature change strand it.
2. **Do not weaken the `#362` pins.** The rewrite touches every test that pins the `--file`/`--require-file` parser, including the must-accept pin (a path beginning with `-` is accepted **and actually read**, exit 1) and the two `=`-spelling pins, which assert in **both** argv positions. Re-run the mutation receipts in #568 after the rewrite — a refactor of the pins is exactly when a pin quietly stops pinning.

feigi:
Closes #569

`cliFixture()` takes the test's `t`, registers its own `t.after(() => rmSync(dir, ...))`, and returns a `cli` already bound to the fixture's `env` and `cwd`. The seven #362 CLI tests stop hand-rolling an identical `try { … } finally { rmSync(dir, …) }` and stop threading `(env, cwd)` through every call; the free-standing `const cli = (args, env, cwd) => …` is gone.

## Measured at this head (base `618d0a7`)

| metric | origin/main | here |
|---|---|---|
| `ledger.test.mjs` lines | 1015 | 1005 (refactor −25, new pin +15) |
| the finding's own boilerplate grep | 21 | **0** |
| `finally` occurrences | 10 | 4 (2 are prose; see below) |
| `./agent-test skills/fleet/scripts` | 1023/1023 | **1023/1023** |
| `ledger.test.mjs` alone | 46/46 | **48/48** |

## Cleanup on failure — the whole point of the `try/finally` being replaced

`t.after()` and `finally` diverge exactly when the body throws, and the suite is blind to it: every CLI test passes on the happy path with or without cleanup. Measured directly, each variant in its own isolated `TMPDIR`, by injecting a failure into all seven CLI test bodies:

| variant | dirs before | dirs after | result |
|---|---|---|---|
| rewrite, all passing | 0 | 0 | 48/48 |
| rewrite, plain `Error` thrown | 0 | **0** | 7 fail, no leak |
| rewrite, assertion failure | 0 | **0** | 7 fail, no leak |
| control — origin/main `try/finally`, plain throw | 0 | **0** | 7 fail, no leak |
| control — origin/main `try/finally`, assertion failure | 0 | **0** | 7 fail, no leak |
| **negative control** — the `t.after` line deleted | 0 | **8** | leak detected |

The negative control is what makes the zero rows evidence rather than a broken counter. 8, not 7, because the new pin below makes an eighth fixture.

## The one pin the file was missing

Nothing above reddened if the removal were dropped, so this PR adds the test that does. It runs as a **subtest**, because a test's own `after` hooks have not fired while its body is still running; the parent's `await` resolves only once they have (probed separately).

Discrimination, per the negative control above: under the deleted `t.after`, **exactly one** test reds — the new one. Two changes it should *not* catch both stay green at 48/48: moving the `t.after` registration to the top of `cliFixture`, and dropping `force: true` from the `rmSync`.

## #362's pins re-verified — every mutation receipt from #568 re-run

Baseline 48/48. Every mutant reddens exactly the test #568 recorded, not merely the same count:

| mutant | pass | fail | which |
|---|---|---|---|
| drop the `--file=` guard | 47 | **1** | the `--file=` test only |
| drop the `--require-file=` guard | 47 | **1** | the `--require-file=` test only |
| widen `--file=` → `--file` (over-broad) | 6 | **42** | every must-accept path, incl. the accept-pin's exit-1 control |
| widen `--require-file=` → `--require-file` | 46 | **2** | both `--require-file` must-accept pins |
| guard only `argv[0]` (the false-claim shape) | 46 | **2** | both `=`-spelling tests, second argv position |
| swap the two guard lines (should NOT be caught) | 48 | 0 | green |
| drop the pre-existing `file.trim() === ""` clause | 47 | **1** | the whitespace test only |

## The two binding constraints

1. **`cliFixture()`'s comment is transplanted verbatim** — the diff touches only the signature line above it and the return below it. It stays true: the `git` symlink is untouched, and the `--file`/`defaultLedgerPath()` reasoning it rests on is unchanged.
2. **`cwd: dir` is safe to bind.** All eight `cli(` call sites already passed it, verified at this head. One test *depends* on it and is left in place to exercise it: the accept-pin hands `ledger.mjs` a **relative** path, which begins with `-` only because the spawn resolves it against `dir`. Point the fixture's `cwd` anywhere else and that test reds.

## Deliberately left

`finally` at `:182` and `:803` (the other two of origin/main's 10 occurrences; the remaining two are prose).

- **`run()`'s `finally` (`:182`)** — a plain helper, not a test, with no `t` to bind to and ~40 callers. Converting it means threading `t` through every one of them, and its `ledgerRepoDir: realpathSync(dir)` is deliberately resolved *before* cleanup.
- **the `ledger-noreop-` test (`:803`)** — the identical shape, but a single non-repeated instance outside the `cliFixture` class, using `run()` rather than `cli`. Converting it saves 3 lines and expands the diff into a test this ticket does not name, while `run()`'s `finally` keeps the file's occurrence count non-zero regardless.

`cli` keeps taking the argv **array** rather than the rest-param `(...args)` the ticket sketched: both loop call sites already hold an array, and the array form is the smaller diff at the other six.


feigi:
## Review adjudication — 0 applied, 2 no-change. **HEAD unchanged at `f9ae87c7`.**

Review returned 0 survived, 0 refuted, 2 suggestions. Both suggestions arrive unchecked (the workflow budgets that band 0 refuters), so each was split by scope and checked here.

**`unverified[0]` — `ledger.test.mjs:998`, `correctness`, `suggestion` — NO ACTION, remedy already satisfied.**
The claim: the 14-line #569 guard test is beyond the ticket's stated deliverable and cuts the promised size win from −26 to −10 net. Its `suggested_fix` asked for exactly one thing: disclose the real net delta in the PR body. That disclosure was already published before the review ran — the measurement table's first row reads `| ledger.test.mjs lines | 1015 | 1005 (refactor −25, new pin +15) |`. The finding read the diff, not the body. Guard kept: the refactor moves cleanup from `finally` to `t.after()`, which changes behaviour only on the throw and assertion-failure paths, and every CLI test passes on the happy path either way — so without the guard the new semantics are unpinned. The finding itself measured the guard discriminating correctly (48 tests, 47 pass / 1 fail on the exact removal it names, other seven CLI tests still green). Not filed: nothing is open.

**`unverified[1]` — `ledger.test.mjs:878`, `comments`, `suggestion` — REFUTED, deferred to #722.**
The claim: the `cliFixture()` comment's clause "which begins with `-` only because the spawn resolves it against `dir`" can be misparsed as attributing the leading `-` to the `cwd` binding, and a reader acting on that reading could unbind `cwd`. In scope (this PR wrote the comment), so one refuter ran it by execution rather than by reasoning.

Mechanism, measured across four spellings of the same spawn — the leading `-` is controlled by staying relative vs `join`ing, `cwd: dir` controls resolvability only, and the axes are orthogonal. So the clause is a compression, which is the finding's factual half and it is correct. Replaying the real accept-pin at `f9ae87c7`, unbinding `cwd` gives `exit=0` against an assertion demanding `exit=1`, with `ledger.mjs` printing `WARNING — ledger file not found: -weird--ledger.md`. The asserted harm is therefore measured-false: the misparse cannot survive one test run, the same sentence's first clause already justifies the binding independently, and the clause's own pointer sends a reader to the precise pre-existing comment at the test site (`:934-938`). Filed `needs-triage` — refuted leaves it unsure anything is broken.

**Gate:** `node --test skills/fleet/scripts/*.test.mjs` from the worktree → `tests 1023 / pass 1023 / fail 0 / skipped 0`. Full suite, not a partial tree.

Nothing staged, nothing committed, nothing pushed — deferring everything is the outcome here, not a stall. `rebase-check` red is behind-count staleness from #718's merge, not a failure.

PR #724, opened 2026-08-20
fix(fleet): claim-ticket.sh's worktree guard misses a dangling symlink

### LINKED ISSUE #188: claim-ticket.sh:30: [ -e ] on a dangling symlink at the worktree path gives the wrong refusal
Deferred from PR #177 review. Nit — it **fails closed**, only the diagnosis is wrong.

## The defect

`skills/fleet/scripts/claim-ticket.sh:30`:

```sh
[ -e "$wt" ] && die "$wt already exists — ticket may already be claimed"
```

`[ -e ]` stats, so it follows symlinks; `git worktree add` refuses on lstat semantics. A **dangling** symlink at `$wt` splits the two exactly as `[ -d ]`/`find` split in #99 — the guard sees nothing, `git worktree add` then refuses anyway:

```
dangling symlink at .worktrees/42-slug:
  exit=2   fatal: '.worktrees/42-slug' already exists
           claim-ticket: worktree add failed

real directory at .worktrees/42-slug:
  exit=2   claim-ticket: .worktrees/42-slug already exists — ticket may already be claimed
```

Both exit 2, nothing is created, no lockfile touched. Only the message differs, and by then the issue is already labelled `in-progress` — as it is for any `worktree add` failure.

## The fix

```diff
-[ -e "$wt" ] && die "$wt already exists — ticket may already be claimed"
+{ [ -e "$wt" ] || [ -L "$wt" ]; } && die "$wt already exists — ticket may already be claimed"
```

`-L` catches the dangling case `-e` misses.

## Why file it at all

The reviewing specialist rated the cost of fixing as exceeding the harm, and I agree — this is filed so the finding is not dropped, not because it is worth a claim on its own. Reasonable dispositions include closing it as `wontfix`, or folding the one-line change into the next PR that touches `claim-ticket.sh` for another reason.

`ready-for-agent` because the fix is settled and one token wide, not because it is urgent. Same `[ -d ]`/`[ -e ]`-vs-`find`/`git` symlink-semantics family as #99 and #186.

Surfaced by the type-design reviewer on PR #177 during a root-cause sweep for other instances of the mismatch #99 fixed.

feigi:
`claim-ticket.sh`'s pre-claim guard used `[ -e "$wt" ]`, which **stats** and so follows symlinks, while `git worktree add` refuses on **lstat** semantics. A dangling symlink at `$wt` splits the two — the same `[ -d ]`/`[ -e ]`-vs-`find`/`git` symlink-semantics family as #99 and #186.

```diff
-[ -e "$wt" ] && die "$wt already exists — ticket may already be claimed"
+{ [ -e "$wt" ] || [ -L "$wt" ]; } && die "$wt already exists — ticket may already be claimed"
```

## Measured, not reasoned (git 2.50.1)

The ticket frames this as fails-closed and cosmetic. Under `--apply` that holds — both spellings reach exit 2 and create nothing, only the diagnosis differs. **The default dry run is where it does not hold:**

| input | mode | base | fixed |
|---|---|---|---|
| dangling symlink | dry run | **exit 0** + receipt naming the path claimable | exit 2 `claim-ticket: … already exists` |
| dangling symlink | `--apply` | exit 2 `fatal: '…' already exists` / `worktree add failed` | exit 2 `claim-ticket: … already exists` |
| real directory | both | exit 2 `claim-ticket: … already exists` | unchanged |
| free path | both | exit 0, worktree created | unchanged |

So base did not merely misdiagnose the path — in the script's **default** mode it predicted a claim that could not be made, and under `--apply` it deferred the refusal until after the `in-progress` label was already on the issue.

## No false refusal added

`-L` is true for *any* symlink, so the widening direction was checked explicitly against `git worktree add`:

| path at `$wt` | `-e` | `-L` | `git worktree add` |
|---|---|---|---|
| dangling symlink | F | **T** | **refuses** |
| symlink loop | F | **T** | **refuses** |
| symlink → real empty dir | T | T | accepts |
| real empty dir | T | F | accepts |

Every path `-L` **newly** admits to the guard is one git refuses too, so it cannot refuse a claim that would have worked. Rows 3-4 are already refused by the shipped `-e` and are untouched here.

## Reuses the predicate this repo already ships

`{ [ -e ] || [ -L ]; }` is byte-identical to `release-ticket.sh:464`'s `occupied()`, which carries the same rationale for residue **this fleet leaves itself**: a symlink pointing *at* the registered worktree directory survives the `git worktree remove` that deletes its target, and is dangling afterwards. `release-ticket.sh:561` already named `claim-ticket.sh`'s `[ -e "$wt" ] && die` as the thing that then "died on it, with nothing left to explain why". This closes that loop.

## `set -eu` safety

`{ …; } && die` restructures an `&&` list under `set -eu`, and this script's header states it uses exit 2 for every refusal and **has no exit 1**. Verified under `sh`, `dash`, `bash`, `zsh` and `ksh`: the guard-false path is exit 0 in all five, refusals are exit 2, and the tail-position status is identical to the old spelling. The guard is not the script's last command, so tail position never applies.

## Tests

One test added, pinning **both** directions — the dangling symlink it must reject, and a free path it must still accept.

- Full `skills/fleet/scripts` suite: **1024 pass, 0 fail**.
- CI-exact `shellcheck -x -S warning`: clean (the two remaining `info` hits are pre-existing, on lines 25 and 406).

### Mutation results

| mutation | result |
|---|---|
| revert guard to `-e` only | **1 fail / 65** — mine, and *only* mine |
| guard always refuses (`\|\| true`) | **62 fail / 65**, incl. the accept assertion `` `-L` must not refuse a path that is simply not there `` |
| reorder to `{ [ -L ] \|\| [ -e ]; }` (equivalent) | **green** — pins behaviour, not spelling |

The first row is also a coverage finding: before this test, the suite was structurally blind to this mode — no existing test changed colour when the guard was reverted.

Closes #188

https://claude.ai/code/session_01AqruLG5fxisLtnwh4pr4rq


feigi:
## Review findings applied — `ffac5e7`

**Suite: 1024 tests / 1024 pass / 0 fail / 0 skipped**, `node --test skills/fleet/scripts/*.test.mjs` from the worktree. The count is unchanged from the previous head because the new coverage is a third case inside an existing test, not a new test.

### Applied

1. **`claim-ticket.sh:43` + `claim-ticket.test.mjs:965` — false claim in the PR's own prose** (`critical`, comments, survived 2-0). The guard comment said an unfixed `-e`-only guard leaves the refusal to fire *"three mutations later"*. Re-measured on git 2.50.1 with the guard reverted, a dangling symlink at `$wt` and a stubbed `gh`, under `--apply`: the label completes, then `git worktree add` dies **inside itself** having already created the branch ref — `install` and the runner write are never reached, 0 worktrees registered. **One** completed mutation, and the refusal lands *during* the next attempt, not after it. The wrong count was duplicated verbatim in the test's assertion message; both sites now say what was measured. Also corrected the adjacent `--apply` clause, which claimed *"only the diagnosis differs"* — base additionally lands the label and strands a branch ref.

2. **`claim-ticket.test.mjs` — the `-e` half was pinned nowhere** (`important` × 2, correctness + tests, both survived 2-0, applied as **one** pin). Deleting the `-e` half outright left the whole suite green, so the *"real directory, both exit 2"* row this PR published was a measurement nothing held.

   Pinned as a **third case inside the existing `#188` test**, not as a test of its own. A separate refusal-only test would survive the widen-everything mutation — the exact vacuity the block's "Both directions" prose exists to prevent — so it shares the accept case already there. The test now covers **dangling-symlink refusal, real-directory refusal, and free-path accept** against one guard.

   Dry run, the mode with no downstream net: `git worktree add` never runs there to refuse the occupied path on its own, so a regression surfaces as exit 0 and a receipt rather than a later refusal. `--apply` stays deliberately unpinned, as the block's prose already argues.

3. **`claim-ticket.test.mjs:933` — line-number citation** (`suggestion`, correctness + types, one finding, survived its refuter). Dropped `:464`, matching the by-name form the sibling `.sh` comment in the same diff already used. Measured: **3 of 3** `file:line` citations into this script family are wrong today — including the one #516's ruling held up as *"explicitly checked and left because it is correct"* (`inflight.sh:436` → `release-ticket.sh:73`, since drifted). #718 applied the same by-name remedy one commit before this head.

### Mutation receipts

Scratch copies of the committed tree; every mutant diffed against the snapshot before running, to rule out one that silently failed to apply.

| mutant | suite | the `#188` test |
|---|---|---|
| delete the `-e` half → `[ -L "$wt" ] && die …` | `1024 / pass 1023 / fail 1` | **RED — and the only failure** |
| semantically equivalent re-spelling → `if [ -L "$wt" ] \|\| [ -e "$wt" ]; then die …; fi` | `1024 / pass 1024 / fail 0` | green |
| widen to refuse everything → `{ [ -e "$wt" ] \|\| [ -L "$wt" ] \|\| :; } && die …` | `1024 / pass 961 / fail 63` | **RED** |

So the test pins behaviour rather than spelling, discriminates the `-e` half specifically, and still fails locally if the guard is widened to refuse everything.

### Deferred and filed

| issue | finding | state | label |
|---|---|---|---|
| #728 | refusal message covers two symlink provenances with **opposite** claim states | refuted | `needs-triage` |
| #729 | tail-block fixture convention undocumented (8/8 on one triple) | refuted | `needs-triage` |

#727 (the EACCES-into-claimable residual, `chmod 000` on an ancestor) is pre-existing — base `[ -e ]` has the identical hole and `-L` uses `lstat`, which fails EACCES the same way — so this diff neither introduces nor widens it. Filed separately by the review controller, cross-referenced to #725.

A fourth finding (delete the test's accept-half as subsumed by the ordinary-slug test) survived refutation but was **ruled against and not applied**: `survived` means a finding withstood refutation, not that the change is worth making. Deleting it would have destroyed the test's local discrimination — the property the third mutation receipt above now demonstrates — and falsified the block's own header prose. Reasoning recorded on #731, closed as not-planned.

Notable from #728's refutation, since it bears on #725 and #727: a dangling link at `$wt` has **two** provenances with opposite claim states. Besides the rc-0 release residue this PR documents, `release-ticket.sh`'s rc-255 halt path leaves a dangling link with the branch **and** the `in-progress` label still alive — so the hedged *"may already be claimed"* is the only wording true of both, and the proposed split message would have told an operator to `rm` evidence of a still-live claim.

### Note on history

`4d2a5dd`'s own commit message carries the same two false claims (*"three mutations later"*, and *"both spellings reach exit 2 and create nothing"*). It is pushed, and rebasing to amend it is not worth an unrelated CI cycle, so `ffac5e7`'s body corrects the record instead.

`rebase-check` is red at behind 2 from #718/#719/#720 merging. That is behind-count staleness, not a failure, and was deliberately not rebased away.

Issue #725, opened 2026-08-20
gone() reads a dangling symlink as established-absent — triplicated in worktree-audit.sh, reap.sh, release-ticket.sh, and only the last compensates


feigi:
Surfaced by `impl-188` while implementing #188, declared-left rather than absorbed. Independently re-measured by the controller before filing. Same `-e`/`-d`-vs-lstat family as **#188**, **#99** and **#186**.

## The defect

`gone()` decides whether a path is *established-absent* (as opposed to unreadable, which must not be treated as absent). It walks up until it finds an existing ancestor, then requires the target itself to be missing and that ancestor to be searchable:

```sh
gone() {
  look=$1
  while [ ! -e "$look" ] && [ "$look" != "${look%/*}" ]; do look=${look%/*}; look=${look:-/}; done
  [ ! -e "$1" ] && [ -x "$look" ]
}
```

`[ -e ]` stats, so it follows symlinks. On a **dangling symlink** `-e` is false while `-L` is true — so `gone()` returns **true** for a path that is, in the sense that matters to `git worktree add`, occupied. That is the same split #188 fixed one layer up in `claim-ticket.sh`.

## It is triplicated, and only one copy compensates

```
$ grep -rn 'gone() {' --include='*.sh' skills/
skills/fleet/scripts/worktree-audit.sh:67
skills/fleet/scripts/reap.sh:58
skills/fleet/scripts/release-ticket.sh:381
```

`release-ticket.sh` compensates at a **call site**, not inside `gone()` — `:694` reads `if [ -n "$wt" ] && { [ -L "$wt" ] || { [ -e "$wt" ] && [ ! -d "$wt" ]; }; }`, and its comments at `:458` and `:692` state the hazard outright. The other two carry no such guard:

```
$ grep -c '\[ -L' skills/fleet/scripts/worktree-audit.sh
0
$ grep -c '\[ -L' skills/fleet/scripts/reap.sh
0
```

Repo-wide, `-L` appears only in `release-ticket.sh` (`:458`, `:464`, `:692`, `:694`). So two of the three copies are exposed and the third is protected by a guard that lives somewhere else entirely.

## Likely symptom

`worktree-audit.sh` — a registered worktree path occupied by a dangling symlink is reported as **`MISSING on disk`** rather than as occupied. That matters because `worktree-audit.sh` is what the fleet controller reads to decide whether a replacement member would *redo* work or *destroy* it.

[...]

PR #726, opened 2026-08-20
simplify(fleet): drop reap.sh's $wt_present — the next `if` re-tests it directly

### LINKED ISSUE #470: reap.sh:101 — the wt_present boolean encodes nothing the next `if` cannot re-test directly
Deferred from PR #466 review. `dimension: simplify` · severity `suggestion`.

`skills/fleet/scripts/reap.sh:101` carries a `wt_present` boolean that encodes nothing the very next `if` cannot re-test directly: the `elif` setting it false immediately falls into a block that skips everything, so the variable exists only to jump one line.

### Remedy the reviewer verified (on the pre-fix tree)

```sh
if [ ! -e "$wt" ] && ! gone "$wt"; then
  keep "$b" "cannot tell whether worktree $wt exists"
  continue
fi

if [ -e "$wt" ]; then
  ... (unchanged body)
```

The unanswerable-probe keep still runs before any `git -C "$wt"`, which is what the comment above it protects. The "Gated on `$wt_present`" sentence further down (the ignored-files block) names the variable and must be updated to name the `[ -e "$wt" ]` branch instead.

Reviewer's measurement: `node --test skills/fleet/scripts/reap.test.mjs` → 17/17 with the remedy applied, full suite 657/657, byte-identical to the unpatched baseline.

### Why deferred rather than applied in PR #466

That block changed underneath the finding. PR #466's fix-apply pass added a main-checkout guard inside the `if [ "$wt_present" = true ]` body (`reap.sh`, commit `8065a00`), so the reviewer's line-level remedy no longer applies verbatim and its 17/17 no longer covers the current shape. Restructuring control flow around a guard added in the same pass, for the gain of one boolean, is not worth folding into a PR whose other three fixes are correctness changes in the same file.

**Re-derive the remedy against current `main` before applying** — do not paste the block above. The reap suite is now 18 tests; the added one (`a [gone] branch checked out in the MAIN checkout is kept, with the reason that is true`) exercises exactly this control flow and must stay green.

feigi:
Closes #470

`reap.sh` carried a `wt_present` boolean that encoded nothing its only reader, one line later, could not ask again. `[ -e "$wt" ]` set it true; `gone "$wt"` set it false into a block that skipped everything; the third state already did its own keep-and-continue. The variable existed only to jump a line.

Folding the reader into the producer makes the three states one `if`/`elif`. The body keeps its exact indentation, so this is a control-flow edit and not a re-indent.

```sh
if [ -e "$wt" ]; then
  ... (body unchanged, from PR #466's main-checkout guard down to the ignored-files case)
elif ! gone "$wt"; then
  keep "$b" "cannot tell whether worktree $wt exists"
  continue
fi
```

Re-derived against current `main`, not pasted from the issue — the issue's own instruction. Its proposed shape (an early `[ ! -e ] && ! gone` keep, then a plain `if [ -e ]`) is semantically identical but tests `-e` twice and reads as a double negative; this spelling tests it once. Both were measured (see M6c below).

## What had to survive

**The unanswerable-probe keep still reaches no `git -C "$wt"`.** The `elif` is *mutually exclusive* with the body rather than merely textually earlier than it, so the guarantee is now structural instead of a consequence of the `continue` above. Every `git -C "$wt"` is syntactically inside `if [ -e "$wt" ]`.

**PR #466's main-checkout guard still gates what it gates**, unmoved, still the first statement of the worktree-present branch. Mutating it to `if false` reds exactly one test (M5).

**Exit status is unchanged.** Before, the body sat in `if [ "$wt_present" = true ]; then … fi`; now in `if [ -e "$wt" ]; then … elif …; fi`. Both take their status from the same trailing `esac`, and an `if` whose condition tests false with no `else` is status 0. The established-absent path still falls out at 0 and reaches `git worktree remove`, which takes a prunable entry at rc 0.

The `Gated on $wt_present` comment over the ignored-files case named a variable that no longer exists. It now names the `[ -e "$wt" ]` branch it sits in — semantically, not positionally (#129).

## Tests

Two added, both written before the refactor.

`reap.sh`'s worktree-present branch owns five keeps. Four had tests — main checkout, missing/dangling `.git` linkage, unreadable status, dirty tree. The **`--ignored` gate had none**, and it is the last thing between a merged, linked, tracked-clean *non-fleet* worktree and `git branch -D`. `git worktree remove` refuses on modified and untracked files but deletes ignored ones silently, so the file that gate protects is exactly the one that exists nowhere else. That is the keep-case test.

Its control — same fixture, no ignored file, opposite verdict — stops the first from passing on a `reap.sh` that keeps every non-fleet worktree unconditionally, and doubles as the exit-status pin for that branch's last command.

Full suite **1025/1025** post-rebase. Shellcheck output byte-identical to `origin/main`, both measured from the real script dir (a scratch copy fabricates SC1091 hits).

### Mutation results

Each mutant applied to a scratch copy and `diff`ed before its result was trusted. Two design-spec prose tests fail in scratch for an unrelated repo-relative-path reason; they are the constant baseline and excluded below.

| # | Mutation | Result |
|---|---|---|
| M1 | present branch → `if false` | 14 red, incl. both new tests |
| M2 | established-absent → `elif true` | 3 red — exactly the reap-an-absent-worktree tests |
| M3 | unanswerable keep → `elif false` | **1 red** — exactly `a worktree behind an unreadable parent is kept` |
| M4 | ignored gate → `if false` | **1 red** — exactly the new keep test |
| M5 | PR #466 main-checkout guard → `if false` | **1 red** — exactly the main-checkout test |
| M6a | `[ -e "$wt" ]` → `test -e "$wt"` | green (equivalence) |
| M6b | `[ -n "$ignored" ]` → `[ "$ignored" != "" ]` | green (equivalence) |
| M6c | re-spelled as the issue's proposed shape | green (equivalence) |

M3 is the safety-critical direction — a probe that could not look must keep — and it is pinned by exactly one test. M6c proves the suite pins behaviour rather than this spelling.

Behavioural equivalence measured directly as well: both new tests pass unchanged against `origin/main`'s `reap.sh`. Same 37 tests, same verdicts on both versions.

## Review fixes (PR #726 review, 6 dimensions / 13 agents)

Two findings applied. **Counts in the sections above are superseded by the counts below.**

**`0820f46` — the `.worktrees/` exemption arm was unpinned** (`tests`, `important`, survived 2-0). M4 above mutated the *ignored gate*; a review specialist mutated the *exemption arm* — `*/.worktrees/*` → a pattern that never matches, routing every fleet worktree through the ignored-file keep and stranding all of them — and `reap.test.mjs` stayed **37/37 green**. Both tests added here build their worktree at `../outside`, so both pin the `*)` arm and nothing pinned the exemption. Added a `.worktrees/`-homed worktree carrying a genuinely ignored file, asserting it IS reaped. Mutation-checked both directions: the never-matching pattern now reds exactly that test, and re-spelling the glob as `*/.worktree[s]/*` (same matched set, verified against three paths) stays green — so the pin is on the behaviour, not the literal pattern. The two existing fixtures, byte-identical to each other for 10 lines, now go through `mergedGoneBranchWithWorktree` via a `wt` parameter defaulting to the fleet home.

**`d782a81` — two comment paragraphs argued against code this PR deleted** (`simplify`, checked by a refuter before applying). They contrasted the new shape against a `wt_present` boolean that `grep -rn` no longer finds anywhere in the tree. Kept the one fact about the shipped source (`-e` is the `if`'s own condition, so existence is settled before any git command runs through `$wt` — measured: the first such call is the status probe, below the `-e`); dropped the `elif`-exclusivity sentence and the "inside the `[ -e ]` branch, and only there" paragraph. The fact the second carried is enforced by a test, not only by prose: moving the `case` block out of that branch reds the #178 no-surviving-ancestor test.

**Current counts: full suite `node --test skills/fleet/scripts/*.test.mjs` → 1026/1026; `reap.test.mjs` → 38/38.** Shellcheck `-x -S warning` rc=0, measured from the real script dir.

**On what the two original tests prove.** They pass unchanged against `origin/main`'s `reap.sh` (stated above), and so does the one added here — all three pin the `--ignored` block, which this refactor does not change. Read them as **additive coverage for the touched block, not as a regression gate for the restructure**: 38/38 is not evidence the control-flow edit was validated. The evidence for that is the M1-M6c mutant table.

https://claude.ai/code/session_01AqruLG5fxisLtnwh4pr4rq


feigi:
## Review adjudication — 12 findings, all dispositions

6 dimensions run, none unrun. **3 applied · 2 handed to the sweep ticket · 5 filed or commented · 1 into the PR body · 1 collapsed into a filing.**

### Applied (2 commits)

| finding | band | disposition |
|---|---|---|
| `tests` · `reap.test.mjs:568` — only the non-fleet arm of the `case` is pinned | **survived 2-0**, important | **`0820f46`** |
| `simplify` · `reap.test.mjs:580` — the two new fixtures duplicate 10 byte-identical lines the helper already covers | suggestion | **`0820f46`** (refuter: could-not-refute, 4 attacks, mutation matrix) |
| `simplify` · `reap.sh:170` — 9 comment lines argue against the deleted `wt_present` | suggestion | **`d782a81`** (refuter: could-not-refute, 5 attacks) |

Each in-scope `suggestion` got one refuter biased to refuse, instructed to verify by running rather than reasoning; both returned could-not-refute with the exact text they had measured, which is what was applied. Gate: **1026/1026**, shellcheck rc=0 from the real script dir.

### Not applied here — the data-loss sweep

`silent-failure` · `reap.sh:213` (**critical**, survived 2-0): the dirty guard reads an rc-0 **empty** `git status --porcelain` as "clean", so under `status.showUntrackedFiles = no` a worktree holding untracked work is reaped and the work destroyed at exit 0. Filed as **#730** by the review controller and deliberately **not** applied here: the PR does not touch that guard, and the same blind shape lives at three more sites (`release-ticket.sh`, `worktree-audit.sh`, `no-undo-audit.sh`) — fixing one of four inside a *simplify* PR ships a half-done sweep.

The `refuted` finding (the `--ignored` probe, `reap.sh:236`) is the **same defect at a fifth site**; its own refutation says so ("whole-script config-sensitivity, pre-existing, not this PR's"). Handed to #730 rather than filed as a near-duplicate.

### Deferred

| finding | dimension | home |
|---|---|---|
| The two unreadable-probe keeps have zero test coverage | correctness | **#732** *(new, ready-for-agent)* |
| `elif ! gone` sits 73 lines from its `if`; the third state has no arm; the guard-clause alternative | correctness · types · simplify | **#733** *(new, needs-triage — 3 findings, one question)* |
| Both status probes send git's diagnosis to `/dev/null` and name the STEP, not the FAULT | silent-failure | **#625** *(comment — already open on these exact two probes)* |
| `gone-walk.test.mjs:5` undercounts `release-ticket.sh`'s `gone()` call sites (4, not 2 — six across three scripts) | comments | **#725** *(comment — its remedy question turns on that inventory)* |
| The new tests pass on the pre-PR script, so they are additive coverage rather than the refactor's gate | correctness | **PR body**, "Review fixes" |

Two of the five deduped onto open issues rather than becoming new ones, per `ledger.mjs check` (exit 3 on three of four subjects).

### One correction carried into the filing

The coverage finding reported `grep -c "could not be read"` → **0**. That is right for `reap.test.mjs`, but repo-wide the string hits **7 times** — in `release-ticket.test.mjs`, `worktree-audit.test.mjs` and `inflight.test.mjs`, every one for a *different script's* message. A repo-wide sweep reads as "covered" and is wrong. #732 states the scope so the next reader does not re-verify it into a false negative.

Issue #218, opened 2026-08-06
review-pr.js: the size tier's comments carve-out is a docs-FILE test, so comment-only src edits get zero comment coverage


feigi:
Deferred from PR #216 review. Dimension: `comments` (comment-analyzer) and `correctness` (code-reviewer) — both reached it independently.

`workflows/review-pr.js` keeps `comments` through the size trim on `stats.kinds?.docs !== 0`. `kinds.docs` counts files whose **path** classifies as docs (`skills/fleet/scripts/diff-stats.mjs:40-51`) — `classify()` returns `src` for any code extension *before* it checks `isDocs`. So a diff whose entire substance is prose living in source comments scores `docs: 0`.

Measured, driving real `computeStats` through the lifted `selectDimensions`:

```
comment-only edit to ONE .js file        profile=single-file docs=0 -> [correctness, silent-failure]
comment-only edit, 2 src files, loc<30   profile=small       docs=0 -> [correctness, silent-failure]
prose in src + 1 config, small           profile=small       docs=0 -> [correctness, silent-failure]
```

The first row is the change class of PR #216's own commits `9bd71b0` and `6e9cac4` — pure comment corrections to `review-pr.js`. A wrong claim in a code comment is exactly the failure comment-analyzer exists for, and it is the one shape that now gets no comment coverage at all.

PR #216 corrected the comment to say what the code checks (a docs-classified file, not "ANY prose"). Widening the check itself is this ticket.

Approach is open, which is why this is not a one-liner: the fix is either a `linesOfComment`-style signal out of `diff-stats.mjs`, or keeping `comments` on any small diff regardless of kind. The second is simpler and costs one extra specialist on small src PRs.

feigi:
## Live instance, measured in a real fleet run — 2026-08-19

PR #682 (issue #604) is a **100% comment correction**: one phrase in one `.js` file, `skills/fleet/scripts/board.mjs:199-200`, zero non-comment change (the implementer proved the rest byte-identical by normalising both revisions).

The review workflow returned:

```
"dimensionsRun": ["correctness", "silent-failure"],
"dimensionsUnrun": []
```

**The `comments` dimension did not run on a pull request whose entire substance is a comment.** That is row 1 of this issue's measured table — `comment-only edit to ONE .js file → profile=single-file docs=0 → [correctness, silent-failure]` — reproduced in production rather than by driving `computeStats` directly.

Two details worth adding to the ticket:

[...]

PR #736, opened 2026-08-21
fix(review-pr): honour the silent-failure floor on every single-file/small diff

### LINKED ISSUE #236: review-pr.js: the silent-failure floor run-team/SKILL.md:379 promises does not exist for any single-file non-src diff
Found by the controller during a `/fleet:run-team` run, reviewing PR #226. Both halves reproduced before filing.

## The divergence

`skills/fleet/skills/run-team/SKILL.md:378-380` states:

> any `single-file` or `small` profile trims to correctness+silent-failure, keeping comments only when a docs file is in the diff and tests only when a test file is

`workflows/review-pr.js` `selectDimensions` does not honour that floor. The `stats.hasSrc === false` branch filters out `types`, **`silent-failure`** and `simplify` **before** the size-tier branch runs, and `SIZE_TIER_DIMS = {correctness, silent-failure}` only re-admits dimensions still present in `dims`. So for any single-file diff that touches no `src` — config, `.yml`, shell — `silent-failure` is already gone and the documented floor does not exist.

## Measured

PR #226 changed exactly one file, `.github/workflows/ci.yml`:

```
$ node skills/fleet/scripts/diff-stats.mjs --pr 226
{"profile":"single-file","docsOnly":false,"hasSrc":false,"hasTests":false,"files":1,"truncated":null}
```

The review returned `dimensionsRun: ["correctness"]` and `agentCount: 2` (one snapshot + one specialist). Five of six dimensions never ran on a change to CI's own gating logic.

## Why it matters, concretely

A CI workflow diff is mostly shell, which is exactly what the silent-failure hunter is for. And on that PR the sharpest finding turned out to be a **comment defect** — surfaced by `correctness`, because the `comments` dimension had not run either. Thin coverage on the diffs least able to afford it.

## Triage's ruling: fix the CODE, not the doc

Honour the documented floor: `silent-failure` is retained for any `single-file`/`small` profile regardless of `hasSrc`. Correcting the doc instead was considered and rejected — it would ratify zero silent-failure coverage on CI-workflow and shell changes, which is where this run actually got bitten. Cost is one extra specialist on config/yml/shell PRs.

## Acceptance criteria

- [ ] A single-file diff touching no `src` (e.g. one `.yml`) runs `silent-failure` in addition to `correctness`
- [ ] The `small` profile behaves the same way
- [ ] Diffs that legitimately have nothing for a dimension to do are unaffected — a docs-only diff still drops `types`/`simplify`/`tests`
- [ ] `SKILL.md:378-380` and the code agree afterwards; whichever wording ships is true of the code
- [ ] A test pins the floor and **mutation-discriminates**: removing the floor must red it, and a benign reword must not. A positive regex over a whole section is a vacuous pin — slice small and pin the behaviour, not the prose
- [ ] Full suite green

## Out of scope

- The `comments` carve-out being a docs-FILE test — that is #218, an adjacent clause in the same function
- Changing `diff-stats.mjs` classification of any file type
- The `docsOnly` branch's own dimension set

feigi:
`selectDimensions` ran the `hasSrc === false` guard **before** the size tier, and the tier only re-admits dimensions still present in `dims`. So on any `single-file` or `small` diff touching no src — config, `.yml`, shell under `.github/` — `silent-failure` was already gone and the floor `run-team/SKILL.md` documents did not exist.

Measured: PR #226 changed one `.github/workflows/ci.yml` and reviewed CI's own gating logic with `dimensionsRun: ["correctness"]`.

Triage ruled the **code** wrong, not the doc: a CI-workflow or shell diff is mostly shell, which is what the silent-failure hunter is for. The `hasSrc` guard now keeps `silent-failure` on a size-tier profile and drops it everywhere else — a 50-loc config-only PR profiles `production`, never reaches the tier, and is unchanged.

## What the floor now covers, and what it deliberately does not

| row | before | after |
|---|---|---|
| `single-file`, config only (`.yml`, `.github/**.sh`) | correctness | **correctness, silent-failure** |
| `small`, config only | correctness | **correctness, silent-failure** |
| `small`, docs + config (not `docsOnly`) | correctness, comments | **correctness, silent-failure, comments** |
| `production`, config only | correctness, comments | unchanged — the guard still gates above the tier |
| `docs` (docsOnly) | correctness, comments | unchanged — returns before the tier |
| `tests-only` | correctness, tests, comments | unchanged — not a size-tier profile |
| every `hasSrc` row | — | unchanged |

`types` and `simplify` are still dropped with no src; they are not in `SIZE_TIER_DIMS` and the ticket does not put them there. A `.sh` **outside** `.github/` is already `classify()`'s `src` residue, so it always had `silent-failure` and was never part of this class.

## The contradicting comment

`review-pr.js` carried a comment defending the old behaviour ("returning early here would hand silent-failure a YAML file — exactly what the guard above drops"). Shipping the fix under it would leave a comment asserting the opposite of the code. It is **rewritten, not deleted**, and the distinction it drew is settled against the tree rather than assumed: after this change the `INTERSECT` form no longer changes any outcome versus a floor over the full set, because every dimension the guards above can remove is one the tier would not have kept anyway (`tests` only when `hasTests === false`, which is the carve-out's own negation; `types`/`simplify` are not in `SIZE_TIER_DIMS`; `comments` and `correctness` are never removed above). The filter form is kept because it stays correct without re-proving that equivalence each time a guard is added above it.

## Mutation discrimination (AC-5)

Measured on a committed tree, restored with `cp` from a scratch backup:

- **Remove the floor** (revert the predicate to `d.key !== "silent-failure"`) → **3 red**, exactly the floor rows.
- **Benign reword** (comment reflow + `d` → `dim` parameter rename) → **31/31 green**.

This also doubles as the positive control that the suite can see the mode changed: the mutation lands inside the region `select-dimensions.test.mjs` lifts from source, and it reddened.

Full suite: **1030 pass, 0 fail**.

## Prose sync (AC-4)

- `run-team/SKILL.md` already promised the floor; its *neighbour* clause needed the fix — "a docs+config diff runs correctness+comments" is now true only above the size tier, which is precisely the row the floor widens.
- `review-and-fix.md` said the tier "intersects what is left" after dropping silent-failure-hunter with no source — self-contradictory even before this change.
- The cost spec argued the composition was what kept `silent-failure` off YAML. Named as superseded, its two matrix rows corrected, a `small, config only` row added because the tests now pin it, and its acceptance criterion 4 struck through rather than quietly rewritten.

Out of scope, per the ticket: #218 (the `comments` carve-out being a docs-FILE test), `diff-stats.mjs` classification, and the `docsOnly` branch's own dimension set.

Closes #236


feigi:
## Review findings adjudicated — 5 applied, 6 deferred

Pushed `5c3353b`. `node --test skills/fleet/scripts/*.test.mjs` → **tests 1030 / pass 1030 / fail 0** (baseline 1030).

`dimensionsRun: ["correctness","silent-failure","tests","comments","types","simplify"]`, `dimensionsUnrun: []`. Every in-scope `suggestion` got one refuter biased to refuse before anything was touched.

### Applied

| finding | dimension | refuter | what shipped |
|---|---|---|---|
| `select-dimensions.test.mjs:152` (two findings, one item) | `tests` + `simplify` | survived | vacuous docs-only pin **replaced** with one that discriminates |
| `select-dimensions.test.mjs:143` | `simplify` | survived | dead `assert.ok`, entailed by the next line's `deepEqual`, deleted |
| `review-pr.js:456` | `correctness` | survived | "a 50-loc config-only PR profiles `production`" → **MULTI-FILE** |
| `survived[0]`, documentation half | `silent-failure` | n/a (survived band) | four docs now name that `tests-only` outranks the floor |

The replaced test was measured vacuous both ways: leaking the size-tier floor into the `docsOnly` filter, **and** deleting that branch outright, each passed all 1030 tests with the old test in place. Both now red the new test and only it; dropping the tier's `tests` carve-out — a change it should not catch — leaves it green and reds the correct row instead.

The `50-loc` clause was false against the block's own cited example: PR #226 is 47+5 across one file = 52 loc, `single-file`, which reaches the tier and gets the floor.

### Deferred

- **#739** (`ready-for-agent`) — `tests-only` shadows the floor: `!hasSrc && hasTests` is tested ahead of `files === 1` and `loc < 30`, so adding one test file to a CI PR **subtracts** the silent-failure hunter. Non-monotonic, and new with this PR. Out of #236's scope (ACs keyed on `single-file`/`small`), so only the wording shipped here. Carries three further findings on the same span — the gated-`if` simplification, the equivalence-comment cut, and the refuted `SIZE_TIER_DIMS` derivation — plus the 1296-case equivalence caveat to re-check when the floor widens. `stats.hasConfig` was verified to exist (`diff-stats.mjs:84`, in the blob at `:95`), so the suggested remedy would at least run.
- **#740** (`needs-triage`) — lockfile false positive, **REFUTED**. Mechanism reproduces, but the class has 0 occurrences in 400 commits and this repo has never held a lockfile or `package.json` on any ref, so the proposed path predicate would be dead code. Filed for the measurement, not for a fix.

PR #737, opened 2026-08-21
test(fleet): build the symlink-over-HEAD and directory-in-HEAD corruption fixtures

### LINKED ISSUE #456: release-ticket.test.mjs: the symlink-over-HEAD and directory-in-HEAD corruption shapes are named in the comments but never built as fixtures
Deferred from PR #450 review (issue #179). Dimension: **tests**. Severity: **suggestion** — 0 refuters by workflow policy.

## The finding

`skills/fleet/scripts/release-ticket.test.mjs`

Two of the four HEAD-corruption shapes PR #450's own comments document — **a dangling symlink over `HEAD`**, and **a directory in `HEAD`'s place** — are named but never built as fixtures, so no test exercises the porcelain output those two specific shapes produce.

## What is measured today, and what is not

- `release-ticket.sh`'s `unresolved_head` comment and the matching test comment both state that only the garbage-content shape is built (PR #450's review corrected that comment: it previously miscounted it as two).
- The two built shapes were probed directly and neither prints a `detached` line.
- `unresolved_head()` does not read `detached` at all — only `HEAD 0+` and `branch ` — so per the PR's own claim the mechanism is shape-agnostic and the missing two shapes should behave identically.
- **That claim is asserted only in prose, never proven by a fixture for those two shapes.** The comment further states that the symlink and directory shapes *additionally* print a `detached` line where the built one does not, so the two unfixtured shapes are the ones whose porcelain differs from what the suite has ever seen.

## Approach

Decided — this is test arrangement, not design:

> a symlink-over-HEAD fixture (`ln -sf /nonexistent .git/worktrees/<n>/HEAD`) is cheap to build cross-platform and would close the gap directly.

`symlinkSync` is already imported in `release-ticket.test.mjs` and already used for worktree-directory fixtures, so the fixture technique is in the file. A directory-in-HEAD's-place fixture (`mkdirSync`) is the same shape.

Whichever fixture lands must be **mutation-verified**: break what it pins, confirm it and only it reds, restore, then confirm it stays green under a mutation it should not catch. A shape-agnostic mechanism means these fixtures can very easily be vacuous — passing because the arm was already right for a reason the fixture never exercises.

## Why deferred rather than added in PR #450

Additive coverage on a feature the PR already pins with three tests plus the sibling-scoping test the review's `survived` findings called for. The finding itself rates it low priority given the documented reasoning and the mechanism's field-agnosticism; growing #450 further buys nothing #179 asked for.

## Related

- #184 — `release-ticket.test.mjs`: the three `0o000` permission fixtures leak on failure, and go vacuous under euid 0. Any new permission-based HEAD fixture inherits that problem; a symlink fixture does not.

feigi:
`unresolved_head` in `release-ticket.sh` names **four** ways an admin HEAD file breaks into the same porcelain shape — the null OID with no `branch` line. Only garbage content was built as a fixture, so the two shapes whose porcelain actually *differs* from anything the suite had seen went unexercised.

## Measured, not assumed (git 2.50.1)

The ticket's central claim was asserted in prose and never proven, so I probed all four shapes directly:

| shape | null OID | `branch` line | `detached` line |
|---|---|---|---|
| `chmod 000` on HEAD | yes | no | **no** |
| garbage content (built today) | yes | no | **no** |
| dangling symlink in its place | yes | no | **yes** |
| directory in its place | yes | no | **yes** |

The claim holds: the two unfixtured shapes are exactly the ones that additionally print `detached`.

## What this adds

- **The two missing shapes**, as one parameterised case. Because they are the only fixtures here whose porcelain carries a `detached` line, they are the only thing that can tell `unresolved_head`'s `branch` key from a `detached` one.
- **The acceptance half**: a symlink over HEAD that *resolves* leaves a real sha and the `branch` line intact, so `unresolved_head` must accept it. A fixture keyed on "HEAD is a symlink" rather than on "HEAD dangles" would refuse a healthy claim.

`chmod 000` stays named rather than built: a permission fixture leaks on failure and goes vacuous under euid 0 (#184). Neither shape added here touches a permission bit, so neither inherits that — measured: a failing run of either leaves **zero** residue in an isolated `TMPDIR`.

## Mutation verification

| mutation | result |
|---|---|
| add `cur&&/^detached$/{hasbranch=1}` to `unresolved_head` | reds the new shapes case **and nothing else**; the pre-existing 87 tests pass under it unchanged |
| make the `rev-parse --show-toplevel` linkage probe fail **open** | reds the new acceptance case **and nothing else**; the pre-existing 87 tests pass under it unchanged |
| De Morgan on `exit !(nullhead && !hasbranch)` → `exit (!nullhead \|\| hasbranch)` | stays green — behaviour pinned, not implementation text |
| reword the branch-mismatch blocker | reds 4 pre-existing tests, both new cases stay green — scoped, not whole-output pins |

The second row was unplanned: that fail-closed guard turned out to have had no coverage at all.

Stale prose in both files (which claimed only one shape was built) updated to name the built routes rather than count them, per this file's own #661 lesson.

Full suite: 1029/1029 green.

Closes #456


feigi:
## Review findings adjudicated — 1 applied, 5 deferred (`e0bf7ac`)

Full six dimensions ran. 0 survived at `critical`/`important`, 2 refuted, 7 unverified (the `suggestion` band gets 0 refuters by policy, so each in-scope one was checked here).

### Applied

- **Two "nothing was touched" assertions** restored to the new symlink/dir case (`release-ticket.test.mjs`), matching its direct sibling. Both mutation-verified independently: a tracker edit in the blocked-exit arm reds `r.calls()`, `rm -rf "$stray"` reds `artefacts()`, and a harmless extra stderr line in the same arm leaves both green. dims `correctness` + `silent-failure`, reached independently.
- **The `chmod 000` justification corrected.** It asserted "a permission fixture leaks on failure" — that was #184's *title*, and **#184 is CLOSED**: the `chmod -R u+rwX` in `repo()`'s `t.after` is the fix that closed it, so a 000-mode fixture does not leak. The true half (root ignores the mode → the route would carry the `EUID0` skip and measure nothing on a root runner, while the other three reproduce as any user) is kept and the leak reason is now named as not applying. dims `types` + `correctness`.

Suite green from the worktree before commit: `tests 1029 / pass 1029 / fail 0`.

### Deferred

| # | Finding | dim | Why deferred |
|---|---|---|---|
| #742 | `:687` pins the linkage guard's phrase, not its fatality | `tests` | Refuted 1-1, upheld on the votes: the refactor the finding *names* (dropping the linkage check as redundant) **reds** the test — the downstream die says `cannot read the status of`. Residual (fatality unpinned when the message is identical) filed honestly with the cheap remedy. |
| #743 | `/^detached$/m` couples the suite to one git build | `correctness` | Refuted. CI run `32460116417` executed this case on **git 2.55.0 / ubuntu-24.04** (log line 1218, `✔ … 152ms`), blob-matched to the head. The proposed loosening to `/^HEAD 0+$/m` was measured to let the *garbage* shape pass green — the exact vacuity the case exists to prevent. Residual portability risk filed. |
| #744 | The die names the wrong sub-fault | `silent-failure` | Out of #456's scope — pre-existing production message. Defect confirmed (the `.git` linkage resolves once HEAD is restored); remedy open, and rewording it reds `:687` unless that pin moves with it. |
| #745 | Two `simplify` suggestions | `simplify` | Both refuted by measurement. The symlink/dir loop **discriminates** (4 mutants red one iteration and not the other, in both directions). The enumeration fix removes 2 of 3 built-vs-named sites and **0 of 6** count sites, so it does not remove the coupling it names. `wontfix`, filed so neither is re-proposed. |
| #746 | `:1375` still cites #184's leak as live | `types` | Out of scope — pre-existing, in a test this diff does not touch. Same root cause as the applied correction: the file now contradicts itself on #184. Also records that the fourth HEAD route is buildable now that the leak reason is gone. |

Note the `rebase-check` red is behind-count staleness against `main`, not a diff failure; the `check` job's own run is green.

feigi:
## Second review adjudicated — 2 more findings fixed (`244a137`)

Both `survived` findings from the second pass are fixed. Suite green at **`tests 1030 / pass 1030 / fail 0`** (baseline 1029 + the one case added below).

### The inert test — fixed by renaming it *and* covering what it claimed

`a symlink over HEAD that RESOLVES is not swept into the unresolved-HEAD arm` **never executed `unresolved_head`**. The resolving symlink leaves the `branch` line intact, so `wt` is set at `release-ticket.sh:217-218` and the whole `[ -z "$wt" ] && [ -n "$stray" ]` block that owns the arm is skipped. Reproduced three ways: a log line as the function's first statement is never written under this case and is written twice under its sibling, and mutating the guard to `END{exit 0}` leaves the case green while reddening four others.

Took **both** halves of the suggested remedy rather than picking one, because converting the test would have unpinned a production die:

- **Renamed and re-prosed** it to `… never reaches the unresolved-HEAD arm — the linkage probe refuses first`, and cut the claim that it is the acceptance case. It is kept because `assert.match(stderr, /cannot read the git linkage of/)` at `:697` is the **only** pin in the suite on the `release-ticket.sh:757` die.
- **Added the acceptance case that does reach the arm** (`:703`): detach the worktree so the `branch` line drops and `stray` is set, then lay a *resolving* symlink over HEAD. The guard must accept it and let the branch-mismatch `else` answer.

Mutation-verified, three mutants:

| mutant | new acceptance case | renamed case |
|---|---|---|
| `END{exit 0}` (guard always fires) | **RED** | green |
| `unresolved_head` keyed on `[ -L "$h" ]` (file type, not fault) | **RED** | green |
| `cur&&/^detached$/{hasbranch=1}` | green | green |

The third is the negative control: whole file `tests 90 / pass 89 / fail 1`, the sole failure being the symlink/dir case. So the new test kills the two mutants it is aimed at and survives the one it should.

### The false `detached` claim — reworded to name the actual gate

The comment said *"every shape the suite has ever seen carries no `detached` line to trip it"*, and `release-ticket.sh` carried the matching overclaim *"the symlink and directory fixtures are the only ones whose porcelain carries the line"*. Both false — **ten genuine `--detach` fixtures** in the same file print it (`:425, :460, :488, :520, :550, :766, :812, :2281, :2303, :2323`), and `release-ticket.sh:434-437`, four lines above and untouched by this PR, says so outright.

The mutation result they gloss is correct and re-confirmed on this tree (only the two new fixtures flip). Both comments now give the real reason: those fixtures carry a **real sha**, so the `nullhead` half never fires and the `detached` trigger has nothing left to flip. The new acceptance case above is itself a live counterexample — a genuine detached fixture that prints the line and is correctly accepted.

### Housekeeping

- **#742** corrected by comment: its premise (that the test exercised `unresolved_head`) was wrong; its residual — the linkage die's *fatality* unpinned when the message is identical — is unchanged, still real, and now sharper.
- **#744** and **#746** updated with post-rename line references.
- The second pass's `unverified[0]` is a third independent discovery of the `chmod 000` leak claim already fixed in `e0bf7ac`; its `unverified[1]` is the same residual recorded in #746. No new tickets.

PR #750, opened 2026-08-21
fix(fleet): stop FORCE_COLOR from corrupting claim-ticket's logic and its test's output

### LINKED ISSUE #514: claim-ticket.test.mjs: nine assertions still fail under a colored reporter
Deferred from PR #510 review (ticket #127, final head `e4e857e`).

- **dimension:** `correctness`
- **file:** `skills/fleet/scripts/claim-ticket.test.mjs`
- **why deferred:** out of #127's scope. #127 owned the five bare `/pass N/` pins; PR #510 fixed those and fixed the color regression its own anchoring introduced. These nine were already color-sensitive at base, for reasons unrelated to pass-count matching.

## Defect

`skills/fleet/scripts/claim-ticket.test.mjs` cannot be run with a colored reporter. `apply()` builds the child env as `{...process.env}` minus only `NODE_TEST_CONTEXT`/`NODE_TEST_WORKER_ID`, so `FORCE_COLOR` reaches the `spawnSync`'d runner, which colors its own summary even into a pipe: `\x1b[34mℹ pass 3\x1b[39m`. Assertions matching the raw bytes then fail.

PR #510 fixed this for the five pass-count pins it touched, using `/(?:ℹ|#) pass N(?!\d)/` in place of `/^(?:ℹ|#) pass N$/m`. **Nine assertions remain color-sensitive, and applying the same `(?!\d)` treatment to every `^(?:ℹ|#) pass N$` site in the file does not clear them** — measured, converting all 22 literal sites still gives `tests 58 / pass 49 / fail 9` under `FORCE_COLOR=1`. They fail for a different reason, so they need a different fix.

## The nine (line numbers at `e4e857e`)

| test | line |
|---|---|
| `install: all four fields empty → true` | generated at :773 |
| `install: empty dependencies object → true` | generated at :773 |
| `install: empty workspaces array → true` | generated at :773 |
| `install: null dependencies → true` | generated at :773 |
| `origin/main beats the local checkout for the runner probe` | :837 |
| `runner: a node_modules in the worktree's own ancestry refuses nothing` | :250 |
| `runner: carries a template stamp on the npm entrypoint` | :723 |
| `runner: scripts.test wins` | :799 |
| `runner: the npm entrypoint is emitted without the directory shim` | :710 |

**Start at :43.** The four `install:` cases are all parsed through `install: out.match(/install: (.*)/)?.[1]`, and a greedy `(.*)` swallows the trailing reset sequence, so the captured value is `true\x1b[39m` rather than `true`. That is a strong candidate for that family — *not verified*, stated as the first thing to check, not as the answer. The other five were not diagnosed at all.

## Measurements

node v26.5.0, all runs piped, on `skills/fleet/scripts/claim-ticket.test.mjs`:

| tree | env | tests | pass | fail |
|---|---|---|---|---|
| base (`3e1260c^`) | `FORCE_COLOR=1` | 58 | 34 | 24 |
| head with the five anchored | `FORCE_COLOR=1` | 58 | 29 | 29 |
| head with the five bounded (`e4e857e`) | `FORCE_COLOR=1` / `=3` | 58 | 34 | 24 |
| head with **all 22** sites bounded | `FORCE_COLOR=1` | 58 | 49 | 9 |
| `e4e857e` | uncolored | 58 | 58 | 0 |

## Scope note

Fixing these improves the file; **it does not make it color-clean** unless every remaining assertion that reads raw reporter bytes is audited, and that audit is the actual work here. CI is unaffected either way — GHA runs a non-TTY bash step and a tree-wide grep for `FORCE_COLOR|NO_COLOR` returns zero hits — so this is developer-environment robustness, not a CI fix.

Untested and worth covering: a real pty (all measurements piped), color forced from outside the repo text (shell profile, org-level GHA env, `NODE_OPTIONS`), and any node old enough to emit the tap `#` prefix — whether tap colors its summary the same way was never exercised.

feigi:
## Summary

`skills/fleet/scripts/claim-ticket.test.mjs` cannot be run with a colored
reporter (#514). Measured on `origin/main` at claim time: `FORCE_COLOR=1
node --test claim-ticket.test.mjs` gave **65 tests / 41 pass / 24 fail**
(the ticket's own "nine" is stale — it was measured against an earlier head
that had 58 tests; the suite grew before this landed).

Enumerated the 24 by root cause rather than by line number. They reduce to
exactly two causes, both a bare `FORCE_COLOR` leaking into a child process
whose plain-text output is then compared literally:

- **16 failures** — every `run()`/`runFrom()` assertion using the bare
  `/^(?:ℹ|#) pass N$/m` anchor. `apply()`'s env copied `process.env`
  verbatim (already scrubbing `NODE_TEST_CONTEXT`/`NODE_TEST_WORKER_ID` for
  an unrelated reason) into the generated `agent-test` runner, so its own
  `node --test` SGR-wrapped its summary (`\x1b[34mℹ pass 3\x1b[39m`) and the
  `$`-anchor no longer matched.
- **8 failures** (4 `install:` cases, `runner: scripts.test wins`,
  `origin/main beats the local checkout for the runner probe`, and the two
  npm-entrypoint tests) — a **real bug in `claim-ticket.sh` itself**, not a
  test artifact. Its `ndeps` count goes through `node -e … console.log(N)`
  for a bare number; Node's console.log SGR-wraps that whenever
  `FORCE_COLOR` is set (`\x1b[33m0\x1b[39m`), so `[ "$ndeps" = 0 ]` never
  matches and a dependency-free `package.json` is wrongly refused with
  `declares 0 dependencies but has no lockfile`. Any developer with
  `FORCE_COLOR` exported hits this against the real script, not just the
  test.

## Approach

Chose **strip color at the source** over per-assertion tolerance. PR #510's
per-site `(?!\d)` treatment, mechanically extended to all remaining
`pass N` sites, was measured (in the ticket) to still leave 9 failing — and
by root-cause, per-assertion tolerance can only ever address the first
family above. The second family is `claim-ticket.sh`'s own internal string
comparison misfiring on colored bytes; no test-side regex change touches
that at all. Source-level stripping is the only approach that reaches
0 failures, and it fixes a real production bug in the process rather than
just hiding it from the test harness.

No test in the file asserts on the presence of ANSI/color bytes (checked
before removing the ability to see them), so nothing is disabled by this.

## What each fix touches, and what it can wrongly refuse

- `claim-ticket.sh`: wrapped the `ndeps` reduce in `String(...)` before
  `console.log`, so it's plain text regardless of environment — the
  strictly-correct fix for that one value. Touches only that one line;
  every other `echo`/message in the script was already plain text.
- `claim-ticket.test.mjs` `apply()`: added `delete env.FORCE_COLOR` beside
  the existing `NODE_TEST_CONTEXT`/`WORKER_ID` scrub. That `env` is shared
  by every `run()`/`runFrom()` call in the file — every test invoking the
  generated runner goes through it. Nothing in the file currently asserts
  on colored bytes, so this can't wrongly refuse an existing assertion; a
  hypothetical future test that deliberately wants to see color from the
  generated runner would need to opt back in explicitly.

Left one control behind: a new test forces `FORCE_COLOR` into its own spawn
(independent of `apply()`/`claim()`'s scrubbing) and asserts
`claim-ticket.sh` still resolves a dependency-free manifest — pins the
script's own robustness rather than an absence of `FORCE_COLOR` in whatever
ran the suite. Verified it fails against the pre-fix script and passes
against the fixed one.

## Test plan

- [x] `FORCE_COLOR=1 node --test skills/fleet/scripts/claim-ticket.test.mjs` → 66/66 pass, 0 fail (65 + 1 new control test)
- [x] `node --test skills/fleet/scripts/claim-ticket.test.mjs` (uncolored) → 66/66 pass, 0 fail
- [x] Same two runs via `./agent-test skills/fleet/scripts/claim-ticket.test.mjs` — same result
- [x] `./agent-test .` (full suite, uncolored, post-rebase) → 1036/1036 pass, 0 fail (1035 baseline + 1 new test, no drop)
- [x] Regression control verified to fail against the pre-fix `claim-ticket.sh` and pass against the fixed one

Closes #514

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

https://claude.ai/code/session_017NPmEus8u7bZ5PTpZGk113


feigi:
### Review findings applied — `3ce8ca5`

`dimensionsRun: ["correctness","silent-failure","tests"]`, `dimensionsUnrun: []`. 1 survived, 1 refuted, 1 unverified-then-checked.

**Applied (2)**

- `derive-testcmd.test.mjs:86` — `withoutNestedTestMarkers()` was the identical helper to the one this PR fixed in `claim-ticket.test.mjs`, left unfixed. It stripped `NODE_TEST_CONTEXT`/`NODE_TEST_WORKER_ID` and nothing else while feeding both spawning tests, so `FORCE_COLOR=1 node --test skills/fleet/scripts/*.test.mjs` was **1036 tests / 1034 pass / 2 fail** — those two the only failures in the whole suite, i.e. the PR did not yet survive the condition it exists to survive. Now **1036/1036** coloured. These two files are the only ones in the repo carrying that helper shape.
- `claim-ticket.sh:148` — the lockfile-mutation guard merged git's stderr into `$dirty` on the **success** path, so chatter from a git that exited 0 was reported as `install mutated the lockfile — wrong command`, *after* `--apply` had labelled the issue, created the branch and built the worktree. The refuter found a **routine** trigger needing no env var at all: a committed `.gitattributes` line with two globs (`*.js *.jsx text`) makes git warn on stderr at exit 0 for every claim, for every member. Confirmed here against a fixture with no env vars set — before, a false refusal with the lockfile provably clean; after, the run completes. `release-ticket.sh:766` already runs this exact capture without the merge and documents why; this was the one site that never got the lesson.

**Deferred (1) → #752**

- `claim-ticket.sh:71` — 1-1 split. The mechanism reproduces (node stderr merges into `$ndeps`; `NODE_DEBUG=module` makes a zero-dependency manifest refuse while naming a count that does not exist), but upheld as refuted on severity and scope: exotic trigger, fails loud at exit 2, and environment-wide rather than specific to this line. Filed rather than dropped because the defect is confirmed and **the remedy is open** — dropping the `2>&1` there would empty `die "… — $ndeps"` on the real parse-failure path, which `install: an unparseable manifest refuses, and says so` pins. #752 carries that warning prominently. That capture is dual-purpose (a reason on failure, a number on success), which is exactly what made `:148` safe to fix and `:71` not.

**Gate:** full suite from the worktree, both ways — `node --test skills/fleet/scripts/*.test.mjs` → **1036/1036**; `FORCE_COLOR=1 …` → **1036/1036**, exit 0 (was 1034/1036). `claim-ticket.test.mjs` 66/66, `derive-testcmd.test.mjs` 13/13 coloured. No `--no-verify`; the repo has no pre-commit hook installed.

PR #751, opened 2026-08-21
refactor(fleet): extract prose-pin's between()/phrase() to a shared module

### LINKED ISSUE #500: skills/fleet/scripts: between() is pasted identically into three test files — extract it to a shared prose-pin.mjs

feigi:
## Summary

`between()` was pasted independently into 7 test files (byte-identical in
all 7), and `phrase()` into 7 more (byte-identical in 6, and semantically
identical in the 7th's line-wrapped form) — deferred across the #498, #526,
#536, and #537 reviews as issue #500, with the copy count growing under each
successive reviewer because each one measured only *definitions*, not call
sites.

Extracted both to `skills/fleet/scripts/prose-pin.mjs`:
- `export function between(text, from, to, what)`, importing `node:assert/strict` itself, carrying over the both-ends-bounded rationale comment verbatim ("an unbounded end lets a later, unrelated occurrence of the same phrase satisfy the assertion with the real clause deleted")
- `export const phrase = (s) => ...`, the regex-escaping/whitespace-collapsing helper

Every inline definition is replaced with `import { between, phrase } from "./prose-pin.mjs";`.

## Call sites: enumerated, converted vs. left

**Converted (definitions, 14 files):** `agent-test-dir-prose`, `dedupe-guard-prose`, `inflight-exit2-prose`, `reaping-prose`, `refuter-scratch-prose`, `review-pr-reads`, `snapshot-runner-audience-prose` (all defined `between()`); `agent-test-dir-prose`, `inflight-exit2-prose`, `liveness-rationale-prose`, `quiet-payload-prose`, `snapshot-runner-audience-prose`, `staleness-qualifier-prose`, `ci-vacuous-green` (all defined `phrase()`; `ci-vacuous-green`'s was a multi-line form — verified semantically identical to the one-liner by normalizing whitespace before comparing, not assumed).

**Converted (hand-rolled call sites without a local definition, 8 sites across 3 files):** the ticket's own comments named `review-pr-unrun.test.mjs`'s `findingsSchema()`, `verifyStage()`, and the specialist-prompt inline slice (3 sites), and `review-pr-snapshot-path.test.mjs`'s `snapshotBlock()` (1 site), as belonging to this class. Independently enumerating every `indexOf`+`assert.notEqual`+`.slice(at, end)` occurrence in `skills/fleet/scripts/` (not just the ones the ticket named) turned up 4 more in `review-pr-testcmd.test.mjs` — one sharing `snapshotBlock`'s exact anchor pair (`"const snap = await agent("` / `"if (!snap"`), three sharing the specialist-prompt slice's exact anchor pair (`"READ ONLY FROM THE SNAPSHOT"` / `"Scratch files go in"`) — hand-rolled duplicates of an already-named site that no prior reviewer's diff happened to touch. All 8 converted.

**Left alone, deliberately:** a `section(source, startAnchor, endAnchor, label)` helper — functionally identical to `between()` (same both-ends-bounded contract, same offset math) but a *different name*, independently duplicated across `fleet-tick-prose.test.mjs`, `implementer-model-tier.test.mjs` (whose own comment already flags the duplication: "Nothing detects drift between the copies"), and `review-path-default.test.mjs`, plus single-purpose siblings (`fallbackSection()`, `specialists()`, `candidates.test.mjs`'s inline slice). None of #500's four comments (across three reviewers) name `section()`, so folding it into `prose-pin.mjs` would be new scope this ticket never asked for — flagging it here as a real, separate duplication class a future ticket could pick up, the same way #533 was carved out for the `new Function` source-lift. Also left: about a dozen other `indexOf`/`.slice()` helpers with genuinely different contracts (single-bound `slice(at)` with no end anchor, an `endsFile` unbounded-fallback branch, `at + X.length`-offset slices that exclude the anchor from the result) — converting these would change behavior, not just deduplicate it.

## What this change could wrongly REFUSE

The 7 `between()` bodies were byte-identical (`diff` empty against the first, both directions) — no behavior risk there. The 8 hand-rolled sites use the *same* `at`/`end`/`slice(at, end)` shape but with the end search starting at `at` rather than `between()`'s `at + from.length`; verified this cannot change any match (none of the `to` anchors are substrings of their paired `from` anchors, so the narrower start bound is equivalent, not just probably-equivalent) and confirmed empirically via the mutation tests below rather than by inspection alone.

## Mutation testing

- **Break, should red:** mutated `run-team/SKILL.md`'s guarded clause ("refuses" → "SUCCEEDS") that `agent-test-dir-prose.test.mjs` pins via the newly-imported `between()`/`phrase()` — reds with the expected assertion diff. Restored via `cp` from a pre-edit backup (never `git checkout --`).
- **Break, should red:** mutated `workflows/review-pr.js`'s `derive-testcmd.sh` line inside the block `review-pr-testcmd.test.mjs`'s newly-converted `between(CODE, "const snap = await agent(", "if (!snap", ...)` site pins — reds with the expected assertion diff. Restored via `cp`.
- **Should NOT catch, stays green:** added an unrelated sentence elsewhere inside the same bounded block (the "Verify it:" paragraph, which no assertion in that test targets) — suite stays green, confirming the extraction didn't turn the pin either too strict or too loose. Restored via `cp`.

Worktree is `git status` clean after all three restores.

## Naming vs. #533

Ticket #500 explicitly named the module `prose-pin.mjs`, so that's what this PR ships. Sibling ticket #533 tracks the same duplication class for the `new Function` source-lift pasted into four test files, and #500's own comment floats "one shared `skills/fleet/scripts` test-helper module ... rather than `prose-pin.mjs` plus a second module" as the possible right home for both. Not implementing #533 here. If #533 lands with that merged-module shape, `prose-pin.mjs` and every one of its 20 importers in this PR would need a rename — flagging the cost now so whoever picks up #533 has it, per #500's explicit naming instruction taking precedence today.

## Test plan

- [x] `./agent-test skills/fleet/scripts` — 1035 pass, 0 fail (same count as `origin/main` baseline; measured fresh, not assumed)
- [x] Mutation tests above (2 break-and-red, 1 harmless-and-green), each restored via `cp`
- [x] `git status` clean after every restore

Closes #500


feigi:
## Review findings applied — `321ce6f`

Review `wf_2a140e08-91c`, six dimensions, 19 agents, `dimensionsUnrun: []`. 6 survived (each 2-0), 0 refuted, 13 unverified.

Full suite from the worktree, `node --test skills/fleet/scripts/*.test.mjs`: **1035 → 1040 pass, 0 fail, exit 0.** The count rose by the five new tests and dropped nowhere.

### Applied

**`survived[0]` (`critical`, `tests`) — `prose-pin.mjs`'s guards had zero failure-path coverage.** Deleting both `assert.notEqual` guards left the suite byte-identical to baseline at 1035/1035. New `skills/fleet/scripts/prose-pin.test.mjs`, synthetic fixtures only. Mutation-verified against the **full** suite:

| mutation | result |
|---|---|
| delete the `from` guard | 1040 tests, **1 fail**, `prose-pin.test.mjs` only |
| delete the `to` guard | 1040 tests, **1 fail**, `prose-pin.test.mjs` only |
| rename the internals (benign) | 1040 pass, 0 fail |
| drop `phrase()`'s metacharacter escape | 10 fail across 7 files, `prose-pin.test.mjs` among them |

Each guard deletion reds exactly one test and nothing else. The discriminating case is the both-ends contract the module's own comment states and nothing verified: a text holding `to` only **before** `from` must throw — unbounded, it returns `""` and throws nothing.

**`survived[1]`/`[3]`/`[4]`/`[5]` — eight staled line citations, converted, not renumbered.** This branch shifted `review-pr-reads.test.mjs` (−7) and `review-pr-testcmd.test.mjs` (+1). Renumbering rots again on the next shift, and this repo already pins the convention (`review-pr-citation-prose.test.mjs` asserts review-pr.js cites review-and-fix.md by section, never by line). Every target re-derived from the post-fix tree; **zero sites needed renumbering**, including the CI workflow comment. Three were pointing at the wrong block even before the shift, and the spec said "refuter prompt" where the source comment says "verifier prompt".

Citations into `select-dimensions.test.mjs` are deliberately left numeric — this branch did not move that file.

**`survived[2]` (`important`, `silent-failure`) — the one nested slice the extraction missed.** Mutation (move `testCmdError` out of `properties`, re-indent the close): raw form **11 pass / 0 fail** against a schema that no longer declares the field; converted form **1 fail**, naming the anchor. The fix is what closed it.

### Deferred and filed

- **#753** `ready-for-agent` — `section()` is a name-invisible twin of `between()`, three copies surviving in `implementer-model-tier.test.mjs:41`, `review-path-default.test.mjs:43`, `fleet-tick-prose.test.mjs:23`. Same two files (same lines) as **#491**, cross-linked there; **#608**'s file-local-fixture refutation is recorded in the body as the argument against. Carries two riders: the reintroduction guard (`unverified[6]`, blocked until the twins go) and a stale count clause in `member-prompt-prose.test.mjs:34`.
- **#754** `needs-triage` — the repeated `specialistPrompt` slice and the one-call `verifyStage()` wrapper. Shape preferences, measured green either way, out of #500's scope.

### Refuted, not applied

Three in-scope `suggestion`-band findings each got one refuter, biased to refuse, verifying by running. All three came back **refuted=true**:

- `unverified[2]` — "the `what` label names the needle, not the haystack". A block name for `what` is the **pre-existing majority convention** (14 filename / 12 block-name at head), present at base in files this PR never touches. The suggested fix would strip the region label from three sites and leave the identical shape at nine others.
- `unverified[7]` — "two anchor-specific messages collapsed into one generic label". Measured base vs head on six mutations: post-diff messages are **more** specific, naming the literal anchor string (`no longer contains "if (!snap" after "const snap = await agent("`); base printed only a prose role and no anchor in any of the six. Single-`what` was 18/18 at base.
- `unverified[11]` — "PR #751 falsified this comment". The file is not in the PR and those lines are byte-identical base to head. One clause *is* false — "two other test files", where there are three — and was false before this PR. Recorded as a rider on #753.

### Noted, not undone

`unverified[0]` — the PR exceeds #500's stated scope: 15 files, an unticketed second helper, eight inline-site rewrites the ticket does not name. Already disclosed by the author and accepted; enumerating call sites rather than definitions is what found four sites nobody had counted. Not a defect, not filed.

`unverified[3]` and `[5]` are the same defect as `survived[1]` and were **folded into that conversion** — `[3]`'s two citers and `[5]`'s three `:23-33` citers are all in the eight converted above. `unverified[4]` is closed by the new test file.

PR #756, opened 2026-08-21
simplify(fleet): migrate reap's five inline PATH shims to withShim

### LINKED ISSUE #628: reap.test.mjs:218,257,301,363,639 — five inline PATH-shim env objects predate `withShim` and were never migrated to it
**Dimension:** simplify (from PR #620's review, `unverified[4]`, second half)
**Deferred from PR #620 review** (issue #391) — the five lines are **outside** that PR's diff.

## The finding

`skills/fleet/scripts/reap.test.mjs` carried six byte-identical spellings of the same env object:

```js
{ PATH: `${bin}:${ENV.PATH ?? process.env.PATH}` }
```

at lines **218, 257, 301, 363, 639** (inline, pre-existing) and 743 (the `withShim` helper #620 added). `grep -Fc` on the exact string returned **6**.

PR #620 fixed the helper's own half — the `?? process.env.PATH` fallback is unreachable (`ENV` spreads `process.env` and never sets `PATH`, so the left operand IS the right one; where `process.env.PATH` is itself undefined, both are). What it deliberately did **not** do is migrate the five pre-existing call sites, because they are pre-existing lines the ticket had no business in.

## Remedy

Replace each of the five `{ PATH: \`${bin}:${ENV.PATH ?? process.env.PATH}\` }` arguments with `withShim(bin)`.

## Verified before filing — it works and drops nothing

- All five pass `PATH` and **nothing else**, and all five already name the variable `bin`, so `withShim(bin)` is a drop-in. (The obvious refutation — a site smuggling extra env keys that a bare `withShim(bin)` would drop — was checked and does not exist.)
- `withShim` is a `const` arrow, not a `function` declaration, so it does **not** hoist above lines 218-639. The migration is still safe, and for a reason worth stating in the fix: `node:test` defers every callback past module evaluation, and all five sites sit inside `test()` bodies (enclosing tests at 212, 237, 295, 358, 619). Proved rather than assumed — a module-eval reference gives `ReferenceError: Cannot access 'withShim' before initialization`; the in-test reference resolves.
- Applied in full in a scratch copy: `node --test skills/fleet/scripts/reap.test.mjs` → `ℹ tests 34 / ℹ pass 34 / ℹ fail 0` (at the reviewed SHA), and `node --check` clean, which is CI's `.mjs` gate.

Line numbers will have shifted since — PR #620 added an arm and reverted another. Re-derive with `grep -n 'PATH: \`'` before editing.

## Not in scope for this ticket

The unreachable-fallback half is already fixed on #620. Only the five-site migration remains.

feigi:
Closes #628

Six byte-identical spellings of this env object in `skills/fleet/scripts/reap.test.mjs` become one:

```js
{ PATH: `${bin}:${ENV.PATH ?? process.env.PATH}` }
```
 PR #620 fixed the helper's own half — the `?? process.env.PATH` fallback is unreachable, because `ENV` spreads `process.env` and never sets `PATH`, so the left operand IS the right one. It left the five pre-existing call sites alone as outside its diff. Those are what this closes.

Line numbers drifted since filing (#738 landed a `gitShim` in this file). Re-derived against `origin/main`: the five sites are at **:233, :265, :297, :359, :717**, and `withShim` at **:804** — not the 218/257/301/363/639 the body names.

## The two claims the ticket rests on, measured rather than assumed

**No site smuggles an extra env key.** All five passed `PATH` and nothing else, and all five already named the variable `bin`. `runReap` spreads `{ ...ENV, ...envOverrides }`, so a second key would have clobbered one of `ENV`'s git-scrubbing entries. Verified by reading all five, and pinned going forward by the new test's `Object.keys` assertion.

**`withShim` is a `const` arrow declared below all five sites**, so it does not hoist and a module-eval reference would be a TDZ error. That all five sit inside deferred `test()` bodies is *measured*: a top-level `const TDZ_PROBE = withShim("/x")` inserted at :186 fails the run outright with `ReferenceError: Cannot access 'withShim' before initialization`. The failure is loud, so a green run is itself the proof.

## What the suite can actually see

A green suite is evidence only about the paths it exercises, so I measured which sites it exercises. Under a mutant `withShim` that drops the shim dir, **10 of 39** tests go red: the new pin, the five sites that already used the helper, and **four of the five** migrated here.

The fifth — `:265`, "a \`+\` inside git's stderr is not a commit line" — survives. It also survives with **no shim passed at all** (measured by replacing its argument with `{}`: still `pass 1 / fail 0`). Its branch is genuinely merged, so real `git cherry` reaches the same verdict the injected warning was meant to test; the arm is vacuous with respect to its own fault injection. **Pre-existing and untouched by this PR** — the same object reaches `runReap` before and after — but now measured rather than assumed. Wants its own ticket.

## The added test

`withShim` had five call sites and no direct coverage; this PR takes it to ten. Its false-positive class is invisible to the suite that depends on it: a PATH it builds wrong does not turn any existing test red, because the shim dir is still first and `git` is still shadowed — what a broken tail costs is `reap.sh`'s access to its *other* tools, which the fixtures do not exercise.

So the pin asserts what the helper must **ACCEPT**: shim on the front, real PATH surviving on the back, `PATH` the only key. Its tail asserts against `process.env.PATH` rather than `ENV.PATH` deliberately — that identity is precisely the licence to spell the tail without the fallback, so the pin makes the licence executable instead of a comment.

Verified red three ways before green: `PATH: bin` (tail dropped), `${ENV.PATH}:${bin}` (order flipped), and a second `GIT_DIR` key. Each fails it; the restored helper passes.

## Deliberately not migrated

The identical spelling also lives at `verify-sha.test.mjs:135`, `release-ticket.test.mjs:155` and `prove-merge.test.mjs:341`. Each is a lone site in a file with no `withShim`, so folding them in means inventing a helper per file or a shared module — a different change from the one this ticket scopes. Declared, not done.

## Verification

- `./agent-test skills/fleet/scripts/reap.test.mjs` → `tests 39 / pass 39 / fail 0`
- `node --test skills/fleet/scripts/*.test.mjs` → `tests 1042 / pass 1042 / fail 0`
- `node --check skills/fleet/scripts/reap.test.mjs` → clean (CI's `.mjs` gate)


feigi:
## Review findings applied — `4c899ea`

**Applied (1):** Finding 1 (`survived`, `important`, correctness) — the comment block above `withShim` asserted a failure mode that does not exist. Rewritten at `4c899ea`; comment-only, no test or source behaviour changed.

**Deferred (2):**

| # | what | label |
| --- | --- | --- |
| #758 | Finding 2 (`suggestion`, correctness) — **REFUTED**: `reap.test.mjs:817` is not redundant | `wontfix` |
| #759 | the `` `+` inside git's stderr `` arm at `:252` is vacuous w.r.t. its own fault injection (pre-existing, `impl-628`'s declared-left item) | `ready-for-agent` |

### Measurements

All against `node --test skills/fleet/scripts/reap.test.mjs`, baseline `tests 39 / pass 39 / fail 0`. Re-derived independently on the post-fix tree (`4c899ea`) — line numbers below are that tree's.

| mutation of `withShim` (`:804`) | tests/pass/fail | red at |
| --- | --- | --- |
| tail dropped — `{ PATH: bin }` | 39 / 28 / 11 | all 10 shim tests + `:821` |
| order reversed — `` `${ENV.PATH}:${bin}` `` | 39 / 29 / 10 | 9 shim tests + `:821` |
| extra key — `+ GIT_CONFIG_GLOBAL: undefined` | 39 / 38 / 1 | `:822` only |
| empty entry — `` `${bin}::${ENV.PATH}` `` | 39 / 38 / 1 | `:821` only |
| duplicated entry — `` `${bin}:${bin}:${ENV.PATH}` `` | 39 / 38 / 1 | `:821` only |
| `ENV.PATH` → `process.env.PATH` | 39 / 39 / 0 | equivalent mutant — green control |

That table is why Finding 1 was applied and Finding 2 was not. The old comment claimed a wrong PATH "fails silently rather than loudly" and that a broken tail shows up as a `reap.sh` that cannot find its tools "not a red assertion here" — the first two rows disprove both. What it got right was the extra-key case (row 3).

The inverse claim — that the PATH equality is therefore redundant — is also false: rows 4 and 5 are caught by that assertion and by **nothing else in the file**, and with it deleted both mutants survive at 39 / 39 / 0. An empty PATH entry is the CWD, so those mutants silently put the fixture worktree on the PATH `reap.sh` runs with. The rewritten comment states that split instead, and both assertions stay.

**Full suite before commit:** `tests 1042 / pass 1042 / fail 0`, from the worktree, no `--no-verify`.

PR #720, opened 2026-08-20
fix(fleet): name the dirty-check comment by construct, not by line distance

feigi:
Closes #715

## What drifted

`reaping-prose.test.mjs` stated the gap between `release-ticket.sh`'s header block and its dirty-check comment as **"545 lines below"**, at two sites (`:33` and `:133`). Re-measured on `origin/main` at `618d0a7`:

```
$ git show origin/main:skills/fleet/scripts/release-ticket.sh | grep -n 'Ignored files are deliberately not a blocker'
760:  # Ignored files are deliberately not a blocker: claim-ticket.sh writes
$ git show origin/main:skills/fleet/scripts/release-ticket.sh | grep -n '^set -eu'
26:set -eu
```

True distance **735** (745 from the header block's first line). Neither is 545.

It drifted rather than being born wrong — at `610f6ef`, the commit that added the test file, the anchor sat at `release-ticket.sh:566`, ~545 below the header. `release-ticket.sh` then grew ~194 lines between the two anchors with nothing pinning the number.

## Remedy: #701's, not a fresh number

Per the ticket's ruling and **PR #713**'s precedent, the number is **dropped, not corrected**. #701 measured that a replacement rots identically: `quiet-payload-prose.test.mjs`'s distance was written 345, measured 353 at filing and 375 at fix — 22 lines of movement during a single fleet run. Both paragraphs are named by construct instead.

Diff is `4 4`, and the file is still 148 lines — **zero line-number drift**, so nothing below either site moved.

## What is deliberately left

- **`release-ticket.sh:15-25`** stays. It measures correct on `origin/main` (line 15 opens the corrected paragraph, 25 closes it, `set -eu` at 26), so #516's ruling leaves it, and the fact that it carries no pin is **#506**'s defect, not this one.
- **`rerun-rebase-check.test.mjs:417`** (`the runs query ~20 lines earlier`) — a fourth member of the same class, found by grepping the general form rather than the two quoted sites. Settled by the #656 review and out of this ticket's stated two-site scope. Worth noting that its two constructs now live in **different files** (the runs-query error at `.github/workflows/rebase-check-refresh.yml:222`, the jobs-query error at `.github/scripts/rerun-rebase-check.sh:119`), so the distance is no longer measurable as written. Not touched here; flagged for a separate ticket.
[...]

Issue #701, opened 2026-08-20
quiet-payload-prose.test.mjs:1 says the two SKILL.md statements are "345 lines apart" — measured 353


feigi:
Deferred from PR #695 review. dimension: `comments` · severity: `suggestion` · `skills/fleet/scripts/quiet-payload-prose.test.mjs:1`

## Finding

Line 1 of `quiet-payload-prose.test.mjs` opens the design comment with:

```
// #677. `run-team/SKILL.md` stated what `--quiet` drops twice, 345 lines apart,
```

The gap is 353, not 345.

Measured on `81b24c9` (PR #695 head):

```
$ grep -n "Own the CI waits\.\|Gate on the \`check\` job" skills/fleet/skills/run-team/SKILL.md
630:**Own the CI waits.** ...
983:Gate on the `check` job ...
# 983 - 630 = 353
```

The number drifted from unrelated `SKILL.md` edits since #677 — nothing in this file or in ci-state.mjs moved it, and nothing reddens when it does.

## Why deferred rather than folded into #695

Out of that PR's scope. Line 1 sits outside every hunk of the #695 diff (`git diff 0008b6b...81b24c9 -- skills/fleet/scripts/quiet-payload-prose.test.mjs | grep -c "^[+-].*345"` → `0`), and the PR neither added the claim nor falsified it. Folding it in would have widened the diff into prose the ticket has no business in.

## Remedy is open

Two shapes, both defensible, neither settled:

1. Update `345` → `353` — cheap, and stale again the next time `SKILL.md` grows a section between the two paragraphs.
2. Reword to drop the numeric distance entirely (e.g. "hundreds of lines apart", or name the two paragraphs by their anchors instead). Removes a number that drifts independently of the file that states it, at the cost of a slightly weaker illustration of why the two copies went out of sync unnoticed.

(2) is the durable one: this is the second-order version of the very defect the file exists to catch — a restated fact with nothing pinning it. Whoever takes this should also check whether any sibling `*-prose.test.mjs` header carries the same kind of hand-maintained distance.

PR #762, opened 2026-08-21
fix(fleet): #160's comment names the two files instead of a distance

### LINKED ISSUE #721: rerun-rebase-check.test.mjs:417 — the "~20 lines earlier" distance is unmeasurable: the two constructs now live in different files
Surfaced by `impl-715` while implementing #715, as a declared-left member of that ticket's class. Re-measured by the controller before filing. `dimension: correctness`.

## The defect

`skills/fleet/scripts/rerun-rebase-check.test.mjs:417` states a hand-maintained distance between two constructs:

```
417:// absence. The runs query ~20 lines earlier already says "This is an API
418:// failure, not a missing run"; the jobs query said nothing, so a run nobody
```

**The two constructs are no longer in the same file, so the distance is not stale — it is unmeasurable.**

- The runs-query error the comment quotes lives in `.github/workflows/rebase-check-refresh.yml`.
- The jobs-query error it contrasts against lives in `.github/scripts/rerun-rebase-check.sh`.

Measured on `origin/main`, with a positive control so the absence is an answer rather than a blind probe:

```
$ grep -rn 'not a missing run' .github/
.github/workflows/rebase-check-refresh.yml:222:              echo "::error::#$PR ($SHA): failed to query CI runs — see gh error above. This is an API failure, not a missing run."
```

One hit, in the workflow. Zero under `.github/scripts/`. The jobs-query error is `.github/scripts/rerun-rebase-check.sh:64`. There is no "~20 lines earlier" relationship to measure, in either file.

## Why this is a different defect from #715 and #717

Both of those were **drift**: a number that was true when written and rotted as a file grew. This one cannot be repaired by re-measuring, because the extraction that split the walk across a workflow and a script destroyed the relation the sentence asserts. A fresh number here would be false on arrival, not merely perishable.

## Suggested remedy — #701's, adapted

Name the two constructs, and name their files, rather than stating a separation. The remedy PR #713 established for `quiet-payload-prose.test.mjs`, and PR #720 applied to `reaping-prose.test.mjs`, is the same one: replace the distance with the construct. The adaptation here is that naming must now carry the file, since the reader cannot find the runs-query error by scrolling.

Do **not** substitute a fresh distance. #701 measured that a replacement number rots the same way — that file's distance was written 345, measured 353 at filing and 375 at fix, moving 22 lines during a single fleet run.

## Why it was not folded into PR #720

Out of #715's stated scope, which was two sites in one file, and widening to a sibling file is exactly the scope creep #701 was itself deferred out of PR #695 to avoid. `impl-715` found this by grepping the general form rather than only the two quoted sites, and declared it left rather than silently absorbing it.

## Related, not duplicates

- **#715 / PR #720** — the same class in `reaping-prose.test.mjs`, drift rather than cross-file.
- **#701 / PR #713** — the source of the remedy.
- **#516** — owns the `file:line` citation sweep. A *distance* is not a `file:line` citation, so this sits outside it, on the same carve-out #715 and #717 used.
- **#506** — a missing pin on a different file's header block; unrelated defect.

## Scope searched

`grep -rn` over `.github/` at `origin/main`, tracked files, plus a repo-wide grep for the general distance form (`[0-9]{2,4} lines (below|above|later|earlier)`) that produced this site among four.

`ledger.mjs check` → exit 0, `verdict: clean`, tracker searched and empty (highest near-miss 0.07, unrelated).

feigi:
`skills/fleet/scripts/rerun-rebase-check.test.mjs` stated a hand-maintained distance — "The runs query ~20 lines earlier" — between two constructs that no longer share a file. The number is not stale; there is nothing to re-measure.

Measured against `origin/main` at `9517310`:

```
$ git grep -n 'not a missing run' 9517310 -- .github/
.github/workflows/rebase-check-refresh.yml:222:  echo "::error::#$PR ($SHA): failed to query CI runs — see gh error above. This is an API failure, not a missing run."

$ git grep -n 'failed to list jobs for run' 9517310 -- .github/
.github/scripts/rerun-rebase-check.sh:64:    echo "::error::#$PR: failed to list jobs for run $RUN — see gh error above."
```

The split is real and it is what destroyed the relation: `a757d0a` created `.github/scripts/rerun-rebase-check.sh` and moved the jobs walk out of the workflow, leaving the runs query behind.

## The change

Names the two constructs and their files, dropping the separation — the remedy PR #713 established for `quiet-payload-prose.test.mjs` and PR #720 applied to `reaping-prose.test.mjs`. The adaptation here is that naming carries the file: a reader cannot reach the runs-query error by scrolling.

No fresh distance substituted, per #701's measurement that a replacement number rots the same way — one moved 22 lines during a single fleet run.

Four insertions, three deletions, one file. The comment's other clauses are untouched.

## Scope

The ticket's own grep for the general distance form is the class:

```
$ git grep -nE '[0-9]{2,4} lines (below|above|later|earlier)' 9517310 -- .
docs/metrics/tier-outcomes.tsv:45
docs/metrics/tier-outcomes.tsv:48
docs/metrics/tier-outcomes.tsv:49
docs/metrics/tier-outcomes.tsv:72
skills/fleet/scripts/rerun-rebase-check.test.mjs:417
```

Only the last is touched. The four `tier-outcomes.tsv` rows are past-tense records of what was measured on a given date — rewriting them would falsify the record rather than repair a citation, and `:49` records the review of PR #656 that examined this very comment. Left deliberately.

Also left, outside the numeric form the ticket searched: `board.test.mjs:35` ("two lines earlier"), `review-pr-snapshot-path.test.mjs:18` ("a few lines above"), `:163` ("the two lines below") — word-form distances, a different sweep.

`1042/1042` green (`node --test skills/fleet/scripts/*.test.mjs`), re-run after the rebase onto `9517310`.

Closes #721


feigi:
## Review deferrals — filed

Both `unverified` `suggestion`-band findings adjudicated. The review budgeted this band 0 refuters, so each was checked here before applying.

| Finding | Dimension | Ruling | Outcome |
|---|---|---|---|
| A — `rerun-rebase-check.test.mjs:419`, the retained "said nothing" clause | comments | **in scope, applied** | `b763fdd` |
| B — four more relative-distance idioms in other test files | tests | **out of scope, filed** | #765 |

**Finding A** was in scope because this PR's whole purpose was making that comment checkable by naming files — doing so exposed an adjacent clause *in the same sentence* that a reader can now trivially disprove, and shipping it would have left the comment more misleading than before the PR.

The clause was factually false. At `47ef697^` (#160's fix's parent) the jobs-listing failure path *did* echo `failed to list jobs for run N` on every candidate before continuing — `47ef697`'s own commit body says so, and running that script under a failing `gh` stub prints the line three times before exiting 4. What the pre-fix path lacked was **classification**, not output: no `JOBS_ERRS`, no counter, no exit status of its own, so the failure fell through the unset `FOUND_JOB` default into `no rebase-check job in any candidate run` and landed in the caller's `no-job` tally where neither guard reads.

The reword was drafted from the tree and then put to a dedicated refuter against all seven of its clauses (including the pre-existing ones and the uniqueness claim in sentence 1). All seven survived, each settled by a pinned-ref read or a run rather than by argument. Past tense preserved; no line number introduced, since a `file:line` rots exactly the way the distance this PR removed did.

**Finding B** is out of scope — different files, none in this diff. Filed as one sweep ticket rather than four because all four are *within-file and small*, merely perishable, unlike #721's cross-file one which was **unmeasurable**. #765 carries the detail that matters: neither enumeration was complete on its own — the review's numeral-form regex missed both word-form sites, and the other pass missed `no-undo-audit.test.mjs:432` — so a sweep needs both forms. A widened sweep confirms no non-test tail.

Suite at `b763fdd`: `tests 1042 / pass 1042 / fail 0`.

PR #768, opened 2026-08-21
test(fleet): build the chmod 000 HEAD-corruption route, the fourth of four

### LINKED ISSUE #763: release-ticket.test.mjs: build the fourth HEAD-corruption route (chmod 000), the last one named rather than built
Split out of #746, whose part 1 (the false `#184` leak clause) shipped separately. #746's part 2 was explicitly optional there; filing it so it survives that PR's `Closes #746`.

## What is missing

`unresolved_head` in `release-ticket.sh` enumerates **four** ways an admin HEAD file breaks into the same porcelain shape — the null object id with no `branch` line. Three are built as fixtures; `chmod 000` is named rather than built.

Re-derived against `origin/main` at `9517310` (the line numbers in #746's body and its follow-up comment have both drifted since):

| site | lines | what it says |
|---|---|---|
| `skills/fleet/scripts/release-ticket.sh` | `417-428` | the four-way enumeration, and why `chmod 000` is named rather than built |
| `skills/fleet/scripts/release-ticket.test.mjs` | `588-593` | garbage-content fixture; points at the enumeration and names the built/named split |
| `skills/fleet/scripts/release-ticket.test.mjs` | `638-642` | symlink and directory fixtures; contrasts them with the `chmod 000` route |

All three state the split, so building the fixture is a three-site lockstep — the same lockstep the sibling `wontfix` ticket #745 measured.

## Why it is buildable now

The reason recorded at `release-ticket.sh:423-428` is **root vacuity only**, and it explicitly refutes the leak reason: `repo()`'s teardown runs `spawnSync("chmod", ["-R", "u+rwX", root])` before `rmSync`, which is what closed #184. Measured directly against that teardown:

```
head-000:   naive rmSync=clean     leaked-without-chmod=false | exists after repo() teardown: false
subdir-555: naive rmSync=ENOTEMPTY leaked-without-chmod=true  | exists after repo() teardown: false
```

A mode-000 HEAD **file** does not even need the teardown — a file's own bits do not gate its unlink, only the parent directory's do. So the `chmod 000` route is the cheap one of the four, not the expensive one.

Root vacuity is real and is the whole remaining cost, and the file already absorbs it nine times over: `grep -c 'if (EUID0) return t.skip(NO_DENIAL);' skills/fleet/scripts/release-ticket.test.mjs` → `9`.

## Shape

Any new fixture must carry the guard, since mode-based denial is vacuous under root:

```js
if (EUID0) return t.skip(NO_DENIAL);
…
chmodSync(head, 0o000);
const { code, json } = release(r, c);
chmodSync(head, 0o644);   // restore before the first assert — artefacts() runs `worktree list` itself
```

`release-ticket.test.mjs:1923`-area already does `chmodSync(gitdir, 0o000)` on a file in the very same admin directory, so the mechanism is in-file precedent rather than new ground.

## Acceptance

- the `chmod 000` fixture exists and produces the null-OID-with-no-`branch`-line shape
- it carries `if (EUID0) return t.skip(NO_DENIAL);` and restores the mode before the first assert
- all three enumeration sites above stop saying `chmod 000` is named rather than built
- `./agent-test skills/fleet/scripts/release-ticket.test.mjs` green, `tests` count up by one

Skipping stays defensible — #746 said so, and the other three routes measure the same arm. This is a coverage addition, not a defect.

## Redundancy check

`ledger.mjs check "build the chmod 000 HEAD-corruption fixture in release-ticket.test.mjs"` → `found: false`, `verdict: tracker-hit`. Tracker hits reviewed: #746 (this ticket's parent), #744 (a `release-ticket.sh` die-message wording defect, unrelated), #453, #381, #471 (unrelated). Near-miss #745 is the `wontfix` record that measured this same three-site lockstep and refused a *simplify* suggestion over it — it does not cover building the fixture.

feigi:
Closes #763.

`unresolved_head` in `release-ticket.sh` enumerates **four** ways an admin HEAD
file breaks into the same porcelain shape — the null object id with no `branch`
line. Three were built as fixtures; `chmod 000` was named rather than built. This
builds it, then corrects the prose in both files that described the split —
including two claims this PR's own earlier passes got wrong.

Measured against `origin/main` `9afea29`. Suite **1042 -> 1043**, 0 skipped —
both counts run here, the baseline with `origin/main`'s copies of both files in
place.

## The four commits

| Commit | What it does |
|---|---|
| `5d37f33` | Builds the `chmod 000` fixture. The only code change in the PR: +49 lines, one new `test()`. |
| `3244204` | The three sites that stated the built/named split, now that all four are built. 4 hunks, both files. |
| `28ed7ef` | Deletes a teardown-cost **ranking** that was never measured, and four counts that rot as the file grows. 8 hunks, both files, plus one added assertion. |
| `ee29e9b` | Deletes a positional reference `28ed7ef` itself introduced, and one count it left behind. |

Apart from the new test, the whole diff is comment-only — verified by filtering
every non-`#`/`//` line out of `git diff origin/main`.

## The fixture

Carries the `if (EUID0) return t.skip(NO_DENIAL);` idiom that every other
mode-based fixture in this file carries — mode-based denial is vacuous under
root, where git reads the HEAD, the `branch` line comes back, the worktree stops
being a stray and the release proceeds. Both reads are taken while the bit is off
and asserted after the restore, since `artefacts()` runs `worktree list` itself.

It also asserts the half of the sibling's parenthetical that nothing in the file
pinned before: `chmod`, unlike the symlink and directory shapes, prints **no**
`detached` line — which is why `unresolved_head` can key on that line in neither
direction.

## Teardown cost, re-measured

Re-derived here and extended to all four HEAD routes — the comparison the earlier
prose asserted but never ran:

```
head-000    : naive rmSync=clean      leaked-without-chmod=false
head-garbage: naive rmSync=clean      leaked-without-chmod=false
head-symlink: naive rmSync=clean      leaked-without-chmod=false
head-dir    : naive rmSync=clean      leaked-without-chmod=false
subdir-555  : naive rmSync=ENOTEMPTY  leaked-without-chmod=true
```

A mode-000 HEAD **file** does not leak even under a naive `rmSync` — a file's own
permission bits do not gate its unlink, only its parent directory's do. So this
route never needs the chmod-back that `repo()`'s teardown runs.

The route it differs from is the mode-555 **subdirectory**, which does leak. It
does **not** differ from the other three HEAD routes: those are ordinary-mode
files and directories and tear down exactly as cheaply. An earlier pass of this
PR called `chmod 000` *"the cheapest of the four to tear down, not the dearest"* —
a ranking measured against `subdir-555` alone and never against the other three.
It is a four-way tie. `28ed7ef` removed that ranking from both files, and this
body no longer makes it either.

What `release-ticket.sh:417-424` states now is the permission-bit split, and
neither the ranking nor the teardown: `chmod 000` is the only one of the four
needing a permission bit, so it alone carries the `EUID0` skip and measures
nothing on a root runner; *"the other three reproduce as any user."*

## Mutation results, re-run against this head

| # | Mutation | Result |
|---|---|---|
| M0 | this fixture's own `chmodSync(head, 0o000)` **removed** | **RED, uniquely this test** — 1042/1043 pass, failing on `fixture: the mode alone must produce the null object id` |
| M1 | `unresolved_head` -> `END{exit 1}` (never fires) | **RED, 3 of 1043** — this test plus both sibling route fixtures |
| M3 | `unresolved_head` also keyed on `detached` (`cur&&/^detached$/{hasbranch=1}`) | **RED, 1 of 1043** — only the symlink/directory sibling; this test stays **green** |

M0 is the fault-injection check: the injection is load-bearing and its red is
uniquely this test's. M3 is the discrimination direction — this fixture keys on
the null-OID/no-`branch` pair, not on `detached`, so it is not a duplicate of the
symlink/directory case. M3 also confirms the claim both files already carry: that
mutation reds those fixtures *and nothing else*.

## A wart the mutation exposed

The mode restore runs before the first assert, so with the fault injection
stripped the release **succeeds** and takes the whole worktree with it — a bare
`chmodSync` then died `ENOENT`, burying the shape assertion that is the real
report. Guarded with `existsSync`, which turns M0's red back into that assertion.
Nothing leaks either way, per the table above.

## The correction that needed correcting

`28ed7ef` removed two positional references from the garbage-HEAD case and
introduced a third at the symlink/directory case — *"unlike the `chmod 000` route
built in the case below"*. `origin/main` carried no positional reference there at
all.

It was accurate the day it was written, which is exactly the trap: the references
it removed were also accurate until this PR inserted a case between them and
their target. Any future case landing between the symlink/directory test and the
chmod test falsifies it silently — same rot mode, same file. `ee29e9b` restates
it by the target's test name, the move `28ed7ef` had already made at the
garbage-HEAD case, and drops the count from *"which the three fixtures above do
not carry"*, restating it as the garbage, symlink and directory routes.

Swept afterwards for the same class. The survivors are deliberate: the
`below`/`above` references **inside** the chmod case point at lines in their own
body, which no inserted test can move, and `the permission fixtures above` is
pre-existing prose that this PR only re-wrapped.

## Deliberately left

Two sites carry similar *wording* about a different *claim*:

- `release-ticket.test.mjs`'s `#208` case (`:1475`) — its unbuilt second shape is
  a mode-555 **subdirectory** mid-delete, a different mechanism (worktree-removal
  refusal at rc 255, not HEAD corruption) and a genuinely leaking route per the
  table above. Its reason still stands.
- `docs/metrics/tier-outcomes.tsv`'s dated run record (the `2026-08-21` row for
  PR `737`), a past-tense historical statement that was true when written.

## Deferred

- #771 — the three unresolved-HEAD cases dereference `json.blockers` with nothing
  proving a payload exists, discarding the stderr that names the fault.
- #772 — the `existsSync` guard cannot go (measured), but the assert reordering
  under it is worth ~11 lines.


feigi:
Review findings applied in 28ed7ef. Suite green from the worktree: `node --test skills/fleet/scripts/*.test.mjs` — `tests 1043 / pass 1043 / fail 0`.

**Deferred, filed:**

- #771 — `ready-for-agent` — the three unresolved-HEAD cases dereference `json.blockers` with nothing proving a payload exists, discarding the stderr that names the fault. Dimension `silent-failure`. Defect confirmed and reproduced; remedy open (1 case vs 3 vs all 57, and the helper-level fix is measured unavailable — it reds 18/91 on a clean tree).
- #772 — `needs-triage` — the `existsSync` guard in the chmod-000 case cannot go (measured: an unguarded restore turns a legible `artefacts()` deepEqual into an ENOENT thrown from the test body), but the assert reordering under it is worth ~11 lines. Dimension `simplify`. Nothing broken, so triage decides whether it is worth re-opening a comment that is correct today.

**Applied:** the false teardown comparative (both readings), its near-verbatim duplicate in `release-ticket.sh` deleted rather than corrected twice, the two positional references replaced with test names, four counts restated by shape, and the garbage route's missing `detached`-absence assertion — which kills its own mutant (fixture rebuilt as the dangling-symlink shape → 1 fail, by its own message) and stays green under a behaviour-preserving SUT change (`/^branch /` → `/^branch refs\/heads\//` → 91 pass).

**No action, ruled:** the two repo-wide test-count sentences in `select-dimensions.test.mjs` and `review-pr-unrun.test.mjs` are past-tense mutation records, not live counts; the `tier-outcomes.tsv` rows are dated historical measurements.

PR #775, opened 2026-08-22
docs(fleet): brace a variable ref in a citation check — quoting is not the fix

### LINKED ISSUE #283: zsh silently rewrites an unquoted `git show $ref:<path>` for most path prefixes — every citation check in this repo is exposed

feigi:
Closes #283

## What the ticket asked, and the one thing that changed

The `## Agent Brief` prescribed **"always quote the entire `ref:path` argument"** — `git show "$REF:path/to/file"`. A later comment on the same issue measured that this does not work, and the controller ruled the comment binding. Quoting does not disable zsh's parameter-expansion modifiers. **The rule shipped here is bracing.**

## Measured on this machine, zsh 5.9, all three fully quoted

```
$ R=$(git rev-parse origin/main)
$ git show "$R:hooks/statusline.js"
fatal: ambiguous argument '.ooks/statusline.js': unknown revision or path not in the working tree.

$ git show "$R:skills/fleet/skills/run-team/SKILL.md" | head -1
commit d6931fabe431b81b1adae82d93e51000ff759d8b     # exit 0, 9 lines — the COMMIT, not the blob

$ git show "${R}:skills/fleet/skills/run-team/SKILL.md" | wc -l
    1593                                            # braced: the real blob
```

The second is the shape this ticket exists to prevent: exit 0, plausible content, so a survival `grep -c` scores a hit and a citation-verification sweep re-anchors a citation that was correct. It occurs **under the remedy the brief proposed**.

A literal ref is unaffected — `git show 8a84402:<path>` and `git show "8a84402:<path>"` hash identically, measured. Literal-ref examples were left alone.

## Changes

- `skills/fleet/commands/review-and-fix.md` — the rule, once, appended to the `git show <sha>:<path>` source-of-truth paragraph in `## Specialists`. That paragraph is what `workflows/review-pr.js` cites as owning this rationale, and it is the repo's only place instructing a citation check. Mechanism kept to one clause; no prefix table — enumerating which prefixes are safe is what produced three wrong accounts on the issue.
- `docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md` — its "quoted and unquoted behave identically" note is true of the literal-ref template it describes and false read generally. Scoped, not deleted.
- `skills/fleet/scripts/review-pr-citation-prose.test.mjs` — the pin, reusing that file's existing `## Specialists` slice helper.

## AC-2 was vacuous, verified

`git grep -nE 'git show\s+"?\$[A-Za-z_]'` over the whole tree returns nothing, before and after. There were no variable-form `git show` examples to convert; every instructional example is a literal ref or a `<sha>`/`<ref>` placeholder. AC-2's conversion set is empty, and the `doesNotMatch` assertion keeps it that way for the section that owns the rule.

## Mutation matrix — AC-4, both directions

Run against a scratch copy at the path depth the test computes `REPO` from. Control 6/6 green.

| mutation | result |
|---|---|
| rule paragraph deleted | RED |
| rule swapped for the brief's quote-only wording | RED (both assertions) |
| exception clause spliced inside the pinned span | RED |
| worked `${SHA}` form dropped from the sentence | RED |
| "quoting is not what fixes it" clause dropped | RED |
| benign reword elsewhere in the `## Specialists` slice | green |
| benign reword of the rule paragraph's tail | green |
| benign reword of the mechanism sentence | green |

The fix, the worked form and the anti-quoting clause are asserted as **one contiguous span** rather than three fragments — two fragments in one sentence still admit a spliced exception, and "just quote it" is exactly the plausible remedy that fails here.

The negative assertion also carries its accept case, fed directly rather than inherited from the file: the braced form must pass it, or the pin would push the next editor back to the unbraced form to get the suite green.

## Suite

`./agent-test .` → `tests 1047 / pass 1047 / fail 0`.


feigi:
Review findings deferred from this PR, filed:

- **#779** `needs-triage` — run-team phase 0's citation sweep has no pointer to the brace rule; decide whether #283's AC-1 allows one. (dimension: `correctness`, from `survived[2]`)
- **#780** `ready-for-agent` — the unbraced-variable-ref pin covers only `git show`/`git cat-file` in one code span; five measured blind spots. (dimension: `correctness`, from `unverified[0]`/`[1]`/`[3]`)

Everything else in the review was applied in 3bb7328. `ledger.mjs check` run before each: `tracker-hit` on #779's subject (#282, CLOSED, a different defect class — bare-numbered citations), `clean` on #780's.

Issue #641, opened 2026-08-18
ledger.mjs: a malformed row anywhere in gh's 6-row window degrades a real tracker hit to unverified (REFUTED as unreachable — close unless a triager disagrees)


feigi:
Deferred from PR #635 review (issue #154). Dimension: **silent-failure**. `skills/fleet/scripts/ledger.mjs:452` / `:471`.

Filed for the record because a finding is deferred, never dropped. **A refuter ran it and refuted it.** Recommendation: close, unless a triager disagrees with the reachability argument below.

## The finding

Raising `--limit` from 5 to 6 widened the array the shape validation guards, so a single malformed row anywhere in the fetched window — the probe row included — now downgrades a genuine tracker hit from exit 3 (blocks filing) to `unverified` exit 0 (filing proceeds), where the same tracker state exited 3 before the diff.

The controller flagged this as material because it tests the implementer's own declaration that the edit "can never turn a hit into a clean bill", and dispatched a refuter.

## Refuter verdict: refuted=true — mechanism real, state unreachable

**1. The flip is real with a synthetic gh.** Measured on a copy of the PR head, subject `"candidates.mjs row states the opposite of its code"`:

| fixture | exit | verdict |
|---|---|---|
| 6 valid rows | **3** | `tracker-hit` (hits=5, truncated=true) |
| 5 valid + 6th missing `number` | **0** | `unverified` |
| 5 valid + `null` 6th | **0** | `unverified` |
| 5 valid + 6th `number` as string | **0** | `unverified` |
| 5 valid + 6th `title: 12345` | **0** | `unverified` (`s.toLowerCase is not a function`) |

Pre-PR file extracted with `git show 6815db6:skills/fleet/scripts/ledger.mjs`: base + 5 valid rows → `exit=3 tracker-hit hits=5`, `--limit 5`. So for a synthetic gh the diff is a real behavioural change.

**2. No real gh can produce the trigger.** It requires a *heterogeneous* list — rows 1-5 well-formed and row 6 malformed — and gh serialises every row through the same typed exporter. Measured against real `gh 2.96.0`:

- **3779 rows, 42 queries, 6 repos** (`cli/cli`, `nodejs/node`, `microsoft/vscode`, `rust-lang/rust`, `golang/go`, `kubernetes/kubernetes`), `--limit 100 --json number,title,state,url` → `malformed=0`.
[...]

Issue #643, opened 2026-08-18
ledger.mjs: the gh row shape guard typechecks `number` only — a non-string `title` crashes scoring and is reported as a failed tracker read


feigi:
Deferred from PR #635 review (issue #154). **Two findings, one guard** — dimensions **silent-failure** and **types**, both on `skills/fleet/scripts/ledger.mjs:452`:

```js
if (!Array.isArray(parsed) || parsed.some((h) => !h || typeof h.number !== "number")) {
  throw new Error("gh returned JSON that is not an issue list");
}
```

Filed together: same line, same root cause, and two tickets would put two agents on it.

## Finding 1 — silent-failure — an internal TypeError is announced as a failed tracker read

A gh row with a non-string `title` passes the `h.number` check, blows up in `norm()` (`ledger.mjs:207`, `(s) => s.toLowerCase()...`), and is reported as a network-ish failure:

```
fixture: 5 valid rows + { number: 6, title: 42, state: "OPEN", url: "u6" }
exit=0 verdict=unverified error="s.toLowerCase is not a function"
stderr: ledger: WARNING — TRACKER NOT CHECKED (s.toLowerCase is not a function)
```

The label blames the tracker for a crash in this script's own scoring code.

**Suggested fix (hypothesis, never run):** extend the guard to the field actually consumed —
`... || (h.title != null && typeof h.title !== "string")` — so the failure reports as `gh returned JSON that is not an issue list`, which is true.

## Finding 2 — types — the same guard, stated generally

`title`/`state`/`url` are accepted whatever type they are, and #635's new ranking/truncation/`truncated` logic now sits downstream of that unchecked trust. Contrast `candidates.mjs:260-269`'s row-shape validator, which typechecks every field (`typeof r.n !== "number"`, `typeof r.t !== "string"`, …) and `die()`s on any mismatch. This finding's own suggested fix is **"None required for this PR"** — align only if the tracker-row validator is ever revisited.

## Why deferred, and the reachability evidence

- **Pre-existing.** The guard line has no `+`/`-` in #635's diff; #635 changed the list caps. Both findings say so themselves.
[...]

Issue #638, opened 2026-08-18
ledger.mjs check: the gh catch loses the cause — clipped to 500 chars with no marker, and emptied entirely by whitespace-only stderr


feigi:
Deferred from PR #635 review (issue #154). **Two findings, one expression** — both land on `skills/fleet/scripts/ledger.mjs:489`:

```js
tracker = { ok: false, query, error: String(e.stderr || e.message).trim().slice(0, 500) };
```

Filed together on purpose: two tickets would put two agents on the same line.

## Finding 1 — dimension: silent-failure — the 500-char clip has no truncation marker

**Claim:** the catch clips its diagnostic to the FIRST 500 characters with no marker, so when gh prints warnings before its real error the cause is discarded and the caller sees a string cut mid-word.

**Repro (measured, PR-head copy).** gh stub emitting 12 warning lines then `gh: FATAL — HTTP 403 rate limit exceeded, resets at 14:02 UTC` on stderr, exit 1:

```
exit 0, verdict unverified
tracker.error length: exactly 500
error tail: "---------\ngh: warning line 6 ---...---\ngh: warning line 7 -------------"
contains FATAL cause? false
```

The rate-limit cause reaches neither stdout nor stderr, and nothing in the string says it was cut. `execFileSync` above sets `stdio: ["ignore","pipe","pipe"]`, so the child's stderr is NOT forwarded — this string is the only copy of the cause.

**Suggested fix (hypothesis, never run):** keep the tail or mark the cut —
`const raw = String(e.stderr || e.message).trim(); const error = raw.length > 500 ? \`…\${raw.slice(-500)}\` : raw;`
The last bytes of a CLI's stderr are where the cause lives; the ellipsis is what stops a reader trusting a clipped string as complete. Line 433's `why` clip has the same shape.

## Finding 2 — dimension: silent-failure — whitespace-only stderr wins the `||` before `.trim()` runs

**Claim:** a whitespace-only gh stderr is truthy, so it wins the `||`, and `.trim()` then empties it. `tracker.error` becomes `""` and the warning prints `TRACKER NOT CHECKED ()` — exit 0 carrying literally no cause.

**Repro (measured, PR-head copy).** gh stub `printf '\n \n' >&2; exit 1`:

```
exit=0 verdict=unverified error=""
stderr: ledger: WARNING — TRACKER NOT CHECKED (). An issue that exists on the tracker but was never recorded in this run is invisible to the answer below.
```

[...]

PR #785, opened 2026-08-22
fix(ledger): validate every tracker field with no defined absent-value

### LINKED ISSUE #232: ledger.mjs check: shape guard validates only h.number, so a partial gh row prints "TRACKER HIT — #114 (undefined) — undefined" at exit 3

feigi:
Closes #232

## What was wrong

`check`'s tracker shape guard tested only that a row's issue number is a number, while the hit line goes on to interpolate the state and the url as well. A row carrying just a number passed the guard and printed a confident hit at the blocking exit code, out of a payload that cannot describe it:

```
STATUS: 3
STDERR: ledger: TRACKER HIT — #114 (undefined)  — undefined
```

That walks straight through an invariant this repo already declares in its own suite — `assert.doesNotMatch(r.stderr, /undefined/, "never print a hit the payload cannot describe")`. The existing wrong-shape test could not reach it: its fixture is bare integers, which `!h` and the issue-number check already rejected.

Fail-closed rather than fail-open — the exit code blocks a filing rather than permitting one — which is why this is a patch.

## The change

The guard now validates the fields with **no defined absent-value**: the issue number, the state and the url.

Deliberately **not** the title. The row builder substitutes `h.title || ""` for that field, in the row and in the score alike, so a row complete but for its title is one this code is written to handle and still describes truthfully. Validating it would trade a real, describable hit for a non-answer over the one field the hit line does not need — the opposite of what a guard against undescribable hits is for.

The failure keeps routing through the existing catch, so it degrades to `unverified` at exit 0 with the not-checked warning, adding no new arm. Which rows are validated is unchanged — every fetched row, the probe row included (#641).

## Tests

Two, one per direction, because a guard test that only ever sees input it must refuse pins half the behaviour:

- a row carrying only an issue number is a failed read
- a row missing only its title is still a tracker hit

**Mutation-checked against the whole suite, both directions:**

| tree | tests | fail | the red one |
|---|---|---|---|
| this branch | 1050 | 0 | — |
| guard reverted to its one-field form | 1050 | **1** | the partial-row test, alone |
| guard widened to include `title` (variant B) | 1050 | **1** | the missing-title test, alone |
| restored | 1050 | 0 | — |

Each mutant kills exactly one of the two new tests and leaves the other green, so neither is a proxy for the other: the partial-row test does not over-fire on the title question, and the missing-title test is what stops a later "tighten the guard" pass from silently becoming variant B.

Run against the full `skills/fleet/scripts/` suite rather than `ledger.test.mjs` alone — sibling test files spawn the script under test, so a mutation claim scoped to one file undercounts.


feigi:
## Review findings applied — 2 applied, 3 deferred and filed, 1 routed to an existing ticket

Head `aea97bd`. Full suite from the worktree, command verbatim: `node --test skills/fleet/scripts/*.test.mjs` → **`tests 1052 / pass 1052 / fail 0`** (1050 on the branch before these commits, + 2 new).

The review ran all 6 dimensions with no coverage gap and returned 8 findings. Two pairs collapsed: `correctness`+`tests` reached one finding independently, and `silent-failure`+`types`+`correctness` are one defect in three dimensions.

### Applied

| # | Band | Dimension | Site | Commit |
|---|---|---|---|---|
| 1+3 | SURVIVED 2-0 (each) | correctness, tests | `ledger.test.mjs:545` | `aea97bd` |
| 2+4+5 | SURVIVED 2-0 / 1-1 split / unchecked | silent-failure, types, correctness | `ledger.mjs:477` | `59d4fcb` |

**`aea97bd` — each half of the guard is now individually pinned.** The branch's negative test fed `{number: 114}`, missing `state` **and** `url`, so either surviving half of the predicate still rejected it: dropping one `typeof` check left the suite green at 50/50 while the CLI printed `TRACKER HIT — #114 (undefined) … — https://x/114` at the blocking exit 3 — the #232 defect itself, with nothing red. Two rows, one field each. Mutation-checked both ways: drop the `state` half → reds **only** `missing only its state`; drop the `url` half → reds **only** `missing only its url`; a behaviour-identical `every()` rewrite of the predicate → both stay green, so they pin behaviour rather than syntax.

**`59d4fcb` — the guard's `title` rationale carried a false clause.** It claimed "a row carrying every other field is one this code is written to handle and still describes truthfully". Measured false for a truthy non-string `title`: it carries every other field, passes the guard, and throws inside the same try (`h.title || ""` keeps `42` → `scoreTokens` → `norm` → `.toLowerCase()`), surfacing as `TRACKER NOT CHECKED (s.toLowerCase is not a function)` — a message naming neither gh, nor the row, nor the field. Reproduced 4/4 on `42`, `{"x":1}`, `["a"]`, `true`. The claim is narrowed to absence, which is what is true; the guard is unchanged.

The **typecheck itself was deliberately not implemented**: #232's Agent Brief rules `title` out of this guard on measured evidence, and the type question belongs to **#643**, which is open and `needs-triage`. Implementing it here would have silently resolved an untriaged ticket. Commented there instead with the independent rediscovery, both measured remedy shapes and the counter-argument from the 1-1 split: #643 (comment `5380332002`). No new issue — #643 already owns it.

### Deferred and filed

Each of the four `suggestion`-band findings arrived unchecked (the workflow budgets that band 0 refuters). Three were in scope and got **one refuter each, biased to refuse**. All three came back `refuted=true`, so none was applied.

| Issue | Dimension | Site | Refutation, in one line |
|---|---|---|---|
| #791 | silent-failure | `ledger.mjs:463` | The degrade is loud and fail-closed; the exit-3→exit-0 half is the diff's stated intent and the remedy does not change it — and the proposed `find()` **re-lands #232** (`[0]`/`[""]` are falsy, so `if (badRow)` never fires → `TRACKER HIT — #undefined` at exit 3), passing 1050/1050 green. |
| #792 | simplify | `ledger.mjs:461` | Rewrite is behaviour-identical (477-case differential, 0 divergences, `[]` included) but not a simplification: 156 chars would be the longest boolean in the tree (file max today 116, tree max 139), and it makes ledger the sole `every`-polarity outlier of three sibling guards `ci-state.mjs:65-74` keeps in deliberate parity. |
| #793 | simplify | `ledger.test.mjs:563` | The duplication is real and the trim is safe (1050/1050), but it is house convention — **69 duplicating comment blocks across 11 source/test pairs**, 16 of 70 in this file, and `ledger.test.mjs:282` shares more (17 grams) than this site. Consolidating "into `ledger.mjs`" would also have kept the half that was false. |

All three `needs-triage` under the repriced bar (#239): each refutation leaves it unclear that anything is broken, which is what that label means. Deduped through `ledger.mjs check` (exit 3 on all three — a capped `gh` window, so the printed hits were read and targeted searches run; nearest neighbours **#786** (same construct in `board.mjs`) and **#638** (the gh catch's clipped cause) are adjacent, not duplicates).

Ordering note: the false-claim fix landed **before** any dedupe of the surrounding prose was considered, so no finding asserted on text another was rewriting.

Issue #660, opened 2026-08-19
inflight.test.mjs: seven denial-dependent chmod fixtures with inline-only restores, and zero euid-0 guards


feigi:
Deferred from PR #658 review (issue #184). **Dimension: `tests`.**

#184 fixed both defects in `skills/fleet/scripts/release-ticket.test.mjs` only. `skills/fleet/scripts/inflight.test.mjs` (1934 lines) carries the same two, untouched — out of #184's stated scope, so PR #658 correctly left it alone.

## Measured against the tree at PR #658's head (`261c92f`)

**1. Zero euid-0 guards.**

```
$ grep -c 'getuid\|geteuid' skills/fleet/scripts/inflight.test.mjs
0
```

**2. Seven denial-dependent chmod fixtures, every restore inline-only.** Each rests on the permission bits being ENFORCED; root ignores them, so under euid 0 each goes green without testing the thing it names.

| fixture | chmod | restore |
| --- | --- | --- |
| `probe 3: an unreadable refs directory is unknown, never a free ticket` | `inflight.test.mjs:643` `refsdir` 0o000/0o400 | `:647` |
| the `refs/heads/fix` one-level-down case | `:674` `typedir` 0o000/0o400 | `:678` |
| `probe 3: an unreadable worktree registry is unknown, never a free ticket` | `:705` `wtroot` 0o000/0o400 | `:707` |
| the unreadable `gitdir` FILE case | `:728` 0o000 | `:733` (0o644) |
| the unreadable registry ENTRY case | `:794` 0o000 | `:797` |
| `accumulate: all three probes unanswerable is exit 2 WITH a payload naming all three` | `:1714` `refsdir` 0o000 | `:1718` |
| the unremovable-capture-dir case | `:1893` `cap` 0o555, set by the `mktemp` stub | `:1900` |

(Seven, not the six the review claimed — the 0o555 `cap` case belongs to the same class: `rm -f` fails on a non-writable directory only because the write bit is enforced.)

**3. The temp-root teardowns are not failure-safe.** Three of them, all the same shape:

- `inflight.test.mjs:119` (inside the `fixture()` helper), `:1797`, `:1923`
- `t.after(() => execFileSync(\"rm\", [\"-rf\", root]));`

A fixture that fails BEFORE its inline restore runs leaves an unsearchable directory under `root`, and `rm -rf` cannot remove it as a non-root user (measured on darwin):

```
[...]

PR #787, opened 2026-08-22
fix(fleet): read the runner's vendored guard spelling only where it is anchored

### LINKED ISSUE #230: agent-test: a suite whose absolute path has a node_modules ancestor is refused, while the relative spelling of the same suite runs

feigi:
Closes #230

The emitted runner's vendored-exclusion guard composed two terms: the caller's
literal spelling, and the resolved path stripped of the ancestor shared with the
runner. The resolved term is anchored; the spelling was not. An absolute
spelling necessarily carries that shared ancestor inside its own text, so a
worktree living under a `node_modules` refused the very directory that ran when
the caller named it relatively.

PR #422 fixed the relative half of this. The absolute half survived, and this is
it.

## The change

`${arg##/*}` empties an absolute spelling — the rule the sibling file branch
already applies. A relative spelling means something at the cwd; an absolute one
names the whole path from the root and says nothing about where the argument
diverges from the runner.

Stripping `$shared` off the spelling instead is **not** equivalent, and was the
first attempt here: that strip is lexical while `$shared` comes from `pwd -P`,
so a caller naming a path through a symlinked ancestor shares no literal prefix
with it and nothing is removed. A macOS `$TMPDIR` is such an ancestor, which is
how the suite caught it.

`$resolved` also falls back to the spelling before the divergence walk, so an
argument `cd` cannot enter still gives the walk something to measure and is
reported as vendored rather than as unreadable.

## Measured

Emitted runners only — the runner is generated text, so every measurement below
ran a runner freshly emitted by the `claim-ticket.sh` under test, never a copy
from an existing worktree.

Against `origin/main`, fixture `<tmp>/anc/node_modules/proj/.worktrees/42-slug`:
`t`, `./t`, `.` and the bare invocation ran; the absolute spelling of those same
directories refused. Positive control — the identical fixture under a
non-`node_modules` ancestor — ran every spelling, so absoluteness alone was
never the cause. After the change all six spellings agree at status 0.

## Acceptance criteria

- Same verdict spelled absolutely, relatively and bare — pinned in
  `runner: a node_modules in the worktree's own ancestry refuses nothing`, the
  only fixture built under a `node_modules` parent.
- A genuinely vendored directory still refused in every spelling — unchanged,
  `runner: a vendored directory argument refuses however it is spelled`.
- A directory `[ -d ]` admits but that carries no search bit still refused, and
  reported as vendored rather than as unreadable — new test, both spellings.
  Nothing exercised that input before, so both the spelling term and the
  fallback were unpinned by construction.
- #186's symlink cases unchanged and green.

## Mutation checks

- Spelling term unanchored again → only the ancestry test reddens.
- Fallback line deleted → only the no-search-bit test reddens.
- The two terms swapped, semantically identical → green, so the pins are on the
  property rather than on the expression's layout.

`./agent-test skills/fleet/scripts/` — `pass 1049`, `fail 0`.

## Deliberately left

- **The file branch.** Out of scope per the brief; verified it has no analogue
  of this defect — its `/*)` arm already exempts absolute spellings and its
  relative arm anchors at `$PWD`.
- **A cwd outside the worktree**, from which a relative argument descending
  through the shared ancestor names that `node_modules` and is refused. That is
  the lexical, cwd-anchored reading node itself applies, which is the precedent
  the brief names. Stated in the guard's comment.
- **Whether the file and directory branches should reach the same verdict on
  the same path** — the brief holds that as the open triage call.


feigi:
## Review findings applied — 12 findings adjudicated

Pushed `615ce6f`. Gate run from the worktree, `node --test skills/fleet/scripts/*.test.mjs`: **tests 1051 / pass 1051 / fail 0 / skipped 0**, exit 0 (1049 at the previous push, +2 new tests).

### Applied

**One defect, three findings** — the `correctness` finding at `claim-ticket.sh:319` and the `silent-failure` finding at the same line describe the same mechanism, and the `silent-failure/suggestion` at `claim-ticket.test.mjs:513` is its acceptance criterion. Re-derived independently of the review's fixture: one NON-vendored directory under a `node_modules` ancestor, three spellings, two verdicts — `t` and the `/private/var/...` realpath reported the permission fault, the `/var/folders/...` spelling of the same directory refused it as vendored. **#230's AC1 failed in exactly the case AC3 names**, so the branch did not close its own ticket. Fixed by anchoring the fallback on the argument's parent; all six spellings of a genuinely vendored directory still refuse.

**One defect, three more findings** — the `correctness` finding at `claim-ticket.test.mjs:500` and the `simplify` findings at `claim-ticket.test.mjs:521` and `claim-ticket.sh:325` all say the spelling term `/${arg##/*}/` is unpinned. It was: deleting it left all 1049 tests green. Pinned by the input the guard's own prose names — a relative argument from a cwd outside the worktree. Measured with the term deleted: 1051 tests, 1050 pass, the one failure being that new test.

**Three false comment claims** corrected — "the resolution is immune because both sides of that comparison are resolved" (false whenever the fallback fired), "the remaining spelling-dependent input is a cwd OUTSIDE the worktree" (it named one of two), and the no-search-bit test's claim to isolate the spelling term (it isolates the fallback).

**Two redundant rows dropped**, each measured redundant only *after* the pin existed: `[["./t"], 3]` (18 mutations tried, none distinguishes it from bare `t`) and the relative spelling in the no-search-bit test (identical failure set with and without it).

Both new tests were mutation-checked in both directions — each reds only under the mutation it pins and stays green under the other's.

### Deferred and filed

| Finding | Dimension | Issue | Why |
|---|---|---|---|
| `claim-ticket.sh:319` — use the POSIX `:-` idiom | simplify | #794 | **Superseded**, not refuted: it targets the very fallback the correctness fix replaced with a multi-line parent-anchored block. One narrower line remains open. |
| `claim-ticket.test.mjs:57` — no standalone entry point for the guard | types | #795 | Out of #230's scope; the finding's own verdict is "not urgent, coverage is solid". |
| `claim-ticket.test.mjs:305` — row loop short-circuits | tests | #796 | **Refuted** by 9 measured mutations: the short-circuit is real but masks nothing, and the proposed remedy changes zero verdicts. |

### Not filed

`claim-ticket.test.mjs:514` — the `process.getuid?.() === 0` root-skip guard is unpinned. Already-known class, deliberately left unfiled at PR #320's review, and #660 covers euid-0 guards in this suite. Commented there instead: https://github.com/feigi/claude-config/issues/660#issuecomment-5380548840

### Surfaced, not filed — pre-existing, outside this PR

Replacing the divergence walk's `$shared` with `$root` in `${resolved#"$shared"}` survives the full suite, so the anchor property the guard's comment calls load-bearing ("anchoring at the worktree root instead is the opposite error") is pinned by nothing. The walk is pre-existing — this PR touches the fallback line and the spelling term, not the anchor.

Issue #698, opened 2026-08-20
release-ticket.sh:178 — an unguarded `git worktree list` exits with git's own code, outside this script's 0/1/2 verdict vocabulary and with no `release-ticket:` diagnostic


feigi:
**Deferred from PR #693 review** (ticket #395). Dimension: `silent-failure`. Two refuters split 1-1 on the finding as originally framed; this issue carries the half that survived both votes.

## The defect

`skills/fleet/scripts/release-ticket.sh:178`

```sh
wt_list=$(git worktree list --porcelain)
```

No `|| die`. Under `set -eu`, a failure here exits the script with **git's own status**. Measured with `git` shimmed so `worktree list` exits 3: the script exits **3** — outside its documented 0/1/2 vocabulary — with empty stdout and no `release-ticket:` prefix on stderr. A caller switching on the exit code has no case for it.

The same shape recurs at the substitutions consuming `$wt_list` (`wt` :209, `main_branch` :211, `main_wt` :216, `stray` :228). There the exit is **1**, which this script's own header defines as a verdict: `#<n> NOT released — nothing was touched`, normally emitted with a JSON blocker payload. An exit 1 with no payload is a tool that never ran wearing the costume of a refusal.

## What was refuted, and why this is narrower than the original finding

A refuter re-ran the finder's probe with a *realistically* broken `awk` (fails unconditionally) instead of one shimmed to fail only on the `refs/heads/` program, and got **exit 2** with `release-ticket: could not count the worktrees git listed for #9`. PR #693's new `|| die` sits on the first awk of the same pipeline, 15 lines above :209, reading the same `$wt_list` — so any *standing* "awk cannot run" fault (missing, not executable, bad PATH) hits the guarded call first and never reaches the unguarded ones. What survives there is only a transient failing invocation N+1 but not N.

That does not reach line 178, which has no guarded canary in front of it: `git worktree list` is called once. The refuter's own counter-evidence records the git case as exit 3 and notes only that git announces itself on stderr — which is true, and is why this is `important`, not `critical`. Direction is safe: nothing is mutated in either case.

## Remedy (open)

```sh
wt_list=$(git worktree list --porcelain) || die "git could not list the worktrees for #$issue"
```

and likewise on the four consumers, converting a silent exit 1 into the labelled exit 2 the script already uses for "unanswerable". Whether the four consumers are worth guarding given the canary above is the open part — line 178 is not.

## Why deferred rather than applied in #693

[...]

Issue #454, opened 2026-08-13
release-ticket.sh:396 — unresolved_head in elif position cannot tell a dead helper from an honest false


feigi:
Deferred from PR #450 review (issue #179). Dimension: **silent-failure**. Severity: **suggestion** — the `review-pr` workflow budgets that band **0 refuters**, so nothing had verified it when it reached the fix-applier.

## The finding

`skills/fleet/scripts/release-ticket.sh:396`

`unresolved_head` is called in `elif` position, where a helper that **DIED** (awk missing, killed, rc != 1) is indistinguishable from one that honestly answered false — the run falls through to the branch-mismatch `else` and asserts "is not on $branch", the exact mislabel this arm was added to stop.

## Measured

```sh
set -eu
locked_like()     { printf 'x\n' | awk 'BEGIN{exit 1}'; }
unresolved_died() { printf 'x\n' | awk 'BEGIN{exit 2}'; }
if locked_like; then echo A; elif unresolved_died; then echo B; else echo "ELSE taken; helper rc was $?"; fi
```

Output: `ELSE taken; helper rc was 2` / `still running after a rc=2 helper: yes`. `set -e` does not fire in condition position and rc 2 collapses into the permissive answer.

**Not reproduced through the real script.** By the time `:396` runs, awk has already succeeded four times in the same process under `set -e`-guarded command substitutions (`:123`, `:126`, `:128`, `:140`), so an awk failure here is close to unreachable — the same "a success in the same run proves the tool was working" argument the file uses elsewhere. That is why the severity is low and why this is not a PR #450 blocker.

## Why deferred rather than applied

The finding's own `suggested_fix` offers two non-equivalent branches and the second one changes behaviour, which #179's brief puts out of scope (outcome, exit code and blocker behaviour explicitly unchanged):

> Either leave it and note the reachability argument in the helper's comment (cheapest, consistent with `locked` directly above it), or make the helper distinguish its own death: capture `unresolved_head "$stray"; rc=$?` and `die` on rc>1 rather than folding it into the `else`.

## Approach still open — hence `needs-triage`

- **Leave it + comment**: consistent with `locked()` immediately above, which has the identical shape and no such guard. Zero behaviour change.
- **`rc=$?` + `die` on rc>1**: converts a currently-unreachable fail-open into an explicit refusal, but introduces an exit-2 path and makes this helper differ from its sibling `locked()` for no measured trigger.
[...]

PR #788, opened 2026-08-22
fix(fleet): release-ticket.sh's worktree lookups refuse in its own voice (#243)

### LINKED ISSUE #243: release-ticket.sh: a newline in <slug> exits 2 through awk's own error, with no refusal line
Deferred from PR #227 review (issue #146). Found while verifying that PR's survived finding 1; out of #146's scope, which is JSON escaping.

**Dimension:** none — this was measured by the fix-applier while checking a review finding's stated reachability, not reported by a review dimension.

## What happens

`release-ticket.sh:48` builds `branch="$type/$issue-$slug"` from argv with no slug validation. At `:78` that branch reaches awk as a `-v` value:

```sh
awk -v b="refs/heads/$branch" '/^worktree /{w=substr($0,10);n++} /^branch /&&$2==b&&n>1{print w}'
```

awk rejects a newline inside a `-v` assignment. Measured:

```
$ sh release-ticket.sh 9 "$(printf 'a\nb')" fix
awk: newline in string refs/heads/fix/9-a
b... at source line 1
rc=2
```

Exit 2 is the right *class* — unanswerable — but the script's whole convention is a `die` carrying a `release-ticket: …` prefix, and here `set -e` propagates awk's status out of the `wt=$(…)` assignment with awk's raw diagnostic and no line of the script's own. A caller grepping for `release-ticket:` on stderr sees nothing, and the receipt contract (no payload on exit 2) is met only by accident.

The same argv reaches a second awk at `:92` (`-v d="/$issue-$slug"`), so a fix has two sites.

## Why this was worth recording

PR #227's review proposed pinning the multi-line half of the `jstr` fix through exactly this route — `sh release-ticket.sh 9 "$(printf 'a\nb')" fix`, asserting `branch === "fix/9-a\nb"`. That test can never pass: the script dies at `:78` before any receipt. The pin was relocated to `no-undo-audit.sh`, whose `wt` is argv `$1` — a filesystem path, and the only input to any of the three `jstr` copies that can hold a newline. Corollary, measured by mutation: deleting the slurp from `release-ticket.sh` or `inflight.sh`, or deleting `s/\n/\\n/g` from `release-ticket.sh`, leaves the whole suite green — those two copies of the multi-line machinery are **unreachable**, not merely unpinned.

## Approach is open, hence needs-triage

Two defensible fixes, and picking between them is a judgement call:

1. **Validate the slug** at the top, alongside the existing `issue must be a number` guard — refuses early with a clear message, but decides a policy (which bytes a slug may hold) that nothing has decided yet.
2. **Guard the two awk calls** — `wt=$(…) || die "cannot list worktrees for $branch"` — keeps the input policy open and fixes only the reporting, but leaves a branch name the rest of the script cannot really handle.

Option 1 also interacts with the unreachability note above: validating the slug would make `release-ticket.sh`'s slurp *provably* dead code, which is #119's business (`jstr` is duplicated three ways) rather than something to settle here.

feigi:
Closes #243

## What

Every awk lookup over `git worktree list --porcelain` in `release-ticket.sh` was a bare `$(...)` assignment. An awk that could not answer propagated its own status out through `set -e`, and the run ended on awk's diagnostic with **no line prefixed `release-ticket:`** — the prefix its callers grep stderr for. Exit 2 was the right class; the reporting was the defect.

Each lookup now carries `|| die`, the convention the worktree counter in the same script already uses: same prefix, same exit 2, still no payload.

## Measured, against `origin/main`

The ticket's own repro, before:

```
$ sh release-ticket.sh 9 "$(printf 'a\nb')" fix
awk: newline in string refs/heads/fix/9-a
b... at source line 1
rc=2
```

After:

```
awk: newline in string refs/heads/fix/9-a
b... at source line 1
release-ticket: could not read the worktree git listed for #9
rc=2
```

awk's own cause is kept; the script's line is added.

## Scope beyond the two sites the ticket names

The ticket names the `awk -v b="refs/heads/$branch"` branch lookup and the `awk -v d="/$issue-$slug"` stray lookup. The main checkout's branch and path are read off the same listing by the same construct and were bare too, so they are guarded here rather than left for a follow-up that would reopen this ticket.

## The undecodable-byte trigger is already neutralised on `main`

The issue's second trigger (binary garbage in a `gitdir` admin file, reaching awk as record data) was measured at `2958bb9`, before `d2624e1` added `export LC_ALL=C` to this script. Re-measured on today's tree, macOS awk 20200816: the same fixture exits **0** under `LC_ALL=C` and reproduces `awk: towc: multibyte conversion failure` only under a UTF-8 locale. That path is therefore covered by the guard rather than reachable through the ticket's fixture, and the test reaches it the way the fixture would — an awk that fails on the listing itself.

## Tests

Four added, each selecting its site **by construct** so they survive the next move (`refs/heads/` for the branch lookup, `length(d)` for the stray one) — never by line number.

Mutation-checked: dropping the branch lookup's guard reds exactly its two tests; dropping the stray lookup's guard reds exactly its one; a guard that refuses an *empty* answer — the false-positive class, since matching nothing is legitimate — reds the acceptance test along with much of the existing suite; and a behaviour-preserving rename of an awk-internal variable stays green.

`./agent-test skills/fleet/scripts` — 1052 pass, 0 fail. The default dry run still exits 0 with its payload unchanged.


feigi:
## Review applied — 18 findings adjudicated

Pushed `74ea82f` (two commits on top of `f6293c4`). Suite **1054 / 1054 pass / 0 fail** via `node --test skills/fleet/scripts/*.test.mjs` from the worktree (baseline was 1052; +2 are the tests added below). No `--no-verify`.

### Applied

| # | Finding | What changed |
|---|---|---|
| 1 | `tests/critical` x2 + `correctness/important` — the `main_branch=` and `main_wt=` guards ship with **no discriminating test** | Two shim cases added. Selectors verified unique first (`grep -cF`): `n==1&&` → 1 hit, `print substr($0,10); exit` → 1 hit |
| 2 | `types/suggestion` — the test block's comment implies systematic per-site verification it did not have | Comment now enumerates all four guarded sites, and names the two predicate lookups it does **not** cover |
| 3 | `silent-failure/important` — *"Every lookup over the listing is guarded"* is **false** (`locked`, `unresolved_head`) | Comment corrected to "every lookup ASSIGNED", with the gap named. Code fix filed, not applied — see below |
| 4 | `comments/suggestion` — the corrupted-`gitdir` aside presented a **dead trigger** as live | Reframed. `export LC_ALL=C` (`:45`) holds it shut — measured both ways: UTF-8 → awk exit 2, `C` → reads it as data |
| 5 | `comments/important` — `claim-ticket.sh:150` cited `release-ticket.sh:766`; **this PR's own +14 lines moved it** | Named by construct, not by a new line number, per the convention this PR's own tests state |
| 6 | `comments/suggestion` — the test header's *"`gh` is the one thing stubbed"* | Corrected; `git`/`sed`/`awk` shims predate this PR and it adds more |
| 7 | `simplify/suggestion` — the awk PATH-shim block written verbatim 3x | Factored to `awkShim(r, marker)` beside the existing `gitShim`. Applied **as** the two new cases were added, not after |
| 8 | `simplify/suggestion` — the newline case re-derives `release()` | Applied — **but not as proposed**: its `suggested_fix` silently dropped the `r.calls()` assertion, which a refuter measured as catching a real tracker call. Kept |

The `#395` counter shim keeps its own inline body deliberately — it models an awk exiting 1 with nothing to say, where `awkShim` exits 2 with a diagnostic. Folding it in would have changed what that test measures.

### Mutation-verified

Every new test kills its own mutant **and only its own**:

| mutation | result |
|---|---|
| strip `main_branch` `\|\| die` | 1 fail — `a main-checkout branch LOOKUP…` |
| strip `main_wt` `\|\| die` | 1 fail — `a main-checkout path LOOKUP…` |
| strip `wt` `\|\| die` | 2 fail — the newline case + `a worktree LOOKUP…` |
| strip `stray` `\|\| die` | 1 fail — `a stray LOOKUP…` |
| **negative control** — strip an *unrelated* guard (`listed=`) | 1 fail — only the `#395` counter case; both new tests stay **green** |

Committed before mutating; restored with `cp`, never `git checkout --`; tree verified byte-identical to the commit afterwards.

### The portability question, now measured

A finding asked whether the newline case pins anything on `ubuntu-latest` and could not answer it. Installed `mawk` and `gawk` and measured:

```
one-true-awk 20200816   rejects a newline inside -v
mawk 1.3.4              ACCEPTS it
gawk 5.4.1              ACCEPTS it
```

**The finding is correct — that case is vacuous on CI.** Demonstrated: with `mawk` first on PATH and the `wt=` guard stripped, it still passes.

Its assertion is **deliberately left as the bare prefix**. Tightening it to the guard's own message, as proposed, would convert a vacuous pass into a **red** on mawk/gawk, since the run walks past the lookup there and dies later, still prefixed. Both comments now say which awk they describe.

The pin that holds everywhere is the shim, which replaces awk's behaviour rather than depending on it — measured under mawk: stripping each of the four guards reds that guard's own case and no other.

### Deferred and filed

| Issue | Finding | Dimension |
|---|---|---|
| #798 | `:517` — `release_outcome` reads an awk that could not run as "not registered", writing a false *Deregistered* into the JSON receipt while git still lists it | silent-failure |
| #799 | `:250` — the stray lookup's `awk -v d=` decodes escapes; a backslash in argv `<slug>` reports a real stray as **absent**, and **this PR's new `\|\| die` cannot fire there** because awk succeeds and returns empty | types |
| #800 | `inflight.sh:436` cites `release-ticket.sh:73`; that is the `json_lib` guard, the check is at `:141`. Pre-existing | comments |
| #801 | `reaping.md:50` repeats the "only `gh` stubbed" claim. Not reached into from this PR | comments |

### Deduped, not filed

- **#698** — the bare `wt_list=$(git worktree list --porcelain)` producer, rediscovered by two dimensions. Already owned there; commented with the line drift 178 → 186 and the measurement. Deliberately left in place by this PR, correctly.
- **#454** — already owns the `locked`/`unresolved_head` status-collapse, option 3 included. Commented with evidence that **contradicts its "close to unreachable" premise** (an awk failing for one program, not the process, reproduces it end-to-end) and recommended `needs-triage` → `ready-for-agent`. Not relabelled here.

### Acceptance criteria

#243 offered two approaches and this PR takes option 2 — guard the awk calls, decide no `<slug>` policy. All four guards intact after these changes; nothing was removed or widened further. Still met.

One narrowing worth recording: the newline trigger #243 measured is **BSD awk's** behaviour. The reporting defect it describes is real wherever awk fails, but that specific trigger does not fire on mawk/gawk.

PR #802, opened 2026-08-22
fix(fleet): reap.sh refuses an argument it does not recognise

### LINKED ISSUE #250: reap.sh guards arity only, so a typo in --apply is silently treated as "not --apply" and reports a clean no-op

feigi:
`reap.sh`'s usage guard checked the argument COUNT and never the argument's
value, so a single argument passed whatever it said and the `--apply` test
beside it demoted anything that was not exactly `--apply` to "not `--apply`".

Measured on the pre-fix script over identical fixtures (a merged `[gone]`
branch and an unmerged one), `reap.sh --aply` and a bare `reap.sh` both exited
0 with **byte-identical stdout and stderr** — same payload, same
`DRY RUN — nothing will be deleted` banner, same per-branch lines. Nothing in
the run said the flag had not been understood.

## The fix

A value guard below the count guard, refusing through the same `die` at the
same exit code, above the fetch — so a rejected invocation reads nothing and
deletes nothing.

## Measured on the fixed script

| argument | rc | stderr |
|---|---|---|
| *(none)* | 0 | honest dry run |
| `--apply` | 0 | unchanged |
| `--aply`, `--nonsense`, `--apply-typo` | 2 | `reap: unrecognised argument '<arg>' — usage: reap.sh [--apply]` |
| `--APPLY`, `--apply=true`, `--`, `--help`, `-h`, `' --apply'`, `''` | 2 | same, naming the argument |
| `back\clue` | 2 | delivered verbatim — `die` is `printf '%s'`, not `echo` |
| `a b`, `--apply x` | 2 | `reap: usage: reap.sh [--apply]` — the count path, unchanged |

## Tests

Four cases in `reap.test.mjs`, mutation-checked:

- **the refusal** — exit 2, no payload, the argument named. Exit 2 and an empty
  stdout do not discriminate on their own (every later refusal in this script
  produces both), so it also asserts the run never reached the fetch it echoes
  and never printed the dry-run banner, and that the branch still stands.
- **the accepted forms** — bare and `--apply`, against a fixture with a branch
  to act on. A new guard's false-positive class is not its false-negative one.
- **the count path** — two arguments still take it, pinned by the *absence* of
  the value guard's own phrase.
- **the spec row** — `docs/specs/2026-07-23-fleet-plugin-design.md`'s
  script-surface row states this script's exit-2 contract, and #114 audited it
  while the guard was count-only. The row now names the new cause, taken from a
  real refusal rather than typed in, the way its two sibling clauses already are.

Mutations run: reverting the guard reds exactly the refusal and the spec-row
pins; neutering the count guard (`-gt 1` → `-gt 9`) reds exactly the count test
and leaves the other three green; rewording the dry-run banner — a change these
tests should NOT catch — leaves the whole suite green.

`./agent-test .` → 1059/1059.

## Note on the ticket's fifth acceptance criterion

It asks for a dry-run fixture on the grounds that "every existing case in
`reap.test.mjs` invokes `--apply`". That was true when the criterion was
written and is no longer true of `origin/main`: the no-argument path already
carries cases pinning the `applied:false` payload, the dry-run advisories
(`would reap`, `would remove worktree`) and dry/`--apply` agreement. Nothing
new was invented for it; the accepted-forms case above covers the bare
invocation as this change's own positive control.

Closes #250


feigi:
## Review adjudication — 15 findings, 6 dimensions, 0 unrun

Pushed `276b3eb`. `node --test skills/fleet/scripts/*.test.mjs` from the worktree: **tests 1059 / pass 1059 / fail 0 / skipped 0**, exit 0 — the same count this branch reported at push time. All four `(#250)` tests present and green in that run.

### Applied (1)

- **`reap.test.mjs:406-411` — the spec-row pin's rationale comment was garbled mid-sentence.** In scope: this diff wrote it. `276b3eb`. Comment only, no assertion changes.

  The `#114` attribution in the same comment was **re-derived after the rewrite** rather than carried over, and stands: #114 is *"fleet-plugin-design: the Script surface table's Non-zero column was written from intent — audit all eleven rows"*, the script-surface table has exactly eleven rows (`docs/specs/2026-07-23-fleet-plugin-design.md:178-188`), and the `reap.sh` row is one of them. #250's own body corroborates it independently — *"Found by the #114 exit-code audit while measuring `reap.sh`"*. The separate finding claiming the citation was misattributed is **refuted**, 2-0 in review plus this check.

### Deferred and filed (2)

- **#804** (`ready-for-agent`) — `claim-ticket.sh:27,32` still carries the exact defect this PR removes from `reap.sh`: arity lower-bound only, no value check on `$4`. A mistyped `--apply` is a byte-identical dry run at exit 0; `--apply extra --whatever` drops the extra args and **enters apply mode**. Falsifies the spec row at `:183`. Three findings (silent-failure, types, correctness), one defect. Not folded into this PR — #250 names `reap.sh` only. Remedy should adopt the siblings' existing wording, not a fourth phrasing.
- **#805** (`needs-triage`) — the design-spec-row read is copied three times in `reap.test.mjs`; a `specRow()` helper survived refutation and measures green at 1059/1059, but two of its three call sites are pins this ticket never wrote. Scope call, filed rather than applied.

### Refuted by measurement, not applied (5)

Each got one refuter, biased to refuse, required to verify by running:

| Finding | Verdict |
|---|---|
| `reap.sh:75` — third wording for one refusal class | **refuted.** Zero consumers grep for either phrase (9 occurrences repo-wide: 3 producers, 5 test assertions, 1 spec row). `die "usage: …"` is the house form in **9 of 9** arg-taking scripts; the sibling `unknown argument` pair exists only where `--apply` is a trailing positional beside a *separate* arity line. `reap.sh:73` and `:75` both end `usage: reap.sh [--apply]` — the sibling phrasing carries no arity and would break that. The remedy measured green (1059/1059) but leaves two dead `doesNotMatch` assertions behind. |
| `reap.sh:73` — arity refusal names no fault | **refuted.** A bare `usage:` line for an arity violation is uniform across all seven scripts measured; the two siblings with a value guard use precisely this two-tier split. The claim that the new test "pins that silence in place" is false — `doesNotMatch(/unrecognised argument/)` forbids the *value* label leaking onto the arity path and constrains nothing about what that path says. |
| `reap.test.mjs:360` — only one malformed shape pinned | **refuted.** The load-bearing claim ("nothing would catch a narrowing to a typo-set `case`") is measurably false: that exact mutant gives `fail 2`. The added cases kill **zero** of six mutants beyond ones no plausible edit produces; the empty-string case kills none. |
| `reap.test.mjs:378` — accept-side test adds no coverage | **refuted — the test stays.** With the diagnosis moved ahead of the decision (`printf` then `[ "$1" = "--apply" ] \|\| exit 2`), the refusal path stays byte-identical while `reap.sh --apply` reaps correctly *while telling the operator it did not understand the flag*. With the test: `fail 1`, on `an accepted invocation must not be refused` alone. Without it, that mutant survives the **whole** suite, 1058/1058. It is the only kill in 1059 tests for that shape. |
| `reap.test.mjs:413` — the repo fixture is unused | **refuted.** `repo(t)` is what keeps this a pin of the spec row: move the repository check above the argument guard (a refactor the spec permits) and the `repo(t)` version stays green and keeps comparing the real label, while `tmpdir()` reds on `fixture must reach the argument refusal: reap: not inside a git repository` — and that red is **machine-dependent**, flipping back to green when `TMPDIR` sits inside a repo. The saving is ~119 ms, ~0.07% of suite CPU and inside run-to-run noise. |

### Confirmed good, no work (4)

Four `tests/suggestion` findings are confirmations that the new tests discriminate correctly and are isolated from each other, each backed by its own RED/GREEN mutation pair. Recorded, nothing to do.

PR #803, opened 2026-08-22
fix(fleet): let ledger.mjs's payload subcommands drain stdout before exiting

### LINKED ISSUE #246: ledger.mjs read: console.log + process.exit(0) truncates the payload at 64 KiB on a pipe — the cockpit silently shows an empty ledger

feigi:
`read`, `row`, `filed` and `ruled` each ended their payload with `console.log`
and then `process.exit(0)`. On a pipe `process.stdout.write` is async and
`process.exit()` discards what is still queued, so the consumer received only
the prefix the kernel had already accepted — at exit 0, which types a corrupt
payload as a successful read.

They now fall out of the dispatch instead of exiting. The dispatch became an
`else if` chain so that not exiting does not carry a good subcommand on into
the unknown-subcommand `die()` at the end; that `die()` is still reached, and
is now pinned for the first time.

## The reported symptom, end to end

Against a self-built oversized ledger, `board.mjs build --ledger <it>`:

| | parse failures | tickets on the board | filed |
|---|---|---|---|
| before | 1 | 81 (all from `gh`, none from the ledger) | 0 |
| after | 0 | 460 | 400 |

`board.mjs` reads through `execFileSync` — a pipe — and accepts its ledger only
as a path, so nothing on the caller's side could have avoided this.

## Enumerate and declare

The class was enumerated from the ticket's prose first, then re-derived from the
mechanism with a scan whose positive control was `origin/main`'s own copy of this
file. The control's first window was too narrow — it found four sites and missed
`check`'s two, which were already known — so it was widened until it reproduced
all six the ticket lists, and only then run over the tree.

**Covered:** `read`, `row`, `filed`, `ruled`. All four measured truncating at
exit 0 before, all four delivering whole payloads after.

**Left knowingly:** `check`'s two exits. Measured to truncate on the same
mechanism, but its codes are the contract callers gate on, and its already-filed
exit sits mid-branch where falling through would run the tracker search that exit
exists to skip. Its payload is bounded by its own argv and gh's reply rather than
by the ledger, so reaching the cut takes an argv no caller sends. Named in the
file's own comment so the sweep is not read as complete.

**Left, ticket's own out-of-scope:** `board.mjs`'s `tryParse` fallback to an
empty ledger — the silent half, which this fix makes far harder to trigger but
does not remove; and whether `read` should be one JSON line at all.

**Checked rather than assumed:** #132 and #299 are both CLOSED, and
`candidates.mjs` is already on `process.exitCode`. They are out of scope because
they are fixed, not because the ticket said so. No other non-test fleet script
carries the shape.

## What this change could newly break, and what pins it

Replacing an exit changes when the process ends and what falls through next, so
each was measured rather than reasoned about:

- **File output** — byte-identical to `origin/main` on the same ledger.
- **Exit codes** — unchanged on every subcommand. `check`'s 0/1/3 are pinned by
  the existing suite, which the restructure leaves green.
- **Unknown subcommand** — still exit 2 with no payload. Nothing pinned this
  before; the `die()` is exactly what a dropped exit falls into.
- **Exit timing** — no lingering; the script holds no async handles.
- **A consumer that closes the pipe early** — same exit code as before.
- **A small ledger through a pipe** — still round-trips.

Each added test was mutation-checked: restoring `process.exit(0)` on `read` reds
that test and only it; removing the final `die()` reds the refusal test and only
it; and an unrelated wording change on a path all three write-subcommands
traverse stays green.

Suite: 1066/1066 on `skills/fleet/scripts`, rebased onto current `main`.

The run-team section that told the controller to redirect this command and never
pipe it is removed — it was a workaround for the defect fixed here, written in
the present tense.

Closes #246


feigi:
## Review applied — 8 applied, 4 deferred and filed

Pushed `45e7c36`. Suite `tests 1066 / pass 1066 / fail 0` locally, mutation-verified both directions.

**CI was red on Linux at the previous head and this fixes it.** Run `32583970394` on `d223459` reported `tests 1066 / pass 1060 / fail 3` — the three new `#246` tests failing `must still exit 0; got null`. The 208 KB tail went over as a single argv element, which exceeds Linux `MAX_ARG_STRLEN`; execve refused the spawn, and `spawnSync`'s `status: null` reads exactly like the truncation these tests exist to catch. darwin caps only the total, so it was green locally. The tail is now split across argv elements — `row`/`filed`/`ruled` join their trailing args, so nothing they echo changes — and the per-element ceiling is asserted in the test rather than assumed in a comment.

### Applied

| # | Change |
|---|---|
| 1 | `ledger.test.mjs` — argv split, the CI fix above |
| 2 | `ledger.mjs:568` — terminal `check` exit becomes `process.exitCode` + fall-through; behaviour-identical, and it was truncating at exit 0 |
| 3 | `ledger.mjs:159` — the comment justifying the `check` exemption was **factually false**. It claimed the payload "is bounded by its own argv and by gh's reply rather than by the ledger". `match` (`:236`) and `near` (`:305`) are both read out of `data.filed`, and `filed` caps nothing. Rewritten to state the ledger bound |
| 4 | `ledger.test.mjs:1071` — "`board.mjs` is that spawnSync consumer" was false; `board.mjs` imports `execFileSync` and calls `spawnSync` nowhere. Names the call it makes |
| 5 | `SKILL.md:381` — "the suite pins that **each** subcommand reaches a pipe whole" was false; `check` is one of the five. Names `read`/`row`/`filed`/`ruled`, and says plainly what is still open |
| 6 | `ledger.test.mjs:1094` — dropped "one measured 65536-byte prefix reads as 64456 characters". That figure came from a different measurement and does not reproduce against this file's fixture. The property is stated with no number to go stale |
| 7 | `ledger.test.mjs:1122` — `parsePayload` labelled every `JSON.parse` failure "the pipe truncated it", a cause it never checks. An empty or polluted stdout at exit 0 got a confident wrong diagnosis pointing at #246. Reports the observation now |
| 8 | `ledger.test.mjs:1142` — four assertions collapse to one whole-payload `deepEqual`. A truncated payload dies in `parsePayload` and never reached them, so the named tail assertion discriminated nothing |

### Deferred and filed

| Issue | Label | Dimension | Why deferred |
|---|---|---|---|
| #807 | `ready-for-agent` | correctness | `board.mjs:77` `tryRun` has no `maxBuffer`, so a payload past 1 MiB still serves a blind cockpit. Out of this ticket's scope — `board.mjs` has its own open tickets |
| #808 | `ready-for-agent` | types | `ledger.mjs:244`'s already-filed exit truncates mid-branch. Falling through would run the tracker search that exit exists to skip, so it needs the branch extracted first |
| #809 | `needs-triage` | comments | REFUTED 2-0 — `ci-state.mjs` carries the same shape, bounded today by the CI job count |
| #810 | `needs-triage` | simplify | REFUTED — hoisting the argv tail to module scope turns 3 named failures into 1 anonymous collection error and hides 56 tests |

### #246's stated symptom is **not** closed by this PR

The ticket's headline is a cockpit whose board is empty at HTTP 200. That still reproduces: `board.mjs:77` reads `ledger.mjs read` through `execFileSync` with no `maxBuffer`, so the default 1 MiB cap applies and `tryParse` falls back to `{rows:[],filed:[],ruled:[]}`. This PR moved the cliff from 64 KiB to 1 MiB rather than removing it — and the `SKILL.md` paragraph it deletes had measured a real **506 KB** ledger, so 1 MiB is roughly twice an already-observed run. Filed as #807; `SKILL.md` now says so in the text rather than claiming "(#246, fixed)".

### Mutation verification

- Reintroducing `process.exit(0)` after the four `console.log` sites → all four `#246` tests fail. The split payload still exercises the real truncation.
- Behaviour-preserving refactor of the same four sites → `1066 / 1066`, green. Discriminates rather than merely failing.
- Extra key in `read`'s payload → the collapsed `deepEqual` kills it; the four per-key assertions let it through. The collapse is stronger, not just shorter.

Not rebased: `behind_by=3` is currency staleness, and `rebase-check` red off a non-zero behind-count never clears by waiting.

Issue #235, opened 2026-08-10
ledger.mjs check: the clean/tracker-hit/unverified discriminant is still written twice, in the stderr chain and the verdict ternary


feigi:
Deferred from PR #225 review (ticket #152). **Dimension:** `code-simplifier` (attribution per controller relay).

## The finding

The same three-way `clean` / `tracker-hit` / `unverified` discriminant was hand-written **three times** in `check`:

1. the stderr `if (!tracker.ok) / else if (tracker.hits.length) / else` chain
2. the `verdict` ternary (`ledger.mjs:351`)
3. the exit-code ternary

## What was already applied on PR #225 (`e5ec1ac`)

Site 3 is gone — the exit code now reads `process.exit(verdict === "tracker-hit" ? 3 : 0)` and is a pure function of the named verdict. A refuter reproduced real drift between sites 2 and 3 before that change (stdout `"verdict":"clean"` while the process exited 3), so it landed rather than deferring.

## What is left, and why it is deferred

Sites 1 and 2 still express the same discriminant independently. Collapsing them means deriving the stderr chain from `verdict`, which requires hoisting the `verdict` computation above the stderr block.

**The maintainer ruled that `ledger.mjs:351` stays byte-identical on PR #225** (see #231, where the semantics of that ternary are themselves under open debate). Moving the line is not compatible with that ruling, so the remainder was not applied.

## Why `needs-triage` rather than `ready-for-agent`

This is sequenced behind #231. If #231 changes the verdict expression — the leading proposal adds a `ledgerRead` term — the right shape for sites 1 and 2 changes with it, and doing this first would mean rewriting the same block twice. Triage should decide whether this is worth doing at all once #231 settles, or whether two adjacent expressions of one discriminant is acceptable for a 90-line subcommand.

**Blocked by:** #231. **Related:** #152 (closed by #225), #232.

feigi:
> *This was generated by AI during triage.*

## Your blocker resolved — and it resolved in the direction that leaves your shape question answerable

#231 has been ruled and promoted. **The ruling does not change the verdict expression.** It adds a separate readability field to the payload and leaves the verdict, the exit mapping and the stderr warning untouched.

[...]

PR #812, opened 2026-08-22
fix(fleet): name the ledger's own readability in the check payload

### LINKED ISSUE #231: ledger.mjs check: verdict "clean" is emitted on a run where the ledger file was never read
Deferred from PR #225 review (ticket #152).

**Dimension:** `silent-failure-hunter` (attribution per controller relay — confirm if it matters for backlog stats).

**Two refuters reproduced opposite conclusions and neither was careless. The maintainer ruled PR #225 ships as is and this gets its own ticket, so the approach is genuinely open — hence `needs-triage`, not `ready-for-agent`.**

## The finding

`skills/fleet/scripts/ledger.mjs:351`:

```js
const verdict = !tracker.ok ? "unverified" : tracker.hits.length ? "tracker-hit" : "clean";
```

`verdict` accounts for the **tracker** half being unread, but not the **ledger** half. When `--file` points at a path that does not exist, the run still emits a positive machine-readable claim of cleanliness.

## The case that it is a real fail-open

The token `clean` is **new in PR #225** — before it, no field in the payload made a positive claim of cleanliness. Reproduced:

```
$ ledger.mjs --file /nonexistent/ledger.md check "..."
# stderr: warns
# stdout: {"tracker":{"ok":true,"hits":[]},"verdict":"clean"}
# exit 0
```

Nothing in the JSON signals the ledger was unread. `grep 'requireFile|not found' ledger.test.mjs` returns nothing — the missing-file arm's verdict is **unpinned by any test**.

The sharp part is the asymmetry: **this very PR insists an unreadable *tracker* must omit `hits` and report `unverified`, while an unreadable *ledger* still yields `clean`.**

## The case against

The missing-file arm, its stderr warning, and the `--require-file` hard-fail opt-in are **all pre-existing** — `grep 'require.file|requireFile|not found'` over the PR diff returns nothing. Pre-PR, the same run already emitted `found:false` at exit 0, so a machine-readable negative claim existed already; #225 only *names* in one field what a consumer previously reconstructed.

And a missing ledger genuinely has no filed rows — **the first `check` of any run is exactly that state**, so `unverified` there would fire on every run's first call.

## Suggested fix to record (the first position's)

Hoist the `existsSync(file)` result into a `ledgerRead` flag:

```js
const verdict = !ledgerRead || !tracker.ok ? "unverified" : tracker.hits.length ? "tracker-hit" : "clean";
```

plus a test asserting `unverified` when `--file` points at a nonexistent path.

**Triage must weigh that against the counter-argument above** — specifically whether "first check of a run" should really report `unverified`, or whether a distinct token (or a separate `ledgerRead` boolean in the payload) is the better shape.

## Why deferred

The maintainer ruled #225 ships to its own ticket's scope; line 351 stays byte-identical. Related to #152 (the tracker half of the same question, which #225 closes) and to the shape-guard ticket filed alongside this one.

feigi:
## What

`ledger.mjs check` emitted a payload that could not distinguish a ledger it read and found nothing filed in from one that was never read at all. Both answered `clean` at exit 0, with every field identical; only stderr told them apart, which a consumer parsing the payload cannot see.

The ledger half now reports its own readability the way the tracker half always has — a boolean of its own, `ledger.ok`, the peer of `tracker.ok`.

Reproduced against `cd41748` before the change (real script, healthy tracker, only the ledger half varying):

```
ledger MISSING:       {"subject":…,"found":false,…,"tracker":{"ok":true,…},"verdict":"clean"}  exit 0
ledger PRESENT+EMPTY: {"subject":…,"found":false,…,"tracker":{"ok":true,…},"verdict":"clean"}  exit 0
```

After:

```
ledger MISSING:       …,"ledger":{"ok":false},"tracker":{"ok":true,…},"verdict":"clean"        exit 0
ledger PRESENT+EMPTY: …,"ledger":{"ok":true},"tracker":{"ok":true,…},"verdict":"clean"         exit 0
```

## What deliberately did not change

The triage ruling rules out folding the flag into `verdict`: `.fleet/` is git-ignored and created lazily by `save()`, so a fresh clone's first `check` legitimately has no file to read, and `unverified` there would fire on the normal case. PR #225 also made the exit code a pure function of the verdict expression, so touching it is touching the exit contract.

- The verdict expression and the `process.exitCode` line are **untouched** — neither appears in this diff.
- The stderr warning is byte-identical, verified by diffing the stderr of a missing-ledger run before and after.
- `--require-file` against a missing ledger still exits 2.
- The field is emitted on the already-filed arm too, where it can only be true: scoped to one arm, its absence reads as falsy — "never read" — on the one answer that proves the ledger was read.

## Tests

`ledger.test.mjs` gains pins for the missing-ledger payload, the tracker-hit arm under an unread ledger (the arm whose exit code is non-zero), and the already-filed arm. The fresh-clone test gains one assertion for the state the ruling protects.

Mutation-checked, each mutant run against the whole file:

| mutant | result |
|---|---|
| field removed from the not-found payload | the three tests asserting that arm red, rest green |
| field removed from the already-filed payload | exactly that one test red |
| field hardcoded `ok: true` (guard untouched) | the three unread-ledger tests red |
| key moved to the end of the payload (negative control) | **green** — the pins are on the field, not on key order |

`./agent-test skills/fleet/scripts/` — 1073 tests, 1073 pass, 0 fail, post-rebase.

## Enumerated and deliberately left

- `read`'s payload has the same missing-vs-empty ambiguity (`load()` returns empty lists for an absent file). Left: its payload is the ledger's own data, and grafting a readability flag into it is a different shape question than this ruling settled.
- `row` / `filed` / `ruled` derive `total` from a ledger that may not have existed, but each writes and `save()` creates the file, so the total they report is true after the write.
- The verdict vocabulary, the exit-code mapping, and #235's stderr/verdict duplication are out of scope per the brief.

Closes #231

---

## Update (`4b09d9e`) — the flag is derived from the parse, not from a second `existsSync`

Review of the first two commits found the field making a false claim of its own. As first written, `ledger.ok` was `existsSync(file)`, so a `--file` landing on an existing file that is **not** a ledger — or on a ledger whose `## Filed` header lost one byte — emitted `ledger:{"ok":true}` with no filed rows, `clean`, exit 0, and nothing on stderr, on the exact subject the intact ledger answers ALREADY FILED at exit 1. The field added to make the machine-readable half honest was itself dishonest on that path.

The flag is now set inside `load()`, past the read and past its early return, so it can only be true of a file that function opened and found a `## Filed` header in. Measured across the fixture set, gh stubbed, one subject throughout:

| `--file` points at | `ledger.ok` | verdict | exit | stderr |
|---|---|---|---|---|
| a path with no file | `false` | clean | 0 | warns |
| an intact ledger holding the subject | `true` | already-filed | 1 | ALREADY FILED |
| an intact ledger holding nothing | `true` | clean | 0 | — |
| a ledger with `##Filed` (one byte) | `false` *(was `true`)* | clean | 0 | — |
| a file that is not a ledger | `false` *(was `true`)* | clean | 0 | — |
| a 0-byte file | `false` *(was `true`)* | clean | 0 | — |

**The existence guard below it is unchanged, deliberately.** The warning answers "there is no file at this path", and `--require-file`'s contract is absence rather than shape, so both still read `existsSync`. Gating either on the parse was measured and rejected: `--require-file` against a mangled ledger flips **exit 0 → exit 2**, and the warning starts printing `ledger file not found` for a file that is sitting right there. Both are pinned by the new test rather than left to drift, and both are filed as **#817**.

`ledger` is also built once and used by shorthand at both emission sites, and the section-header regex moved out of `section()` to a single definition the parser and the flag share.

### Narrower, not airtight

Parse-derived `ok` **narrows** the false-claim surface; it does not eliminate it. A ledger truncated **after** the `## Filed` header still reports `ok:true`, with the rows below it lost (measured). Truncation *before* the header reports `ok:false`. This is stated in the code comment, in `SKILL.md` and here rather than claimed closed.

### How reachable is any of this

The corrupt / not-a-ledger state is **operator-reachable only**: `save()` cannot emit a structurally invalid ledger, and no fleet caller passes `--file` to `check`. The fix is worth making because the claim is machine-readable and false, not because the state is likely.

### Docs moved with it

The design spec's script-surface row never named the `ledger` field at all, and its `check` enumeration was otherwise exhaustive — so the row was stale from the first commit of this branch. The row now names the field and carries the field-completeness pin its `no-undo-audit.sh` and `worktree-audit.sh` siblings already had, derived from a real `check` run. All three prose sites that described the field as an existence probe moved together: the comment above the flag, the Run ledger section of `run-team/SKILL.md`, and that spec row.

### Tests

`node --test skills/fleet/scripts/*.test.mjs` — **1075 tests, 1075 pass, 0 fail** (1073 on the branch point).

Each new test mutation-checked against the **whole** suite:

| mutant | result |
|---|---|
| flag reverted to `existsSync(file)` | exactly the new present-but-unparseable test red |
| `ledger` key moved to the end of the payload (negative control) | **green** — the pin is on the field, not on key order |
| `ledger` dropped from the spec row's `check` enumeration | exactly the new field-completeness test red |
| prose reworded inside that same spec clause (negative control) | **green** |

The field-completeness pin could not be copied verbatim from its siblings: measured, their word-boundary match over the whole Out cell reports `ledger` present against the unfixed row, because the cell's prose calls the ledger a ledger. Scoping to `check`'s own clause and to the cell's backticked tokens is what makes it discriminate.

### Filed from this review

- **#816** (`needs-triage`) — `read` keeps the missing-vs-empty ambiguity, and `read` is what `board.mjs` consumes. Already listed under "Enumerated and deliberately left" above.
- **#817** (`ready-for-agent`) — the corrupt case is now correct in the payload and silent on stderr; plus the `--require-file` question, recorded and explicitly out of scope there.


feigi:
**Review findings applied — `cee90fd`**

Applied:

- **spec row, `docs/specs/2026-07-23-fleet-plugin-design.md:188`** (correctness + comments, one defect on one sentence, one edit): the script-surface row typed `check` without the `ledger` field this branch emits. The field now sits in the base group, which is also what scopes it — the "besides unless already filed" qualifier reaches only `near`/`nearTotal`/`tracker`, while `ledger` rides the already-filed arm too. Gloss added for `ledger.ok`, settled by running `check` both ways.
- **field-completeness pin, `skills/fleet/scripts/ledger.test.mjs`**: the guard its `no-undo-audit.sh` and `worktree-audit.sh` siblings carry, which ledger.mjs had no equivalent of. Not a verbatim copy — measured, the siblings' word-boundary match over the whole Out cell reports `ledger` present against the row as it stood before this commit, because the cell's prose calls the ledger a ledger. Scoped to `check`'s own clause and to backticked tokens, the same match reports it absent. Mutation-checked both ways against the whole suite: field dropped from the row → 1073/1074, exactly this test red naming `ledger`; prose reworded inside the same clause (negative control) → 1074/1074 green.
- **`skills/fleet/scripts/ledger.test.mjs`** — the five `ledger?.ok` assertions drop their optional chaining; the field is unconditionally present on both payloads.

`node --test skills/fleet/scripts/*.test.mjs` from the worktree: **tests 1074, pass 1074, fail 0** (1073 before, +1 for the new pin).

Deferred and filed:

- **#816** (`needs-triage`) — `read` keeps the missing-vs-empty ambiguity #231 removed from `check`, and `read` is what `board.mjs` consumes. Out of #231's scope and named in this PR's own "Enumerated and deliberately left" section.

Held, not applied: the cluster of findings that rewrite `ledger.mjs:220` and the two `ledger: { ok: … }` emission sites. They propose mutually exclusive changes to the one expression — stop meaning existence, thread the flag out of `load()`, collapse to `const ledger = { ok: existsSync(file) }` — so none can be applied before the conflict is settled. That line and both emission sites are untouched by this commit.

feigi:
**Held-cluster ruling applied — `4b09d9e`**

`ledger.ok` is now derived from the parse inside `load()` rather than from a second `existsSync`. The full write-up, the measured fixture matrix and the mutation table are in the PR body under **Update (`4b09d9e`)**.

Applied:

- **`survived-1`, narrow form.** Flag set inside `load()`; the existence guard, the warning and `--require-file` are untouched. A mangled header, a non-ledger file and a 0-byte file report `ok:false` where they reported `ok:true`; the intact ledger still answers `already-filed` at exit 1.
- **`unverified-3` + `unverified-4`** — one `const ledger`, shorthand at both emission sites, initialised from the parse-derived flag rather than from `existsSync` (applying the finding verbatim would have cemented the expression the fix above replaces).
- **Test** for the present-but-unparseable case, which the branch's other tests structurally could not reach — they only exercised absent vs. well-formed. Mutation-checked against the whole suite both ways.
- **Three prose sites moved together:** the comment above the flag, `run-team/SKILL.md`'s Run ledger section, and the design spec's script-surface row. Leaving one behind is the drift this PR's own review already caught once.

Rejected on measurement, filed as **#817** (`ready-for-agent`):

- Emitting the existing warning whenever `ledger.ok` is false prints `ledger file not found` for a file that exists — a fresh false claim on the human-readable half.
- Gating `--require-file` on the parse flips exit 0 → exit 2 against a mangled ledger, which #231 puts out of scope.

Both suites stay green through either, which is why they are pinned by the new test rather than left to a future reader's judgement.

Closed without change: the `load()`-threading remedy. Applied verbatim it leaks a fourth key into `read`'s payload — `read` is `JSON.stringify(data)` — reddening two existing #246 tests and contradicting the spec's `{rows[], filed[], ruled[]}`; kept out of `data` it is a measured no-op. Its actual intent is subsumed by the fix above, which reaches "one observation, from the pass that produced `data`" without touching `read`'s contract.

`node --test skills/fleet/scripts/*.test.mjs` — **1075 tests, 1075 pass, 0 fail**.

Deferrals from this review: **#816** (`needs-triage`), **#817** (`ready-for-agent`).

Issue #199, opened 2026-07-31
workflows/review-pr.js has no args decode guard, so Workflow({args}) throws at 0 agents every time — third recurrence


feigi:
Not from a PR diff — hit while running `/fleet:review-and-fix #195` on 2026-07-31.

`Workflow({name: "review-pr", args: {pr, branch, worktree, scratch, ...}})` delivers `args` to the script as a **JSON string**, not an object. Every `args && args.pr` reads `undefined`, and the guard at `workflows/review-pr.js:127`

```js
if (!pr || !worktree) throw new Error("review-pr: args.pr and args.worktree are required");
```

fires in 7 ms with `agent_count: 0`. The error names the missing fields, not the type, so it reads as a bad invocation rather than a serialization bug.

Known since 2026-07-23 (same failure on `#413`, reproduced twice there). `grep -n "typeof args" workflows/review-pr.js` → no match, so the fix identified then was never applied.

**Fix** — decode defensively at the top, before the `const pr = args && args.pr` block:

```js
const A = typeof args === "string" ? JSON.parse(args) : args || {};
```

then read every field off `A` instead of `args`. Keeps working when `args` does arrive as an object.

**Workaround until then:** every Workflow invocation persists its script and returns `scriptPath`; edit that file to hardcode `pr` / `branch` / `worktree` / `scratch` / `dimensions` as literals and re-invoke with `{scriptPath}`. Costs one wasted launch.

Related: #113 (`args.dimensions` keys are not normalized) and #142 (default `testCmd` is repo-specific) are also `args`-handling defects in this file — worth fixing in one pass.

feigi:
Closing as already fixed.

The `args` decode guard exists on `main`: `decodeArgs()` at `workflows/review-pr.js`, which returns the object untouched when `args` is already an object and `JSON.parse`s it when it arrives as a string, throwing a named error if that parse fails.

It landed **2026-08-05 in `b243b4e`** — a commit whose subject is *"fix(fleet): the default review path had no way to label a PR with nothing wrong with it"*. It closed this issue incidentally, which is why nobody noticed. This ticket was filed 2026-07-31, five days earlier.

Verified rather than inferred — a `/fleet:run-team` review run today invoked `Workflow({name: "review-pr", args: {...}})` against PR #224 and spawned **11 agents** (1 snapshot + 6 specialists + refuters), not the 0 this issue describes.

[...]

Issue #187, opened 2026-07-31
recall-lib.sh:146: the stale-recall prune silently no-ops when TMPDIR is unset, because /tmp is a symlink


feigi:
Deferred from PR #177 review. **Same defect class and same one-character fix as that PR** — worth landing while the reasoning is fresh.

## The defect

`hooks/lib/recall-lib.sh:146`:

```sh
find "$dir" -maxdepth 1 -name 'agent-brain-recall-*' -mtime +3 -delete 2>/dev/null
```

`$dir` derives from `${TMPDIR:-/tmp}` (`recall-lib.sh:116`). On macOS `/tmp` is a symlink:

```
$ ls -ld /tmp
lrwxr-xr-x@ 1 root wheel 11 /tmp -> private/tmp
```

`find` does not descend a symlinked **argument**, so with `TMPDIR` unset the prune walks nothing and stale ledger files accumulate forever. Reproduced directly, with `/usr/bin/find`:

```
$ ln -s real link      # real/ holds an old agent-brain-recall-* file
$ find link  -maxdepth 1 -name 'agent-brain-recall-*' -mtime +3 -print   ->  (nothing)
$ find link/ -maxdepth 1 -name 'agent-brain-recall-*' -mtime +3 -print   ->  link/agent-brain-recall-old
```

`2>/dev/null` means it cannot even complain.

## The fix

```diff
-  find "$dir" -maxdepth 1 -name 'agent-brain-recall-*' -mtime +3 -delete 2>/dev/null
+  find "$dir/" -maxdepth 1 -name 'agent-brain-recall-*' -mtime +3 -delete 2>/dev/null
```

Exactly the fix PR #177 applied to `claim-ticket.sh`, for exactly the same reason: stat-style predicates follow symlinks, `find` on a bare argument does not.

## Severity

Low, and it **fails safe** — the cache lingers rather than the wrong thing being deleted. macOS-only in practice, and only when `TMPDIR` is unset (macOS normally sets it to `/var/folders/...`, so this bites where it is cleared — cron, some CI shells, `env -i`).

`ready-for-agent`: defect verified, fix is a single character with a directly analogous precedent now in the tree. A test is optional given the deletion is guarded by `-mtime +3`; if one is wanted, the shape is the symlink fixture in `claim-ticket.test.mjs`.

Surfaced by the code reviewer on PR #177 during a root-cause sweep for the same `[ -d ]`/`find` mismatch; reproduced independently by the reviewer before filing.

[...]

PR #813, opened 2026-08-22
fleet: a phase-0 liveness probe whose third answer is "could not check"

### LINKED ISSUE #238: fleet: a ticket fixed as a side effect of an unrelated commit is never reconciled, so it stays ready-for-agent as live bait
Found by the controller during a `/fleet:run-team` run. Reproduced and then acted on before filing.

## What happened

**#199** — *"workflows/review-pr.js has no args decode guard, so `Workflow({args})` throws at 0 agents every time — third recurrence"* — was filed **2026-07-31**.

It was fixed **2026-08-05** by `b243b4e`, whose subject is *"fix(fleet): the default review path had no way to label a PR with nothing wrong with it"*. The commit added `decodeArgs()` while solving a different problem, and closed #199 as a side effect. Nobody noticed.

The issue stayed **OPEN** and **`ready-for-agent`** for five days and reached this run's phase 0 candidate list as apparently-live supply.

Verified before closing it, rather than inferred:

- `git log -S"decodeArgs" -- workflows/review-pr.js` → `b243b4e`, and `git merge-base --is-ancestor b243b4e origin/main` → true
- a live `Workflow({name: "review-pr", args: {...}})` in this run spawned **11 agents**, not the 0 the issue describes

## Why it matters

`candidates.mjs` cannot tell a fixed ticket from a live one — both are open and labelled. The cost is not cosmetic: phase 0 counts it as supply, and an implementer can claim it, take the `in-progress` label, create a worktree and consume a slot before discovering there is nothing to do. In this run I read it in full during the shortlist, which is the expensive read the label is supposed to have earned.

It also inflates the supply figure the queue-depth table uses to decide whether to suggest `/triage`.

## Why needs-triage: the approach is genuinely open

Four candidates, each putting the work in a different place, none obviously right:

1. **A phase-0 staleness probe** on shortlisted tickets only — cheap, but only covers tickets that reach the shortlist, and adds a per-candidate cost to the hot path.
2. **A periodic sweep job** over all open `ready-for-agent` issues — catches everything, but needs a way to decide "is this still true?" without a human, which is the hard part.
3. **A triage step** — reliable, but it is human time on every pass and it scales with the backlog rather than with the defect rate.
4. **Require commits to name every issue they close** — fixes it at the source and costs nothing to check, but it is a process rule with no enforcement point today, and `b243b4e`'s author had no reason to know it was closing #199.

Note that (1) and (2) both need a general "is this ticket still reproducible?" oracle, which does not exist and may be the real ticket hiding behind this one.

## Acceptance criteria

- [ ] A ticket whose defect no longer reproduces stops being offered as `ready-for-agent` supply, by whichever mechanism triage picks
- [ ] The mechanism does not close a ticket on a *guess* — closure requires evidence the defect is gone, of the kind gathered for #199 above
- [ ] A ticket that is merely hard to reproduce is not silently dropped
- [ ] The chosen approach is stated in the fleet docs where the next reader will find it

## Related

- #199 (closed not-planned during this run, with the evidence in a comment) is the worked example

feigi:
Closes #238

Triage ruled option 1 — the phase-0 staleness probe. Nothing reconciles a
backlog ticket against the tree, so a ticket fixed as a side effect of an
unrelated commit stays open and `ready-for-agent`: #199 was closed by `b243b4e`
while that commit was solving something else, and phase 0 counted it as live
supply until someone claimed it.

## What lands

`skills/fleet/scripts/staleness.mjs --path <p> (--gone <s> | --present <s>)`,
run at phase 0 step 4 once the survivor read has supplied the string. It does
the mechanical half — what `origin/main` says about one string, and which
commit changed it — and leaves choosing the string to the read that already
happened, which is why it is on the shortlist and not in the candidate scan.

| Exit | Verdict | Supply |
|---|---|---|
| 0 | still reproduces | offer it |
| 1 | provably fixed | do not offer; close citing the payload's `commit` and `subject` |
| 2 | could not check | offer it, saying the probe could not check |

Three values because two is the defect. Exit 2 is `arg.mjs`'s `die()` code, so
every guard and every failed git call lands on the verdict that keeps the
ticket in the queue; the outer catch is there because Node's own exit code for
an uncaught throw is the one this script reserves for "provably fixed".

`fixed` never rests on the string's presence alone — it carries the commit
`git log -S` names at that path, reachable from `origin/main` by construction
since that is where the walk starts. A `--gone` string absent from the file
that also never changed count there is exit 2: the positive control, without
which a typo in the string closes a live ticket.

It reads `origin/main` and nothing else, never the working tree and never
anything on disk. A path this repo does not track is exit 2 rather than a
clean read — #187's case and the generated-artifact case at once, `agent-test`
being written into a claimed worktree by `claim-ticket.sh`'s heredoc, which
makes every copy under `.worktrees/` a snapshot frozen at claim time.

## Verification

Run against the ticket's own worked example, on `origin/main`:

    staleness.mjs --path workflows/review-pr.js --present decodeArgs
    → fixed, b243b4e, "fix(fleet): the default review path had no way to
      label a PR with nothing wrong with it"

and against the decoy:

    staleness.mjs --path agent-test --gone <anything>
    → could not check, tracked: false

Every mutant of a pinned branch was killed by its own test and no other —
untracked-reads-as-gone, the missing positive control, `HEAD` in place of
`origin/main`, an unreadable object carried past, and both catches removed so
an uncaught throw exits 1. A reworded `why` string stays green, so the suite
discriminates rather than merely reddening.

No reproduction runner and no backlog-wide oracle — option 2, which the ruling
declines. No auto-closing: the probe informs supply, a human or a triage pass
closes. `fleet-tick.mjs`'s supply count is unchanged, since correcting it
would mean probing every open ticket.


feigi:
Review findings applied and deferred. Pushed `c0c2b54`.

**Applied** (two commits): the sweep moved below the value guards so its own comment is true and `arg.mjs` keeps `board.mjs` as its one documented exception; `--path` anchored to the repo root and normalised to the spelling `ls-tree` prints; the `cat-file` catch no longer asserts a cause it cannot tell apart, and the read has real headroom; the payload goes out through `writeSync`, so a verdict nobody received is a could-not-check instead of a silent exit 1; the `unknown()` shape comment now says what every branch actually emits, measured; and the `--`-prefixed-needle restriction is stated in the usage block and in `arg.mjs`'s header. Four new pins — the origin/main-scoped citation walk, the empty-file guard, the one-answer guard, and the root anchoring — each mutation-tested, killed by its own case only, with three no-op changes confirming they stay green. Full suite `node --test skills/fleet/scripts/*.test.mjs`: **1085 pass / 0 fail**.

**Deferred**, filed one issue each:

| Issue | Label | Finding |
|---|---|---|
| #818 | `needs-triage` | `--`-prefixed needle: defect refuted, capability gap remains |
| #819 | `needs-triage` | spec script-surface table lists none of the scripts added since |
| #820 | `needs-triage` | `tracked:false` for a pathspec that matched nothing — refuted |
| #821 | `needs-triage` | exit-1 row and the subject caveat — refuted |
| #822 | `needs-triage` | outer catch backstop unpinned, unreachable from the CLI |
| #823 | `ready-for-agent` | no prose pin over the new SKILL.md block |
| #824 | `needs-triage` | verdict exit-code lookup unguarded — refuted |
| #825 | `ready-for-agent` | nothing pins that `--gone`/`--present` route through `has()` |
| #826 | `needs-triage` | redundant `lines` array; its count check is an equivalent mutant |

Five `suggestion`-band findings were dispatched to a refuter each before any edit; four came back refuted and are filed with the refutation rather than applied.

Issue #263, opened 2026-08-10
cksum exit status hidden behind the pipe emits a blank runner template stamp at exit 0


feigi:
Deferred from PR #247 review. **One issue for one defect** — the review returned it four times across three bands (`unverified-2` correctness, `unverified-3` silent-failure, and `refuted-1` silent-failure, which is the same defect restated). Filing it once rather than duplicating.

**dimension:** correctness + silent-failure
**severity:** suggestion (both `unverified` entries); `important` in the refuted band
**file:** `skills/fleet/scripts/claim-ticket.sh:118` and `:121`

**Finding.** `tmpl_stamp=$(cksum "$0" | cut -d' ' -f1)` takes the pipeline's status, which is `cut`'s, so `set -eu` (no `pipefail`, POSIX sh) never fires when `cksum` cannot read `$0`. The runner is then written with an empty stamp line and the claim exits 0. The second half: under the read protocol the shipped comment prescribes ("diff the stamp against a fresh `cksum` of this script"), a blank stamp could be misread as "old template" when the truth is that stamping never ran.

**The mechanism is real and was reproduced three times independently.** `sh -s 42 slug fix --apply < claim-ticket.sh` (so `$0` is `sh`): stderr `cksum: sh: No such file or directory`, exit 0, and `od -c` of line 2 shows `# a g e n t - t e s t   t e m p l a t e :  \n` — nothing between the colon-space and the newline. The absolute-path form emits `# agent-test template: 3887412534` normally.

**Why deferred — refuted by both refuters, on four grounds:**

1. **Unreachable from every invocation form that exists.** `SKILL.md:156` prescribes the absolute path `~/.claude/skills/fleet/scripts/claim-ticket.sh …`; `settings.json` uses a path; both test call sites are `spawnSync("sh", [SCRIPT, ...])` with `SCRIPT` absolute. No stdin/pipe/`sh -s` form exists anywhere in the repo. The probe is an artifact, not a route.
[...]

PR #828, opened 2026-08-23
docs(fleet): point a runner's reader at the template stamp, and at what it cannot tell them

### LINKED ISSUE #252: document the agent-test template stamp where an operator hits a stale runner
Deferred from PR #247 review.

**dimension:** correctness
**severity:** suggestion
**file:** `skills/fleet/skills/run-team/SKILL.md:171`

**Finding.** PR #247 (#124) stamps every emitted runner with `# agent-test template: <cksum>`, a feature whose entire value is human legibility — nothing reads it, nothing refuses on a mismatch. Nothing outside `claim-ticket.sh` and its test mentions it. `grep -rn "cksum|template stamp|agent-test template|tmpl_stamp" --include="*.md" .` over the 23 `.md` files in the snapshot exits 1: no doc anywhere points at the mechanism.

Suggested fix was: add one line to the SKILL.md reused-worktree paragraph — the runner's line 2 carries `# agent-test template: <cksum>`; compare it with `cksum ~/.claude/skills/fleet/scripts/claim-ticket.sh` (or `git show <rev>:skills/fleet/scripts/claim-ticket.sh | cksum`) to tell whether the runner predates a template fix.

**Why deferred — refuted, on the landing spot, not on the gap.** A refuter ran the recipe and the greps:

- The recipe is sound. A real `--apply` claim emitted stamp `3887412534`; `cksum <claim-ticket.sh>` → `3887412534`, and `git show ad283f1:skills/fleet/scripts/claim-ticket.sh | cksum` → `3887412534`. Both compare forms reproduce the emitted value.
- The doc gap is real: no `.md` mentions the stamp.
- **But `SKILL.md:171-179` is the wrong place.** Verbatim it covers a worktree that "predates the marker and has no `./agent-test` — a member told to use it stalls on a missing script": an **absent** runner. The stamp exists for a **present** runner written from an older template. Different failure.
- That section's remedy is also already prescribed unconditionally for every not-freshly-claimed worktree — "either re-materialize the runner first or tell the member the runner is absent" — and re-materializing overwrites a stale-template runner too.

So the gap stands but the fix as specified lands in a paragraph about a different scenario, and the obvious host section already prescribes the remedy. `needs-triage`, not `ready-for-agent`: the defect is verified, the approach is open — decide *where* the stale-template case gets documented (a new SKILL.md paragraph of its own? `references/isolation.md`, which describes the runner's contents? nowhere, on the ground that re-materializing already covers it) before anyone writes the line.

Note the shipped comment in `claim-ticket.sh` was corrected in PR #247's fix commit to say plainly that the stamp is a whole-script checksum that over-reports, so any doc line written from here must not re-import the old "changes exactly when the template does" claim.

feigi:
Closes #252

`references/isolation.md` gains one paragraph, in the section that describes the runner file. No code changes.

> **This description was corrected after review.** Two of its original bullets asserted a calibration the review then disproved by measurement. They are rewritten below to match the shipped text, and what was wrong is recorded at the end rather than quietly dropped.

## What it says

- The runner is untracked and written once at claim time, so every copy under `.worktrees/` is a snapshot — reading or measuring one measures that build.
- Line 2 carries `# agent-test template: <cksum>`, and both compare forms reproduce it.
- The stamp checksums the **whole emitting script**, not the runner template inside it, so it **over-reports**: an edit anywhere in `claim-ticket.sh` moves the number while the runner it produces can stay byte-identical.
- It also **under-reports**, which is why a match is not the freshness proof it looks like. The stamp covers `claim-ticket.sh` and nothing else, while the entrypoint the runner `exec`s — and whether the directory-expansion shim is emitted at all — comes from the sibling `derive-testcmd.sh`, which is not checksummed. Read a match as "this runner came from these bytes of `claim-ticket.sh`", nothing wider.
- A mismatch means the emitting bytes differed. Not "the template changed", and not "older" either: `cksum` carries no order, and over this script's own history both the checksum and the byte count fall as well as rise.
- Nothing reads it and nothing refuses on a mismatch — it is legibility, not a gate.

## Settled by running, not by quoting the ticket

| Claim | How |
|---|---|
| Both compare forms agree with a real emitted stamp | `cksum ~/.claude/skills/fleet/scripts/claim-ticket.sh`, `git show origin/main:skills/fleet/scripts/claim-ticket.sh \| cksum`, and line 2 of a real emitted runner all agree |
| `agent-test` is generated, not tracked | `git ls-tree -r origin/main --name-only \| grep agent-test` returns only the out-of-scope note and a prose test — no runner (positive control: the same grep for `claim-ticket` returns the script and its suite) |
| `isolation.md` mentioned nothing of the stamp | `git grep -iE 'cksum\|template stamp\|agent-test template\|tmpl_stamp\|checksum' origin/main -- <that file>` exits 1, with `agent-test` as the positive control on the same invocation form |
| The stamp under-reports | Two copies of `claim-ticket.sh` with an identical `cksum`, in fixture repos differing only in the sibling `derive-testcmd.sh`, emit runners carrying the **same stamp** with different `exec` lines and different sizes |
| The over-report direction is real | Appending a comment to a third copy moves the stamp while the emitted template compares identical under `cmp` |

## What the review corrected, and why it is worth recording

The original description — and the paragraph as first written — claimed *"the error runs one way, so a match means fresh"*. That is **false**, and the review proved it by construction rather than by argument: the runner's entrypoint comes from a sibling script the stamp does not cover, so a change confined to that sibling yields a byte-different runner under an identical stamp.

The claim was not invented here. It was inherited from `claim-ticket.sh`'s own comment, which still states it — that half is routed to #263, which owns and has moved that block, so this PR deliberately leaves `claim-ticket.sh` untouched. The lesson the mistake carries: confirming that a source *says* something is not confirming that it is *true*, and the original evidence table cited reading the source as if it were verification.

## Landing spot

Per the triage ruling on #252. The reused-worktree paragraph in `run-team/SKILL.md` is **not** touched — that covers an *absent* runner; the stamp is for a *present* runner from an older emitter.

Placed as a second paragraph under the existing runner section rather than as a new `##` section: every heading in that file is a slice boundary for a prose pin, and each existing heading has a counterpart bullet in `SKILL.md`, which a new heading would not.

## Tests

Full suite green from the worktree. The added paragraph widens the slice `agent-test-dir-prose.test.mjs` takes between two headings, so that pin was mutation-tested at the widened size: replacing the pinned clause reds exactly the `references/isolation.md` rows and leaves the `run-team/SKILL.md` rows green, and rewriting the new sentence — which the pins must not catch — leaves both prose suites green. Restored by `cp`. No pin was added; the ticket's acceptance criteria say no code changes.

A finding about "nothing reads the stamp" was refuted by its own refuter and deferred to #830.

https://claude.ai/code/session_01P82HSsEj2q98cFBxBxZsNG


feigi:
Review findings adjudicated — 2 applied in `0a2847f`, 1 deferred as #830.

| finding | band | checked by | outcome |
|---|---|---|---|
| `survived-0` — "The error runs one way, so a match means fresh" is false | important | 2 refuters, both reproduced it by running | **applied** |
| `unverified-1` — "a mismatch means this runner predates some change" asserts an ordering `cksum` cannot carry | suggestion | 1 refuter dispatched, **survived** (refuted=false) | **applied** |
| `unverified-0` — "Nothing reads the stamp" is false | suggestion | 1 refuter dispatched, **refuted** (refuted=true) | **deferred → #830** |

Both `suggestion`-band findings arrived unchecked (the workflow budgets that band 0 refuters), so each got its own refuter before anything was touched.

**Re-measured here rather than taken from the findings.** Two copies of this worktree's `claim-ticket.sh`, `cksum 782616158 28102` for both, in two fixture repos differing only in the sibling `derive-testcmd.sh`, claimed with `--apply` and `gh` stubbed: both runners carry `# agent-test template: 782616158`, one `exec`s `node --test` with the directory-expansion shim, the other `node --test --experimental-test-isolation=none` without it, and they differ from the fourth line on. The over-report direction still holds and is kept in the text: appending a comment to a third copy moved the stamp to `2648065824` while the emitted template stayed byte-identical under `cmp`. For the ordering claim, `cksum` over `claim-ticket.sh`'s own history moves both fields in both directions (checksum `1921791372` → `1577731428`, byte count `23788` → `23526`), so a difference cannot say which side is older.

The paragraph now states what the stamp **covers** rather than what it guarantees, which is the shape that failed.

**Left alone deliberately:** `claim-ticket.sh`'s own comment carries the identical overclaim. PR #829 (#263) owns that file and has already moved that block, so editing it here would conflict on relocated lines. Not filed either — already routed.

**Tests:** `node --test skills/fleet/scripts/*.test.mjs` → `tests 1090, pass 1090, fail 0`, unchanged from the pre-edit baseline.

**Prose pins mutation-tested both ways.** The edited paragraph sits inside `agent-test-dir-prose.test.mjs`'s `between()` slice (confirmed: the slice contains the new text). Breaking a pinned phrase reds exactly one test — `references/isolation.md says a directory expands to the test files under it` — while its `run-team/SKILL.md` twin stays green. Rewriting the new sentence, which the pins should **not** catch, leaves both prose suites green at 12/12. No heading added or renamed. Tree restored with `cp`, not a git discard.

PR #829, opened 2026-08-23
fix(fleet): derive the runner template stamp as one command, before the claim

### LINKED ISSUE #263: cksum exit status hidden behind the pipe emits a blank runner template stamp at exit 0

feigi:
Closes #263

## What this is, and what it is not

Not a reachable defect. Grounds 1-3 of the refutation stand: every call site
this repo has — the skill text, `settings.json`, and both test call sites —
names the script by an absolute path, so `$0` always resolves; a blank stamp
is not confusable with an old one, since the only parser in the repo rejects
the blank line; and nothing reads the stamp or refuses on a mismatch.

What this applies is the convention `inflight.sh` already states in its own
comments — one command, not a pipeline that hides every status but its last —
to the site that was still carrying the exception.

## The change

`tmpl_stamp=$(cksum "$0" | cut -d' ' -f1)` reported `cut`'s status, so
`set -eu` (no `pipefail`, POSIX sh) never saw a `cksum` that could not read
its operand. It is now `cksum` alone with its own `|| die`, and the trailing
fields come off by parameter expansion.

Reading the status is only half of it. The derivation ran after
`gh issue edit --add-label in-progress` and after `git worktree add`, so a
guard added where it sat would refuse a ticket that was already half-claimed.
It depends only on the script's own path, so it moves up with the derivations
that run before the apply branch, where a refusal has nothing to clean up —
the same property the `json.sh` guard is placed for.

## Measured

Against a fixture repo with a stubbed `gh` and a stubbed failing `cksum`:

| script | mode | exit | receipt | issue labelled | worktree | branch ref | stamp |
|---|---|---|---|---|---|---|---|
| base | `--apply` | 0 | `"applied":true` | yes | on disk | created | nothing after the colon |
| base | dry run | 0 | `"applied":false` | no | none | none | n/a |
| this PR | `--apply` | 2 | none | no | none | none | n/a |
| this PR | dry run | 2 | none | no | none | none | n/a |

With a working `cksum`, both modes are byte-identical to base: the dry run's
stdout and stderr diff clean, the `--apply` run's stdout and stderr diff
clean, and the emitted runner diffs clean apart from the stamp value itself,
which moves because the script's own bytes moved — which is what the stamp is
for. The value is digits-only and equals a fresh `cksum` of the script.

## What this restructuring newly does

- **The default dry run now reaches the derivation.** Where the stamp cannot
  be computed, the dry run refuses (exit 2) instead of exiting 0 and
  predicting a claim. Every other derivation ahead of the apply branch already
  behaves this way, and a dry run promising to write a runner it could not
  stamp would be predicting a claim it cannot make. With a working `cksum` the
  dry run is unchanged, byte for byte.
- **`set -eu` coverage changed shape**: the pipeline is gone, so the
  derivation's status is now the script's own to read, and it reads it
  explicitly rather than relying on `-e`. The `|| die` keeps the refusal on
  exit 2 — this script's only failure code, with no exit 1 — where a bare
  `set -e` abort would have introduced one.
- **The script's last command is unchanged**: the receipt `printf` still is,
  so exit status on the success path is untouched.
- **`tmpl_stamp` is now set but unused in the dry run.** Harmless under
  `set -u`, which fires on unset, not unused.
- The stamp's comment still says it is a checksum of the whole script, not of
  the emitted template.

## Test

One test added, `--apply` and the dry run in one fixture. Exit 2, an empty
stdout and a matching stderr are each reproducible by another guard, so the
assertions that carry it are the `gh` log, the worktree registration and the
branch ref — the state the guard-in-place form left behind.

Mutation matrix, run on copies under scratch:

| mutant | result |
|---|---|
| pipeline restored, placement kept | red, and the only red |
| status read, placement reverted into the apply branch | red, and the only red — killed by the `gh` log; the exit code, the empty stdout and the stderr match all still held |
| same fix, a different position ahead of every mutation | green |
| an unrelated stderr message reworded | green |

Suite: 1091 pass, 0 fail across `skills/fleet/scripts/` after rebase onto
`main`, which is every test file in the repo. `shellcheck -s sh` reports the
same notes as base.

## Out of scope, per the ticket

Nothing is made to read or enforce the stamp. Where the stamp is documented
for a reader is #252. A scan of `claim-ticket.sh` outside the emitted runner
finds no other shell pipeline, so no adjacent instance of this class is left
in the file.


feigi:
## Review findings applied — `557f88c`

Seven findings. Applied 4 (three defects, two of which were one defect each stated twice), refuted 1, filed 1 record.

| Finding | dimension | Outcome |
|---|---|---|
| `survived-0` + `unverified-0` — the guard reads `cksum`'s status only | silent-failure / correctness | **applied** |
| `survived-1` + `unverified-1` — `.out-of-scope` record describes code this PR moved | comments / correctness | **applied** |
| `refuted-0` — the spec table's refusal row omits the new `die` | comments | not applied — the row summarises refusal *classes*, and already omits a pre-existing unnamed `die` at `claim-ticket.sh:25` independent of this diff |
| `unverified-2` — three "nothing claimed" assertions are dominated | simplify | **applied**, after one refuter measured it |
| `unverified-3` — the two rationales are each written out twice | simplify | **refuted** by one refuter — record filed as #831 (`wontfix`, closed) |

### The guard now reads the value as well as the status

A `cksum` exiting 0 printing nothing left `tmpl_stamp` empty — `set -u` catches an unset variable, never an empty one — so the blank stamp line still shipped at exit 0 with the issue labelled, the branch ref made and the worktree on disk. That is verbatim the outcome this PR's own comment says the fix removes, so the code was changed to make the comment true rather than the comment softened. Two parts now, in the shape the lockfile guard already uses.

### The fix needed a third stub, or it unpinned what this PR is named for

Measured after committing the guard: with only the failing and the empty stub, restoring the pipeline form `cksum "$0" | cut -d' ' -f1` — the #263 bug itself — left **both tests green**, and so did dropping the `|| die` to a bare `|| true`. A failing `cksum` prints nothing, `cut` succeeds on empty input, and the new value check refuses what the status check was meant to.

A `cksum` that exits nonzero *after printing a plausible line* separates them. Whole-suite mutation matrix at `557f88c`:

| mutant | reds |
|---|---|
| value check deleted | `comes back empty`, and only it |
| status check → `\|\| true` | `fails despite printing`, and only it |
| pipeline form, both checks gone (the #263 bug) | all three stubs |
| benign comment reword (control) | nothing — 1093/1093 green |

### The stamp's error is not one-way

`claim-ticket.sh`'s comment claimed a match means fresh. It does not: the runner's body is chosen by `$testcmd`, which comes from the sibling `derive-testcmd.sh`, and the checksum does not cover that file. Measured — two byte-identical copies of `claim-ticket.sh`, differing only in that sibling, emitted runners of very different sizes under one stamp. Corrected here, and annotated on `.out-of-scope/runner-template-stamp-hardening.md`, which carried the same claim. That record is annotated, never rewritten: its analysis and tense are left as written. It also notes that the repo's line-distance prose gate walks `skills/fleet/scripts/` and keys on `.sh`/`.mjs`/`.js` markers, so nothing scans that directory — measured, while its idiom pattern does match the stale sentence.

`node --test skills/fleet/scripts/*.test.mjs` → **tests 1093 / pass 1093 / fail 0**.

Issue #784, opened 2026-08-22
phase-3 CI watcher reports an API outage as silence — "no event" is indistinguishable from "not green yet"


feigi:
> *This was generated by AI during triage.*

Split out of #262. That ticket carries four open governance questions about the REST budget; **this is its one decided half** — the failure being *silent* — and it has a fix measured working in two separate live runs.

#262 stays open for the budget question. Do not close it with this.

## The defect

The natural way to write the phase-3 CI watcher is the one that hides the outage:

```sh
st=$(node skills/fleet/scripts/ci-state.mjs --pr "$pr" 2>/dev/null) || continue
```

`ci-state.mjs` fails closed and says so on stderr — correct on its own terms. The watcher discards it, so a rate-limited probe emits **nothing**, and "no event" is byte-identical to "no terminal state reached yet". The controller's standing rule that *a monitor event is a wake-up, never a verdict* gives no protection, because the failure mode is the **absence** of events.

Consequence: every open PR stalls unlabelled and the run reads as merely quiet.

Observed live at least twice. In the second instance the controller was waiting on an event that could never arrive, for a PR that was already green.

## The fix, measured working

Check the budget **before** polling — `gh api rate_limit` is not itself rate-limited, so the probe is free — and make the outage a transition event rather than silence:

```sh
rl=$(gh api rate_limit --jq '.resources.core.remaining' 2>/dev/null || echo "ERR")
if [ "$rl" = "ERR" ] || [ "${rl:-0}" -lt 200 ] 2>/dev/null; then
  if [ "$lastfail" != "budget" ]; then
    echo "WATCHER DEGRADED: core REST budget=$rl — CI polling paused, silence is NOT green"
    lastfail=budget
  fi
  sleep 120; continue
fi
[ "$lastfail" = "budget" ] && { echo "WATCHER RECOVERED: core budget=$rl, CI polling resumed"; lastfail=""; }
```

with the same treatment one level down: an empty `ci-state.mjs` payload emits a DEGRADED line rather than `continue`.

Both fired correctly in a live run:

```
WATCHER DEGRADED: core REST budget=0 — CI polling paused, silence is NOT green
WATCHER RECOVERED: core budget=4952, CI polling resumed
```

## Agent Brief

**Category:** bug
**Summary:** The phase-3 CI watcher reports an API outage as silence, which is indistinguishable from "no CI result yet"

**Current behavior:**
[...]

PR #833, opened 2026-08-23
fix(fleet): name a quota refusal as its own cause in ci-state's payload

### LINKED ISSUE #262: fleet: the phase-3 CI monitor exhausts the GitHub REST rate limit, and the natural watcher shape reports the outage as silence

feigi:
Closes #262

Triage ruled one of this ticket's four open questions in scope: make `ci-state`
distinguish a rate-limit 403 from every other cause landing on the same failure
arm. The other three (cadence/backoff, fan-out throttling, duplicate-monitor
detection) were ruled not-now with re-open triggers, and the silence half is
#784. None of them is touched here — in particular there is **no backoff**, per
the ticket's own measurement that a fixed pause stalls the fleet longer than the
outage does.

## The defect

Every `gh` read failure reached one arm that reported the exit code and nothing
a caller could act on, so an exhausted REST quota, an unresolvable repo and a
revoked token were one undifferentiated cause. The correct responses differ: a
quota refusal clears on its own and is worth re-probing, the others need someone
to look.

## What the edit does

A refusal whose forwarded `gh` stderr carries the quota wording now prints
`{pr, verdict: "rate-limited", reasons[]}` on stdout before dying. Detection
reads the child's stderr off the thrown error — measured: `execFileSync` both
forwards it to fd 2 **and** captures it — so no extra API call is spent
establishing that the API is refusing calls, and nothing is reprinted.

Matched on the quota wording rather than on the 403 status, because 403 also
carries refusals no amount of waiting clears.

## What it does NOT change

- **Exit codes.** Still 2. The not-green-exits-non-zero contract is untouched.
- **Every other cause.** Unchanged: same stderr line, same exit, still no
  payload. Pinned by a test, and proven by the always-emit mutation below.
- **stderr.** No diagnostic text moved to stdout; `gh`'s own 403 line already
  reaches stderr on its own, so the refusal is not restated.

## What it newly does to existing callers

- **This arm previously printed nothing at all**, so any payload here is a new
  shape. `board.mjs`'s `mapCi()` was measured against it: `status` is absent, so
  it stays on the `!== "completed"` arm and returns `unknown` — never green,
  never a false red. That matches the cockpit design's standing rule for this
  case.
- **Nothing can read as green.** `verdict` is a value no gate matches for green.
  A gate keyed `verdict != "not-green"` would see a new value — flagged rather
  than silently relied on.
- **Unobserved fields are absent, not null.** A quota refusal is a probe that
  could not look. `jobs`/`missing` absent is what lets a fail-closed default
  (`// ["absent"]`) fire instead of reading an empty array as "nothing missing";
  emitting nulls or empty arrays would let unread state read as read. Pinned by
  its own test, killed by its own mutation.
- **`prHead == runHeadSha` stays vacuously true** on this arm, as it already was
  against the empty payload — absent fields compare equal either way. Not a
  regression, and not fixed here; a gate resting on that field alone was already
  vacuous during an outage.
- **`board.mjs` no longer logs its own `ci-state ... failed` line** for a quota
  refusal, because it now receives a payload instead of an empty capture. The
  cause moved into the payload rather than being lost.
- **For #784:** a rate-limited probe is no longer an empty payload. A watcher
  keyed on "empty output means degraded" will need to read `verdict` instead.
  Flagged, not acted on — that ticket owns the watcher.

## Tests

The `gh`-subprocess-failure arm had **no test at all** before this: every other
exit-2 case dies in workflow discovery or in a shape check. The stub now carries
a failure *message*, since the cause lives in `gh`'s stderr and not its exit
code, which is what makes both causes drivable.

`./agent-test skills/fleet/scripts/` — **1096/1096**, 0 fail.

Mutation table (each applied and verified applied, restored by `cp` from a
pristine copy, never a git discard):

| Mutation | Result |
|---|---|
| never emit — collapse quota into the generic cause | 2 red: the quota test and the no-CI-state test. Control green |
| always emit — collapse the generic cause into quota | 1 red: the control only |
| report unobserved state as nulls/empty arrays | 1 red: the no-CI-state test only |
| drop the quota wording from the reason | 1 red: the quota test only |
| reword the reason, quota wording kept (non-catch) | all green |
| change the refused-query format (non-catch) | all green |

Each new test kills its own mutant and only it; two non-catch controls stay
green, so the tests discriminate rather than merely fail. One earlier non-catch
run was discarded as vacuous — its replacement had silently not applied, which
reads exactly like a passing control.


feigi:
Review applied at `b1d6a77`. Full suite `tests 1100, pass 1100, fail 0` from the worktree (baseline 1096, plus the four tests added here).

**Deferred:** #840 — `ci-state.mjs` serialises a non-numeric `--pr` as `pr: null` in both payloads. Dimension `types`, `ready-for-agent`. Pre-existing on the main path, where it ships at exit 0 with `verdict: green`, so it is out of this PR's scope; remedy is settled and recorded on the issue.

Everything else in the review was either applied here or refuted with the refutation recorded in the commit messages.

PR #319, opened 2026-08-10
docs(spec): re-anchor the three surviving bare-numbered citations, revert the narrowing

feigi:
Finishes the sweep #117 left incomplete and #272 deliberately did not widen into.

Three live, present-tense, bare line-numbered citations survived in
`docs/specs/2026-07-23-fleet-plugin-design.md`. All three resolved, so this was
fragility rather than rot — but two point at `.gitignore` in this repo, where the
numbers rot on any edit above them.

Each now names a **quoted fragment and its section**, per the spec's own citation
convention marker, and drops the number: a fragment survives renumbering, a
number does not. That matches the form #117 used for the `run-team` re-anchors
elsewhere in the same document.

| was | now |
|---|---|
| `` `${CLAUDE_PLUGIN_ROOT}` (`command-development/SKILL.md:564`) `` | `command-development/SKILL.md`'s `### CLAUDE_PLUGIN_ROOT Variable`, "an environment variable that resolves to the plugin's absolute path", plus the `vendored outside this repo` note its sibling already carried |
| `` `.gitignore:40` `` | `.gitignore`'s `plugins/` under `# Auto-managed by Claude Code` |
| `` `.gitignore:49` ignores `skills/*` `` | `.gitignore`'s `# Skills:` comment block ("Trailing-slash ignore blocks re-inclusion, so exclude contents and negate") ignores `skills/*` |

Re-measured at pickup, not taken from the ticket. All three of the issue's line
numbers were still accurate:

```
$ sed -n '39,40p;47,49p' .gitignore
# Auto-managed by Claude Code
plugins/
# Skills: symlinks to local skill repos aren't portable config; own skills are.
# Trailing-slash ignore blocks re-inclusion, so exclude contents and negate.
skills/*
$ sed -n '562p;564p' ~/.claude/plugins/marketplaces/claude-plugins-official/plugins/plugin-dev/skills/command-development/SKILL.md
### CLAUDE_PLUGIN_ROOT Variable
Plugin commands have access to `${CLAUDE_PLUGIN_ROOT}`, an environment variable that resolves to the plugin's absolute path.
```

## The narrowing, reverted — with one carve-out

#272's clause was narrowed to `every live citation **into `run-team` SKILL.md**
from this spec`. It reverts to the document-wide form here, but names the
exemption the document already states elsewhere: the vendored
`plugin-structure/SKILL.md:431` keeps its bare number, exempt where it is cited.

The issue predicted the document-wide form "becomes true" after these three. It
[...]

PR #842, opened 2026-08-23
docs(spec): mark the gitignore step done without rewriting the plan

### LINKED ISSUE #323: fleet-plugin-design spec :110-118: the gitignore paragraph narrates a completed migration in the present tense
Deferred from PR #319 review (closes #282). **dimension: `correctness`** — severity `suggestion`, from the `unverified` bucket (the review workflow budgets that band 0 refuters, so nothing adversarially checked it; see "Verification status" below).

## Finding

`docs/specs/2026-07-23-fleet-plugin-design.md:110-118` — the "One gitignore edit is mandatory" paragraph describes the **pre-migration** `.gitignore` in the **present tense**. PR #319 re-anchored its citation from bare `.gitignore:49` to the live `` `# Skills:` comment block `` , so an auditor following the new anchor now lands on a block that contradicts the paragraph.

Current prose (`:110-118`):

```
**One gitignore edit is mandatory.** `skills/` is not wholesale tracked either:
`.gitignore`'s `# Skills:` comment block ("Trailing-slash ignore blocks
re-inclusion, so exclude contents and negate") ignores `skills/*` and
re-includes individual directories by negation (`!skills/next-ticket/`,
`!skills/sizing-a-ticket/`, `!skills/caveman-compress/`). Without adding
`!skills/fleet/`, `git add skills/fleet` stages nothing and every packaging
commit is silently empty — the same class of failure as `/clean_gone` exiting 0.
Verify with `git check-ignore -v skills/fleet` before and after. The two
negations for the moved skills are retired once `!skills/fleet/` covers them.
```

What the anchored block actually contains at `origin/main`:

```
$ sed -n '47,53p' .gitignore
# Skills: symlinks to local skill repos aren't portable config; own skills are.
# Trailing-slash ignore blocks re-inclusion, so exclude contents and negate.
skills/*
# Kept after uninstalling the caveman plugin — now our own copy, not vendored.
!skills/caveman-compress/
!skills/fleet/
# Vendored skills stay ignored whether installed as symlink or copy — list them:
```

**Three clauses are pre-migration state, not two.** The finding named the first; the other two are the same defect:

1. `!skills/next-ticket/` and `!skills/sizing-a-ticket/` — **gone**. Both skills moved under `skills/fleet/`.
2. "**Without adding** `!skills/fleet/`" — `!skills/fleet/` is **already there** (`:52`). The mandatory edit was made.
3. "The two negations for the moved skills **are retired once** `!skills/fleet/` covers them" — already retired. Future tense over completed work.

The plan executed exactly as written (the two moved-skill negations went, `!skills/caveman-compress/` stayed because it was not moved into `fleet`), so nothing here is a *wrong* claim about history — it is a completed migration plan still narrated in the present.

**Out of #282's scope.** #282's stated job was to give three bare-numbered citations a quoted fragment **and** its section. The negation list is **pre-existing prose that PR #319 did not introduce** — it appears in both the `-` and `+` sides of the diff:

```
$ git diff origin/main...c801541 -- docs/specs/2026-07-23-fleet-plugin-design.md
-`.gitignore:49` ignores `skills/*` and re-includes individual directories by
-negation (`!skills/next-ticket/`, `!skills/sizing-a-ticket/`,
-`!skills/caveman-compress/`). Without adding `!skills/fleet/`, ...
+`.gitignore`'s `# Skills:` comment block ("Trailing-slash ignore blocks
+re-inclusion, so exclude contents and negate") ignores `skills/*` and
+re-includes individual directories by negation (`!skills/next-ticket/`,
+`!skills/sizing-a-ticket/`, `!skills/caveman-compress/`). Without adding
+`!skills/fleet/`, ...
```

The PR changed the **anchor**, not the claim. It also did not change the claim's truth value: the old subject `.gitignore:49` is the single line `skills/*`, so "re-includes individual directories by negation" already reached past its own anchor into the lines below. The re-anchor makes the mismatch **more discoverable**, not newly false.

**And the suggested fix as written would ship a half-corrected paragraph.** The review proposed marking only the parenthetical (`"(then `!skills/next-ticket/`, …)"`). That leaves clauses 2 and 3 above in the present tense two lines below, so the paragraph would then assert the mandatory edit is still pending while flagging its consequence as historical — the "correction mints a new wrong claim" failure mode this repo has now recorded four times.

## The open decision — why `needs-triage`, not `ready-for-agent`

This is a **design spec describing a migration plan**, and the file already carries a convention marker (`:392`) declaring certain line numbers "pre-fix" — but that marker is scoped to the **Site** column of the two tables in the migration-debt section and the prose between them. It does **not** reach `:110`. So the approach is genuinely open, and two competent implementers would build materially different things:

- **(a)** Leave it. A dated design spec legitimately records the state it was written against; present tense is the document's voice throughout.
- **(b)** Mark the whole paragraph pre-migration (all three clauses), matching the retirement sentence's framing.
- **(c)** Extend the `:392` convention marker to cover design prose as well as the tables, and mark this paragraph under it.
- **(d)** Rewrite the paragraph in past tense as a record of a completed step.

Pick the policy first; the edit is trivial once the policy is decided. Torn → surfaced rather than guessed, per run-team phase 0.

## Verification status

**No refuter was dispatched.** The scope split settles this before adversarial verification: out of the ticket's scope → defer and file, never apply. The *factual* half is confirmed by the `sed` output above; what is unsettled is whether the document should say something different, which is exactly the open decision.

## Sequencing — do not collide

`docs/specs/2026-07-23-fleet-plugin-design.md` currently has three owners on distinct sites:

- **this issue** — `:110-118`
- **#305** (gated behind PR #319 merging) — `:180` and `skills/fleet/commands/run-merge-bot.md`
- **#48** — `:208`, `:225`, `:428`

Filed separately rather than appended to #48 because #48 is `ready-for-agent` with three fully-settled fixes; folding an undecided item into it would break its decided status.

feigi:
Closes #323

## What the defect is

`docs/specs/2026-07-23-fleet-plugin-design.md`'s **"One gitignore edit is mandatory"** paragraph narrates the pre-migration `.gitignore` in the present tense — across its negation list, its condition on adding `!skills/fleet/`, and its sentence predicting the moved skills' negations retire. PR #319 re-anchored the paragraph's citation from a bare line number to the **live** `` `# Skills:` comment block ``, so an auditor following the anchor now lands on content that contradicts the prose, and can come away believing `!skills/fleet/` is missing and packaging commits are silently empty.

## What this does

Adds one annotation paragraph immediately ahead of the plan paragraph, naming it by its own lead-in construct so the mark scopes over the whole step rather than one clause. The plan's wording is **not** edited.

Per #257's class ruling — these specs are dated records, corrected by **annotation, never a silent rewrite** — a past-tense rewrite (option d) would make the spec read as though it always said the corrected thing. Marking only the parenthetical, as the originating review suggested, was rejected in triage: it leaves the "without adding" conditional and the retirement sentence reading as current, which is the "correction mints a new wrong claim" failure the ticket itself names.

The file's existing pre-fix convention marker was **not** extended to design prose — that is option (c), a separate decision, and out of scope. Its *form* (declare the scope, say the lines record a past state, give the reader the resolving command) is reused; its declared scope is untouched.

## Claims settled by running commands against `origin/main`

The issue body and triage comment were treated as leads, not citations. Every factual claim the diff makes was re-derived:

| Claim in the added prose | Settling command | Result |
|---|---|---|
| the `# Skills:` comment block it cites is live | `git show origin/main:.gitignore` | block present |
| now carries `!skills/fleet/` | same | present |
| no negation for the skills that moved under it | same | `!skills/next-ticket/`, `!skills/sizing-a-ticket/` both absent |
| …and those skills did move under it | `git ls-tree -d --name-only origin/main skills/fleet/skills/` | `next-ticket`, `run-team`, `sizing-a-ticket` |
| `!skills/caveman-compress/` is correctly still there | `git ls-tree -d --name-only origin/main skills/` | `skills/caveman-compress` — never moved into `fleet` |
| `git check-ignore -v --no-index skills/fleet` names the negation | that command, in a scratch repo built from this repo's `.gitignore` with `skills/fleet/` tracked | `.gitignore:53:!skills/fleet/`, rc 0 |
| with the negation deleted, the same command names `skills/*` instead | same fixture, negation removed, files still tracked | `.gitignore:50:skills/*`, rc 0 |
| the plain form can no longer witness the edit | `git check-ignore -v skills/fleet` in both fixtures | rc 1, no output — **identical** in both |
| the plain form did discriminate before the tree was tracked | same command, untracked fixtures | `skills/*` before the edit, `!skills/fleet/` after; both rc 0 |

### Correction after review

The first version of this annotation closed on *"so `git check-ignore -v skills/fleet` exits non-zero with no match"*. Review reproduced that as wrong twice over, and re-measurement confirms it: the sentence is **false** while `skills/fleet` is untracked (exit 0, printing the negation), and in this repo — where the tree *is* tracked — the command is **non-discriminating**, because `check-ignore` skips paths in the index without `--no-index`. It returns exit 1 and no output whether or not `!skills/fleet/` is present, i.e. it reports the same thing in the broken state the paragraph exists to warn about.

The annotation now cites `--no-index`, where the **printed pattern** is the signal and the exit status is not (both cases exit 0). The paragraph's conclusion is unchanged: the path is not ignored, and the plan executed as written. Only the cited evidence was replaced.

The plan's own *"Verify with `git check-ignore -v skills/fleet` before and after"* line is deliberately **left as written**, per #257. It asserts no exit code, and it genuinely discriminated in the untracked tree it was written for. It went stale when `skills/fleet` became tracked — a change in the world, not an error in the record — so the annotation says that outright rather than leaving a reader to choose between two commands inside the same step.

The ticket's own line numbers had drifted (it quotes the `# Skills:` block at `:47-53`; at `origin/main` it starts at `:48`), so nothing here rests on them.

## Blast radius

- **No prose pin slices this region.** Every test that reads this spec anchors a **script-surface table row** — `startsWith("| \`ledger.mjs\` |")` and its siblings in `ledger.test.mjs`, `no-undo-audit.test.mjs`, `reap.test.mjs`, `worktree-audit.test.mjs` — none uses a `between()` slice over this file, and none matches any string in the paragraph or the annotation.
- **No heading added or renamed.** Headings are `between()` slice boundaries in these files; the annotation sits inside the existing section.
- **No live line-numbered citation into this file below the insert point.** Checked both directions; the in-file `:84` self-reference sits above it, and the pre-fix numbers in the migration-debt tables resolve at the ref their own marker names.
- **Line numbers below the insert shift.** #305 (`:180`) and #48 (`:208`, `:225`, `:428`) own other sites in this file and cite them by number; those are now offset. The repo convention already says to cite this file by construct, and both tickets re-derive against `origin/main` before implementing.
- `.gitignore` is not touched, and no other paragraph of the spec is edited.

## No test added, and why

Not "specs are dated records so pinning them would be wrong" — this repo pins this very file. The measured reason: every existing pin into it anchors a table row stating a **live contract with a script**, and is asserted against that script's real behaviour. This paragraph states no contract with any script, so a pin over it could only assert its own text. Pinning the annotation against the live `.gitignore` would be worse — it converts a dated record into a live claim needing maintenance, which is what #257's ruling exists to prevent.

Review re-opened this as a suggestion; a refuter tested the proposed pin by writing it and mutating. It goes red when `!skills/fleet/` is deleted from a copy of `.gitignore`, but stays **green** both when the annotation paragraph is deleted outright and when its claim is inverted to a falsehood — so it pins `.gitignore`, not this prose, and cannot tell a correct paragraph from a missing or false one. Finding refuted; deferred to #844 rather than applied.

## Tests

`node --test skills/fleet/scripts/*.test.mjs`, from the worktree at the pushed head → **tests 1104, pass 1104, fail 0**.


feigi:
Review findings adjudicated at `a3f6e1e`.

**Applied** — `survived-0` + `survived-1`, one defect: the annotation's closing clause cited `git check-ignore -v skills/fleet` as "exits non-zero with no match". Re-measured in throwaway repos built from this repo's own `.gitignore` (git 2.50.1): false while `skills/fleet` is untracked (exit 0, printing the negation), and non-discriminating here where the tree is tracked — exit 1 / no output whether or not `!skills/fleet/` is present, because `check-ignore` skips paths in the index without `--no-index`. Now cites the `--no-index` form, where the printed pattern is the signal and the exit status is not. The paragraph's conclusion is untouched.

**Not fixed, by decision** — the plan's own "Verify with `git check-ignore -v skills/fleet` before and after" line. It is pre-existing and asserts no exit code, and it did discriminate in the untracked tree it was written for (`skills/*` before the edit, `!skills/fleet/` after). Editing it would rewrite the record, which #257 forbids; so the annotation now says outright that it predates the tree being tracked, rather than leaving the step holding two commands that contradict each other.

**Deferred** — #844, `unverified-0` (dimension `correctness`, severity `suggestion`): the annotation is held by no test. Refuted — a refuter wrote the suggested pin and mutated against it; it goes red when `!skills/fleet/` is deleted, but stays green when the paragraph is deleted outright *and* when its claim is inverted. It pins `.gitignore`, not this prose.

Tests: `node --test skills/fleet/scripts/*.test.mjs` → tests 1104, pass 1104, fail 0.

PR #843, opened 2026-08-23
fix(fleet): the near-miss floor comment denied being the threshold it is

### LINKED ISSUE #642: ledger.mjs: nearTotal's floor is a 0.005 score threshold, not "shares a content word" as its comment says (REFUTED as unreachable — the comment, not the code, is the candidate fix)
Deferred from PR #635 review (issue #154). Dimension: **silent-failure**. `skills/fleet/scripts/ledger.mjs:300` (comment at `:293-296`).

Filed for the record because a finding is deferred, never dropped. **A refuter ran it: the claim is TRUE, the state is unreachable.** Recommendation below is to fix the comment rather than the code, or close.

## The finding

`nearTotal` is not the exact total the comment claims: the filter at line 300 runs on the ROUNDED score, so a filed row that DOES share a content word but scores below 0.005 is dropped from both `near` and `nearTotal`, and no line reports it. The comment at `:293-296` states the floor is "shares at least one content word, not a score threshold"; `round2` before `> 0` makes it one.

## Refuter verdict: refuted=true, on reachability only

**The mechanism reproduces exactly.** Subject = 251 distinct content tokens, one filed row = 251 different distinct content tokens, one shared word:

```
"near":[],"nearTotal":0 — exit 0, no near-miss line, no further-near-miss line
```

Control, same subject vs a short filed row sharing the same word: `near-miss 0.33`, `nearTotal 1`.

**Boundary measured precisely.** The floor is `round2(overlap) > 0`, i.e. `shared / min(a,b) >= 0.005`. `Math.round(0.5)` rounds half up, so with one shared word:

- 200 distinct content tokens on **both** sides → `score 0.01`, `nearTotal 1` (surfaced)
- 201 on both sides → `near []`, `nearTotal 0` (silent)

`overlap` divides by `Math.min`, so BOTH sides must be ≥ 201. `scoreTokens` dedupes via a Set, drops tokens < 3 chars, drops the 50-word stoplist, folds a trailing `s`, and does not cap set size — nothing bounds the mechanism, only the input does.

**Nothing real gets near it.** 400 real issue titles from this tracker (`gh issue list --state all --limit 400 --json title`), scored through the file's own pipeline:

```
max distinct content tokens: 25   top10: 25,22,22,22,21,21,21,20,20,20   median: 13   count >= 201: 0
```

And the largest prose blobs that actually exist in `.fleet/ledger.md` (`## Ruled` rows, far longer than any `filed` subject): 109, 86, 76 distinct content tokens. The trigger needs ≥ 201 **simultaneously on both sides** — 8x the longest real issue title. For any realistic pair (`min <= 25`) a single shared word scores ≥ 0.04, 8x above the floor, so the rounding is a no-op there.

## The suggested fix works but is not clean

Applied verbatim to `:299-301`:

```js
.map((row) => ({ row, raw: overlap(scored, scoreTokens(subjectOf(row))) }))
.filter((n) => n.raw > 0)
.map((n) => ({ row: n.row, score: round2(n.raw) }))
.sort((a, b) => b.score - a.score);
```

- Suite: `tests 921 / pass 921 / fail 0`, identical to baseline (measured pre-#635-review-commit).
- Differential over 65 cases (the file's 5 real fixtures + 60 randomised ledgers built from the 400 real titles), full stderr + stdout + exit code: **no difference**.
- It does fix the probe case: 251/251 one shared word → `nearTotal 1`.

**But** the row then surfaces at a display score of `0.00`, so the output reads `near-miss 0.00 — …` and `1 further near-miss(es) not shown — highest withheld 0.00; the cap dropped them, not the score`. It trades the comment's imprecision for a two-decimal display that cannot represent what it now admits. If this is ever applied, the honest version is a wider display precision or a `<0.01` rendering, not just the raw filter.

## Recommendation

Either close, or take the cheap half: reword `ledger.mjs:293-296` so it stops promising "not a score threshold" when `round2` makes it one below 0.005. That is a one-line comment fix with no output change and no display-precision problem.

feigi:
## What

`ledger.mjs`'s near-miss comment claimed the floor is "shares at least one content word, **not a score threshold**". `rankedNear`'s filter tests the **rounded** score, which makes it exactly a score threshold. Comment corrected; no code touched.

## Measured, not inherited

Driven through the real `check` CLI against a two-row fixture — N distinct content tokens on both sides, exactly one shared, so raw overlap is `1/N`:

| raw overlap | `near` | `nearTotal` | stderr |
|---|---|---|---|
| `0.005` | `[{score: 0.01}]` | 1 | `near-miss 0.01 — …` |
| `0.004975…` | `[]` | 0 | *nothing* — no near-miss line, no further-near-miss line |

So the effective floor is `round2(overlap) > 0`: a raw overlap below `0.005` rounds to `0.00` and the row leaves both the list and the total, unreported. Confirmed against `round2` directly — `0.005` → `0.01` (kept), `0.0049999999` → `0` (dropped).

The `0.005` figure is a property of `round2`'s two decimals, not of any sampled data, which is why the comment names it and frames it as an artifact rather than a tunable knob.

I also re-measured the alternative triage rejected: filtering on the raw score admits that row at a display score of `0.00`. The comment now says so, so the next reader does not re-litigate it as a free fix.

## No behaviour change

- Diff is comment-only — every changed line starts with `//` (checked mechanically).
- Stdout, stderr and exit codes **byte-identical** to `origin/main` across 8 invocations (11420 B stdout, 7089 B stderr on both sides), comparing against a copy of `origin/main:skills/fleet/scripts/ledger.mjs`.
- Suite: `tests 1104 / pass 1104 / fail 0` before **and** after.

Closes #642

https://claude.ai/code/session_01P82HSsEj2q98cFBxBxZsNG


feigi:
## Review adjudication — no-op, HEAD unchanged at `6fc61ae`

Four findings, **zero applied**. All three in-scope findings were dispatched to refuters biased to refuse and all three came back `refuted=true`; each refutation was then re-derived from the tree independently rather than taken on report.

| finding | dimension | verdict | outcome |
|---|---|---|---|
| `refuted-0` | correctness | refuted 2-0 at review time | not applied, not filed (ruled) |
| `unverified-0` + `unverified-1` | correctness + silent-failure | refuted | deferred -> #845 |
| `unverified-2` | silent-failure | refuted | deferred -> #846 |

`unverified-0` and `unverified-1` were adjudicated as one defect — the same claim from two dimensions — so one refuter and one issue cover the pair.

### Why nothing was applied

**#845** wanted the comment to record that the round-to-`0.00` drop is unreachable. The intended phrasing was a durable property — both sides needing *orders of magnitude* more content words than any real subject. Measured over 876 real subjects (283 `## Filed` rows + 593 tracker titles) through the file's own pipeline, the largest real subject is **52** stems against a floor of **201**: **3.87x**, not orders of magnitude. Of 101,168 real pairs sharing at least one content word, **0** are dropped. The drop is genuinely unreachable today, but no true-and-durable qualifier fits a ~4x tail margin, and the harm it guards against — a reader reaching for the raw-score "free fix" — is already foreclosed verbatim at `ledger.mjs:341-343`.

**#846** wanted the comment to name `contentWords`'s `>= 3` floor as a second silent drop. A pair sharing only `CI` shares **zero** content words (`contentWords("CI gate misreports") = ['gate','misreports']`), so the comment's sentence — scoped to a row that *does* share a content word being dropped anyway — makes no claim about it. The scoring path's subjection to that floor is already stated at its definition site, `ledger.mjs:311-316`, and pinned at `ledger.test.mjs:279` and `:754`.

### Verification

- Suite: `tests 1104 / pass 1104 / fail 0` — full suite, run from the worktree via `node --test skills/fleet/scripts/*.test.mjs`. No edit was made, so there is no separate before/after.
- Diff re-confirmed comment-only, mechanically: **0** non-comment lines added, **0** non-comment lines removed (11 comment adds, 4 comment dels).
- PR body re-read against the final diff — every claim in it still holds, so no reconciliation was needed.
- Every measurement carried a **positive control** proving the probe could see a matching row, and printed `ledger.ok` — the review's own finder went wrong by probing outside a git repo, where every check returns `nearTotal 0` regardless of input.

PR #849, opened 2026-08-23
docs(fleet): the verify-sha guard named one outcome and left exit 2 unhandled

### LINKED ISSUE #558: run-team/SKILL.md: the "Verify every reported SHA" guard names no exit-2 branch — all five of verify-sha.sh unanswerable paths are undocumented on the consuming side

feigi:
Closes #558

## What was wrong

`verify-sha.sh` answers "is this SHA on `origin/<branch>`" with exit 0 or 1, and says "I could not tell" with exit 2. The only consuming-side rule — the **Verify every reported SHA** guard in `run-team/SKILL.md` — named just the reachable/not-reachable half:

> Not reachable → flag, do not enqueue, do not return the ticket to the pool until the maintainer rules.

In `sh`, both `cmd || handle` and `if cmd; then` treat 1 and 2 alike, so the natural reading files an unanswerable probe under "not reachable" by accident rather than by rule.

That accident is fail-safe in **one** direction only. Exit 2 emits no JSON, so a controller cannot read `reachable:true` off it and cannot wrongly enqueue. In the other direction it is not fail-safe: an unanswerable probe recorded as a real verdict flags a member for a stray commit that was never demonstrated, and holds its ticket out of the pool "until the maintainer rules" over a failure a re-run would have cleared.

## The fix

One rule keyed on the **exit code**, not on a list of causes, plus the exit-code annotation `run-team` already uses for `inflight.sh`. Every `die()` in the script reaches the caller as the identical shape — no stdout, cause on stderr — so a cause list adds nothing a controller could act on differently, and it rots on the next commit that adds or removes one.

The prose deliberately states no count, and this ticket's own history is why. `verify-sha.sh`'s exit-2 set has already grown twice — the `merge-base --is-ancestor` guard, then #119's JSON-escaping guards — and #558 was filed **before the last of those landed**. Measured: the issue was created `2026-08-17T20:44:06Z`, against five `die()` sites, which is exactly what the tree held at `72a8a6f^`; `72a8a6f` took it to eight the next morning at `2026-08-18T11:12:44Z`. The ticket's count was accurate when written and stale by the time this fix was. A number written into the guard would age the same way.

*(An earlier revision of this PR asserted the reverse — that the issue was written against an already-stale count. That was wrong in the direction that matters, and the header comment carrying the same claim is corrected in this diff.)*

## Verification

`skills/fleet/scripts/verify-sha-exit2-prose.test.mjs` — five pins over a slice bounded to the guard itself, mutation-verified **both** directions. The matrix stands at **33 distinct cases, every one as expected**:

- **17 defect mutants, each redding exactly the pin that owns it.** The overshoot is in there in seven spellings across both directions — exit 1 collapsed into exit 2's remedy and the reverse, each cased, lowercased and bold-split — because that is the false-positive class this change could itself introduce. A guard that teaches "non-zero is unanswerable" destroys the distinction the script exists to make, so each branch is pinned as an **adjacency** with the other exit code excluded from the gap. Also covered: `is not a verdict` deleted outright, the negation flipped, the re-run remedy deleted, the maintainer's ruling deleted, the flag verb swapped, a written tally in four phrasings, and a genuine second `exit 2` code path in the script.
- **16 benign edits, all green:** the guard reflowed to a single line, ten in-place bold and italic moves across both pinned phrases, a comment in `verify-sha.sh` that merely mentions exit 2, an `exit 22`, two non-tally sentences that use a counted noun near an exit code, and a `ways?`-widened tally pin measured against the guard's unmodified text.

Three pins were corrected by that matrix rather than shipped as written:

- **The gap exclusions were case-sensitive.** `(?!exit\s+2)` did not see `Exit 2`, which is what sentence-initial prose writes — so the exact overshoot the file exists to forbid stayed green. Measured with a byte-identical control: `Exit 2 counts as not reachable too.` passed 5/5, `An exit 2 counts as not reachable too.` reddened. Now `[Ee]xit`, matching every other literal in the file.
- **The bold tolerance covered one seam per pin, not the ceiling the file claims.** The header promises a moved `**bold**` stays green; in fact `Exit 2 is **not a verdict**` and `Not reachable **(exit 1)**` — both ordinary editing — reddened. `**` is now allowed at any word boundary inside a pinned phrase, including inside the gap lookaheads, where `**Exit** 2` had been drifting through the exclusion undetected.
- **The tally pin's noun list missed `cases`**, measured: `three unanswerable cases` passed where `three unanswerable failure modes` reddened. The regex is hoisted to a `TALLY` const so the green case is pinned against the same object — the guard's own required `2 unanswerable` annotation is asserted non-matching, so a future widening that reaches it reds with an honest message instead of claiming a count was written into the guard. `ways?` was measured and left out: it reds `Treat exit 2 the same way in all cases`, which is prose. The residual is #856.

One pin is cross-file: `verify-sha.sh`'s header still declares the codes the guard quotes, and `die()` is still the sole way it exits 2. That is what lets one rule cover every cause; a second exit-2 spelling would need the guard revisited. That count now strips whole comment lines first — the script is two-thirds prose about its own exit codes, and counting the raw substring reddened on a comment that merely *mentions* exit 2, with a message sending the reader after a code path that does not exist.

## Deliberately left

- **#267**, the same verdict-collapse class in the sibling `prove-merge.sh` — its own ticket.
- **#855** — `docs/specs/2026-07-23-fleet-plugin-design.md`'s exit-2 cause rows. An earlier revision of this PR claimed that table "already documents every exit-2 cause, accurately and without a count". The first half does not hold: `json.sh missing or unreadable` is one clause for two distinct `die()` sites (`verify-sha.sh:20` and `:23`), and the same collapse is in all seven script rows. Out of this PR's diff, so filed rather than fixed. The second half stands — the table does falsify the issue's claim that `run-team/SKILL.md` is the only place naming `verify-sha`.
- **#856** — the tally pin still misses a `ways`-phrased count, and the obvious widening reds correct prose. The remedy needs a different axis, not more nouns.


feigi:
**Review findings applied and deferred** — seven findings, pushed at `b766d4d`. Full suite `1109/1109`.

### Applied (5)

| finding | dimension | outcome |
|---|---|---|
| `survived[0]` — case-sensitive gap exclusions | correctness | applied; mutants m1/m3-cap now red, lowercase controls still red |
| `unverified[0]` — raw `exit 2` substring count | correctness | claim (a) applied, comment lines stripped first; claim (b) **refuted** — see below |
| `unverified[1]` + `refuted[0]` — tally noun list | tests / correctness | narrowest measured widening applied (`cases?`); residual → #856 |
| `unverified[2]` — bold tolerance narrower than the declared ceiling | tests | applied, plus a bold-split hole in the lookaheads the finding did not name |
| `unverified[3]` — the stale-count clause in the header comment | comments | corrected; the comment inverted the actual sequence |

### Deferred and filed (2)

- **#855** `needs-triage` — dimension `comments`. `docs/specs/2026-07-23-fleet-plugin-design.md`'s `json.sh missing or unreadable` is one clause for two distinct `die()` sites, and re-measurement shows it is all **seven** script rows, not just `verify-sha.sh`'s. Out of this PR's diff. Records the disagreement with impl-558's assessment so it is not re-litigated.
- **#856** `ready-for-agent` — dimension `tests`. The tally pin still misses a `ways`-phrased count; the obvious widening was measured to red correct prose, so the remedy needs a different axis.

### Refutations worth recording

- **`unverified[0]` claim (b) — refuted.** The `TypeError` on `.length` of a null match is unreachable: the `assert.match(SCRIPT, /^die\(\)\s*\{[^}]*exit 2;?\s*\}/m)` three lines above carries the same `exit 2` literal, so anything satisfying it guarantees a non-null match. Brute-forced over 813,615 candidate strings built to break that implication — zero violations. No `?? []` guard added; it would be dead code.
- **`refuted[0]`'s refuting vote — refuted.** It held that widening the noun list must red the guard's own required annotation. Measured noun-by-noun: none of the eight proposed nouns matches anywhere in the guard slice. The two that *do* red correct text are `reachable` and `unanswerable` — the annotation's own adjectives — and those are now pinned against, so the boundary is enforced rather than remembered.
- **`ways?` — measured and rejected.** It reds `Treat exit 2 the same way in all cases`, which is prose, not a tally.

PR #850, opened 2026-08-23
docs(specs): the candidate-scan step named a flag candidates.mjs refuses

### LINKED ISSUE #240: docs/specs/2026-07-22-run-team-agent-fleet-design.md:201 names `--label ready-for-agent` — a flag candidates.mjs now refuses, and the string is hand-copied across three files

feigi:
Closes #240.

## What the ticket asked, and what measuring changed

The ticket prescribed swapping `--label ready-for-agent` for `--require-label ready-for-agent` at the run-team design spec's Phase 0 candidate-scan step. Applying that literally would have minted a false claim, so the remedy shape changed — the correction itself is what the ticket asked for.

**Measured:** at `02bfd59`, the commit that added this spec, the `next-ticket` step 1 that the spec *cites* was a raw `gh issue list`, and its text read "Add `--label ready-for-agent` first". So `--label` was **correct the day it was written** — the same `gh` flag the ticket warns not to sweep. It became wrong at `c3310fe` the next day, when that query moved into `candidates.mjs`, which has read `arg("require-label")` in every commit of its life.

That makes it the exception the dated-spec rule carves out — a behavioural claim the code contradicts, not wording that merely aged — and that rule asks for **annotation, never a silent rewrite**. So this adds a supersession note in the voice of the notes already at the head of that document, recording that the scan became a script call and why the flag changed, and then corrects the step in place, which the packaging note above it already licenses ("Paths and command names below have been updated in place").

Settling commands run against the tree, not quoted from the issue:

| Claim | Result |
|---|---|
| `--label` refused | exit 2, `candidates: Unknown option '--label' — accepted: --require-label, --allow-fallback, --limit` |
| the refusal is about the flag NAME, not blanket | `--require-label ready-for-agent` passes the name guard into the real query |
| the accepted spelling has never changed | `arg("require-label")` at every commit touching `candidates.mjs` |
| the plugin spec specifies it | its `candidates.mjs` contract row, `[--require-label L]` |

The reflow of the corrected line is whitespace only, asserted word-identical before it was written.

## The pinning question the ticket left open

**Pinned:** the corrected step, in `candidates.test.mjs` beside the existing `run-team/SKILL.md` pin. The flag is **derived from `candidates.mjs`'s own `OPTIONS` declaration**, not restated, so a rename reddens instead of minting the next stale copy. The derivation is a guarded helper called from the test body — see *Review fixes applied* below for why it is not a module-scope expression. Scoped to the step rather than the file, for the reason the neighbouring pin gives and one more: the new supersession note cites `--label` deliberately, as the spelling the script refuses, so a file-wide negative would fail on the correction's own prose.

**Also pinned — the half that guards against a fix for this class:** the cockpit design spec's `gh issue list --label ready-for-agent`, where `--label` is gh's own flag and correct. That is the input the check must ACCEPT, and it now reddens if a sweep rewrites it.

**Left unpinned, deliberately:** the `--require-label` mentions in the trust-the-label design spec. Correct prose in a dated record that self-disclaims its citations; a pin there would assert a historical document must track the code, firing on a future *correct* rename to demand history be rewritten.

## Sites examined and why each was or was not changed

The ticket's table was measured at `e840bc5` and the tree has moved, so the class was re-derived against `origin/main`.

| Site | Verdict |
|---|---|
| run-team design spec, Phase 0 candidate-scan step | **changed** — the defect |
| cockpit design spec, `gh issue list --label ready-for-agent` | left — gh's flag, correct; now pinned |
| `board.mjs`'s `gh issue list` call | left — **live code**, gh's flag, correct. Not in the ticket's table; a blind sweep would have broken production, not just prose |
| trust-the-label design spec, both `--require-label` mentions | left — correct, dated record |
| `next-ticket/SKILL.md` | left — correct, live, unpinned (see report) |
| `run-team/SKILL.md` phase 0 rule | left — correct, already pinned |
| `candidates.mjs`, `arg.mjs`, `candidates.test.mjs` comments citing `--label` | left — deliberate citations of the refused spelling |
| `candidates.test.mjs` cases feeding `--label` to the script | left — they assert the exit-2 refusal |
| `issue-tracker.md`, `review-and-fix.md`, `run-merge-bot.md` | left — `gh` flags, correct |
| `docs/fleet-rule-manifest.txt` | **no longer exists** — retired at `7d1cd9f`; the ticket's table row is stale |

## Review fixes applied

The review's one `survived` finding was applied in full, both halves.

The `OPTIONS` derivation originally ran at module scope and indexed two unguarded regex matches.

- **First index, fails loudly.** The regex needs a literal `\n};` terminator, so reformatting the `OPTIONS` literal — zero behavioural effect — threw at import: `TypeError: Cannot read properties of null (reading '0')`, `tests 1 / pass 0 / fail 1`, against a 62/62/0 baseline for the file. Node names the file and line, so a diagnostic exists; what was missing is a *test-authored* one naming `OPTIONS`. Now guarded and called from the test body: the same mutation gives `tests 62 / pass 61 / fail 1`, one named red, `candidates.mjs' OPTIONS block no longer matches — update this test`.
- **Second index, fails silently — the more dangerous half.** It took the FIRST `*label` key. With `"exclude-label"` declared above `"require-label"` and the spec step rewritten to name `--exclude-label` (a sentence the script refuses at exit 2), the old form went **green at 62/62/0 while pinning the wrong flag**; the new form reddens with `candidates.mjs' OPTIONS declares 2 \`*label\` flags, not the 1 this test derives`. The match is now global and asserted unique rather than anchored to the literal `require-label`, which would have re-introduced the restated string this PR removes.

Mutation coverage is unchanged — renaming the declared option still reddens the spec test by name — and a change it must not catch (a comment line above `OPTIONS`) stays green at 62/62/0.

The review's one `suggestion` was checked with a refuter and **not** applied: deriving the cockpit-spec pin's flag from `board.mjs`' `gh issue list` argv is strictly worse than the literal, because an indiscriminate `--label` → `--require-label` sweep rewrites `board.mjs` and the spec together, and the derived pin then compares the swept spec against the swept authority and goes green where the literal reddens. Filed as #857 with the measured matrix; the separate placement objection is #858.

## Tests

`./agent-test skills/fleet/scripts` — **1106 pass, 0 fail, 0 skipped**, and the log was grepped for both new test names to prove the run actually saw them.

After the review fixes, re-run from the worktree: `node --test skills/fleet/scripts/*.test.mjs` → **tests 1106, pass 1106, fail 0**, and `candidates.test.mjs` alone → **tests 62, pass 62, fail 0** (the per-file count matters here: the finding above was about that file silently collapsing to 1).

The spec pin was run against the **uncorrected** spec first and failed on the flag assertion, so it is not green before the fix. Mutation-checked: renaming the declared option to `required-label` fails it by name against the corrected spec, proving the derivation binds rather than restates.


feigi:
Review findings adjudicated. Applied: `survived[0]` in full, both halves (`e2f5672`). Deferred and filed:

- **#857** — `needs-triage` — derive the cockpit-spec `--label` pin from `board.mjs` instead of hardcoding it. Given one refuter, biased to refuse; **refuted** with a mutation matrix. The disqualifying row: an indiscriminate `--label` → `--require-label` sweep rewrites `board.mjs:395` and the cockpit spec together, so the derived pin compares the swept spec against the swept authority and goes green (62/62/0) exactly where the literal reddens (61/1). `board.mjs` is a peer consumer of gh, not an authority over gh's flag names. dimension: correctness.
- **#858** — `needs-triage` — the placement half of the same finding: both new pins sit in `candidates.test.mjs`, where a red reads as "candidates.mjs broke". The objection is sound but the destination the finding names (beside `board.mjs`' tests) is refuted by the same evidence, and no convention settles where a pin derived from a script's declaration belongs. Deferred rather than moved blind. dimension: correctness.

Not deferred: the finding's second remedy branch, "drop the test as speculative" — refused on the merits. The cockpit pin is the measured accept-side control for this sweep class.

Suite from the worktree at `e2f5672`: `node --test skills/fleet/scripts/*.test.mjs` → tests 1106, pass 1106, fail 0; `candidates.test.mjs` alone → tests 62, pass 62, fail 0.

Issue #539, opened 2026-08-17
review-pr.js:622 — a snapshot agent that DIED reports as 'returned no tree', naming a missing tree when only the schema handshake failed


feigi:
Deferred from PR #536 review. **dimension: `silent-failure`** · severity: `suggestion` · `workflows/review-pr.js:622`

## The claim

`snapshotMissing`'s first branch collapses two different failures into one message:

```js
if (!snap || !snap.path || !snap.head) return "the snapshot agent returned no tree — nothing to review";
```

`workflows/review-pr.js:50-53` records that structured-output exhaustion "emits `Failed to provide valid structured output after <n> attempts` and `agent()` then returns null". So a snapshot agent that **died, or whose structured output was rejected**, reports as *"returned no tree"* — which reads as *the tree is missing* when the tree on disk may be perfectly good and only the schema handshake failed. The reader is pointed at `git archive` when the fault is the agent contract.

The finding notes #536 widens that surface by adding a third required field (`pathVerified`) to the same schema: one more way for the handshake to fail into a message about a missing tree.

## Suggested remedy (unverified)

Split the falsy-`snap` branch from the missing-field branch, e.g.

> "the snapshot agent returned nothing (it died, or its structured output was rejected) — no tree to review"

## Why deferred rather than applied

**Out of #140's scope.** #140 is "refuse a snapshot whose path was never confirmed to exist" — the `pathVerified` branch. This is the *pre-existing* falsy/missing-field branch, which `review-pr-snapshot-path.test.mjs:48-49` explicitly marks as behaviour #140 must not regress. Rewording it is a separate concern.

**And the approach is open, which is why this is `needs-triage` rather than `ready-for-agent`.** The branch currently covers three distinct inputs — `snap` falsy (agent died / schema rejected), `snap.path` absent, `snap.head` absent — and the finding only proposes splitting the first off. Whether the other two deserve their own wording, and whether `review-pr-snapshot-path.test.mjs:50-56`'s six-input loop (which pins `/no tree/` across all of them) should then be split too, is a maintainer call. That test would need updating either way.

The claim itself is directly readable in the source and I did not dispute it; the finding's evidence was reasoned from source rather than run against a live agent, which is why it landed at `suggestion` with zero refuters — the review budgets that band none.

feigi:
[...]

PR #852, opened 2026-08-23
docs(review-pr): record the measured node:fs verdict, and close #538

### LINKED ISSUE #538: review-pr.js: does the Workflow sandbox permit node:fs? If yes, the whole pathVerified round-trip is unnecessary machinery
Deferred from PR #536 review. **dimension: `correctness`** · severity: `suggestion` · `workflows/review-pr.js:617`

## The question

`pathVerified` is a four-part round-trip: a prompt paragraph, a shell line, a schema field, and a `required` entry. All of it exists because the caller supposedly cannot touch the filesystem. **Nobody has executed a test to confirm that.**

If a Workflow script may `import`/`require` `node:fs`, the entire round-trip can be replaced by a real caller-side check inside `snapshotMissing`:

```js
// instead of reading a boolean the agent typed
try { if (!readdirSync(snap.path).length) return `the snapshot at ${snap.path} is empty …`; }
catch { return `the snapshot at ${snap.path} does not exist …`; }
```

That version is genuinely untrustable by the agent, needs no prompt text to keep in sync, and would let four pins added in #536 be deleted.

## Why it is not just assumed

The repo documents the sandbox as having no filesystem/Node API access in three places, and #140's implementation relied on that documentation:

- `skills/fleet/scripts/review-pr-reads.test.mjs:7-13` — "extraction to a module would need `import` to resolve inside the Workflow sandbox ('no filesystem or Node.js API access'), which nothing in `workflows/` does, and a failed import bricks the fleet's DEFAULT review path"
- `grep -n '^import|require(' workflows/review-pr.js` returns nothing
- `grep -rn 'node:fs' workflows/` returns nothing

Three documents agreeing is not a measurement. The finding's own evidence concedes it: *"I could NOT confirm whether a workflow file may import node:fs … This finding is therefore reasoned from the source, not executed."*

## What to do

**Execute it.** Add a throwaway `node:fs` import (or `require`) to a scratch copy of a workflow and run it through the harness. Then:

- **Imports work** → replace the `pathVerified` round-trip with the `readdirSync` check, and delete the prompt paragraph, the shell probe, the schema field, the `required` entry and the four pins #536 added for them. Note the ordering fix #536 landed (probe above the `ln -s` line) becomes moot, since `readdirSync` would run in the caller after the agent returns — but it would then need to exclude the `node_modules` symlink itself, which is the same trap in a new place.
- **Imports do not work** → close this, and record the measurement somewhere durable so the next reviewer does not re-derive it. The current machinery stays as-is.

## Explicitly out of scope — do NOT re-file

The finding's *other* branch — softening the comments at `workflows/review-pr.js:578` and `:618-622` to say the agent reports a mechanical check rather than that the caller performs one — **is refuted 2-0, twice**, at `:578` (`types`) and `:618` (`silent-failure`). Both refuter pairs RAN the probe and found both comment blocks already self-scope in their trailing clauses: "`snapshotMissing` below is what turns a false report into a refusal" and "the field cannot be silently omitted, only reported false". The comments do not overstate. That half is settled; this issue is only about the `node:fs` question.

feigi:
Closes #538

#538 asked whether a Workflow script may `import`/`require` `node:fs` — and pointed out that the repo answered it with three documents that agree with each other and no execution. This runs the experiment and records the result where the next reader will stand.

## The measurement

Run against the Workflow harness, not plain `node`. A dispatched subagent has no `Workflow` tool, so the probes were driven through a `claude -p` session, which does; probe workflows lived in a scratch project's `.claude/workflows/`, so nothing was written to the checkout.

| mechanism | result |
|---|---|
| `import { readdirSync } from "node:fs"` | `SyntaxError: Unexpected token '{'. import call expects one or two arguments.` |
| `await import("node:fs")` | `SyntaxError: import() is not available in workflow scripts.` |
| `require("node:fs")` | `Error: require is not defined` |

A workflow body compiles as a **function body** inside the harness VM — which is also why the existing top-level `return` is legal — so a static `import` is parsed as the dynamic call form and never reaches a module loader.

Enumerating the scope: ECMAScript builtins plus the harness injections (`agent`, `args`, `budget`, `log`, `parallel`, `phase`, `pipeline`, `workflow`, `console`, timers). `process`, `Bun`, `require`, `module` and `fetch` are all `undefined`, and a filter for filesystem-capable names comes back empty.

**Verdict: imports do not work.** The documented claim was correct; it is now measured rather than asserted. So this takes #538's second branch — the `pathVerified` round-trip stays exactly as it is, and nothing about it is deleted.

## Controls

A probe that errors for an unrelated reason is indistinguishable from a sandbox refusal, so both verdicts were made observable before either was trusted.

- **Positive** — a workflow importing nothing returned `{"marker":"CONTROL_OK_RETURNED","value":42}`, the `42` computed from `41+1`, so the harness demonstrably executes rather than echoes.
- **Negative** — a deliberate `throw` surfaced as a distinct error with a stack, so a genuine failure is visible and attributable.
- **Mechanism vs. resolution** — `import()` of a module that does not exist returns the *same* `import() is not available` error, at the same stage. The block is at syntax level, before the specifier is resolved, so the refusal cannot be blamed on `node:fs` in particular.
- **Confound caught** — a first attempt read the static-import file's absence from the registry as the import being rejected. A twin with the import removed was *also* absent, exposing an unrelated discovery rule: a statement placed before `export const meta` drops the file from the registry silently. Re-running with `meta` first isolated the import and produced the syntax error above. Without that control this PR would have recorded the right verdict with the wrong mechanism.

## What changed

Comment prose in three places, and the pins over it. `workflows/review-pr.js` gains no executable line.

- `workflows/review-pr.js` — the verdict and its controls recorded beside `snapshotMissing`, which is where a reader who notices that `pathVerified` is a boolean the agent types will be standing when they reach for `readdirSync`, exactly as #538 did.
- `skills/fleet/scripts/review-pr-reads.test.mjs` — the lift rationale cited the sandbox's documented "no filesystem or Node.js API access" as its reason. That quote is one of the three agreeing documents #538 was filed against, so it can no longer be the citation for the claim #538 settled; it now points at the measurement.
- `skills/fleet/scripts/select-dimensions.test.mjs` — the same citation, at the same superseded vintage, in the file `review-pr-reads.test.mjs` names as the same-reason precedent. Now a pointer to the measured record rather than a second restatement of it. Rewrapped to an unchanged line count, because `review-pr-reads.test.mjs:10` cites `select-dimensions.test.mjs:23-40`.
- `skills/fleet/scripts/review-pr-snapshot-path.test.mjs` — five pins over the record, one generated test each, in the sibling idiom at `review-pr-citation-prose.test.mjs:48-63`.

## The pins, and what they are worth

Three pins hold the **evidence** — that the verdict was measured rather than read off the docs, that the refusal is of the mechanism and not `node:fs`, and that `require` was covered too. Two hold the **verdict** itself, at both places the record states it. Evidence without verdict is not a pin on the conclusion: at the three-pin vintage, inverting `It cannot be dropped.` to `It can be dropped.`, `So the round-trip stays,` to `goes,`, and softening to `It could arguably be dropped, but … for symmetry …` each left the file 8 pass / 0 fail — the record telling the next reader the opposite of what was measured.

Two pin shapes were measured against that battery rather than argued: a contiguous evidence-to-verdict span with a negative lookahead, and plain independent phrases. Identical on every acceptance row; the contiguous form then falsely reds on a legitimate long sentence inserted between evidence and verdict, and on relocating the verdict clause. The plain form ships.

What the pins do **not** do is stated where they live: a pin is a substring test, so prose *contradicting* a fragment, added around it, passes all five — verified on this tree. No positive assertion reaches that, so nothing here claims to. For the same reason the failure messages name the phrase that stopped matching rather than the misreading they cannot actually detect, and `prose-pin.mjs` is untouched: case- and backtick-sensitivity is the repo's prose-pin idiom, and loosening it for this file would weaken every other pin in the suite.

## Safety on the default review path

`workflows/review-pr.js` is the fleet's default review path, so a failed import there would brick it. This PR adds no import and no executable change to it: `workflows/review-pr.js` comment-stripped and blank-filtered is byte-identical to `origin/main`, and its half of the diff is zero non-comment lines. Everything executable in this PR is test code. If the sandbox verdict is ever wrong, the only thing wrong is the prose — nothing breaks, because nothing executable moved.

Suite: 1109 pass, 0 fail.


feigi:
Review findings applied at `b9e8acd`. One deferral, and it deduped rather than filing:

- **`snapshotMissing`'s first branch collapses distinct causes into one message** (`silent-failure` · `workflows/review-pr.js:713`) — deferred as pre-existing: this PR is comment-only in `review-pr.js`, so the branch logic is untouched by it. `ledger.mjs check` exited **3** on an existing tracker row, so this went to **#539** as a comment, not a second issue — #539 is the same subject and already names the three-input split as its open maintainer question. The comment adds that a second reviewer independently proposes the full split, and corrects two stale details in the issue (line `:622` → `:713`; "four causes" → three fields).

Settled without action, recorded here so it is not re-filed:

- **"the harness claims cannot be verified from this repository alone"** (`comments` · `review-pr.js:692`) — right that the tree cannot verify it, wrong that it is unverified. The measurement ran *inside* the real harness via a nested `claude -p`, with a positive control (computed marker), a negative control (deliberate `throw`), and a mechanism-vs-resolution control (`import()` of a nonexistent module returning the identical error at the same stage). The suggested "second pass by anyone who can run inside the actual Workflow harness" describes what was already done.
- **prose-pin `phrase()` case/backtick sensitivity** (`tests` / `silent-failure` · `:224`) — not loosened. It is the repo's prose-pin idiom across the suite; a local override would diverge this one file for no measured defect. The half of that finding that *was* actionable — failure messages asserting a semantic loss a pin cannot detect — is fixed in this commit.

PR #866, opened 2026-08-24
docs(fleet): inflight.sh's probe 3 cited release-ticket.sh by a line that moved

### LINKED ISSUE #800: inflight.sh:436 cites release-ticket.sh:73 for the worktree-registry read — line 73 is the json_lib sourcing guard; the check is at :141
Deferred from PR #788 review (issue #243). Dimension: **comments**. Severity: **suggestion**.

**Pre-existing, and not caused by PR #788.** Surfaced only because that review audited every `release-ticket\.sh:[0-9]+` cross-reference in the tree. Filed rather than fixed there, because #243's PR has no business editing `inflight.sh`.

## The finding

`skills/fleet/scripts/inflight.sh:436` cites `release-ticket.sh:73` for "its own worktree-registry read (#84)".

Line 73 of `release-ticket.sh` is the `[ -r "$json_lib" ] || die ...` sourcing guard — not a worktree-registry readability check. The check actually being referred to is `[ -r "$wtroot" ] && [ -x "$wtroot" ] || die ...` at **:141**.

Both lines sit well above PR #788's first hunk (`@@ -214,...`), so that PR neither introduced nor shifted this mismatch.

## Remedy

Two options, and the second is what this repo's own convention now asks for:

1. `73` → `141`.
2. Better — name the target by construct rather than by line, e.g. "the `-r`/`-x` guard on `$wtroot` in `release-ticket.sh`". PR #788 had to fix a *different* cross-file citation for exactly this reason: its own +14 lines moved the content out from under `claim-ticket.sh:150`'s `release-ticket.sh:766`. A positional citation corrected today breaks again on the next insertion, and `release-ticket.test.mjs`'s own comment already states the rule: "The sites are named by construct throughout, never by line: they have moved every time this file was touched."

Prefer (2). A sweep for other `release-ticket\.sh:[0-9]+`-style citations while in there would be in keeping — this is the second stale one found in one review.

Deferred from PR #788 review

feigi:
Closes #800

`inflight.sh`'s probe 3 borrows its establish-absence rule from
`release-ticket.sh`'s worktree-registry read and cited that check as
`release-ticket.sh:73`. Re-derived against `origin/main`: line 73 is inside the
`--apply` argument-validation prose. The check is the `-r`/`-x` guard on
`$wtroot`.

Remedy is the ticket's preferred option — name the target by **construct**, not
by line. `release-ticket.test.mjs` already states the rule this repo works to:
"The sites are named by construct throughout, never by line: they have moved
every time this file was touched."

The comment keeps its line count, so nothing below probe 3 shifts.

## The ticket's own claim, re-derived

The ticket says line 73 is the `[ -r "$json_lib" ]` sourcing guard. Against
`origin/main` that guard is **not** on 73 — the number had already drifted
again between the ticket being filed and this branch. The half that settles the
fix holds and was measured: the guard the comment means is the `-r`/`-x` pair on
`$wtroot`, and it is the only such pair in the file.

## Why a test came with it

Nothing read `inflight.sh` as text, so the drift could not redden a run — it
surfaced only when a review audited line-numbered references into
`release-ticket.sh` across the tree.

`inflight-citation-prose.test.mjs` pins probe 3's comment slice with a **pair**
of assertions, because neither holds alone: the positive pin stays green beside
a stray line citation, and the negative pin stays green with the whole clause
deleted. Both were red before the fix.

Mutation-tested in both directions, on the tree:

| Mutation | Expected | Result |
|---|---|---|
| baseline | green | green |
| line citation added **outside** the slice (probe 2's comment) | green — no over-refusal | green |
| bare, un-numbered `release-ticket.sh` added **inside** the slice | green — the accept case | green |
| construct reworded away, no line number added | positive pin reds | 1 fail |
| clause deleted entirely | positive pin reds | 1 fail |

The negative pin forbids a citation *form* (path, colon, digit), not a
vocabulary token, so a legitimate un-numbered mention does not trip it — pinned
above rather than assumed.

## In-family sweep

A strip-and-join sweep (not a line-anchored grep — a citation can wrap across
two comment lines and hide from one) found exactly one other family member:
`reaping-prose.test.mjs` cites `release-ticket.sh:15-25` at two sites.

Measured: that range is **still correct** — those lines are the script's header,
and the wording they are cited for sits on the last line of the range. So this
is fragility removal, not a second correction: the next insertion anywhere in
that block moves the target out from under the number while the prose still
reads as precise. Both sites already name the construct ("the script's header");
dropping the range is a strict deletion that changes no claim.

`claim-ticket.sh` was checked too — PR #788's re-anchoring held; it names
`release-ticket.sh`'s sites by construct and carries no line numbers.

**Deliberately left:** `docs/metrics/tier-outcomes.tsv` cites
`release-ticket.sh:757` and `:766`. That is an append-only record of what was
true during those runs; re-anchoring it would falsify the history. The wider
`<file>:NNN` citation population in `docs/specs/*.md` is also left — out of the
ticket's stated family, and that directory carries its own convention marker
with rows explicitly exempted.

After this, no line-numbered citation into `release-ticket.sh` remains outside
`docs/metrics/`.

## Review fixes

Applied after the six-dimension review, in `42fa84e`.

**The slice-bounds comment minted a false cross-file claim.** It said `inflight.sh`
names `release-ticket.sh` elsewhere "and an unbounded slice would let one of
those satisfy the positive pin, or trip the negative one". "Those" points at the
sibling mentions the previous clause just named, which makes it a present-tense
claim about them — and it is false of every one. Measured: applying the real
`phrase()` to the whole flattened file with probe 3's naming line removed gives
`false`, the control with it gives `true`, and `grep -nE 'release-ticket\.sh:[0-9]'
inflight.sh` exits 1. On a ticket whose whole subject is removing an untrue
cross-file claim, that is the defect the ticket exists to prevent, minted in the
fix's own prose.

Reworded to the property rather than the hazard: the siblings satisfy neither
pin today, and the bounds are what keep a future one from doing so. No count is
written into the prose — a count is false on the next commit that adds or removes
a mention. The structural fact the bounds rest on is measured too: the slice runs
from probe 3's heading to `probe_local() {`, and the other mentions sit on both
sides of it, which is why both ends are bound rather than one.

**Dropped the `REPO` round-trip.** The constant walked three directories up and
the next line walked the same three back down, only to reach `inflight.sh` — a
sibling of the test file. `join(import.meta.dirname, "inflight.sh")` resolves to
the identical path and is the idiom `inflight.test.mjs` already uses for the same
script.

Both mutation-checked on a scratch copy, never in the worktree: reverting probe
3's comment to the pre-PR `release-ticket.sh:73` reds both tests (pass 0 / fail
2), and a zero-behaviour reflow moving the wrap point inside the pinned phrase
stays green (pass 2 / fail 0). Mutating the copy's own `inflight.sh` is what
reddens it, which is also the proof the shortened path still reads the intended
file.

**Deferred, filed, not applied here:** #870 (the construct citation is unpinned on
the `release-ticket.sh` side — a zero-behaviour `$wtroot` rename leaves the whole
suite green; the remedy is a cross-file pin, which is a different deliverable
from re-anchoring one comment), #871, #872, #873 (three findings a refuter
measured against and refuted, filed as records), and a comment on #516 for the
line-numbered citations in `no-undo-audit.sh` and `printf-die-sweep.test.mjs`,
which that issue already owns.

## Tests

`./agent-test skills/fleet/scripts` — 1127 pass, 0 fail, post-rebase, and
`node --test skills/fleet/scripts/*.test.mjs` re-run after the review fixes —
tests 1127, pass 1127, fail 0, skipped 0. Confirmed
the directory expansion discovers the new file rather than only the targeted
run. `shellcheck` output on `inflight.sh` is byte-identical before and after.

https://claude.ai/code/session_01YBWPegMBtbh9BrZtbENDU4


feigi:
Review deferrals filed:

- **#870** `ready-for-agent` — the construct citation is unpinned on the `release-ticket.sh` side. Dimensions **correctness**, **silent-failure**, **types**, which reached it independently; filed as one issue. A zero-behaviour `$wtroot` rename leaves the full suite green while `inflight.sh` names an identifier that file no longer has, and `inflight.sh`'s own identically-shaped guard makes the rotted citation misdirect rather than dangle. Defect confirmed, remedy open — a cross-file pin is a different deliverable from re-anchoring one comment. Also folds in the `types` finding that this PR's `reaping-prose.test.mjs` citation fix is likewise unpinned.
- **#871** `needs-triage` — the negative pin forbids only the colon spelling (**correctness** + **silent-failure**). Refuted: both proposed widenings miss mutants their own findings list, and no non-colon form exists in the tree.
- **#872** `needs-triage` — `stripHashGutter` duplicated across the two prose tests (**simplify**). Refuted on the seam: hoisting it into `prose-pin.mjs` adds an unpinned third export to the module whose convention is that every export is pinned, and 4 of its 6 mutations already survive green.
- **#873** `needs-triage` — the bare `(release-ticket.sh)` parenthetical in `reaping-prose.test.mjs` (**correctness** + **silent-failure**). Refuted: "the script's header" is already a construct name, which is #800's preferred option, and the proposed quoted-sentence remedy was measured to rot silently.
- **#516** — commented rather than duplicated, for the **types** finding on `no-undo-audit.sh:117`/`:146` and `printf-die-sweep.test.mjs:97`. That issue already owns the remaining-citations sweep.

One `types` finding needed nothing: "no casts, type assertions, or conformance-erasing constructs exist anywhere in this diff" is a null result.

Applied in `42fa84e`: the slice-bounds comment reword and the `REPO` round-trip removal, both described in the body.

Issue #808, opened 2026-08-22
ledger.mjs:244: check's ALREADY-FILED exit is the last process.exit() that truncates its payload on a pipe — mid-branch, so it needs the branch extracted


feigi:
`ledger.mjs:244` — `check`'s ALREADY FILED branch — still does `console.log(JSON.stringify(...))` at `:239` then `process.exit(1)` at `:244`. On a pipe `process.stdout.write` is async and `process.exit()` abandons the queued write, so the payload arrives cut. This is the same mechanism #246 named, and PR #803 removed it from `read`/`row`/`filed`/`ruled` **and** from `check`'s terminal exit at `:568` — this mid-branch exit is the one site left.

## Measured

`filed 9999 <~230 KB subject>` (exit 0), then `check <same subject>` through a real OS pipe → **exit 1, stdout exactly 65536 bytes, `JSON.parse` fails: "Unterminated string in JSON at position 65536"**. Reproduced twice independently against the PR snapshot.

The payload is **ledger-bounded, not argv-bounded**: `match` at `:236` is `data.filed.find(...)`, i.e. a row read straight out of the ledger, and `filed` applies no length cap to the subject it stores. So a four-word `check` against a ledger holding one wide filed row reaches the cut.

## Severity — deliberately stated narrowly

Unlike the sites #803 fixes, the exit code here stays **1**, the strong ALREADY FILED signal, so a caller gating on the status is not misled into filing a duplicate. What is lost is the JSON payload, including the `verdict` field #152 added for parsing consumers. That makes this a corrupt payload under a *correct* code, not #246's corrupt-payload-typed-as-success. It is why this is filed rather than being treated as a #803 blocker.

## Remedy is open

The `:568` fix does not transfer. That exit was the **last statement of its branch**, so it became `process.exitCode = ...` with fall-through. This one sits **mid-branch**: falling through would go on to run the near-miss ranking and the tracker search that this exit exists to skip. Removing it needs `check`'s body extracted into a function so `process.exitCode = 1; return;` can jump correctly — a restructure outside PR #803's scope.

A `check`-on-a-pipe oversized-payload test should land with it; the suite has none (`grep -rn '#246' skills/fleet/scripts/ledger.test.mjs` → 1063/1126/1154/1181 only, and the `for (const cmd of [...])` sweep omits `check` by construction).

**Confirmed defect; remedy open.**

dimension: `types`
Deferred from PR #803 review.

PR #868, opened 2026-08-24
fix(board): an oversized ledger read was killed, not truncated, and served a blank board at 200

### LINKED ISSUE #807: board.mjs:77: tryRun's execFileSync has no maxBuffer, so a ledger past 1 MiB still serves a blind cockpit at HTTP 200 — #246's symptom survives PR #803
`board.mjs:77` `tryRun` calls `execFileSync(cmd, args, { encoding: "utf8" })` with **no `maxBuffer`**, so node's default 1 MiB cap applies to the `ledger.mjs --file <F> read` it runs at `board.mjs:384`. Past that the child is killed with `ENOBUFS`, `tryRun` returns `null`, `tryParse` falls back to `{rows:[],filed:[],ruled:[]}` (`board.mjs:77-78, 384-385`), and the cockpit is served an **empty board at HTTP 200** with only a stderr line — the exact #246 symptom.

**PR #803 does not close this.** It removed `ledger.mjs`'s 64 KiB `console.log`+`process.exit()` pipe cut, which moves the cliff from **65536 B to 1048576 B**; it does not remove it. The consumer-side buffer was never touched.

## Measured

Two independent reproductions against the PR's own snapshot, using `board.mjs:77`'s exact call and options:

- direct: a ledger whose JSON payload exceeds 1 MiB → `spawnSync node ENOBUFS`; a 364851 B ledger → OK.
- through the real consumer, importing `gather()` from the snapshot's `board.mjs`: `big.md -> rows: 1600 filed: 1600 ruled: 1600` vs `huge.md -> rows: 0 filed: 0 ruled: 0`, stderr `board: node .../ledger.mjs --file .../huge.md read failed: spawnSync node ENOBUFS`.

The threshold is on **payload bytes, not ledger-file bytes** (one reproduction's first fixture parsed fine at 1192281 file bytes because its payload was 912031 B); it takes roughly 6000 rows/filed/ruled to cross.

**1 MiB is about twice an already-observed run.** The `SKILL.md` paragraph PR #803 deletes had measured a real **506 KB** ledger in production use, so this is not a hypothetical ceiling.

## Remedy (one line, one place)

```js
execFileSync(cmd, args, { encoding: "utf8", maxBuffer: 64 * 1024 * 1024 })
```

`tryRun` is the single point every board read routes through, so this covers the `gh` reads at the same time. `grep -rn maxBuffer skills/fleet/scripts` returns nothing today (rc=1, pattern verified to fire on a fixture containing `maxBuffer`).

## Why deferred rather than applied in #803

PR #803's ticket scope is `ledger.mjs`'s own exit shape. Widening it into `board.mjs` — which already carries several open tickets (#468, #586, #603, #605, #606, #686, #786) — would edit a file that ticket has no business in.

**Confirmed defect, remedy settled and one line.**

dimension: `correctness`
Deferred from PR #803 review.

feigi:
Closes #807

## What was wrong

`tryRun` ran every board read through `execFileSync` with no `maxBuffer`, so
node's default stdout cap applied to the `ledger.mjs read` behind the cockpit.
Past it the child is **killed**, not truncated — and in `tryRun` that throw is
indistinguishable from an unreachable tool. The read returns null, `tryParse`
falls back to the empty ledger, and the cockpit renders a board with nothing on
it at **HTTP 200**, one stderr line the only trace. That is #246's own symptom:
#803 moved the cliff up from the pipe buffer rather than removing it.

Reproduced through the real CLI on this branch's base, before any edit: a
ledger whose `read` payload clears the default builds a board with no tickets
and no filed rows, exit 0, `spawnSync node ENOBUFS` on stderr.

## What changed

A shared options object carrying an explicit `maxBuffer`, taken by **both**
child reads in the file — the same headroom `staleness.mjs` already takes, for
the same reason. Bounded rather than `Infinity`: a runaway child should still
be stopped rather than allowed to exhaust the box.

The ticket's remedy said `tryRun` is "the single point every board read routes
through, so this covers the `gh` reads at the same time". The `gh` half is
right; the "single point" half is **not**. `runCiState` spawns its own
`execFileSync` — it needs the exit code a plain `tryRun` discards — so a fix
applied only inside `tryRun` would have left that read uncapped. Its payload is
small in practice (`--quiet` drops `jobs`), so this is closing the class rather
than fixing a second live case, and it costs one identifier.

## Coverage

`grep`-confirmed on the base: nothing in the suite drove a board read at a size
node would cap, so the mode this changes was invisible to a green run. The new
case drives it through the real CLI and pins the **accept** side — every row
arrives. Asserting the failure shape instead would stay green on a `board.mjs`
that refused the oversized read outright, and a refusal is not the fix: the
ledger grows for the life of a run, so the read has to keep working.

The fixture calibrates itself. Before asserting anything it runs the pre-fix
call shape against the same file and requires `ENOBUFS`, so a fixture that
drifted under the cap — or a node whose default rose above it — goes red naming
that, instead of passing over a read that was never at risk.

Mutation, both directions: dropping `maxBuffer` turns exactly this test red and
nothing else in the file; changing the constant to a different ample value
leaves the file green, so the test pins the property rather than the number.

## Also here

- The rig's own `spawnSync` needed the same headroom for an unrelated reason: a
  board built over this ledger prints past the **reader's** default, which kills
  the child and reports `status: null` — indistinguishable from the subject
  crashing. No production consumer reads `board.mjs build` stdout through a
  capped reader; `serve` calls `gather()` in-process and writes the file.
- `run-team/SKILL.md`'s "do not read this as #246 closed" note listed this
  defect as still open. That item is now rewritten rather than dropped. Review
  caught the first draft of it over-claiming — it said the consumer side no
  longer adds a cliff of its own, and that is not what this change does. The cap
  is deliberately bounded, so a payload past it still reaches a blank board, and
  a ledger that cannot be read at all does so at any size: the read fails open to
  an empty ledger and `board.json` carries no field saying so, while `spend` in
  the same payload does carry its own error. The note now says that, and points
  at #874, which tracks closing the fail-open. The `check` half (#808) stands
  unchanged.
- One restated options literal in a neighbouring test comment, which this change
  had made inaccurate; the property it names is unaffected.

## Tests

`./agent-test skills/fleet/scripts` — 1126 pass, 0 fail.


feigi:
## Review adjudication — applied, deferred, filed

Review ran 6/6 dimensions, none unrun. One apply, three deferrals filed, one record ticket.

**Applied here** (`5a1999f`) — the `run-team/SKILL.md` sentence this PR itself added. It claimed the consumer side no longer adds a cliff of its own; the cap is deliberately bounded, so a payload past it reaches the same blank board, and an unreadable ledger reaches it at any size. Settled against this tree before rewriting, not taken from the finding: an unreadable ledger builds at exit 0 with no tickets, no filed rows, and no field in `board.json` saying why, while `spend` in the same payload carries its own error. Two review dimensions reached that conclusion independently.

**Deferred and filed**

| Issue | Label | What |
|---|---|---|
| #874 | `ready-for-agent` | `board.json` carries no ledger-read error — a failed ledger read renders as a legitimately empty pipeline at exit 0. Confirmed 2-0 and reproduced a third time here, by two triggers: an unreadable ledger (no size involved) and a payload past the new cap. Pre-existing; changes the payload shape and `board.html`, so a different deliverable from #807. |
| #875 | `ready-for-agent` | `runCiState`'s exit-code gate accepts a truncated or signal-killed read as a CI verdict, discarding the `prevCi` carry-forward — the #262 regression its own comment says the gate prevents. The `maxBuffer` framing of this one was **refuted and measured unreachable**; the residual is not, and was measured reachable at exit 1 with unparseable bytes. Gate predates this PR. |
| #876 | `ready-for-agent` | `ci-state.mjs` truncates its own verdict JSON at the pipe buffer (`console.log` + `process.exit`) — the #246 pattern still live, and the upstream half of #875. Surfaced by a refuter working a different finding; not itself one of the review's findings. `maxBuffer` cannot fix it. |
| #877 | `wontfix`, closed | Record ticket for the `suggestion` band. |

**Not filed, by ruling:** the claim that raising the cap widens the duplicated stderr re-emitted through `${e.message}`. The duplicating line is untouched by this diff; the repro used a stderr below node's default, so it shows only pre-existing duplication with no diff-attributable delta; and in the one regime where this diff changes anything the pre-diff behaviour is a killed read, which is the bug #807 exists to fix. Trading a killed read for a large log line is the intended fix.

**The `suggestion` band, checked.** The workflow budgets that band no refuters, so a refuter was dispatched against each in-scope finding, biased to refuse and required to verify by running something. All five rulings came back refuted with measurement, each with zero blast radius — nothing applied. Detail and the measurements are in #877 so they are not redone; highlights: the shared-`maxBuffer`-constant extraction was built and run green and still rejected (three sites cap three different payloads); the `doesNotMatch` assertion claimed redundant is the **sole** detector in two constructed states, one of them an equally correct alternative fix for this very issue; and the suggested `runCiState` test would have pinned a property false of the real system.

**Tests:** `node --test skills/fleet/scripts/*.test.mjs` from the worktree — `tests 1126 / pass 1126 / fail 0`, exit 0. Matches the baseline four refuters established independently against the pre-fix head, so this is the whole tree, not a partial one.

PR #869, opened 2026-08-24
fix(fleet): a non-numeric --pr shipped an unidentifiable payload at exit 0

### LINKED ISSUE #840: ci-state.mjs: a non-numeric --pr serialises as `pr: null` in BOTH payloads — an unidentifiable report at exit 0, verdict green
Deferred from PR #833 review (issue #262). Finding `unverified-3`, dimension **types**, severity `suggestion`. Refuted as a defect *in this PR*, filed because the underlying behaviour is real and pre-existing.

**file:line:** `skills/fleet/scripts/ci-state.mjs:73` (the #262 outage payload) and `skills/fleet/scripts/ci-state.mjs:501` (the pre-existing normal-path payload).

## What was measured

`--pr` is validated for truthiness only (`if (!pr) die(...)`), never for numeric shape, so `--pr abc` survives. Both payloads then build `pr: Number(pr)`, and `JSON.stringify(NaN)` is `null`.

Outage payload, stub `gh` refusing on quota:

```
$ node ci-state.mjs --pr abc --quiet
exit=2
{"pr":null,"verdict":"rate-limited","reasons":["gh pr view was refused by the GitHub API rate limit — ..."]}
```

Main path at the PR base commit `b7cc345` (so: pre-existing, not introduced by #833) — and this is the worse one, because it ships an unidentifiable payload at **exit 0 with `verdict: green`**:

```
$ node ci-state.mjs --pr abc --quiet
{"pr":null,"branch":"fix/1",...,"verdict":"green","reasons":[]}
exit=0
```

## Why it was deferred rather than applied

- **Not this PRs regression.** The same `Number(pr)` shape predates #833 on the main path; #833 only added a second site with it.
- **The claimed contradiction does not hold.** The finding argued this violates `emitRateLimited()`s stated invariant that unobserved fields are absent rather than null. That invariant is scoped to *observations* — the PRs own test enumerates `status`, `conclusion`, `jobs`, `missing`, `runId` and deliberately excludes `pr`, and the design-spec row documents `pr` as present. `pr` is the echoed argument, not a reading.
- **No consumer reads it.** The only programmatic caller, `board.mjs`, passes `String(pr)` from a numeric board record and its `mapCi()` reads `status`/`verdict` only.

## Remedy (verified, not applied here)

Validate once near `arg("pr")` and `die()` there, which fixes both payloads:

```js
if (!/^[0-9]+$/.test(pr)) die(`--pr needs a number, got ${pr}`);
```

Measured against the PR head: full suite `tests 1096, pass 1096, fail 0` with the guard in, and no existing test passes a non-numeric `--pr` deliberately. Note nothing then covers the new guard either — a test would need to come with it.

Remedy is settled; scope is the only reason this is not in #833.

feigi:
`--pr` was validated for truthiness alone, so `--pr abc` reached every payload site in `ci-state.mjs`. Each builds `pr: Number(pr)`, and `JSON.stringify(NaN)` is `null` — so the normal path emitted a payload carrying no identifying field **at exit 0 under `verdict: "green"`**, which is the verdict the fleet gates PR-green on, and the #262 outage payload did the same at exit 2.

One guard beside the existing usage die covers them all and refuses before the first `gh` read.

## Re-derived against `origin/main`, not the ticket

The ticket's line numbers have drifted; the mechanism has not. `pr` is the only argument reaching `Number()`, and the sites the ticket names are the only ones writing a payload to stdout — `--base`/`--workflow`/`--workflow-file` share the truthiness-only shape but stay strings bound for `gh`, where a non-numeric value is a legitimate ref. So the class is closed by a single guard upstream of all of it, and nothing is deliberately left. The guard refuses at the argument, so it does not depend on how many sites there are.

Reaching `gh` at all is the second harm, and it is not hypothetical: the mutation run below caught the script mid-flight asking `gh pr view abc --json …`. `gh pr view` resolves a non-numeric ref as a **branch**, so the old behaviour could return a genuine verdict for whatever PR that branch belonged to, reported under `pr: null`.

The sibling scripts reached by the same shape are **not** closed here — `diff-stats.mjs` and `pr-overlap.mjs` still build `pr: Number(…)` behind a truthiness-only check, under `--pr` and under `--a`/`--b` respectively. That is #840's own scoping, and the class is filed as its own ticket rather than widened into this PR.

## What the guard can wrongly refuse

Digits-only forfeits the branch and URL spellings `gh pr view` itself accepts — the same trade `arg.mjs` documents for its `--`-prefixed values, and it is recorded in the comment rather than worked around. Nothing passes one: `board.mjs`'s `runCiState()` sends `String(pr)` off a numeric record, and every documented invocation is `--pr <N>`.

That direction is a test, not an assurance: **the accept case is pinned**, with `assert/strict` so `pr` reading back as the string `"42"` fails too.

## Placement

Below the usage die, never merged into it. `RegExp.test` coerces a null argument to the string `"null"`, so a guard hoisted above it answers an omitted `--pr` with a complaint about a number and never prints the usage line — both spellings exit 2, so the exit code cannot tell them apart. That path is now pinned; nothing covered it before.

Above `sweep()`, per `arg.mjs`: where both would refuse, the more specific wording wins. `--pr 42 --basee main` still refuses as the unknown flag, so `arg.test.mjs`'s stray-flag row is unchanged.

## The receipt the tests rest on

The refusal tests' load-bearing assertion is that `gh` was never asked: exit 2, an empty stdout and the stderr line are all reproducible by a *later* guard, so only "gh was never asked" separates the two. Two things were needed to make that assertion measure something rather than decorate the block.

The block first carried its own stub helper, which interpolated the log path into the shell **unquoted**. Under a `TMPDIR` holding a space the redirect split, the receipt file was never written, and `assert.equal(r.log, "")` passed while `gh` had in fact been called — the pin measured nothing on exactly the paths where a spaced temp directory is in play. The file's existing `run()` already passes that path through the environment and quotes it, so parameterizing its `--pr` value retires the helper rather than repairing it, and the pin reddens under a spaced `TMPDIR` again.

The refusal cases pass `prView: null` with it. Under `run()`'s green fixture a downgraded guard reaches a `gh` that *answers*, the run diverges on its exit code before the receipt is read, and the assertion that reddens is the status one — leaving the receipt argument above asserted but untested. A `gh` that fails when reached restores the signature a later guard reproduces, putting the receipt back in the only position that separates them.

The comment naming `arg.test.mjs`'s `stubGhBin()` drops its ordinal: the same two jobs are enumerated in one order here and the other order there, so "the second job" read correctly against this block and backwards against the file it cited.

## Tests

Suite re-measured on current `main` — the ticket's own figure predates it. Baseline green, and green again with the guard and the added tests.

Every mutant below was applied to a copy outside the worktree and run whole-suite:

- **guard downgraded to advisory** → the refusal tests redden, each on the `gh` receipt. Exit 2, empty stdout and the stderr match all still passed under the mutant — a later guard reproduces that whole signature, so "gh was never asked" is the assertion carrying it.
- **either anchor of `/^[0-9]+$/` dropped** → the mutant then matches a substring instead of the whole value. Dropping `$` is caught by `--pr 42x` alone, dropping `^` by `--pr x42` alone; each value refuses only while its own anchor is present, so neither mutant survives the pair. Before these, every `--pr` the suite fed the guard was all digits or none, which both mutants agree with the real guard about — so both survived the whole suite.
- **normal payload's `pr` switched to a string** → only the accept test reddens; nothing else pins that field's type. Switching the outage payload's as well additionally reddens the quota test, which is the sibling assertion covering that one.
- **guard hoisted above the usage die** → only the usage test reddens.
- **payload keys reordered** (should not be caught) → whole suite green, so the added tests pin values and fatality rather than serialization order.
- **guard respelled `/^\d+$/` with string concatenation** (should not be caught) → whole suite green, so nothing is pinned to the guard's own syntax.

Closes #840


feigi:
## Review adjudication — findings applied and deferred

The review ran all six dimensions with none unrun. Every in-scope `suggestion` was checked by a refuter briefed to refute and to verify by running something, since that band gets no refuters by policy.

### Applied here

- **`tests`** — neither anchor of the guard's `/^[0-9]+$/` was pinned; dropping either left the whole suite green. `--pr 42x` and `--pr x42` now each refuse, and each kills only its own anchor's mutant.
- **`correctness`** and **`simplify`** — these prescribed opposite remedies on the same helper (repair its unquoted gh-log path vs. delete it and reuse `run()`), so they went to a single refuter briefed with both. Both claims held; the deletion subsumes the repair, because `run()` already passes that path through the environment and quotes it. The helper is gone and `run()`'s `--pr` is parameterized. The refusal cases pass `prView: null` with it — under the green fixture a downgraded guard diverges on its exit code before the receipt is read, which would have left the receipt argument asserted but untested.
- **`comments`** — the ordinal in the `stubGhBin()` reference is dropped. The two votes split on which enumeration it read against, and both conceded the sentence was ambiguous between them; dropping it makes both readings true.

### Deferred

- **#878** (`ready-for-agent`) — the `silent-failure` and `types` findings that survived refutation, plus the `correctness` and `simplify` findings pointing at the same defect from the unverified band. All four are the one defect and prescribed three different remedies, so they are filed as one ticket with one remedy. `#840` scopes itself to `ci-state.mjs`, so widening this PR was rejected. Note the class is **not** bounded by the `arg("pr")` spelling — `pr-overlap.mjs` has the same fail-open under `--a`/`--b`, which a grep for `arg("pr")` misses.
- **#879** (`needs-triage`) — the three suggestions refuted by measurement, recorded so they are not re-derived and re-applied. Recommendation is to close. The band's `tests` entry was a null result and is accounted for there too.
- **#880** (`ready-for-agent`) — found while sweeping for the retired helper's template: `arg.test.mjs` and `claim-ticket.test.mjs` build their gh stubs the same unquoted way, and both assert on the receipt's *non-existence*, so under a `TMPDIR` with a space those pins pass vacuously. Out of this PR's scope to edit.

### Verification

Full suite green at the committed tree, and green at the same count with each should-not-catch mutation applied. Mutation results for every claim the PR body makes are in the body itself, re-measured against this tree rather than carried over.

Issue #867, opened 2026-08-24
tier-outcomes.tsv's PR #802 row says one sibling garbled site survived; #806 closed two — record it in #865's OWN row, do not rewrite the historical one


feigi:
Deferred from the PR #865 review (issue #806). Dimension `comments`, severity `suggestion`, band `unverified` — the workflow budgets a suggestion 0 refuters by policy, so it reached the controller unchecked. **Controller-verified before filing**; scope-ruled out of PR #865 because it targets a file that PR does not touch.

## The observation

`docs/metrics/tier-outcomes.tsv` line 94 (the PR #802 / issue #250 row) reads:

> A garbled sentence was repaired (grammar, not a false claim); its **sibling one block down** survives and is filed as #806.

Singular. PR #865 then closed **two** sites, not one — the exit-0 keep-reason pin and the refusal-states pin. A future reader skimming that row would underestimate what #806 covered.

Verified at `origin/main` before filing:

```sh
sed -n '94p' docs/metrics/tier-outcomes.tsv        # → "its sibling one block down survives"
gh pr diff 865 --name-only                          # → skills/fleet/scripts/reap.test.mjs ONLY
```

## The remedy is NOT to edit line 94 — read this before implementing

**`tier-outcomes.tsv` is an append-only record of what was true, and known, during each run.** Line 94 records what the #802 audit found at the time. It is **incomplete, not false**: the sibling one block down genuinely did survive, and the row nowhere claims it was the *only* survivor.

Rewriting it would falsify the record of what was known then — the failure mode where a suggested fix demands a past-tense historical record be rewritten as a live count, and applying it mints the very false claim the correction exists to prevent. The same reasoning already governs this file elsewhere: PR #866's implementer deliberately left `release-ticket.sh:757`/`:766` citations in this file untouched for exactly this reason, calling it "the one place the numbers are *supposed* to rot".

**Do the additive thing instead:** ensure PR #865's own row in this file records that it closed two sibling sites rather than one. That puts the correct count in the same file a future reader is skimming, in the row that owns the claim, without touching history.

## Why this was worth filing at all

Not because the row is wrong — it is not. Because the durable-home rule says a fact living only in a PR body does not outlive the merge, and PR #865's body is currently the only place carrying the two-site scope (it has a dedicated section, `## The ticket named one site; the tree had two`). This ticket is that fact's home until #865's tsv row exists.

## Also worth knowing for whoever picks this up

[...]

Issue #864, opened 2026-08-24
run-team SKILL.md tier-guard figures are ~1.6x stale, and the date confound it describes has since weakened — three sonnet rows now break it, not one


feigi:
Found during a `/fleet:run-team` phase-3 tier-guard recount, 2026-08-24, immediately after PR #863 merged (`docs/metrics/tier-outcomes.tsv` at `b8186f7`).

`skills/fleet/skills/run-team/SKILL.md`'s tier-guard section instructs the reader to **recount before citing** its figures. Doing so is what surfaced this: every figure in that section is now stale, and one of its stated *conclusions* no longer follows from the data.

## The stale figures

The section's snapshot is stated "As of PR #747, 2026-08-21, 50 rows". Recounted at `b8186f7`:

| claim in SKILL.md | measured now |
|---|---|
| 50 rows | **80** |
| 33 `class=routine` PRs | **49** |
| across 8 distinct `run_date`s | **11** |
| 5 routine rows with `closed_own_ticket` `no` | **6** |
| 9 of 33 routine rows with `minted_false_claim` `no` | **13 of 49** |
| "3 failures in 9 `routine`/`sonnet` rows against 2 in 24 `routine`/`opus`" | **3 in 11** against **3 in 38** |

The named failure list is also incomplete. SKILL.md enumerates `#452, #466, #536 at sonnet; #693, #707 at opus`. The sixth is **PR #787** (2026-08-22, `routine`, `opus`), which no sentence in that section accounts for.

Commands used, so the recount is reproducible rather than asserted:

```sh
grep -vc '^#' docs/metrics/tier-outcomes.tsv
awk -F'\t' '!/^#/ && $4=="routine" {n++; d[$1]=1; if($6=="no") no++} \
  END{print n, length(d), no+0}' docs/metrics/tier-outcomes.tsv
```

## The part that is more than a stale number

SKILL.md argues the same-class cross-tier comparison is uninformative because tier is confounded with calendar date:

> 8 of the 9 `sonnet` rows fall on 2026-08-13 to 08-17 and 23 of the 24 `opus` rows on 08-18 to 08-21, so tier is very nearly confounded with calendar date and therefore with prompt evolution

and names **PR #714** as "the one row that breaks the confound".

That is no longer the picture. Measured now, `routine`/`sonnet` holds 11 rows, and **three** of them post-date the prompt-evolution window:

```
2026-08-20  PR#714  closed_own=yes
[...]

Issue #823, opened 2026-08-23
the new run-team SKILL.md staleness block has no prose-pin test


feigi:
Deferred from PR #813 review (issue #238).

**dimension:** `tests`

## The finding

The prose PR #813 adds to `skills/fleet/skills/run-team/SKILL.md` — the `staleness.mjs` usage block, the exit-code table, and the "Annotate every survivor whose liveness probe came back could not check" instruction — has no prose-pin regression test, though this repo's convention is to pin exactly this kind of load-bearing instructional clause.

Verified on the pushed head:

```
$ grep -rln 'could not check' skills/fleet/scripts/*.test.mjs
(no matches)
```

`staleness-qualifier-prose.test.mjs` exists but pins the `review-and-fix.md` / SKILL.md finisher-dispatch twins from #196, not this block.

## Why deferred

Out of PR #813's scope: the ticket asks for the probe and the phase-0 instruction, and the pin is a separate piece of test work with its own design questions (which clauses to anchor, where the slice boundary goes). Adding it inside the review-fix pass would widen the diff into a file the pins would then have to be re-derived against.

## The remedy is known

Mirror `staleness-qualifier-prose.test.mjs`'s slice-anchored `phrase()` pattern. Pin at minimum the exit-code table's three verdicts and the "annotate … could not check" instruction. Two traps that file already documents apply: the slice anchor is what does the anchoring — a positive regex matched over the whole document is satisfiable from outside the clause it guards — and the pin must be mutation-tested both ways, gutting the clause to see it red and reflowing the paragraph to see it stay green.

`ready-for-agent`: the gap is measured, the pattern to follow is in the repo.

PR #881, opened 2026-08-24
chore(fleet): this run's tier rows, and the two rules it found by getting them wrong

feigi:
Controller-authored close-out for one `/fleet:run-team 2 5` run. Four tickets shipped: #806, #800, #807, #840.

**Deliberately not labelled `ready-to-merge`.** That is a reviewer's label and this is controller-authored work; it is left for the maintainer, per the precedent of PR #863.

## Four tier-outcome rows

One per implementer PR, appended when the controller ruled each review, as the file's header requires. `run_date` is the date of the **ruling**, not of this PR landing — three rows are 2026-08-24 and one is 2026-08-25.

| PR | issue | class | tier | closed_own_ticket | minted_false_claim |
|---|---|---|---|---|---|
| #865 | #806 | correction | opus | yes | no |
| #866 | #800 | correction | opus | yes | no |
| #868 | #807 | routine | opus | yes | no |
| #869 | #840 | routine | opus | yes | no |

**These rows carry no tier signal.** Every class dispatched at the session tier — `model` omitted on the `Agent` call — so `tier` is uniform across all four. They are data for the accumulated file, nothing more. Restoring a cheaper tier for `class=routine` still needs a deliberate control, which is the maintainer's call.

Field-count check clean: `awk -F'\t' '!/^#/ && NF!=8'` returns nothing.

## Two rule corrections, both from a controller error

Neither came from a review. Both came from members refusing to take the controller's word.

### `solely` is the load-bearing word

`review-and-fix.md` states the rule correctly: a `rebase-check` red is staleness **solely** from a non-zero behind-count. SKILL.md said "off a non-zero behind-count", which reads as an implication rather than a discriminator — and a controller paraphrasing from there told four members outright that a behind-count settles it.

It does not. `ci.yml`'s `rebase-check` exits 1 on five conditions: base fetch failure, unresolvable base, no common ancestor, merge-base ≠ base tip, and merge commits above base ≠ 0. Only the fourth is staleness.

There is a sharper trap underneath, found by a finisher this run: **the staleness condition exits before the merge-commit condition is ever evaluated**, so a job log showing condition 4 does not clear condition 5. The log is silent on it by construction. It has to be measured separately with `git rev-list --merges --count "$BASE..HEAD"`.

Two members declined the controller's assertion and measured it properly. One took it on trust — correctly, as it happened, but only by luck.

### The merge wave can fire while the finisher is still live

The phase-3 edge read "`ready-to-merge` appears → merge-bot wave" with no caveat. But the label lands at the finisher's **duty 3** and its report is **duty 4**, so the gap between them is the normal case, not a rarity — it fired on every labelled PR this run. And the wave's first act on a behind PR is a rebase, which is destructive to a worktree someone is in: the same hazard the reaping rule already names for the same reason.

The wave was held by hand all three times. Waiting is nearly free — a PR already labelled blocks nothing, and its behind-count is expired on arrival regardless.

## What this does not do

**No pin was added for either rule.** Pins in this file are added deliberately, and #823 already owns the question of what in the new prose wants one. Adding an unrequested one here would pre-empt that ticket.

**Partially pre-empts two open tickets, neither closed:**
- **#864** asks for the tier-guard figures to be refreshed and its date-confound paragraph corrected. This PR touches neither — it only edits the two `rebase-check` sites and the phase-3 edge. #864 remains entirely open, and its row count is now further out of date, not less.
- **#867** asks that PR #865's own row record it closed two sibling sites rather than one. **That is satisfied here** — #865's row says so. #867 also asks that line 94 be left unchanged, which this PR honours. Worth a triager confirming rather than my closing it myself.

## Verification

- Suite green at this head: `node --test skills/fleet/scripts/*.test.mjs` → **tests 1133, pass 1133, fail 0**.
- A prose pin **caught the first draft** of the finisher-facing edit — it broke the pinned phrase `a \`skipped\` heavy job is behind-count staleness`. The edit was reworded to preserve that phrase and add the qualifier around it, rather than editing the test to accommodate the prose. Re-run green.
- Only two files are in this PR. `settings.json` and `probeA.mjs` were dirty in the checkout before this run began and are deliberately untouched.

## Note on the commit trailer

This commit carries `Co-Authored-By` / `Claude-Session` trailers because a standing harness instruction directs it. **Every other commit in this repository uses `Refs #NNN` and carries neither.** A member this run flagged the same conflict and matched the repo instead; the controller upheld that for its case, on the narrower ground that mixing styles inside one PR is worse than either style. Which convention should win repo-wide is a maintainer decision, not one to settle from inside a chore PR.


feigi:
## Three more tier-outcome rows — from the `/fleet:run-team 2 5` run of 2026-08-25

Posting here rather than opening a second close-out PR: this PR already edits `docs/metrics/tier-outcomes.tsv`, and two PRs appending to the same file end conflict by construction. Fold these in when you take this PR, or tell me to open a separate one once it lands.

Tab-separated, `run_date pr ticket class tier closed_own_ticket minted_false_claim note`, field count 8 each:

```
2026-08-25	882	580	routine	opus	yes	no	verify-sha fatality pins. Review 3/6 dims (tests-only profile), unrun=none, 4 agents: 0 survived / 0 refuted / 1 unverified. Nothing applied, no fix-applier dispatched. Implementer swept all 8 die sites rather than the 2 its ticket named and found a third whose mutant fails OPEN where the ticket's two fail closed, filed as #884. minted_false_claim=no per the diff-scoped definition: its stated mechanism for that third guard was wrong (claimed exit 0 plus malformed payload; measured exit 1 via short-circuit leaving sha_j unassigned then set -eu abort) but appeared only in the PR body, never in the diff. Controller relayed that wrong mechanism into #884 at filing and corrected it after the review measured it. Ticket body claim that an equivalent progress marker exists for each guard below is false for the cat-file guard; implementer escalated, got no ruling, proceeded by shimming git cat-file, ruled accepted after the fact.
2026-08-25	883	808	routine	opus	yes	no	ledger check already-filed payload cut at the pipe. Review 6/6 dims, unrun=none, 17 agents: 2 survived 2-0 / 3 refuted / 11 unverified. Headline survived finding was reviewability not behaviour: runCheck body indented 4 spaces against the file's 2, inflating the diff from 59 real changed lines to 839; finisher proved the fix whitespace-only by git show -w yielding zero changed lines. Second survived finding was docs rot in run-team SKILL.md, a file the PR does not touch, which told agents check is pinned on a pipe nowhere. minted_false_claim=no: the one falsified comment pre-existed and was falsified BY the new test rather than added by it. Fix-applier caught a methodology defect in the review itself, an unescaped issue reference in --test-name-pattern acting as a capture group so the subsumption check matched the file and reported success having run nothing named. Hang claim refuted by measurement: blocking pre-exists at merge-base for read and check's clean arm at the same byte threshold, and there are no programmatic consumers of check.
2026-08-25	885	876	routine	opus	yes	no	ci-state verdict payload cut at the pipe. Review 6/6 dims, unrun=none, 17 agents: 1 survived 2-0 / 4 refuted / 8 unverified. Three of four refutations were 1-1 splits. Controller reversed one after reading the votes: its two refuters measured contradictory results for the same operation, and the refuting vote's control did a bare writeSync on an fd never stream-initialised, so it measured a blocking fd the script never has. Member confirmed the reconciliation by experiment, console.error first then writeSync short-writes and returns while bare writeSync blocks. Defect real and in scope; the loop closes it and also closed a second finding the controller had ordered filed, so that filing was correctly not made. Member improved the prescribed remedy with a 1ms wait over a busy continue, measured 2891ms CPU down to 71ms for identical bytes, and caught its own false comment before pushing after measuring that a shadowing declaration is a hard SyntaxError rather than an invisible definition.
```

### These rows carry no tier signal

All three dispatched at the session tier with `model` omitted on the `Agent` call, so `tier` is uniform. They are data for the accumulated file, nothing more. Restoring a cheaper tier for `class=routine` still needs the deliberate control the guard names, which is a change to the dispatch rule and therefore the maintainer's call.

### Guard state, recounted rather than cited

Recounted at `8fc60ff` before these rows: **49 `class=routine` rows across 11 distinct `run_date`s, 6 carrying `closed_own_ticket=no`.** SKILL.md's own figures (33 / 8 / 5) are stale — PR #787 is the row its list omits. Floor and trigger are both met, but the guard **fired on 2026-08-16 and that revert already happened**, so it is in its fired state with no further move. Full recount and method posted to #864, which owns the stale-figures ticket.

### Also relevant to this PR

Its two SKILL.md rule corrections were adopted by hand for the whole run rather than read from the file, since they are not in `main`. Both earned their place:

- **`solely`** — every `rebase-check` red this run was attributed by measuring `git rev-list --merges --count "$BASE..HEAD"` separately, never off the behind-count. It came back 0 every time, so all were pure staleness, but it was measured rather than assumed on each of five occasions.
- **The merge-wave/finisher gap** — no merge bot was dispatched until the finisher's report had landed, not merely its label.

While this PR stays open it also costs two things beyond the stranded rules: it inflates `fleet-tick`'s review backlog by one permanently (it can never be labelled by the fleet), which pins implementers at HOLD; and it blocks #288 from being staged, since both edit `run-team/SKILL.md`. It is currently **behind 14** and `check`-green.

feigi:
## ⚠️ Superseded — do NOT fold the rows above in

My earlier comment on this PR (09:35 today) posted three tier rows for **#882, #883, #885** and asked you to fold them into this PR. **Disregard that.** The rows are now carried by close-out PR **#909**, which is green and open.

The reasoning that produced that comment was sound at the time — this PR already edits `docs/metrics/tier-outcomes.tsv`, and two PRs appending to the same file tail conflict by construction. What changed is that the run kept going and produced six more rows (#891 #892 #900 #902 #904 #908) plus two rule fixes to `run-merge-bot.md`, which needed a PR of their own rather than a third comment.

**Folding both in would append #882/#883/#885 twice.** #909 is the single source for all nine rows.

### What still holds from that comment

- The append conflict is real and unresolved. #909 and this PR both append to the tail of `docs/metrics/tier-outcomes.tsv`. Whichever lands second needs a keep-both resolution in chronological order — no content is lost either way, the rows just need ordering.
- This PR's two `run-team/SKILL.md` corrections are still outside `main`, so every run re-adopts them by hand. Both earned their keep again this run: `solely` on `rebase-check` held on five separate occasions (each attributed by measuring `git rev-list --merges --count "$BASE..HEAD"` separately rather than off the behind-count, 0 every time), and the merge-wave/finisher-report gap held throughout.
- It still inflates `fleet-tick`'s review backlog permanently — it can never be labelled by the fleet — which pinned implementers at HOLD for this entire run despite `supply=79`. #909 now does the same, so the backlog reads 2.
- It still blocks #288 from being staged, both editing `run-team/SKILL.md`.

### One figure in that comment has gone stale

It said *"currently behind 14 and `check`-green"*. Now **behind ~37**, with `rebase-check` red purely on the behind-count and `check` still green.

Issue #876, opened 2026-08-24
ci-state.mjs truncates its own verdict JSON at the pipe buffer — console.log + process.exit, the #246 pattern still live


feigi:
**Dimension:** silent-failure · **Surfaced during PR #868 review** by a refuter dispatched against a different finding; not itself one of the review's findings.

`skills/fleet/scripts/ci-state.mjs`

## The defect

`ci-state.mjs` emits its verdict payload with `console.log(JSON.stringify(payload))` and then calls `process.exit(...)`. On a pipe, `process.exit` discards the queued remainder, so a payload larger than one pipe buffer is guillotined mid-JSON. The exit code survives; the payload does not.

This is #246's exact shape — that issue was closed for `ledger.mjs read`, and the same `console.log` + `process.exit` pattern is still live here.

## Measured

Driving the real script with a workflow file large enough to make `missing` long:

```
real ci-state --quiet payload written to a FILE: 1,680,315 bytes
read through a pipe:  status=1, stdout=65536 bytes, JSON.parse ok? false
```

Independently reproduced with a stub on the same payload path:

```
{ "threw": true, "status": 1, "signal": null,
  "stdoutBytes": 65536, "gateReturnsThisAsVerdict": true, "parses": false }
```

`maxBuffer` cannot fix this — the cut happens in the child before the parent's buffer is ever consulted. PR #868's `maxBuffer` change makes no difference to it: fixed and unfixed trees produced byte-identical results.

## Consequence

The truncated bytes reach `board.mjs`'s `runCiState`, whose gate accepts them as a CI verdict; `mapCi` cannot parse them and answers `"unknown"`, and the non-null return skips the `prevCi` carry-forward. Filed separately as the companion `runCiState` gate issue — that one is the consumer half, this is the producer half. Either fix alone narrows the window; both together close it.

## Suggested remedy (open — not settled, never run)

Replace the payload's `console.log` with a synchronous `writeSync(1, ...)` before the exit, the way `emitRateLimited()` already writes. That is the fix #246 established for this pattern.

## Reachability

Bounded rather than urgent: it needs a payload past one pipe buffer, which today wants a workflow file with a very large job count. But the gate downstream turns it into a lost CI state rather than a loud failure, so it degrades silently when it does happen.

Issue #793, opened 2026-08-22
REFUTED: ledger.test.mjs:563 — the title-exempt rationale duplication is house convention (69 blocks, 11 pairs), and consolidating kept the false half


feigi:
Deferred from PR #785 review (issue #232). Dimension: **simplify**. Severity: **suggestion**. Site: `skills/fleet/scripts/ledger.test.mjs:563`.

Filed as a record because a finding is deferred, never dropped. **It was refuted by a dispatched refuter that ran the trim** — the duplication is real, but it is this repo's house convention, and the proposed consolidation direction keeps the wrong half.

## The claim as submitted

The `title`-is-exempt rationale is written twice, near-verbatim: `ledger.mjs:453-460` (8 comment lines) and `ledger.test.mjs:563-568` (6 comment lines) make the same argument — same clause `in the row and in the score alike`, same `#232` citation — so 14 lines of prose attach to a 2-line predicate change. Proposed remedy: keep the rationale once in `ledger.mjs`, trim the test block to its unique content.

## Why deferred: REFUTED

The `suggestion` band gets 0 refuters from the review workflow, so this arrived unchecked. It edits the PR's own new comment, so it was ruled in scope and one refuter was dispatched, biased to refuse. `refuted=true`, on three measurements.

1. **The duplication is genuine** — the two blocks share 14 verbatim 6-grams including the `h.title || ""` clause, and both cite #232.
2. **The trim is safe but arbitrary.** Nothing asserts on that prose: the trimmed copy runs **1050/1050**, byte-identical counts to baseline. But a shared-phrase sweep found **69 test comment blocks across 11 source/test pairs** in `skills/fleet/scripts/` that restate their source's rationale — 16 of 70 in `ledger.test.mjs` alone:

```
arg 2/18 · board 9/34 · candidates 19/80 · ci-state 5/20 · compute-board 0/4
compute-spend 5/12 · diff-stats 5/16 · fleet-tick 6/20 · ledger 16/70
pr-overlap 1/4 · prose-pin 1/7        →  pairs=11  duplicating_blocks=69
```

Site 563 is not an outlier or even the worst: `ledger.test.mjs:282` shares **17** 6-grams with `ledger.mjs:303-314` and is a deliberate section banner. The finding gives no reason this one site is special. Actioning it once is arbitrary; actioning it consistently is a 69-site sweep nobody asked for.

[...]

PR #883, opened 2026-08-25
fix(fleet): check's already-filed answer abandoned its payload at the pipe

### LINKED ISSUE #808: ledger.mjs:244: check's ALREADY-FILED exit is the last process.exit() that truncates its payload on a pipe — mid-branch, so it needs the branch extracted

feigi:
`check`'s ALREADY FILED answer wrote its payload with `console.log` and then
called `process.exit(1)`, which abandons the queued async write on a pipe. The
payload arrived cut, under an exit code that was correct throughout.

Closes #808

## Re-derived against `origin/main`

The ticket's line numbers had drifted; the sites were re-located semantically,
never by the quoted numbers. The claim that this was the last `process.exit()`
truncating a payload in `ledger.mjs` **holds**: sweeping the file for a
`console.log` followed by a `process.exit()` on any path finds every other
payload arm already falling out of the dispatch chain, and `check`'s terminal
exit already an `exitCode` assignment. The already-filed arm was the only one
left, and it is the one this changes.

Measured before, through a real pipe, 3/3: exit 1, stdout stopping at the pipe
buffer, `JSON.parse` refusing an unterminated string. The same command
redirected to a file delivered the whole thing and parsed — that pair is what
identifies the pipe rather than the payload as the cause. After: the pipe
delivers the byte count the file already did, at the same exit 1.

## Why a restructure

The fix `read`/`row`/`filed`/`ruled` and `check`'s terminal exit took does not
transfer. Those exits were the last statement of their arm, so falling out
reached the same code. This one sits mid-branch, where falling through would
run the near-miss ranking and the tracker search the exit exists to skip. So
`check`'s arm is now `runCheck()`, and `process.exitCode = 1; return;` leaves
it. The extracted body is re-indented to the file's own convention, so it comes
back byte-identical to what it stood as inside the chain and the diff shows the
extraction rather than the whole moved body.

**The exit code does not move.** 1 remains the ALREADY FILED signal callers
gate on, and nothing on that path assigns `exitCode` after the return.

## What this newly does, beyond what it stopped doing

- The `return` reaches the process's normal exit, so the stderr ALREADY FILED
  line — queued before the payload and abandoned by the same `exit()` — now
  drains too.
- `die()` is unaffected: it is `writeSync`-based and never returns, so it
  behaves identically from inside a function.
- Nothing follows the dispatch chain, so returning early runs nothing new.

## Enumerated, covered and left

**Covered** — the already-filed arm (the ticket's case), and the arm past it,
which the restructure could break without the ticket ever mentioning it.

**Deliberately left** — `die()`, which is pipe-safe by construction rather than
by sweep, and the sibling scripts, which are outside this ticket and which
`ledger.mjs`'s own rationale already records as carrying the same reasoning.

## Tests

The suite could not see this mode at all: the pipe sweep beside these drives
the subcommands whose payload echoes an argv tail, and `check`'s does not — its
wide field is a row read back out of the ledger, so the fixture that makes it
oversized is a wide ledger, and no ceiling on an argv element applies.

Three tests, each mutation-tested to a mutant it kills **alone** (one failure
file-wide) and against a zero-behaviour rename of the extracted function that
must stay green, which it does across the whole file:

| Pins | Killed alone by |
|---|---|
| the already-filed payload arrives whole at exit 1 | the defect restored |
| the already-filed payload carries its own shape, ranking and tracker skipped | that arm picking up a field of the arm it departs before |
| the not-filed payload arrives whole and still reaches the tracker | that arm's exit reverting to `exit()` |

The second is the half `gh` cannot witness: the near-miss ranking runs entirely
in-process, so a run that never invoked `gh` is no evidence the ranking was
skipped. The third is the input this must still ACCEPT — every other case here
stops early, and a restructure that stopped early always would satisfy them
all.

## Applied from review

- The extracted body was first left at the nesting level it had inside the
  dispatch chain, so every line of it read as changed and a reviewer had to
  read all of them to establish that nothing did. Dedented to the file's own
  convention, the body is byte-identical to the pre-image.
- `skills/fleet/skills/run-team/SKILL.md` — the one file here outside the
  ticket's own two — told agents that `check` is pinned on a pipe nowhere and
  that its ALREADY FILED exit still cuts its payload. True when written, false
  once this lands. It also contradicted that same file's `check` exit-code
  section, which tells readers to gate on `verdict` precisely because a
  searched-and-clean tracker and one that was never read both leave at exit 0.
- The `run()` helper rested its stderr assertions on `check`'s stderr staying
  far under the pipe buffer. The fixtures added here drive it well past, so the
  premise is gone even though the conclusion holds. The comment now names the
  ceiling that is actually left, `spawnSync`'s own `maxBuffer`.
- The not-filed case asserted "an unreachable tracker hit" over a fixture whose
  `gh` stub is reachable and returns nothing — a state that assertion could
  never report, since a tracker hit leaves at exit 3 rather than the 0 asserted.
- The not-filed arm gained the shape pin its sibling already had. The
  design-spec row that reads both arms checks the keys that are emitted, so it
  catches a field leaking in and cannot catch one going missing; `subject` is
  the field of that arm no other test reads, and dropping it survived the whole
  suite before this.


feigi:
Fix-applier pass on the review of this PR. Applied at `8c96958`; suite `node --test skills/fleet/scripts/*.test.mjs` from the worktree reports `tests 1136 / pass 1136 / fail 0`.

**Applied** — both `survived` findings, plus three `suggestion`-band findings that survived a dispatched refuter:

| finding | dimension | what |
|---|---|---|
| survived | correctness / simplify | `runCheck()`'s body re-indented to the file's convention — same defect found independently by two dimensions, applied once |
| survived | comments | `run-team/SKILL.md`'s "`check` is pinned on a pipe nowhere" caveat, false once this merges |
| unverified | correctness | `run()`'s stderr comment rested on a size this PR's own fixtures exceed |
| unverified | correctness | the not-filed assertion named a tracker state its fixture never builds |
| unverified | types | the not-filed arm's shape pin — a dropped `subject` survived the whole suite before it |

**Deferred** — three refuted by a dispatched refuter, three out of scope:

| # | disposition |
|---|---|
| #886 | REFUTED — the already-filed arm's undrained-pipe blocking is pre-existing file-wide, at the same byte threshold as `read`, and no undrained consumer of `check` exists |
| #887 | REFUTED — no subsumption: a one-row ledger is a boundary the new fixture cannot reach, and an off-by-one mutant kills the old test alone |
| #888 | out of scope — `isMatch` re-runs an identical subset scan on same-size near-misses of four tokens or more |
| comment on #516 | out of scope — two stale `file:line` citations, deferred to the issue that already owns that sweep |
| comment on #793 | REFUTED — rationale duplication within `ledger.mjs`; deferred to the issue whose ruling already covers the class |

One review finding (`ci-state.mjs`'s surviving `console.log` + `process.exit()` payload site) was neither applied nor filed: it is issue #876, owned by open PR #885.

PR #885, opened 2026-08-25
fix(fleet): ci-state.mjs cut its verdict payload at one pipe buffer

### LINKED ISSUE #876: ci-state.mjs truncates its own verdict JSON at the pipe buffer — console.log + process.exit, the #246 pattern still live

feigi:
`ci-state.mjs` emitted its verdict payload with `console.log(JSON.stringify(payload))` and then called `process.exit(...)`. On a pipe that write is asynchronous: the kernel takes one pipe buffer's worth synchronously and `process.exit()` discards the rest rather than draining it. The payload arrives cut mid-JSON while the exit code arrives whole, so `board.mjs`'s gate accepts it as a CI verdict, `mapCi` cannot parse it, and the state degrades to `"unknown"` with nothing failing loudly. `emitRateLimited()` already wrote its payload with `writeSync` for exactly this reason; the verdict emit did not.

Producer half only — the companion `runCiState` gate issue is the consumer half and `board.mjs` is untouched here.

## What changed

Both of the tail's writes now go through one `emit(fd, text)` helper backed by `writeSync`. `emitRateLimited()` uses the same helper instead of repeating the try/catch.

**`emit()` loops on the count `writeSync` returns rather than calling it once.** A single call was not enough, and the review caught it: initialising a stream for an fd — which `vlog`'s `console.error` does to fd 2 — puts that fd in `O_NONBLOCK`, and a non-blocking write to a full pipe **short-writes**. It returns the count it managed and throws nothing at all, so the `catch` never fires, nothing is logged, and the exit code is delivered as if the write had succeeded. That is a failure the catch structurally cannot cover, because the write reports success.

Measured against this file's own pipe-clearing fixture, before and after the loop:

| | verdict line on stderr | trailing newline |
|---|---|---|
| single `writeSync` | cut at 65535 bytes | gone |
| looped `writeSync` | 131165 bytes, whole | intact |

Both in the `--quiet` mode the controller's CI Monitor polls and in the verbose one. stdout and the exit code are unchanged either way. The deciding control: a child that calls `console.error` first short-writes 65536 of 200000 bytes and returns normally, while a child that calls only `writeSync` on an untouched fd blocks until the whole write lands — so consuming the return value makes a non-blocking fd finish the way a blocking one already does.

`EAGAIN` then means the buffer is momentarily full rather than that the write failed, so it waits for the reader and retries; every other code returns and loses the bytes, which is the behaviour that was already there. Reporting a green PR as failing CI would be new, and worse than the bug being fixed. The wait is not decoration — against a reader asleep three seconds, a bare retry burned a full core for the whole stall where the wait burned almost none, both delivering the same bytes.

## Enumerating the class

Swept every `console.log`/`console.error` in the file that precedes a `process.exit()` on any path:

| site | state |
|---|---|
| verdict payload, stdout | **fixed** — the ticket's subject |
| verdict summary, stderr | **fixed** — same exit, unbounded growth via `reasons`, and the channel the short write actually cut |
| `emitRateLimited()` payload, stdout | already `writeSync`; now shares the helper |
| `die()` in `arg.mjs` | `writeSync`, but unlooped — carries the same short-write defect, filed as #889 |
| `vlog` diagnostic stream | **deliberately left** |

`vlog` is left because it is `--quiet`-gated, nothing acts on it, and the script's own contract says a caller loses nothing it decides on when it goes. Converting it would mean reimplementing `console.error`'s varargs formatting across its call sites for output no consumer parses.

`die()`'s site is `arg.mjs`'s, not this ticket's, and its exit code survives regardless — it loses the message, not the verdict. Measured and filed rather than fixed here, together with the open question of whether the guard should be hoisted into `arg.mjs` and shared.

## Verification

Pins on a real OS pipe. `spawnSync`'s default stdio is one — measured, the same payload reaches a file fd whole and a pipe cut, so a harness capturing to a file would have passed whether or not the script was ever fixed.

- a not-green payload past one buffer under `--quiet`, where `reasons` alone does the growing
- a green payload past one buffer still exiting 0, which is the direction this fix could newly break
- the bytes on a payload that never reaches the buffer, pinned exactly rather than as "parses", so a missing or doubled newline cannot pass
- the verdict line on stderr past one buffer, asserted whole by content — it must reach the last job it names — before any size guard, because a cut line is exactly one buffer long and would fail a size guard while blaming the fixture
- the guard executed, not merely present: the child's stdout is `/dev/null` opened read-only, so the first `writeSync` raises `EBADF` deterministically instead of by racing a reader
- the loop itself, as a source shape through `stripComments()`

Each new test was mutation-checked against a copy outside the worktree, and each mutation names only what it should:

| mutation | result |
|---|---|
| revert `emit()` to a single `writeSync`, keep the catch | stderr pin and shape pin red |
| drop the guard entirely | guard test red, with the stdout survival pin |
| catch rethrows instead of returning | **only** the guard test red |
| reword and add a comment line inside `emit()` | green — the control that must not fire |
| correct shape parked in a block comment, real `emit()` reverted | shape pin still red — `stripComments()` refuses the escape |

## Known limit

The `EAGAIN` throw itself does not reproduce on darwin; on this platform the same saturated fd short-writes instead. The guard test therefore forces `EBADF` through the same arm rather than racing for `EAGAIN`, which is how `arg.test.mjs` already pins `die()`'s structurally identical guard. The `EAGAIN` retry arm rests on the Linux measurement from #299/#322.

## Side effect worth recording

Folding stderr into the payload (`2>&1`) is usage `run-team/SKILL.md` forbids outright, and it was separately measured to truncate the payload to one buffer and lose it entirely. Under the loop that pipeline now delivers the payload whole and parseable — 462648 bytes against 65536 before, exit code unchanged. The prohibition stands; the failure mode behind it is gone.

Closes #876


feigi:
Review deferrals filed.

| # | subject | label | dimension |
|---|---|---|---|
| #889 | the `writeSync` short-write fixed here is still live in `arg.mjs`'s `die()` and `staleness.mjs` | `ready-for-agent` | `types` |
| #890 | the rate-limited payload's trailing newline is unpinned — dropping it leaves the suite green | `ready-for-agent` | `types` |

Both deduped through `ledger.mjs check` before filing: clean ledger, tracker searched, no match.

#889 carries a correction to the finding it came from. That finding asserted a sibling site in `ledger.mjs` was "confirmed still open and unfixed"; that half is no longer true — it is #808, fixed by PR #883. Only the architectural half survives, now with a measurement behind it: `die()` cut its message at one pipe buffer and lost its trailing newline where the looped write delivered all of it, exit 2 intact in both.

Two review findings were **applied** rather than deferred: the short-write in `emit()` (the loop), and the missing coverage on `emit()`'s guard. The rest were refuted — including, on measurement, the proposal to revert `emit()` in favour of `process.exitCode`: its headline advantage was that it un-truncated stderr, and the loop already does that, so at `07fd709` the two are byte-identical on stdout, stderr, verdict line and exit code in the `--quiet` mode that matters.

One deferral the review expected did **not** need filing. The `2>&1` shared-fd truncation is fixed as a side effect of the loop — measured at `f1ac062` the payload was absent in 65536 delivered bytes, and at `07fd709` it parses whole in 462648.

Issue #580, opened 2026-08-18
verify-sha.sh:27 and :35 — fatality is unpinned, because a LATER fatal guard reproduces the same exit-2/null-JSON/stderr signature


feigi:
Surfaced by `impl-574` while implementing #574, from a per-guard sweep it ran rather than inheriting. #574 pinned the **fetch** guard's fatality; two sibling guards in the same file have the identical hole and were deliberately left out of that diff.

Filed by the fleet controller. Guard line numbers **verified against `origin/main`** before filing; the mutation results are `impl-574`'s.

## The class

A guard whose *message* is pinned by a test, but whose *fatality* is not — because downgrading it from `die` to a warning lets execution continue to a **later** fatal guard that produces the same observable signature (exit 2, no JSON on stdout, and the earlier guard's own message still on stderr). Every assertion the suite makes still holds, so the mutant survives.

This is the defect #574 fixed for `verify-sha.sh:24`. It is not specific to that guard.

## Per-guard sweep of `verify-sha.sh` (measured by `impl-574`: each `die` downgraded to a warning one at a time, full suite re-run each time)

| guard | line on `origin/main` | fatality pinned? |
|---|---|---|
| argc | `14` — `[ $# -eq 2 ] \|\| die "usage: verify-sha.sh <branch> <sha>"` | YES — killed by *a wrong argument count is exit 2* |
| fetch | `24` — `\|\| die "cannot fetch origin/$branch"` | **was NO** — fixed by #574 / PR #579 |
| **rev-parse** | `27` — `\|\| die "origin/$branch does not resolve after fetch"` | **NO — mutant survived 9/9** |
| **cat-file -e** | `35` — `\|\| die "cannot resolve $sha to a commit in this repository"` | **NO — mutant survived 9/9** |
| merge-base status | `47` — `[ $? -eq 1 ] \|\| die "git merge-base --is-ancestor failed …"` | YES — killed by the #266 test |

## Mechanism, per site

- **`:27` advisory** → `tip` is empty, the script continues, and the `cat-file -e` guard at `:35` kills it. Exit 2, null JSON, and the echoed `origin/$branch does not resolve after fetch` all still present.
- **`:35` advisory** → `merge-base --is-ancestor` on an unknown object exits **128**, and the `[ $? -eq 1 ]` guard at `:47` kills it. Same signature again.

Each is masked by a later fatal guard, so its existing test pins the **message**, not the **fatality**.

## The shape of a fix that works

[...]

PR #891, opened 2026-08-25
test(fleet): the payload-escaping guard's downgrade lied exit 1, unseen

### LINKED ISSUE #884: verify-sha.sh: the JSON-escape guard is unpinned AND fails OPEN — downgraded it fabricates the exit-1 "not reachable" verdict for a sha that IS reachable
Surfaced twice independently: by `impl-580` during a per-guard sweep while implementing #580, and by the `correctness` dimension of the PR #882 review. Filed by the fleet controller.

**Filing correction, 2026-08-25.** This issue was first filed with the wrong failure mechanism, taken from the implementer's sweep and not re-derived. The PR #882 review measured it with a purpose-built fixture and its account is the correct one. The original text is superseded, not merely refined — the exit code is different, and so is what the caller does with it.

## The defect

`verify-sha.sh`'s payload-escaping guard — `|| die "could not escape the payload fields for origin/$branch"` — has no test pinning its fatality. Downgrading its `die` to a message-preserving warning leaves the whole `verify-sha.test.mjs` suite green.

## Mechanism — measured, and NOT what this issue first said

The guard reads:

```sh
branch_j=$(jstr "$branch") && sha_j=$(jstr "$sha") && tip_j=$(jstr "$tip") \
  || die "could not escape the payload fields for origin/$branch"
printf '{"branch":"%s","sha":"%s","reachable":%s,"tip":"%s"}\n' \
  "$branch_j" "$sha_j" "$reachable" "$tip_j"
```

When `jstr` fails, the `&&` chain short-circuits and `sha_j` / `tip_j` are **never assigned**. With the `die` downgraded, execution reaches the `printf`, which references them, and `set -eu` aborts the shell.

Measured by the review against a real fixture (bare origin + clone, one pushed commit, a `sed` shimmed to exit 1 first on PATH, since `jstr`'s escaping is a `sed`/`tr` pipeline), for a sha that IS genuinely reachable:

| | exit | stdout | stderr tail |
|---|---|---|---|
| unmutated | **2** | empty | `verify-sha: could not escape the payload fields for origin/main` |
| mutant | **1** | empty | `… IS reachable on origin/main` / `sed: outage` / `sha_j: unbound variable` |

**Exit 1, not exit 0.** This issue originally claimed the `printf` runs with empty fields and emits a malformed payload at exit 0. That describes the *rejected* design the comment block above the guard warns against — an inline `$()` in the `printf` argument list, which would sit outside the `|| die` and contribute an empty argument. It is not what the current code's mutant does.

## Why the corrected mechanism is worse than the one first filed

`verify-sha.sh`'s contract is 0 = reachable, 1 = NOT reachable, 2 = unanswerable. A malformed payload at exit 0 would be a parse failure the caller notices. **Exit 1 with empty stdout is a well-formed, confident, wrong verdict.**

`run-team/SKILL.md` reads exit 1 as "the SHA is not on the expected branch" and prescribes: flag, do not enqueue, do not return the ticket to the pool until the maintainer rules. So the regression this missing pin would allow costs a member its ticket and demands a human ruling, for a SHA that was on the branch all along.

That also distinguishes it from #580's two guards, which fail **closed** — downgraded, each still reached exit 2 with no JSON, so the observable signature was unchanged.

## Suggested shape

The review wrote the fixture already: shim `sed` on PATH to exit 1 for a genuinely reachable sha, then assert exit 2, null JSON, and `doesNotMatch(stderr, /unbound variable/)`. The progress marker is the `IS reachable on origin/<branch>` echo, already pinned verbatim by *a healthy run stays quiet*. Same shape as the `cat-file` test PR #882 adds.

Derive the assertion from the tree at pickup rather than copying it from here.

## Also unpinned, and benign — footnote, not scope

`. "$json_lib" || die "$json_lib failed to load"` survives the suite too. The review measured why it is harmless: with `json.sh` unloaded, `jstr` is undefined, so the escaping guard above fires and the script still exits 2. Record it; it does not need its own fix.

## Worth sweeping

`json.sh`'s sibling consumers — `inflight.sh`, `prove-merge.sh`, `no-undo-audit.sh`, `release-ticket.sh` — also read exit 1 as a verdict, so the same site in each deserves the same check. Not scoped here.

## Out of scope

- `verify-sha.sh`'s rev-parse and `cat-file -e` guards — covered by #580 / PR #882.

feigi:
Closes #884

## What was unpinned

`verify-sha.sh`'s payload-escaping `|| die` had no test over its fatality. Every case around it asserts what a **working** escaper produces, so none of them reaches that guard. Measured against `origin/main`: downgrading it to a message-preserving warning left `verify-sha.test.mjs` entirely green.

## The mechanism, re-derived here

The ticket was filed once with the wrong mechanism and corrected; this PR re-derived the corrected one rather than inheriting it. Fixture: bare origin + clone, one pushed commit, `sed` shimmed to exit 1 on PATH (`jstr` escapes through a `sed`/`tr` pipeline), for a sha that **is** genuinely reachable.

| | shell | exit | stdout | stderr tail |
|---|---|---|---|---|
| unmutated | `/bin/sh` (macOS bash 3.2.57) | **2** | empty | `verify-sha: could not escape the payload fields for origin/main` |
| unmutated | `/bin/dash` | **2** | empty | `verify-sha: could not escape the payload fields for origin/main` |
| guard downgraded | `/bin/sh` (macOS bash 3.2.57) | **1** | empty | `… IS reachable on origin/main` / `sed: outage` / `sha_j: unbound variable` |
| guard downgraded | `/bin/dash` | **2** | empty | `… IS reachable on origin/main` / `sed: outage` / `sha_j: parameter not set` |

The escaped fields are assigned by one `&&` chain, so the first `jstr` that fails leaves the rest unassigned; with the guard advisory the `printf` it protects reads an unassigned name under `set -u` and the shell aborts. Under a bash-family `sh` that lands on **exit 1 with empty stdout**, not the malformed payload at exit 0 the shape of the code suggests — and exit 1 is the code this script's contract reserves for "the sha is NOT reachable". `run-team/SKILL.md` answers that by flagging the member, withholding the enqueue and holding the ticket until a maintainer rules. So the missing pin allowed a confident wrong verdict rather than a parse failure the caller would notice.

## What ships

- **`an escaper that cannot run is exit 2, never the exit 1 that means `not reachable`.** The guard has no progress marker of its own — nothing is printed between the escaping and the `printf` it protects — so the `IS reachable` trace and the shim's own line bracket it instead, placing the failure at this guard and nowhere else. Both integrity assertions run before the verdict, so "the shim broke something else" and "the guard is not fatal" cannot be reported as each other.
- **`a shadowed `sed` that works is answered normally`.** The accept-side leg: same fixture, same shadowed name, passthrough body, byte-identical ordinary payload at exit 0. Without it the case above could be measuring a PATH it broke wholesale and still read green.

## Every guard in the file, measured

Each `die` downgraded to a message-preserving warning, one at a time, against a scratch copy:

| guard | before this PR | after |
|---|---|---|
| `[ -r "$json_lib" ]` | killed by `a missing json.sh is exit 2` | unchanged |
| `. "$json_lib"` | **survives** | **left, deliberately** — see below |
| `[ $# -eq 2 ]` | killed by `a wrong argument count is exit 2` | unchanged |
| `git fetch` | killed by `a failed fetch is fatal` (#574) | unchanged |
| `git rev-parse` | killed by `a rev-parse … is fatal` (#580) | unchanged |
| `git cat-file -e` | killed by `a sha the cat-file guard rejects is fatal` (#580) | unchanged |
| merge-base status | killed by `a merge-base that fails … is exit 2` (#266) | unchanged |
| payload escaping | **survives** | **killed by this PR** |

`. "$json_lib" || die` is left because its benignity was verified rather than inherited from the ticket. With `json.sh` as a directory — the case that passes `[ -r ]` and still fails `.` — and that `die` downgraded: exit 2, empty stdout, `jstr: command not found`, and the escaping guard above catches it. Same observable signature as unmutated, so it has nothing of its own to pin.

Out of scope and confirmed already landed: the rev-parse and `cat-file` guards (#580 / PR #882). Not swept: `json.sh`'s sibling consumers, which the ticket names and scopes out.

## Mutation evidence

- **Both shells, corrected case.** Downgraded guard: red under `/bin/sh` (the exit-code assertion) and red under `/bin/dash` (the `sha_j` assertion), one test failing in each. Unmutated: green under both. Guard `die` message reworded — the change the pin must **not** catch: green under both. Run against a scratch copy; the worktree never held a mutated script. For contrast, the case as first written: red under `/bin/sh`, **green under `/bin/dash`** with the mutant in place.
- **Kills its own mutant, and only it.** Escaping guard downgraded → `an escaper that cannot run …` reds, 17/18 pass. Every other guard's mutant reds only its own owner; neither new pin reds under any of them.
- **Neutral control.** The `&&` chain's operands reordered — behaviour preserved — 18/18 green.
- **Reword control.** Every `die` message in `verify-sha.sh` reworded, the escaping guard's own included: **both new pins stay green.** They name no `die` string; their brackets are the `IS reachable` echo (pinned verbatim by `a healthy run stays quiet`), the shim's own line, and the `sha_j` the shell's nounset abort names. The pre-existing cases that do name a `die` string red under that reword, which is their existing design and untouched by this PR.

## The pin was measured on one shell only — corrected in review

The case as first written asserted `doesNotMatch(stderr, /unbound variable/)` beside its exit-2 assertion, and the rationale block stated the exit-1 abort as an unqualified measured fact. The table above is why that was wrong: `verify()` spawns a bare `sh`, and ci.yml's `check` job runs on `ubuntu-latest`, where that name resolves to **dash**. On the runner that gates the merge the mutant exits **2** — the status a firing guard returns — and says `parameter not set`, so the exit-code assertion and the wording assertion were **both satisfied by the mutant**. The pin did not discriminate on the platform that runs it, and the implementer's mutation check could not see that because macOS `/bin/sh` is bash.

`sha_j` is the shell-agnostic discriminator: it is named in the nounset abort and nowhere else. No `die` message carries it, and it is absent from stderr on a healthy run, on a genuine `not reachable`, on an ordinary `die` and on this guard firing — measured under `/bin/sh` and `/bin/dash`. The assertion now matches the name; the rationale block states its measurement per shell rather than generalizing one.

## Tests

`./agent-test skills/fleet/scripts` — **1146 tests, 1146 pass, 0 fail**, `verify-sha-exit2-prose.test.mjs` included.


feigi:
## Review findings — applied / deferred

**Applied (1, the survived finding):** `verify-sha.test.mjs` — the `#884` pin discriminated on bash and waved the mutant through on CI. Assertion now matches `/sha_j/` rather than `/unbound variable/`; the rationale block states its measurement per shell. Commit `2961885`.

**Deferred and filed:**

| finding | where | outcome |
|---|---|---|
| escaper exiting 0 with empty output fabricates a payload at exit 0 | #893 (`needs-triage`) | refuter: mechanism reproduces on both shells, reachability refuted, proposed remedy sentence factually wrong |
| `worktree-audit.sh:194` escape guard's fatality unpinned | #894 (`ready-for-agent`) | out of #884's scope; re-measured here, mutant green on both shells |
| `verify-sha.test.mjs:47` "the one test that shadows `git`" — two tests use it | #895 (`ready-for-agent`) | pre-existing, untouched by this diff |
| `printf-die-sweep.test.mjs:97` cites `verify-sha.sh:33`; the trace is at `:42` | comment on #516 | pre-existing; #516 is the open ticket collecting exactly this class |

## Mutation evidence, both shells

`sh` is `/bin/sh` (macOS bash 3.2.57); `dash` is a PATH where bare `sh` resolves to `/bin/dash`, which is what `ubuntu-latest` gives the bare `sh` these suites spawn. Mutants applied to a scratch copy; the worktree never held a mutated script.

| `verify-sha.sh` | test | `/bin/sh` | `/bin/dash` |
|---|---|---|---|
| unmutated | as first written | `pass 18 fail 0` | `pass 18 fail 0` |
| guard downgraded | **as first written** | `pass 17 fail 1` | **`pass 18 fail 0`** ← the defect |
| unmutated | corrected | `pass 18 fail 0` | `pass 18 fail 0` |
| guard downgraded | corrected | `pass 17 fail 1` | `pass 17 fail 1` |
| guard `die` message reworded | corrected | `pass 18 fail 0` | `pass 18 fail 0` |

The reworded-`die` row is the change the pin must **not** catch. Under `/bin/sh` the mutant reds on the exit-code assertion; under `/bin/dash` on the `sha_j` assertion — one test failing in each, and it is the same test.

## Why `sha_j` and not a wording

`sha_j` is named in the nounset abort and nowhere else. Measured on a real fixture under `/bin/sh` and `/bin/dash`, it is absent from stderr on a healthy run, on a genuine `not reachable`, on an ordinary `die` and on this guard firing; present in the mutant's abort under both. bash writes `sha_j: unbound variable` at exit 1, dash `sha_j: parameter not set` at exit 2 — the name is the part they share.

## Gate

`node --test skills/fleet/scripts/*.test.mjs` from the worktree: `tests 1146 / pass 1146 / fail 0`.

PR body reconciled against the final diff: the mechanism table is now per-shell, the reword control no longer rests on a wording, and the correction has its own section.

Issue #889, opened 2026-08-25
fleet scripts: the writeSync short-write #885 fixed in ci-state.mjs is still live in arg.mjs die() and staleness.mjs


feigi:
Deferred from PR #885 review.

**Dimension:** `types`

**file:line:** `skills/fleet/scripts/arg.mjs:54` (`makeDie`), `skills/fleet/scripts/staleness.mjs:117`

## The defect

PR #885 fixed `ci-state.mjs`'s `emit()` to consume `writeSync`'s return value, because a single `writeSync` **short-writes** on a pipe whose fd has had a stream initialised for it: it returns the count it managed and throws nothing at all, so a `try`/`catch` around it never fires and the message is cut with no diagnostic.

Every remaining `writeSync` site in `skills/fleet/scripts/` still calls it once and discards the count, so the class is live in both of them:

- `arg.mjs:54` — `die()`'s refusal message.
- `staleness.mjs:117` — the verdict JSON on stdout.

`ci-state.mjs` is now the only one that loops.

## Measured

`die()` driven with an fd 2 that a `console.error` had already touched, against a reader asleep three seconds, versus the loop shape writing the identical bytes:

```
child.mjs   (arg.mjs die())  -> EXIT=2  stderr_bytes=65536   ends_with_newline=False
control.mjs (looped write)   -> EXIT=2  stderr_bytes=200017  ends_with_newline=True
```

`die()` stopped at one pipe buffer and lost its trailing newline. The exit code is **intact** in both — this is message loss, not the #299/#328 exit-code inversion, which is why it is lower severity than the bug #885 fixed.

## Reachability

Latent rather than live. No `die()` message in `skills/fleet/scripts/*.mjs` interpolates unbounded text today — `ci-state.mjs`'s own `run()` deliberately names the cause rather than quoting the child's stderr. It becomes live the moment a caller forwards a large message through `die()`. `staleness.mjs:117` writes a JSON verdict whose `needle` and `extra` fields are caller-supplied, so its exposure depends on the caller.

## Why deferred, and the remedy

Out of #876's scope — that ticket is `ci-state.mjs`'s verdict payload, and #885 touches only that file and its test.

[...]

PR #892, opened 2026-08-25
refactor(fleet): a same-size near-miss was walked twice for one answer

### LINKED ISSUE #888: ledger.mjs isMatch runs the identical subset scan twice on same-size near-misses of four tokens or more
Deferred from PR #883 review (issue #808). **Deferred as out of scope, not refuted** — the defect is confirmed. #808 is about `check`'s already-filed exit abandoning its payload on a pipe; this is a redundant scan in relocated matching code. PR #883 moved and re-indented `isMatch` into `runCheck()` without authoring it, and its own review finding notes that. Applying it there would have added a behavioural edit to a PR whose point is that the extraction is pure.

`dimension: simplify` · severity: `suggestion` · site: `skills/fleet/scripts/ledger.mjs:269-273` (position re-derived after PR #883's re-indent; the review pinned it pre-indent)

## The defect

```js
const isMatch = (filedTokens) => {
  if (filedTokens.size === target.size && [...filedTokens].every((t) => target.has(t))) return true;
  const [small, big] = filedTokens.size <= target.size ? [filedTokens, target] : [target, filedTokens];
  return small.size >= 4 && [...small].every((t) => big.has(t));
};
```

When the two token sets are the same size, the destructure takes the `<=` branch, so `small` is `filedTokens` and `big` is `target` — and the final `every` is then byte-for-byte the scan the first line already ran, re-run on an answer already known to be false.

## One correction to the finding as submitted

The finding says this happens "whenever the two token sets are the same size". It is narrower than that: `small.size >= 4` short-circuits ahead of the second scan, so equal-sized sets below four tokens never re-scan. Instrumented with a counting wrapper:

```
equal size 4, non-match   current=2 scans   proposed=1
equal size 4, match       current=1 scans   proposed=1
equal size 3, non-match   current=1 scans   proposed=1
```

So the condition is: sizes equal, **at least four tokens**, and the subset test failing.

## Proposed remedy — verified equivalent, not yet applied anywhere

```js
const isMatch = (filedTokens) => {
  const [small, big] = filedTokens.size <= target.size ? [filedTokens, target] : [target, filedTokens];
  if (![...small].every((t) => big.has(t))) return false;
  return small.size === big.size || small.size >= 4;
};
```

Same two acceptance rules — subset-and-equal-sized at any size, or subset-and-at-least-four — each stated once. Equivalence checked exhaustively rather than argued: every subset of a six-token universe as `target` against every subset as `filedTokens`, `cases=4096 disagree=0`.

The remedy has **not** been run against the suite by this deferral; the equivalence proof and the scan counts are what is settled here. Whoever picks this up should run `node --test skills/fleet/scripts/*.test.mjs` with it applied.

## Why it is worth doing at all

Small: the scan is over a filed row's token set, and the redundant pass happens only on same-size near-misses of at least four tokens. The value is that the two acceptance rules currently read as three conditions across three lines, one of which is dead on the path that reaches it.

feigi:
Closes #888.

## What was wrong

`isMatch()` in `ledger.mjs` tested set equality with its own subset walk, then fell through to a second walk over the smaller set. When the two sizes are equal the destructure makes the filed set the smaller one, so the second walk is the first walk again, re-deriving an answer already known — and the four-token floor guarding it is dead on the path that reaches it, because equality has already been settled by then.

The ticket's own correction to the original finding is **confirmed by measurement, not restated on trust**: the repeat is not "whenever the two token sets are the same size". Instrumenting the real predicate, the repeat needs sizes equal *and* at least four tokens *and* the subset test failing — below the floor the short-circuit stops ahead of the second walk, and a match returns before reaching it.

## The remedy — deliberately not the shape the ticket proposes

Both acceptance rules are size tests, so this states them once as a guard on a single walk:

```js
const [small, big] = filedTokens.size <= target.size ? [filedTokens, target] : [target, filedTokens];
return (small.size === big.size || small.size >= 4) && small.isSubsetOf(big);
```

The ticket proposes walking first and testing the sizes after. That removes the repeat, but adds a walk everywhere the walk was previously skipped: a pair of unequal sizes whose smaller side is under the floor is refused today **without walking at all**, because the floor short-circuits ahead of the spread. Ordering the size tests first keeps that skip and still collapses the repeat.

Measured over every pair of subsets of an eight-token universe (65025 cases): both shapes agree with the original on every case, `disagree=0`. Counting walks over the same space, the ticket's shape raises the total from 39275 to 65025; this shape lowers it to 30553. It is the only one of the three that never walks more than the original on any input, and it states the two rules once each — the readability the ticket is actually asking for.

## Both directions of the predicate

A wrong `isMatch` fails in two different ways, and they are not symmetric: a false accept files nothing and silently loses a finding, a false refuse files a duplicate someone closes. Two pins, each killing its own mutant and only it:

- **must not accept** — a same-size four-token near-miss differing in one token. This is exactly the mode the repeat lived on. Accepting without the subset walk reds it.
- **must accept** — equal token sets below the subset floor. The floor is a condition of the subset rule alone, so equal sets qualify without it. Dropping the equality rule reds this pin **and nothing else in the suite** — the gap was real, and folding two rules into one expression is precisely the edit that loses it.

A third pin was written and then removed: a same-size *match* satisfies both rules at once, so no single-rule mutation can red it. It discriminated nothing and an existing test already covers an equal-set match.

## Review follow-ups

Two review findings were verified and applied on top of the original diff:

- The subset test asked its question by spreading the smaller set into an array for `.every()`. `Set.prototype.isSubsetOf` answers it off the set itself, walking the same set once without the copy. Equivalence measured over every ordered pair of subsets of an eight-token universe, unequal sizes and the empty set included, with no divergence from either the spread form or the pre-merge predicate; the mutants that reach this line are caught by the same tests under both forms, so the swap moves no coverage. It raises this script's runtime floor to Node 22, which `.nvmrc` and CI already clear.
- The rationale comment claimed the old form "re-walked a same-size near-miss". It re-walked only those that cleared the subset floor — below it the second branch was gated away before walking, so such a pair was walked once under the old form too. Narrowed to the case that re-walked.

The review's remaining findings were measured and deferred rather than applied. Two proposed opposite edits to the same-size refusal test and were refuted together (#897); one proposed trimming the rationale this diff narrows, and was refuted because the edit it guards against is behaviour-preserving and therefore untestable (#898); one was self-declared not actionable and is noted on #889 instead. The same measurement surfaced a gap on the other side of this predicate — the subset floor is pinned only from below — filed as #899 and not addressed here.

## Verification

- Equivalence re-proved from scratch rather than inherited from the ticket — 65025 cases, `disagree=0`.
- Instrumented the shipped predicate to confirm the suite reaches the same-size mode at all: it does, but only through rows built for other questions (a stoplist scoring case, and the pipe-payload test whose repeated words dedupe to a set the size of its subject). Nothing asserted the boundary before this.
- Mutation matrix over five semantic mutants and two benign rewrites: every semantic mutant reds, both benign rewrites stay green.
- `./agent-test skills/fleet/scripts/` — **tests 1146, pass 1146, fail 0**.


feigi:
Review adjudication, pushed as `22446f6`. Every finding arrived in the `unverified` band as a `suggestion`, which the workflow budgets no refuters — so each was checked here before anything was touched.

**Applied**

- `simplify` — the subset test spread the smaller set into an array to run `.every()`; `Set.prototype.isSubsetOf` answers it off the set. Survived refutation: equivalence over every ordered pair of subsets of an eight-token universe with no divergence, the full suite green, and the mutants reaching this line caught by the same tests under both forms.
- `correctness` — the rationale claimed the old form "re-walked a same-size near-miss"; it re-walked only those clearing the subset floor. Narrowed to the case that re-walked.

**Deferred, with the refutation recorded**

- #897 (`wontfix`) — the same-size refusal test drew two opposite proposals: delete it as redundant, and extend it with a `verdict` assertion. Sent to one refuter as a single ruling, since they cannot both hold of one assertion set. Both refuted. Under the fold-the-rules mutant the test reds beside tests that die on a precondition rather than their subject; reshaping only those fixtures leaves it the sole red, so it is not redundant. The `verdict` property belongs to the tracker arm, which owns it already, and this test returns before `gh` is ever spawned.
- #898 (`wontfix`) — trimming the rationale as changelog. The supporting claim that nothing pins the prose is true; that is why the trim loses. The edit those lines guard against is behaviour-preserving, so no test can catch it — restoring the pre-PR body under the trimmed comment leaves the suite green.
- #889 (comment) — hoisting `isMatch` to a named top-level function. Self-declared not actionable; recorded there as a measured negative rather than filed as a rival issue, since it has one definition and one call site.

**Surfaced by the measurement, not by the review**

- #899 (`ready-for-agent`) — the subset floor is pinned only from below. Raising it (`>= 4` → `> 4`, and `>= 5`) survives the entire suite, because no fixture builds a strict subset of exactly four tokens: the test named for that case has a subject that normalises wider. Out of scope for this diff, filed with the remedy open.

Full suite green before each commit — `tests 1146 / pass 1146 / fail 0`.

PR #903, opened 2026-08-25
test(fleet): pin the subset floor from above, where nothing was watching it

### LINKED ISSUE #899: the subset floor acceptance boundary is unpinned above four tokens
Deferred from PR #892 review (issue #888) — surfaced by the adversarial measurement, not by any of the review's own findings. Dimension: `tests`.

`isMatch` in `skills/fleet/scripts/ledger.mjs` accepts on either of two rules: equal token sets at any size, or a subset from at least four tokens. PR #892 pinned both rules at the boundary where the *other* cannot cover for them. The subset floor is nonetheless pinned in one direction only.

**Lowering the floor is caught. Raising it is not.**

```
# raise the floor: >= 4  ->  > 4
$ perl -pi -e 's/small\.size >= 4/small.size > 4/' skills/fleet/scripts/ledger.mjs
   ℹ tests 1146 / pass 1146 / fail 0     <-- mutant SURVIVES the whole suite
```

`>= 5` survives identically. Lowering it (`>= 3`) reds several tests, so the floor is anchored from below and free to drift upward.

## Why the mutant survives

No fixture in the suite builds a strict subset of *exactly* four tokens. The test named for that case, `a strict four-token subset of a filed row is still a match, exit 1`, is named for the literal four-token phrase in its subject, but the subject normalises wider:

```
$ node -e 'const norm=(s)=>s.toLowerCase().replace(/[^a-z0-9]+/g," ").trim();
  const ts=(s)=>new Set(norm(s).split(/\s+/).filter(Boolean));
  console.log(ts("Non-zero column audit 11 rows").size)'
6
```

Six tokens against a fifteen-token filed row — comfortably above the floor, so it keeps matching when the floor rises. Every other accepting fixture sits further above it.

The consequence is the direction that matters: `check`'s bias is deliberately asymmetric, because a false "already filed" silently loses a finding while a false "not filed" only creates a duplicate someone closes. A floor that drifts upward turns real near-duplicate rewordings into "not filed" — the cheap direction — but it also means the acceptance boundary the ticket set can move without anything noticing.

## Remedy (open)

Add an accepting fixture whose smaller token set normalises to exactly four tokens and is a strict subset of the filed row, so `> 4` and `>= 5` both red. The existing same-size refusal test covers the refusing side of the same boundary and should stay as it is.

Confirm both directions on whatever is added: the new pin must red under `>= 5` and stay green under a change it should not catch.

feigi:
`isMatch` in `skills/fleet/scripts/ledger.mjs` accepts on either of two rules — equal token sets at any size, or a subset from at least four tokens. The refusal beneath the floor anchors it from below. Nothing anchored it from above: raising the floor left the whole suite green, so the acceptance boundary the guard sets was free to drift upward unobserved.

Reproduced on this tree before changing anything. Raising the floor by either spelling — dropping the equal case from the comparison, or moving the constant up one — left every test passing. Lowering it reds existing tests, which is what made the asymmetry visible rather than the floor simply being unpinned.

The reason nothing noticed: no fixture built a strict subset sitting exactly at the floor. The test named for that case has a subject whose punctuation normalises it wider, and the other accepting fixtures are equal sets — four tokens against four, and three against three — which match under the equal-size rule that the floor does not gate, so a raised floor kept matching everywhere the suite could see. (Issue #899's own body says these sit above the floor; measured, they do not, and that is why nothing noticed.)

## What this adds

An accepting fixture whose subject normalises to exactly the floor and is a strict subset of a longer filed row. Strictness is the load-bearing part: an equal pair qualifies under the equal-size rule too, so the floor would no longer be what decides and a mutation of it would have nothing to move.

Review then found another member of the same class equally unpinned, and it is carried here as well — see **The ordering member**.

## The rest of the predicate

Enumerated and measured rather than assumed, since the floor is one boundary among several the expression carries:

- **Subset floor, lowered** — already pinned, by the existing below-floor refusal.
- **Subset floor, raised** — this PR.
- **Equal-size rule removed** — already pinned, by the existing below-floor acceptance of an equal pair.
- **Equal-size rule widened until it is always true** — already pinned, by the existing below-floor refusal.
- **Subset conjunct dropped** — already pinned, by the existing same-size near-miss refusal.
- **Pair ordering dropped** — this PR. Absent from this enumeration as first written, which is the part worth flagging: the list was offered as complete and was not.

Each was mutated and observed to red. The raised floor and the dropped ordering are the members that survived unpinned, so those are the ones that get a test here.

## The ordering member

`isMatch` orders the pair by size before either rule reads it, so both rules always measure the smaller set. Hardcoding that ordering away — the pair taken as `(subject, filed)`, whichever set arrives first — left the whole suite green.

What that leaves uncovered is the mirror orientation: a filed row that is a strict subset of a **longer** checked subject. Six fixtures already build that orientation — one is even named "a filed row contained in a longer subject scores on the smaller set, not the union" — but none of them reaches a *match* through it, so the ordering never decides their answer — that is the narrowing that matters, and it is exactly what the mutation establishes: dropping the ordering reds this test alone. Measured separately, each of the six is refused at three tokens against the four-token floor; that is a stronger claim than the mutation supplies, so it is cited here rather than asserted in the test comment, which comments only on what the mutation settles.

The `#NNN` strip is not what makes the orientation reachable — disabling it leaves the filed row smaller than the subject either way. What it decides is whether that row is a **subset**, since the issue number otherwise survives as a token the subject never carries, which is what `ledger.mjs`'s own rationale says.

Its fixture sits clear of the floor rather than on it. At the floor a raised floor reds it too, and it would then be pinning a boundary its neighbour already owns instead of the ordering.

The `#888` section header claimed the tests in its fence pin each acceptance rule. The ordering is not either rule, and the empty-subject usage failure inside that same fence pins neither — it dies on the empty-subject guard before `isMatch` is reached. Nor does the same-size refusal: at four tokens against four both size rules pass, and it turns on the subset conjunct alone, which is what it actually pins. The header now names the two acceptances that pin the rules and names both tests that pin neither, so every test in the fence is accounted for rather than covered by a bound the reader has to infer.

Two further clauses in that block were corrected with it. "An equal set under the floor is **the** acceptance only one rule can explain" was true while it was the only such acceptance; the strict subset at the floor is a second, resting on the subset rule alone, so the article is now indefinite. And "above the floor the ordering is the only thing left that can move it" contradicted the strip-disabled measurement recorded below, so the comment now says what this body says.

## What the new pins do not catch

Deliberately narrow, and checked in that direction too.

The floor pin stays green when the floor moves the other way — that direction belongs to the refusal test — when the near-miss reporting cap changes, and when the issue-number strip on a filed row is disabled.

The ordering pin stays green under a raised floor, which is what keeps the two separable: measured, a raised floor reds the floor pin alone, and dropping the ordering reds the ordering pin alone. It does red when the issue-number strip is disabled, alongside the equal-sets test — that strip is what lets the filed row be a subset at all, so the sensitivity is the point rather than a leak.

A match returns before the tracker query is built, so nothing in term selection or near-miss scoring is on either path.

Both fixtures are plain words on both sides, carrying no punctuation for `norm()` to fold, so the token counts the pins rest on cannot shift with it. The floor fixture was additionally checked against several folding variants, and holds where the pre-existing fixture's count moves.

Closes #899


feigi:
Closing and immediately reopening to resync GitHub's PR object.

The branch ref advanced to `221f4e9` after a successful `gh pr update-branch --rebase`, but this PR object stayed on the pre-rebase head `8bc2cc4` — confirmed on both REST and GraphQL, with `mergeable_state: unknown` and zero CI runs created for the new head. No `synchronize` event fired, so nothing can bind and the merge gate cannot evaluate it.

The branch content is correct and unchanged by the rebase (`git cherry` against the branch returns all `-`; `merge-tree` clean). This is purely to make GitHub re-read the ref.

feigi:
Superseded by #908 — same branch, same commits, same head `221f4e9`, nothing re-authored.

This PR could not be reopened: the `gh pr update-branch --rebase` that advanced the branch **orphaned** the head this PR records (`8bc2cc4`), and GitHub cannot reopen a pull request whose recorded head is unreachable from the branch. Confirmed with `git merge-base --is-ancestor 8bc2cc4 origin/test/899-subset-floor-boundary` → not an ancestor.

Recorded for the next person who hits a PR/branch desync: **check head reachability before closing.** Close-and-reopen is the usual remedy and is normally reversible, but it is not reversible once a rebase has orphaned the recorded head — the close succeeds and the reopen cannot. Opening a replacement PR from the same branch is the recovery.

Issue #901, opened 2026-08-25
ci-state.mjs: the stderr verdict summary is the one emit() call site with no trailing-newline pin, and the existing test cannot see one by construction


feigi:
Surfaced by `impl-890` while implementing #890, from a per-call-site sweep it ran rather than inheriting. Filed by the fleet controller.

## The gap

`ci-state.mjs`'s `emit(fd, text)` appends no newline of its own — every call site supplies its own trailing `\n`. That convention now has a byte-exact pin at two of its three call sites:

- verdict payload → stdout: **pinned** (pre-existing).
- rate-limited payload → stdout: **pinned** by #890 / PR #900.
- **verdict summary line → stderr: unpinned.**

## Measured

`impl-890`, against the tree at `698b000`: dropping the trailing `\n` from the stderr verdict summary leaves `ci-state.test.mjs` fully green.

The reason nothing catches it is structural rather than an oversight. The completeness assertion PR #885 added for that line asserts by **content**, after splitting stderr on `\n` — which destroys terminator information by construction. So the existing test cannot see a missing terminator no matter how it is worded.

## Why #890 did not cover it

Scope, and a real difference in shape. #890 is the rate-limited payload on stdout. This line is on **stderr**, which also carries `vlog` traces, so the whole-stream byte comparison #890's pin uses (`stdout === JSON.stringify(payload) + "\n"`) has no equivalent here — there is no single expected byte string for the stream.

It therefore needs its own assertion shape, and its own mutant matrix covering **both** a dropped and a doubled terminator, plus a leading-newline case the stdout sites do not have.

## Note for whoever takes this

Two controls, both cheap, both earned this run:

1. **Prove the mutation was applied.** A `sed -i` whose replacement contains the delimiter applies nothing and every mutant reads green. Separately, shell over-escaping can apply a *different* mutation — `impl-890`'s first doubled-newline mutant wrote `\\n\\n`, an escaped backslash rather than two newlines, and it reddened three tests, which reads like success. Assert the file changed **and** read the diff.
2. **Prove the harness can see an applied mutation.** Mutate a site already pinned and confirm it reds.

Neither substitutes for the other.

Also check both shells. `ci-state.test.mjs` gates the merge on CI, `ubuntu-latest`'s `/bin/sh` is dash, and a sibling PR this run shipped a pin that discriminated under bash and waved its own mutant through under dash. `impl-890` verified #890's pin on both and found identical discrimination; do the same here rather than assuming it carries over.

Issue #895, opened 2026-08-25
verify-sha.test.mjs:47 — the REAL_GIT comment says "the one test that shadows git"; two tests use it


feigi:
Deferred from PR #891 review (issue #884). **Dimension:** `comments`. Severity as filed: `suggestion`.

**file:line:** `skills/fleet/scripts/verify-sha.test.mjs:47`.

## The defect

The comment reads:

```js
// Absolute path to the real git, for the one test that shadows `git` on PATH.
const REAL_GIT = execFileSync("sh", ["-c", "command -v git"], { encoding: "utf8" }).trim();
```

`REAL_GIT` serves more than one test. Measured against PR #891's head:

```
$ grep -n 'REAL_GIT' skills/fleet/scripts/verify-sha.test.mjs
48:const REAL_GIT = execFileSync("sh", ["-c", "command -v git"], { encoding: "utf8" }).trim();
134:    `#!/bin/sh\n[ "$1" = merge-base ] && …\nexec "${REAL_GIT}" "$@"\n`
323:    `#!/bin/sh\n[ "$1" = cat-file ] && …\nexec "${REAL_GIT}" "$@"\n`
```

Enclosing tests, resolved from the file:

- use at `:134` sits in `test("a merge-base that fails rather than answers is exit 2, never reachable:false", …)`, opening at `:111`
- use at `:323` sits in `test("a sha the cat-file guard rejects is fatal — the script stops rather than asking merge-base about it", …)`, opening at `:294`

## Why deferred rather than applied on PR #891

Pre-existing and untouched by that diff — `verify-sha.sh` and both of those tests are unchanged by it. #884 scopes itself to pinning the escaping guard's fatality.

It is nonetheless more conspicuous now: the comment sits directly beside the `REAL_SED` comment PR #891 adds, which says "the git shims below" — plural, and correct.

## Remedy

Change `the one test that shadows \`git\`` to `the tests that shadow \`git\``, or name both. The sibling `REAL_SED` comment already uses the plural form.

Related: #516 (the open question on `file:line` citation rot in these same scripts — a different shape: this is a stale count in prose, not a dead line pointer), #847 (a stale count in a `ci-state.mjs` comment, same class as this one).

Issue #890, opened 2026-08-25
ci-state.mjs: the rate-limited payload's trailing newline is unpinned — dropping it leaves the suite green


feigi:
Deferred from PR #885 review.

**Dimension:** `types`

**file:line:** `skills/fleet/scripts/ci-state.mjs:132` (the `emitRateLimited` payload), pinned from `skills/fleet/scripts/ci-state.test.mjs`

## The gap

`emit(fd, text)` appends no newline of its own, so every call site supplies its own trailing `\n`. That convention is pinned for the main verdict payload by the byte-exact test "a payload that never reaches the buffer is emitted byte for byte as before: one line, one trailing newline", and PR #885 added a completeness assertion for the verdict line on stderr. The **rate-limited** payload at `ci-state.mjs:132` has neither.

## Measured

Each `emit(...)` call site mutated in turn, dropping and then doubling its trailing `\n`, running `node --test skills/fleet/scripts/ci-state.test.mjs`:

- rate-limited payload — drop `\n`: **green**. Double `\n`: **green**.
- verdict payload — drop `\n`: **red**, one named test. Double `\n`: **red**.

The surviving mutation produces genuinely different bytes on the wire — the rate-limited payload came back a byte shorter with its terminator gone:

```
xxd (unmutated) | tail -1   ->  74 22 5d 7d 0a    t"]}.
xxd (mutated)   | tail -1   ->  74 22 5d 7d       t"]}
```

That payload is what a caller polling a rate-limited PR parses, and a missing terminator concatenates it with whatever the caller prints next.

## Why deferred

Out of #876's scope. #876 is the verdict payload's survival across a pipe; this is a different call site whose newline was already unpinned before #885 and is not affected by it. Filed rather than fixed so the PR's diff stays the ticket's.

## Remedy

A byte-exact assertion on the rate-limited payload, in the shape the verdict payload's existing test already uses. The rate-limited path has fixtures in `ci-state.test.mjs` already, so this needs no new harness.

PR #909, opened 2026-08-25
chore(fleet): this run's nine tier rows, and the rebase poll that watched the field that desyncs

feigi:
Close-out for this `/fleet:run-team 2 5` run. Nine tickets merged: #580 #808 #876 #884 #888 #890 #901 #895 #899.

## Left deliberately unlabelled

`ready-to-merge` is a reviewer's label and this PR is controller-authored, so it is the maintainer's call — the #863 and #881 precedent. **Disclosing explicitly that this is not data-only:** it changes `run-merge-bot.md`, a file the fleet executes, not just the metrics rows.

## 1. The rebase poll read the field that desyncs

`run-merge-bot.md` step 1 polled `gh pr view <pr> --json headRefOid` after `gh pr update-branch --rebase` — the one field that desyncs from the branch it is meant to be watching.

Measured on #903: the rebase **landed** and moved the ref to `221f4e9`, while `headRefOid` stayed on the pre-rebase `8bc2cc4` with `mergeable_state: unknown` and no CI run on the new head. The poll read a landed rebase as un-landed for all sixty iterations, then fell through to the local-rebase fallback — which would have replayed commits the remote already carried.

Now polls `git ls-remote` on the branch ref. The `headRefOid` read stays in the `printf`, where its *disagreement* with the ref is the desync signal rather than the gate.

The `[ -n "$pre" ]` guard is load-bearing: `git ls-remote` on a nonexistent ref prints **empty at exit 0** (measured), so a mistyped branch leaves `pre` and `post` both empty, comparing equal forever and reporting "the rebase did not land" while never having looked.

## 2. The precondition the run paid for

Close-and-reopen is the usual remedy for a desynced PR and is normally reversible — but a rebase orphans the recorded head *by construction*, and that is the same event that produced the desync.

#903 was closed without checking. GitHub refused to reopen it three times (`Could not open the pull request`), and recovery was PR #908 from the same branch, same five commits, same head, nothing re-authored — losing the review thread and every label. **The desync itself had already reported the condition.**

Step 1 now requires `git merge-base --is-ancestor <pr_head> origin/<branch>` before any close, with non-ancestor → do not close.

## 3. Both rules pinned, mutation-tested five ways

| mutation | reds |
|---|---|
| revert only the poll line inside the loop | ✔ |
| drop the authoritative-source sentence | ✔ |
| weaken the ancestry command | ✔ |
| invert the ancestry verdict | ✔ |
| **sensitivity control** on an already-pinned site | ✔ |

Every mutation's application was proven by printed diff before its result was read, and the file restored green after.

The first row is why the poll assert is anchored to the loop body rather than to any `ls-remote` in step 1: `pre=` uses the same command, so a bare presence match **stayed green under the partial revert**. Measured, then tightened.

The last row is the inertness control — without it, five greens-turned-red prove nothing about a harness that might not be reading this document at all.

## 4. Nine tier rows

PRs #882 #883 #885 #891 #892 #900 #902 #904 #908. `#885`'s was recorded late — its review ruling was written but the row fields never were, found by checking coverage per-PR rather than by a total.

## Conflict note — #881 touches the same file tail

Open PR #881 appends the *previous* run's four rows to `docs/metrics/tier-outcomes.tsv`. Both PRs are pure appends to the file tail, so whichever lands second needs a keep-both resolution in chronological order. Flagging rather than working around it: #881 is held for the maintainer.

## Verification

- Full suite on this branch: **1154 pass / 0 fail** (`origin/main`'s 1152 + the 2 new prose pins).
- Working tree carries exactly three files; the maintainer's uncommitted `settings.json` and `probeA.mjs` are untouched and absent from the diff.


feigi:
## Third rule added — `63b22f8`

merge-bot-10's report on #908 surfaced a defect in this same file that I had walked straight into, so it belongs in this PR.

**Step 4 named the flag and not the reason.** `gh pr merge --merge` always writes a two-parent merge commit, even where the branch is trivially fast-forwardable — and step 5's proof *requires* that. `prove-merge.sh:107` is `[ "$parents" -ge 2 ] || die "... has no second parent — not a merge commit"`, and `die()` exits **2**. Exit 2 is "could not evaluate the claim at all", not "the claim is false", so a genuine fast-forward leaves the member with **no proof and a halt** rather than a failed proof.

I found it by being wrong: my dispatch brief told merge-bot-10 to expect **"a fast-forward"**. That was true of the *content* — merge tree == pin tree == `988f29d1`, `git diff 221f4e9 e9d31d2` empty — and false of the *shape*. Nothing broke, because `gh` wrote the two parents regardless of my prediction. But a member who reads the doc, concludes a fast-forward is expected, and reaches for one gets exit 2 from the step whose entire job is to produce a proof.

*"The merge adds nothing"* is a claim about content. Writing it as *"fast-forward"* makes it a claim about shape, and the two come apart exactly where the proof lives.

Pinned and mutation-tested four ways — dropping the load-bearing claim, dropping the die-string, weakening exit 2 to exit 1, and an inertness control on a site pinned by a *different* test all red; restored green.

**Suite is now 1155/1155** (was 1154 in the PR body, before this commit).

Credit where due: I did not catch this. merge-bot-10 did, in a report it had to re-send because the first delivery was lost in transit.

PR #912, opened 2026-08-25
fix(board): a transcript line that fails to parse away from the tail warns instead of vanishing

### LINKED ISSUE #606: board.mjs:244: the per-line catch is position-blind — a mid-file corrupt transcript line drops a turn's whole cache spend as silently as a torn last line

feigi:
Closes #606

## What was wrong

`readAgent`'s per-line catch was position-blind:

```js
let j; try { j = JSON.parse(line); } catch { continue; }
```

The comment above it justifies the skip with exactly one cause — a transcript
appended to WHILE we read it has a torn last line, every tick. That reason is
correct and reaches the final element of the split. The catch applied it to
every line, so a line damaged anywhere else was dropped just as silently, with
`skipped` still reading 0 and nothing on stderr.

## The fix

Walk the split by index. A parse failure at the final element stays silent — it
is the documented torn-by-append case. Anywhere else it warns once per
transcript path, the hand-rolled idiom `warnedMeta` and `warnedSkips` already
use in this file. No new mechanism, and deliberately not `warnOnce` — that
helper does not exist here and is #603's work.

Known ceiling, carried over from the ticket: a transcript whose writer has
already exited has no legitimate torn last line either, but `readAgent` cannot
tell a live writer from a finished one, so that line keeps passing in silence.
Strictly better than warning on none. The review challenged that sentence as an
overstatement — `gatherSpend` has the file's mtime in hand, so why not use it —
and the challenge was **refuted by building it**: an mtime-staleness
discriminator warns on a live writer whose last write is old and stays silent
on a dead writer whose tear is fresh. It discriminates write recency, never
liveness, and nothing on disk records a writer's pid. `cannot` is the right
word. The residual — a crashed agent's last turn is still lost in silence — is
#917.

## Measured, before → after

Three-turn transcript, `cache_creation_input_tokens` 100/200/300, re-measured
against this PR's base rather than quoted from the ticket:

| shape | row | before | after |
|---|---|---|---|
| one line per turn | CLEAN | 600, stderr empty | 600, stderr empty |
| one line per turn | MID-TORN | 400, `skipped` 0, stderr empty | 400, `skipped` 0, **one stderr line** |
| one line per turn | TAIL-TORN | 300, `skipped` 0, stderr empty | 300, `skipped` 0, stderr empty |

Every spend number is byte-identical across the change; only stderr moves, and
only in the mid-torn row. The `for...of` → indexed walk changes nothing about
which lines are billed or how `turnById` groups them.

**One correction to the ticket's own analysis, found by re-measuring.** The
body says losing the first line of a `message.id` makes "the entire turn's
cache spend vanish". That holds only when the turn is a single jsonl line. With
a turn spanning several lines — how Claude Code actually streams them — a later
line carrying the same `message.id` is then the first one seen and bills the
full usage, so the same mid-file tear costs 0 cache spend (measured 600 → 600).
The tear is still a real fault worth warning about; its cost is just
shape-dependent.

**And a correction to that correction, from the review.** An earlier draft of
this body and a comment in the test file said the multi-line tear "only clips
the `output_tokens` snapshot" / "costs nothing". Both understate it. Measured on
one turn written as three lines, `output_tokens` 1/1/300, blocks
thinking/Bash/Read:

| | cacheWrite | output | tools |
|---|---|---|---|
| intact | 1000 | 300 | Bash, Read |
| tear the Bash line | 1000 | 300 | **Read only** |
| tear the out-300 line | 1000 | **1** | **Bash only** |

`cache_creation`, `cache_read` and `maxCtx` repeat on every line of a turn, so
they survive any single tear. Two things do not: the torn line's own `tool_use`
blocks leave the by-tool panel, and tearing the line holding the largest
`output_tokens` snapshot drops the turn's output to the largest that survived.
`board.html:119` renders `sp.totals.output` in the spend header lede, so that
one is operator-visible. The test-file comment now says this instead.

## Two comment claims the review falsified, both corrected here

- `readAgent`'s comment blamed a specific cause: a non-final malformed line
  "was written whole and damaged afterwards". The transcript writer lives
  outside this repo, and the review produced two filesystem-level routes that
  land in the same branch without it — a crash-then-resume appender, and a
  single healthy writer taking a short write it does not loop on. Replaced with
  the property that actually justifies the warn: a line that is not last can
  never be completed by a later append, so it is still malformed on every tick
  after. Measured — a tail tear heals on the next tick (cacheWrite 1000 →
  3000, stderr stays empty), a mid-file tear never does.
- The same sentence cited "(the rule `warnedMeta` and `warnedSkips` already
  follow)" for its warn-once gate. That is not why the gate is safe here.
  `warnedSkips` suppresses a message whose payload is a **count**, which is
  exactly why its own comment has to point at `skipped` reaching the browser
  every tick to keep that number live. This message carries no count, so there
  is no number that can go stale. The comment now states its own justification
  instead of borrowing a sibling's. (The FULL-PATH keying comment still cites
  `warnedSkips` — that borrow is legitimate, it is the same argument.)

## Coverage

The ticket's coverage claim was grep-level. Confirmed by mutation: with the
per-line catch **deleted outright**, the whole suite was **1155/1155 green** —
this mode had no coverage at all.

**Five** tests added, each mutation-verified against the whole suite:

- unfixed tree (base `board.mjs`) → the two real-fault tests red, tail-silent test green
- discriminator removed (warn on every bad line) → **only** the tail-silent test red
- warn-once gate removed → **only** the warn-once test red
- `warnedLines.add(file)` moved OUT of the position check → **only** "a tail tear
  that later moves mid-file" red (1160 → 1159 pass, 1 fail)
- `warnedLines` Set replaced by a module-level boolean → "two damaged transcripts
  in one dir" red **in isolation**, which is the point of it
- stderr message reworded, behaviour identical → all green (discriminates behaviour, not prose)

The last two exist because the review showed the first three could not see them.
The warn-once test holds ONE transcript path constant across two ticks, which a
module-level boolean satisfies identically — under that mutation it still went
red, but only because an earlier test in the same file had already set the flag,
so it discriminated by execution order rather than by anything it built. Two
damaged transcripts in one directory is the shape that actually pins per-path
keying. Likewise all three original tests keep the file's SHAPE constant across
ticks, so none of them could see where `warnedLines.add` sits: moving it out of
the position check, so a legitimate tail tear consumes the file's one warning,
left the whole 1158-test suite green while permanently silencing the real fault.

`MIDFILE_TEAR` is hoisted to a module const: three tests feed the same input and
two exist only to re-run the first's, but spelled out per site the copies can
drift and the whole suite stays green — measured.

Full suite: **1160 pass, 0 fail**.

## Review deferrals

Filed rather than applied: #916 (a damaged line still has no browser channel —
pre-existing, and the same open-remedy shape as #602), #917, #918 (a real
unpinned mutant: `turnById.clear()` in the per-line catch survives the whole
suite), #919, #920, #921, #922, #923. #603 has a comment recording that the
review re-derived the `warnOnce` collapse independently and that `warnedLines`
is now a third call site.


feigi:
## Review applied — 4 findings applied, 12 deferred and filed

**Applied** (commit `77b179e`):

| finding | dimension | what landed |
|---|---|---|
| `survived[0]` | tests | Two damaged transcripts in one dir, each asserting its own warning — pins that `warnedLines` is keyed on the FULL PATH. The existing dedup test holds one path constant across two ticks, which a module-level boolean satisfies identically; it went red under that mutation only via cross-test state leakage from an earlier test in the file. |
| `unverified[0]` | correctness | Tick 1 = legitimate torn tail (silent), tick 2 = same tear after the transcript grew (warns). Pins where `warnedLines.add` sits — moving it out of the position check left the whole 1158-test suite green while permanently silencing the real fault. |
| `unverified[1]` | correctness | The test comment claimed a tear inside a multi-line turn "costs nothing". Measurably false, and worse than the finding said: the torn line's `tool_use` blocks leave the by-tool panel too. Comment rewritten to what measures true. |
| `unverified[5]` | comments | `readAgent`'s comment blamed one causal mechanism ("written whole and damaged afterwards") for a writer that lives outside this repo. Replaced with the property that actually justifies the warn. The suggested replacement was itself rejected as circular; see the commit body. |

Also corrected, on the controller's ruling on `refuted[1]`: the warn-once gate cited "(the rule `warnedMeta` and `warnedSkips` already follow)", which is not why it is safe here — `warnedSkips` suppresses a message carrying a **count**, this one carries none. The comment now states its own justification.

**Mutation matrix for the two added tests**, whole suite (1160 tests):

| mutant | result |
|---|---|
| `warnedLines.add(file)` moved out of the position check | 1159 pass / **1 fail** — only "a tail tear that later moves mid-file" |
| `warnedLines` Set → module-level boolean | 1157 pass / 3 fail; **isolated**, "two damaged transcripts in one dir" is 0 pass / 1 fail |
| stderr message reworded, behaviour identical (control) | **1160 pass / 0 fail** |

**Deferred and filed** — every one carries its dimension, its refutation where it was refuted, and why it was deferred:

| issue | label | finding |
|---|---|---|
| #916 | `ready-for-agent` | `refuted[0]` + `unverified[2]` (same subject, one home): a damaged mid-file line has no browser channel. Both refuters verified the mechanics; the disagreement was scope. Pre-existing — this PR strictly improves it. Same open-remedy shape as #602. |
| #917 | `needs-triage` | `unverified[3]`: refuted. mtime is not a liveness discriminator — the remedy was **built** and it warns on a live writer with an old last write while staying silent on a dead writer with a fresh tear. |
| #918 | `ready-for-agent` | `unverified[4]`: refuted as filed (the FIRST-line shape kills nothing new), but the refuter found `turnById.clear()` in the per-line catch **survives the whole suite**. Verified replacement test in the body. |
| #919 | `needs-triage` | `unverified[6]`: refuted. The `!line` guard is not what makes the final split element the torn write; and the read-path risk it names is already caught by this PR's own tail-silent test. |
| #920 | `needs-triage` | `unverified[8]`: out of scope (pre-existing helper, 2 of 4 sites predate this PR), so unchecked — no refuter dispatched. |
| #921 | `needs-triage` | `unverified[9]`: refuted. Nets **−1 line** and leaves the "sibling of `fixture()`" comment false. |
| #922 | `needs-triage` | `unverified[11]`: refuted. 850/850 real usage objects carry both fields; the proposed shape occurs in 0 of 850. |
| #923 | `needs-triage` | `unverified[12]`: refuted. Net **zero** lines, the catch is the coldest branch (0 executions on a healthy 21.8 MB transcript), and `lines.entries()` measured slower. |

`unverified[13]` (collapse the three warn-once Sets) was **not** filed — #603 already owns it, and now carries a comment recording that the review re-derived it independently, that `warnedLines` is a third call site, and the shared-key caveat an implementer needs.

`node --test skills/fleet/scripts/*.test.mjs` → **tests 1160, pass 1160, fail 0**.

Issue #286, opened 2026-08-10
no test pins the no-ci finisher gate in any of the three documents that state it — the drift this repo pins everywhere else


feigi:
**Deferred from PR #273 review** (ticket #111). `dimension: tests`, severity `suggestion`.

`skills/fleet/skills/run-team/SKILL.md:330` and `:557` · `skills/fleet/commands/review-and-fix.md` step 6 · `skills/fleet/commands/run-merge-bot.md:91-98`

The no-ci finisher gate is now stated in **three** documents, and no test pins any of them.

Measured: `grep -rl "no-ci\|declare-no-ci" skills/fleet --include="*.test.mjs"` returns only `board.test.mjs` and `ci-state.test.mjs` — both script-level, neither prose-level.

This repo has an established practice of pinning exactly this kind of cross-document prose consistency (e.g. "the fix commit is gated on a test run, in both files", "the in-scope-suggestion refuter carries its anti-rubber-stamp clause, in both files" — both present and passing). Without one, a future edit can drop or contradict the no-ci instructions in one of the three docs — SKILL.md's finisher gate stops requiring the reviewer's suite run while `review-and-fix.md` still does — with nothing catching the drift.

**Suggested fix:** add a source-text pin analogous to the existing cross-file consistency tests, anchored to the actual clause text rather than the bare `no-ci` keyword, which is shared with surrounding prose.

**Two traps, both already paid for in this repo:**
1. A positive regex over a whole markdown *section* is a vacuous pin — the slice size is what does the anchoring. Mutation-test the pin **both ways**: it must go red when the clause is removed AND stay green under an unrelated edit inside the same slice.
2. Anchor the clause the reader acts on, not a restatement of it. The three docs now say different-but-compatible things (`review-and-fix.md` step 6 is the reviewer's own attestation; SKILL.md:557 is the finisher's check; `run-merge-bot.md` reads the label), so a naive "same sentence in three files" pin will not fit — pin the invariant each site must keep.

**Why deferred:** writing a non-vacuous prose pin across three documents whose wording deliberately differs is its own design problem, larger than #111's diff.

PR #913, opened 2026-08-25
ci-state simplify pass: the items that still reproduce, and every guard clause the suite could not see

### LINKED ISSUE #288: ci-state.mjs/.test.mjs + SKILL.md simplify pass: 4 measured reductions from the #273 review (plus 1 refuted, recorded)

feigi:
Closes #288.

Every item was re-measured against `origin/main` before being applied. Three of the four non-refuted items still reproduce; one does not, and one is now contradicted by the tree.

## Applied

**Item 3 — `expected` derived on the arm that reads it.** The ternary and the `if (!noCi)` guard on its `vlog` existed only to give a value to the arm that never reads it. Moved to the non-no-ci arm's first statement, which keeps `expectedJobs()`'s `die()` ahead of the run query. Measured against the previous revision on a green fixture and a no-ci fixture under a stubbed `gh`: stdout, stderr and the gh call log byte-identical on both, at the same exit codes.

That ordering was pinned first, in its own commit, because nothing in the suite reached the derivation's `die()` at all — relocating the derivation below the run query reds the new test, and reds the explicit `--workflow-file` sibling with it.

**Item 1 — the unreadable-workflow cases moved onto `run()`.** Both hand-rolled the tmpdir, gh stub and env `run()` already builds. Kept as two cases: the routes refuse at different depths, and removing discovery's refusal reds the scan case alone while the explicit sibling stays green on its later death.

**Item 6 — every clause of the reply shape guards.** The ticket names two. Mutating each clause of each guard found the class is wider: the type halves of both identifying fields and `isObject`'s `typeof` half survived the file too. Each field now gets an empty value and a non-string one — an absent field, which is all the suite fed before, is refused by both clauses at once and so pins neither. `isObject` gets an array and a scalar at both the row and the job site; every malformed row and job fixture in the file was `null`.

**Item 7 — the reported row and job index.** Both fixtures put the malformed element first, so a position-blind guard was right by accident. Each site now carries a well-formed element ahead of the bad one.

## What the review of this PR sent back

**Item 1's collapse dropped a pin, and this PR restores it.** The two hand-rolled tests used a bare `mkdtempSync`; `run()` `git init`s its fixture. So the deleted test was the only fixture in the tree that ran the script outside a repo, and `workflowsPath()` made eager — hoisted above the `arg("workflow-file") ||` short-circuit, against the sentence in `ci-state.mjs` saying an explicit `--workflow-file` answers without a repo root and must not die for want of one — survived the entire suite here while killing a test before this PR. `run()` now takes `git: false` and the explicit route uses it, so the same two cases carry that discrimination alongside the refusal-depth one.

**That case was also passing for a reason it did not mean.** It asserted `cannot read`, which is equally how a path that never existed is refused, so drift between its `--workflow-file` argument and its fixture would have left it pinning absence — the one shape its own comment says must never be declared away. Asserted down to `EACCES`, bounded to one line of stderr.

**The run-list row guard's non-null clauses had no fixture.** Every malformed row in the file was `null`, so weakening `!isObject(r)` to `r === null` survived the whole suite while the mirror weakening at the job site reds. Both sites now take the same two shapes from one table — the review's own remedy proposed the array alone, which leaves the `typeof` clause still unpinned.

**A comment above the not-green section claimed every fixture above it was green.** Measured: fixtures above it also reach `no-ci` and exit 2. Reworded to the property the section rests on, and without the tally — the `else` arm holds more `reasons.push` branches than the sentence counted, and the extra one is unpinned. That gap is filed as #930 rather than fixed here, to hold this diff to the reviewed items.

## Not applied

**Item 2 — no longer reproduces.** `prView`, `runList`, `runView` and `fixtureFile`'s `content === null` branch all have call sites now. The ticket already flagged `runView` as stale after #273; the other three have since joined it.

**Item 4 — contradicted by the tree.** The item rests on "no prose pin covers it". `skills/fleet/scripts/no-ci-gate-prose.test.mjs` (filed as #286, after this ticket) pins both of the clauses the trim would delete, in exactly that bullet, and its header states the design: the no-ci sites word the gate differently on purpose, and each one that can refuse to label carries the refusal. Applying the trim reds two tests there. Not implemented, and the ticket's premise is stale rather than the code.

Items 5 and 8 stay refuted, per the ticket. The review re-raised item 8's remedy three more times; all three were re-refused with the measurement, recorded as a comment on #288.

## Verification

`node --test skills/fleet/scripts/*.test.mjs` — tests 1167, pass 1167, fail 0.

Every added test was mutation-checked against the whole suite, and each mutant's unified diff was read before its run:

| mutant | reds |
| --- | --- |
| `workflowsPath()` hoisted above the `--workflow-file` short-circuit | the explicit-route case, alone |
| the explicit route's argument pointed at a path that never existed | the explicit-route case, alone |
| discovery's unreadable refusal removed | the discovery-scan case, alone |
| row guard `!isObject(r)` → `r === null` | both run-list row cases |
| row guard weakened to null-or-array only | the scalar row case, alone |
| job guard `!isObject(j)` → `j === null` | both job-entry cases; both row cases stay green |
| a comment inserted beside the workflow-file lookup | nothing |

The accept-side cases — a well-formed run list including a superseded row, and an in-progress job whose `conclusion` is legitimately `null` — stay green throughout, which is the direction these guards are there not to catch.


feigi:
## Review deferrals filed

Applied here (one commit, `50d63f8`): the two `survived` findings — which are one defect found independently by `correctness` and `silent-failure`, applied once — plus three `suggestion`-band findings that survived a refuter each. Details in the PR body.

Deferred, each with its `dimension` and why:

| # | dimension | label | what |
| --- | --- | --- | --- |
| #926 | silent-failure | `ready-for-agent` | the `gh pr view` shape refusal says "missing" for a field that is present but empty or non-string — measured, three distinct faults print one line |
| #927 | silent-failure | `needs-triage` | the no-ci payload emits `"jobs":[]`/`"missing":[]` for a repo whose workflow set was never read — pre-existing, unchanged by this diff |
| #928 | simplify | `needs-triage` | the one-CI-workflow `repoFiles` literal is respelled at every call site — measured at +3 lines, lowest value of its batch |
| #930 | correctness | `ready-for-agent` | found while fixing, not by the review: the `no run whose headSha equals the PR head` branch is unpinned, and deleting it reports `green` at exit 0 |

Three further `simplify` findings re-raised the remedy #288's item 8 already refused. `ledger.mjs check` returned `verdict: tracker-hit` on #288 for two of them, so all three are [recorded there](https://github.com/feigi/claude-config/issues/288#issuecomment-5416734928) with the re-refusal and its measurement, rather than as new tickets — including a live residual from the third that keeps the reply's shape literal.

One finding the review sent as an interaction warning — that the explicit-`--workflow-file` case was passing for the wrong reason — was applied together with the pin it shares a test case with, not separately, and the combined result was mutation-checked in both directions: the regression it exists to catch reds it, and the ENOENT path that used to satisfy it now reds it too rather than passing it.

Issue #622, opened 2026-08-18
reap.sh:147 — a dying `git worktree list` reaps merged [gone] branches blind; making it fail closed is a control-flow ruling #391 reserved


feigi:
**Dimension:** correctness (from PR #620's review, `unverified[0]`)
**Deferred from PR #620 review** (issue #391). `skills/fleet/scripts/reap.sh:147`

## What is true

`wt=$(git worktree list --porcelain | awk …)` takes **awk's** exit status, never git's — the same swallow #264 fixed for `git cherry` one branch above. When that `git worktree list` dies, `$wt` is empty, the whole `[ -n "$wt" ]` block is skipped (the dirty check, the ignored-files check and the removal), and flow falls through to `git branch -D`.

PR #620 originally made this arm fail closed (`keep` + `continue`). The review flagged it as a **behaviour** change rather than a message change, and #391's Agent Brief forecloses it:

> No control-flow change. Still `keep`, still `continue`. Do not abort the sweep, do not escalate

…with #391's triage ruling being *report the state, do not change the behaviour*. So the control-flow half was reverted on that PR and is filed here for the maintainer call the brief reserved. The **message** half stayed: a branch that does have a worktree still reaches `git branch -D`, still refuses, and now carries git's own `error: cannot delete branch 'x' used by worktree at '…'` instead of a bare `branch delete failed`.

## Measured, not reasoned

Fixture: bare `origin`, clone `w`, `feature/merged` merged into `main` and dropped on the remote (so `[gone]`), **no worktree**. PATH shim fails `git worktree list` only (`fatal: worktree list exploded`, rc 128) and `exec`s the real git for everything else. Scripts taken from the object store at each ref.

Pre-#620 `reap.sh` (merge base `fca74fd`), `sh reap.sh --apply`:

```
$ git fetch --prune origin
fatal: worktree list exploded
    REAPED feature/merged
{"applied":true,"reaped":["feature/merged"],"kept":[]}
```

`git branch --list 'feature/*'` afterwards → empty. **The branch was force-deleted while the sweep could not see the registry at all.**

The same fixture under the fail-closed spelling kept it:

```
    KEEP feature/merged — worktree lookup failed — cannot tell whether this branch has a worktree: fatal: worktree list exploded
```

[...]

PR #914, opened 2026-08-25
fix(reap): enumerate by %(refname), and hand the merge probe a qualified rev

### LINKED ISSUE #634: reap.sh:113 — %(refname:short) emits heads/<name> for a branch with a same-named tag, bypassing every worktree guard
`reap.sh:113` enumerates `[gone]` branches with:

```sh
git for-each-ref --format='%(refname:short) %(upstream:track)' refs/heads |
  awk '$2=="[gone]"{print $1}'
```

`%(refname:short)` is **ambiguity-aware**: when a tag shares a branch's name it does not shorten to the bare name, it emits `heads/<name>`. PR #620's body left this as a known remainder ("a refname defect, not a discarded-reason one") without a ticket behind it; this is that ticket.

## Measured

git 2.50.1 (Apple Git-155), throwaway repo, branch and tag both named `feat-x`:

```
$ git for-each-ref --format='%(refname:short) %(upstream:track)' refs/heads
heads/feat-x
main

$ git branch -D heads/feat-x
error: branch 'heads/feat-x' not found        # rc=1
```

## Why it is more than a cosmetic name

`$b` is not only printed — it is the key for the worktree lookup:

```sh
wt=$(git worktree list --porcelain |
     awk -v b="refs/heads/$b" '/^worktree /{w=substr($0,10)} /^branch /&&$2==b{print w}')
```

With `b=heads/feat-x` that searches for `refs/heads/heads/feat-x`, which matches nothing, so `$wt` comes back **empty** and the whole `if [ -n "$wt" ]` block is skipped — the main-checkout guard, the `.git` linkage guard, the dirty check, and the ignored-files check all stand down for this branch.

Control then falls to `git branch -D "$b"`, which fails "branch not found" as measured. So **nothing is destroyed** — the fail-closed outcome is luck, not design: the same ambiguous name that bypasses the guards is also the one `-D` cannot resolve. Observable damage is bounded to a merged `[gone]` branch that is never reaped, reported as a keep whose reason names a branch that does not exist.

`git cherry "$base" "$b"` is unaffected — `heads/feat-x` resolves fine as a *rev* (`refs/<refname>`); it is only `git branch -D` and the literal `refs/heads/$b` string that break.

## Fix

Ask for the unambiguous full refname and strip the prefix, rather than asking git to shorten:

```sh
git for-each-ref --format='%(refname) %(upstream:track)' refs/heads |
  awk '$2=="[gone]"{sub(/^refs\/heads\//,"",$1); print $1}'
```

`%(refname)` is always `refs/heads/<name>`, so the strip yields the bare name for every branch, ambiguous or not. The `$2=="[gone]"` field test is unchanged.

## Test

`reap.test.mjs` has no fixture with a tag, so the current spelling passes the whole suite. A regression test creates a branch plus a same-named tag with a `[gone]` upstream and asserts the branch is reaped (or, at minimum, that `reaped[]`/`kept[]` carries the bare name and never `heads/<name>`).

Found finishing PR #620 (issue #391) — an unbacked deferral in the PR body. Not a duplicate of #596, which is prove-merge.sh's pre-existing ambiguous-refname *warning* on its merge-base calls.

feigi:
Closes #634

## The defect, as filed

`%(refname:short)` is ambiguity-aware: where a tag shares a branch's name it stops shortening and emits `heads/<name>`. That string names no branch, so `git branch -D` refuses it, and it does not build the `refs/heads/$b` key the worktree lookup matches on either — `$wt` comes back empty and the main-checkout, `.git`-linkage, dirty and ignored-files guards all stand down for exactly that branch. Reproduced on the fixture the issue describes (git 2.50.1, Apple Git-155).

## The half the issue did not have

The issue reasons that `git cherry "$base" "$b"` is **unaffected**, because `heads/feat-x` resolves fine as a rev. That is true of the code as it stands — and false of the code once fixed. Restoring the bare name makes an ambiguous branch ambiguous *as a rev* again, and git prefers `refs/tags/` over `refs/heads/` when resolving one, so the merge probe answers about the **tag** while `git branch -D` deletes the **branch**. `-D` is authorized by that probe and by nothing else.

That sentence bounds what **else** can authorize `-D` — `git branch -D` has one call site and every other gate in the loop can veto but never authorize. It is **not** a claim that the probe cannot be wrong, and this PR's review established that it can be: `$base` still reaches the same `git cherry` exactly as `BASE_REF` spells it, so a local tag carrying that spelling outranks the remote-tracking ref and the probe answers about the tag. Measured on a fixture: an unmerged `[gone]` branch `REAPED` at exit 0 with an empty `kept[]`. That hole is **pre-existing, not introduced here, and not fixed here** — `BASE_REF` may legitimately name a tag, a sha or a local branch, so there is no prefix to prepend the way `refs/heads/` was correct on the branch side. It is open as #924, and the cherry-probe comment in this diff now says so rather than promising protection the check does not give.

Measured end-to-end through the real script, on a fixture whose tag sits on a merged commit and whose branch holds a commit that exists nowhere else:

| variant | outcome for the ambiguous unmerged branch |
|---|---|
| `origin/main` today | `KEEP … unmerged commits` — kept |
| enumeration fix **alone** (the issue's prescription) | `REAPED`, exit 0, empty `kept[]` — **branch destroyed** |
| this PR | `KEEP … unmerged commits` — kept, under its bare name |

So the enumeration change ships together with a qualified rev at the merge probe. Qualifying costs nothing for an ordinary branch — both spellings name the same commit — and it is the same key the worktree lookup already builds.

## Selection set is unchanged

The `$2=="[gone]"` field test is untouched and still selects the same branches. A space is not legal in a branch name (measured — git refuses `has space`), so the name is always the whole of `$1`; a branch with no upstream leaves `$2` empty; and the tracking forms that *do* contain a space (`[ahead 1]` and siblings) put `[ahead` in `$2`. None of those match `[gone]` before or after.

## Tests

Two, one per direction — the suite had no fixture with a tag at all, so the old spelling passed it whole:

- an ambiguous **merged** branch is reaped and reported under its bare name
- an ambiguous **unmerged** branch is still kept, and the merge probe reads the branch rather than the same-named tag

Mutation-tested on a scratch copy:

| mutant | merged test | unmerged test | other reds |
|---|---|---|---|
| revert enumeration only | RED | RED | none |
| revert the qualified rev only (*the issue as prescribed*) | green | **RED** | none |
| reword an unrelated KEEP reason (control) | green | green | the two dirty-worktree tests |

Full scripts suite: 1157 tests, 1157 pass, 0 fail. `shellcheck -x` clean.

## Siblings, deliberately left

`inflight.sh` enumerates with `%(refname:short)` too, but its consequence is a different class and it deletes nothing: its filter regex still matches the prefixed form (verified), so the "ticket is taken" hit is still recorded and only the operator-facing `local branches:` line names `heads/<name>`. Filed as #915. `release-ticket.test.mjs` uses the short form in assertions; test-only, and no fixture there makes an ambiguous name.

The line-number citation fixed in `printf-die-sweep.test.mjs` had already drifted off all four sites at `origin/main`; this diff moves them further, so they are named instead (#129). Both counts in that comment re-derived and unchanged.

## Follow-ups filed from this PR's review

Every one is **pre-existing or out of this ticket's scope**; none is a defect this diff introduces, and none is fixed here.

- **#924** — `git cherry` takes `$base` unqualified, so a tag named `origin/main` authorizes deleting an unmerged `[gone]` branch. The sibling half of the argument this PR qualifies on the branch side; both refuters reproduced it end-to-end and both called it pre-existing. The remedy is genuinely open.
- **#925** — `git for-each-ref` drops an unreadable ref at exit 0, so a branch missing from the sweep leaves no trace in `reaped[]` or `kept[]`. This diff rewrites that statement's format string and awk program without touching it; both spellings drop identically.
- **#915** — the `inflight.sh` sibling described above.
- **#622** — the pipeline on the enumeration takes `awk`'s status, never `git`'s. Commented onto the existing issue, which already records that making it fail closed is a control-flow ruling #391 reserved.
- **#929** — the adjudication record for the review's `suggestion` band: two mutually-exclusive enumeration rewrites that survived refutation but were declined on cost, one refuted finding, and three confirmations.


feigi:
Review adjudication — `unverified` band (severity `suggestion`, **0 refuters by policy**, so each in-scope one was checked here).

| finding | scope | outcome |
|---|---|---|
| `reap.sh` cherry-probe comment | in scope | **applied** as `e24cfa9` — but not the remedy as proposed; see below |
| `%(refname:lstrip=2)` rewrite | in scope | refuter returned `refuted=false`; **declined on cost** → #929 |
| full-refname loop rewrite | in scope | refuter returned `refuted=false`; **declined**, mutually exclusive with the above → #929 |
| the two #634 test-discrimination checks | in scope | confirmations, no defect → #929 |
| tests survive an equivalent-mechanism swap | in scope | confirmation — behaviour-pinning, working as intended. **Not a coverage gap; do not "fix" it** → #929 |
| `for-each-ref` drops an unreadable ref at exit 0 | out of scope | deferred → **#925** |
| `inflight.sh` `%(refname:short)` sibling (×2) | out of scope | already **#915**; the one thing it lacked is commented there |

**Applied: 1. Deferred and filed: 1 (#925). Recorded: 6 (#929). Commented onto an existing issue: 2 (#915, and #622 by the controller).**

### What `e24cfa9` changes, and what it does not

The duplication finding was **refuted**: the sentence *"-D is authorized by this check and by nothing else"* does appear twice in that comment run, and this PR added the second — but the two carry different conclusions (probe gives *no* answer → KEEP; probe gives a *confidently wrong* answer → nothing downstream catches it), and the repo has a separate test for each (#264 and #634). The remedy's other half was wrong on the facts: the parenthetical it proposed dropping measures `refs/tags/` precedence, not the `%(refname:short)` fact the enumeration block measures. **Neither occurrence was deleted.**

Checking it surfaced a real defect in the same sentence. It is accurate about control flow, but read as reassurance it overclaims — #924 establishes the check is defeatable via the unqualified `$base`. Re-measured independently before writing the comment:

```
git rev-parse origin/main       -> 555889b   (the TAG, not the remote-tracking ref)
refs/remotes/origin/main        -> 6d39d11
git cherry origin/main              refs/heads/feature/solo -> no + line   (reads CLEAN)
git cherry refs/remotes/origin/main refs/heads/feature/solo -> + 555889b   (keeps)

$ sh reap.sh --apply
    REAPED feature/solo
{"applied":true,"reaped":["feature/solo"],"kept":[]}   exit=0
$ git rev-parse --verify refs/heads/feature/solo
fatal: Needed a single revision
```

The comment now names that residual and points at #924. **The `$base` hole is not fixed here** — `BASE_REF` may legitimately name a tag, a sha or a local branch, so there is no prefix to prepend the way `refs/heads/` was correct on the branch side. The PR body carried the same overclaim and has been reconciled.

### One correction to the record

The full-refname rewrite was expected to red `reap.test.mjs`'s *"an ordinary run's payload is byte-for-byte what it has always been (#391)"* by leaking qualified refnames into the payload. **It does not** — `b=${ref#refs/heads/}` keeps `$b` bare, and the pin passes. That pin is not vacuous (deleting the strip outright reds it loudly). What declined the rewrite was different: it adds a line rather than removing one, and it splits a seven-site convention `release-ticket.sh` and `claim-ticket.sh` share.

`node --test skills/fleet/scripts/*.test.mjs` from the worktree: **`tests 1157, pass 1157, fail 0`**, before and after.

PR #931, opened 2026-08-25
fix(ledger): carry the cause when check's gh probe fails — keep the end of an overrunning stderr, fall through a blank one

### LINKED ISSUE #638: ledger.mjs check: the gh catch loses the cause — clipped to 500 chars with no marker, and emptied entirely by whitespace-only stderr

feigi:
Closes #638

Both findings land on one expression in `skills/fleet/scripts/ledger.mjs`. The ticket cites it at `:489`; at `origin/main` `919269b` it is `:584`, and the `why` clip the ticket names in passing as "line 433" is `:510`. Re-derived, not trusted.

## Reproduced first, against this tree

Both defects were driven through the real script before anything was edited, with the suite at `70/72`:

- **overrunning stderr** — a gh stub emitting warning lines then `gh: FATAL — HTTP 403 rate limit exceeded`: `tracker.error` ended `"gh: warning line 7 -"`, exactly 500 chars, no FATAL line, no marker.
- **blank stderr** — a gh stub doing `printf '\n \n' >&2; exit 1`: `tracker.error` came back `""`, and the warning read `TRACKER NOT CHECKED ()` at exit 0.

A third defect of the same family surfaced during review and is fixed here too:

- **a gh Node killed** — a gh stub printing one cached-credentials notice and then hanging: after 20.0 s, `tracker.error = "gh: warning: using cached credentials for github.com"`, with the timeout named nowhere in the payload or the operator warning. The same stub printing nothing came back `"spawnSync gh ETIMEDOUT"` — the abort was named only where the child happened to be silent.

## The fix

One helper, `cause()`, because the first two defects are one mechanism and the repository probe in the same arm had the same shape:

- **choose on the trimmed text.** A whitespace-only stderr is truthy, so it won a choice made on the raw values and then trimmed away to nothing. Now it falls through to the thrown message, which names the command that failed.
- **keep the end, and mark it.** gh prints its warnings ahead of the error that killed it, so the leading bytes are the ones worth dropping. Still capped, marker included — the field ships on stdout inside a machine-parsed contract.

And one ordering rule at the gh catch, for the third:

- **when Node aborted the child, ask the message first.** A timeout is named in `e.code` and `e.message` and NOWHERE in `e.stderr`, which on a timeout holds whatever gh printed before the kill — so stderr-first reported an unrelated notice as the cause of a 20-second stall. `e.code` is the discriminator, measured rather than assumed on Node v26.7.0: Node sets it on the aborts it performs (ETIMEDOUT, ENOENT, EACCES) and **not** on an ordinary non-zero exit, where the child explained itself on stderr. So the flip reaches only failures whose cause the child could not have printed, and it moves exactly one of them — the other aborts leave `e.stderr` `undefined`, so the message already won there.

The comment above that line named the wrong class, and the correction is in this diff. It grouped a timeout with a spawn Node aborted and a refused parse as "failures that carry no stderr field at all". Measured, stderr holds no cause in **three** shapes and only one of them is absence: no `stderr` property at all on the refused parse, the property present but `undefined` on a spawn never started, and present but **empty** on a child killed before it printed. A timeout is the third, never the first.

## Both prescribed remedies were run, and neither survived as written

- Finding 1's `` `…${raw.slice(-500)}` `` takes the payload to 501 characters. The comment above the line argues the cap as a payload-size constraint, so the marker goes *inside* it.
- Finding 2's `String(e.stderr ?? "").trim() || String(e.message ?? "").trim() || "gh failed with no diagnostic"` is right about the trim order and wrong about the shape: written as three independent expressions it duplicates the cap logic per call site, and the repository probe — which has the same two defects — is left out entirely.
- The review's remedy for the timeout composed a string off `e.signal`. That field is not read here at all: a `spawnSync` timeout does not reliably imply `SIGTERM`, and `e.code` alone settles it.

## What the change is not allowed to move

`check`'s payload is what the fleet's dedupe guard reads. The differential ran twice.

Originally, over 33 `check` input shapes — every tracker arm, each against an unfiled, an already-filed and a near-miss ledger — comparing the whole payload with `tracker.error` removed: **0 of 33 moved anything outside `tracker.error`.**

Re-run after the ordering change, over **48** shapes across three subjects — every tracker arm plus both hang shapes, the at-cap and astral stderr fixtures, a gh that crashes silently, gh absent, and a ledger outside any repository: **identical again**, `verdict` / `tracker.ok` / `tracker.hits` / `truncated` / exit code unmoved on every shape. Backed by a positive control: the same harness against a tree with a mutated `verdict` string reports DIVERGED on every row, so the identical result is a measurement and not a probe that could not look.

## Tests

`node --test skills/fleet/scripts/*.test.mjs` → **1180 pass, 0 fail.**

Six pins on what `tracker.error` CARRIES, plus a generic `GH_STDERR` escape on the gh stub so a test can drive the stderr *shapes* that matter — the existing `GH_FAIL` emits one tidy line, which is the shape that hides the first two defects — and a `GH_HANG` escape for the third.

Each mutant ran against the whole suite, since sibling test files spawn these scripts:

| mutant | expected | result |
|---|---|---|
| head clip, no marker | red | **red**, only the overrun pin |
| choose on raw values, trim after | red | **red**, only the blank-stderr pin |
| mark unconditionally | red | **red**, only the pass-through pin |
| revert the abort flip | red | **red**, only the hang pin |
| cap `>` to `>=` | red | **red**, only the at-cap pin |
| revert the repository probe to its pre-PR expression | red | **red**, only the probe pin |
| slice the tail by code POINTS | red | **red**, only the overrun pin |
| drop the last-resort literal | green | **green** — confirms the comment's claim |
| `!raw` → `raw === undefined` (equivalent) | green | **green** |
| `find(Boolean)` → `find((s) => s.length > 0)` (equivalent) | green | **green** |

The blank-stderr pin **failed its first mutant**: asserting `tracker.error` is non-empty read green off the last-resort literal while the reader had lost the command. It now asserts the argv gh was given, which only the thrown message supplies, and refuses the literal by name.

The pass-through pin is the must-ACCEPT half: a stderr short enough to fit must arrive trimmed and otherwise byte-identical, with no marker on a string that was never cut.

The at-cap pin closes a boundary every other fixture here misses by hundreds of characters — the overrunning one is ~3400, the accept-unchanged one ~54 — so `>` to `>=` marked an uncut cause as cut and passed all 1177 tests.

The hang pin costs the full 20 s the script waits and cannot cost less: that timeout is `execFileSync`'s own, on the parent side, so no stub shortens it. It buys the only failure class where this catch's ordering is observable at all.

The repository probe had no pin: reverting only that line to its pre-PR expression left the whole suite green. It is reachable in production and not merely constructible — `ledger.mjs` scrubs `GIT_DIR` and `GIT_WORK_TREE` off the probe and nothing else, so a malformed ambient `GIT_CONFIG_*` reaches it, and real git then puts ~640 bytes of echoed input ahead of the line saying what actually killed it.

The overrun pin gained a code-unit guard. The cap is on `.length`, which counts UTF-16 code units; a tail sliced by code *points* keeps up to twice as many, measured 999 under a remedy proposed during review, and the ASCII fixture cannot see it because there one code point is one code unit.

Two assertions added earlier in this PR were dropped again, both measured unable to fail independently: the `doesNotMatch(/…/)` after the accept-unchanged strict equal can never execute (`node:assert/strict` throws on the equal first, and no mutation reds one without the other), and the `/\S/` in the blank-stderr test is the residue of the superseded pin `a8c112a` replaced.

## Enumerate and declare

**Covered**

- `check`'s gh catch — all three defects.
- `check`'s repository probe — the `why` clip the ticket names in passing. Same clip, same raw choice; routed through the same helper, and now pinned at that site. An empty result stays tolerable there and the comment says why: the message it interpolates into still names the probe and the path.

**Deliberately left, to a ticket of its own**

- `ci-state.mjs`'s and `diff-stats.mjs`'s `raw.trim().slice(0, 120)` — the unmarked-clip half only, not the blank-cause half. Different scripts, and the string clipped is gh's *stdout* on a not-JSON path, where the noise a `die()` is explaining (a login banner, a deprecation notice) leads and the head is the informative end — the opposite of stderr. Deciding it needs its own measurement. **Not already filed**: the dedupe guard's tracker hit scores 0.00 and is a review record, and broader single-term searches surface nothing on it.

**Filed from this PR's review**

- #935 — four `cause()` micro-refactors, each measured behaviour-preserving and each refused on worth (`wontfix`).
- #936 — the lone surrogate the tail-keep can leave, refuted as cosmetic; its prescribed code-point remedy ships a 999-character cause and is now pinned against in-tree (`wontfix`).
- #937 — `GH_FAIL` and `GH_STDERR` are two stub hatches doing one job, but `GH_FAIL`'s shape is what documents #638 (`needs-triage`, unchecked).
- #938 — the tracker payload is a hand-rolled discriminated union with no static enforcement (`needs-triage`, unchecked, and out of #638's scope by the finding's own conclusion).

**Scanned and ruled out**

- The blank-cause defect has no other site in the repo. Every other consumer of a child's error uses `e.message` alone or `??`.
- `ledger.mjs`'s `.slice(0, 3)` on the query terms is term selection, not a diagnostic clip; #153 owns its ceiling.
- `fleet-tick.mjs` already keeps the tail of a failed child's stderr rather than its head — prior art in this repo for the choice made here.
- The review's `String(c ?? "")`-as-erasing-cast finding was refuted 2-0 and is not filed: `cause()` is module-private, the script has no exports, and every value the two call sites can pass is `string | undefined` by construction.

## Note for review

The helper's first placement put `const CAUSE_MAX` after the module-level dispatch chain that calls `runCheck()`, so every `check` invocation died in the temporal dead zone. The suite caught it: 12 red. Moved up beside `die()`.


feigi:
**Review findings applied — deferrals filed.**

All 17 findings from the review were adjudicated. Every `unverified` finding in #638's scope got one refuter, dispatched biased to refuse and told to verify by running.

**Applied** (commit `1c816b2`):

- the gh catch reported a 20-second hang as whatever gh last warned about — `e.code` now decides which field is asked first, and the timeout is named. Pinned end to end.
- the comment above it named the wrong class: measured, a timeout leaves `e.stderr` present and *empty*, never absent.
- the cap's own boundary (`>` vs `>=`), previously unpinned — the mutation passed all 1177 tests.
- a code-unit guard on the cap, because a proposed remedy ships a 999-character cause under a green suite.
- the repository probe at the second call site #638 names — reverting only that line left the whole suite green.
- two assertions dropped, both measured unable to fail independently.

`node --test skills/fleet/scripts/*.test.mjs` → **1180 pass, 0 fail.** Every added pin mutation-verified: its own mutant reds it and only it, and an equivalent refactor leaves the suite green. The payload differential re-ran over 48 shapes with a positive control — nothing outside `tracker.error` moved.

**Deferred and filed:**

| issue | label | finding |
|---|---|---|
| #935 | `wontfix` | four `cause()` micro-refactors — each measured behaviour-preserving, each refused on worth |
| #936 | `wontfix` | the lone surrogate after the `…` marker — cosmetic, and its code-point remedy ships a 999-char cause |
| #937 | `needs-triage` | `GH_FAIL`/`GH_STDERR` are two stub hatches doing one job — but `GH_FAIL`'s shape is what documents #638 |
| #938 | `needs-triage` | the tracker payload is a hand-rolled discriminated union with no static enforcement |

**Ruled, nothing to file:** the `|| "gh failed without saying why"` last resort (three findings, one subject) — a refuter measured it reachable in **0 of 13** public failure modes because every `execFileSync` failure class carries a non-empty `.message`, which is exactly what the comment above it already claims; no test can reach it and deleting it would remove the guard against the defect #638 was filed for. The `String(c ?? "")`-as-erasing-cast finding was refuted 2-0 by the review itself.

Issue #739, opened 2026-08-21
review-pr.js:463: the tests-only profile shadows the #236 silent-failure floor — adding a test file to a CI diff SUBTRACTS the hunter


feigi:
Deferred from PR #736 review (closes #236). Five findings converge on the same span; the confirmed defect is section 1, the rest are the same lines' neighbours and are recorded here so they are settled together rather than five times.

`dimension`: `silent-failure` (§1), `correctness` (§1 twin), `simplify` (§2, §3), `types` (§4).

---

## 1. CONFIRMED DEFECT — `tests-only` outranks the floor, so adding a test file SUBTRACTS a reviewer

**`workflows/review-pr.js:463`** — `dimension: silent-failure`, severity `important`, band `survived` (2 refuters, both `refuted=false`).

`computeStats`' else-if chain (`skills/fleet/scripts/diff-stats.mjs:88-93`) tests `!hasSrc && hasTests` → `tests-only` **before** `files.length === 1` → `single-file` and `loc < 30` → `small`. So a config+test diff never gets a `SIZE_TIER_PROFILES` profile, the new carve-out at `review-pr.js:463` (`d.key !== "silent-failure" || SIZE_TIER_PROFILES.has(stats.profile)`) evaluates false, and the `hasSrc === false` guard still strips `silent-failure`.

Reproduced independently by both refuters, lifting `DEFAULT_DIMENSIONS`/`SIZE_TIER_PROFILES`/`selectDimensions` from the snapshot's own `review-pr.js` (the technique `select-dimensions.test.mjs` already uses) and driving real paths through the real `computeStats`:

```
.github/workflows/ci.yml alone                  -> single-file -> correctness, silent-failure
.github/workflows/ci.yml + one .test.mjs        -> tests-only  -> correctness, tests, comments
2x .yml + one .test.mjs                         -> tests-only  -> correctness, tests, comments
.github/scripts/release.sh + its test           -> tests-only  -> correctness, tests, comments
```

**The inversion is NEW with PR #736.** Pre-PR, `ci.yml` alone was `[correctness]` and `ci.yml`+test was `[correctness, tests, comments]` — a superset. Post-PR it is not: adding one file to the diff **subtracts** the silent-failure hunter. Coverage is non-monotonic, and #236's own motivating class (CI gating logic, no src) is the one that loses it.

[...]

Issue #571, opened 2026-08-17
PR #564 review: the suggestion band, checked — 5 refuted with evidence, 1 survived but declined, 2 applied


feigi:
**Deferred from PR #564 review** (implements #376). Record ticket, same shape as #483 and #542.

The `suggestion` band gets 0 refuters from `review-pr` by policy, so the fix-applier dispatched **one refuter per finding, biased to refuse, each required to verify by running something**. Outcome across the band: **2 applied, 5 refuted with evidence, 1 survived but declined, 1 no-action, 2 split off to #570.**

Nothing here needs doing. It exists so the refutations are not re-derived and the refuted remedies are not re-applied.

---

## Applied in 7ebb721 — do not re-file

**`unverified2` · `silent-failure` · `no-undo-audit.sh:344`** — the PR's own new prose stated a ceiling that does not measure: `core.logAllRefUpdates=false` does **not** leave the probe with nothing to look at. Measured (git 2.50.1): the setting suppresses the HEAD reflog (`.git/logs` absent after the first commit) but `git stash push` force-creates `logs/refs/stash` regardless, so such a repo probes normally and reports `unknown`. The clause is gone; the unreadable-`logs/refs`-DIRECTORY half is real, was re-measured, and stays.

**`unverified8` · `simplify` · `no-undo-audit.sh:372`** — the 4-line relative-to-absolute `case` fix-up is what `rev-parse --path-format=absolute --git-path` already returns, the spelling this same script uses unconditionally at :211 and :213. Verdict-identical across 16 measured layouts (main/linked worktree, cwd in and out of `$wt`, relative `$wt`, symlinked `$wt`, `--separate-git-dir`, `GIT_DIR`/`GIT_COMMON_DIR` set); five rows differ in string only, and `$stash_reflog` is consumed by exactly one expression, `[ -s ]`.

---

## Refuted by measurement — do NOT re-apply

[...]

Issue #542, opened 2026-08-17
PR #537 review: the suggestion band, checked


feigi:
Record of the `suggestion` band from the PR #537 review, all of it checked. The workflow budgets that band 0 refuters, so every finding arrived `unverified`; the review returned `survived: 0, refuted: 0, unverified: 8`. Five refuters ran against the in-scope seven; the eighth went to #500. Nothing here is queued work — it is the evidence, so the next reviewer meets it instead of re-raising.

Applied in `c9f19a2` (PR #537): `correctness:50`, `simplify:46`, and the regex widening below. Deferred to #500: `types:36`. Everything else is refuted, below.

## Refuted — the three-way regex conflict, `snapshot-runner-audience-prose.test.mjs:76/:80`

Three findings targeted the one regex in the closing guard and were mutually exclusive. One refuter took all three; **all three remedies were refuted** and a synthesis was applied instead.

- **`correctness:80`** — proposed `/\.\/agent-test\s+\S*\//`. Its claim (the old `/\.\/agent-test\s+skills\//` misses the `./`-prefixed, bare-relative, quoted-glob and no-`./` spellings) is **true and reproduced**. Its remedy is not safe as written: `\s+` crosses newlines unguarded, so correct prose that merely ends a line on the runner's name reds. Measured on `agent-test-dir-prose.test.mjs`'s untouched header, reflowed at widths 30–100: **red at 7 of 71 widths**, against 0 for both the old regex and the one applied. That contradicts the subject file's own stated invariant, "Reflow stays green by design".
- **`silent-failure:80`** — proposed `/agent-test\s+\S*\.test\.mjs/`. Claim true, remedy **drops a catch the old code already had**: the pattern requires the argument to end in `.test.mjs`, and the directory form `./agent-test skills/fleet/scripts` never does. Confirmed by running both: directory form is **red** under the old regex and **green** under this one. `./agent-test <file-or-dir>` is a supported form (`agent-test-dir-prose.test.mjs`, PR #75). Neither reviewer could see this — `correctness` measured the directory form as caught, `silent-failure` never tested it.
- **`types:76`** — "no action needed, the check is self-verifying and will fail loudly the moment a comment collides". Refuted on its reasoning: that describes the false-**positive** direction. This guard's failure mode is a false **negative** — the offending comment ships green and a snapshot-bound specialist gets a command that is not there. Nothing fails loudly. Its evidence (0 matches today) proves non-vacuity, not coverage.

[...]

PR #933, opened 2026-08-25
test(review-pr): derive the declared-exactly-once names instead of listing three

### LINKED ISSUE #653: review-pr-reads.test.mjs: the declared-exactly-once guard does not cover unrunReason/unrunEntries/unrunCrashed, so a duplicate-after-real declaration stays invisible
**Deferred from PR #647 review** (issue #143, head `b393dff`). Dimension: **tests**. Location: `skills/fleet/scripts/review-pr-reads.test.mjs:397`.

**Why deferred:** out of #143's ticket scope — the guard is pre-existing and its file is not in PR #647's diff. Not refuted; the review's own measurement stands unchallenged (a `suggestion`, so the workflow budgeted it 0 refuters and the fix-applier's scope split, not a refutation, sent it here).

## Claim

The structural guard `test("each function is declared exactly once at top level")` hardcodes the name list `["usableDiff", "readRules", "snapshotMissing"]`. It does not cover `unrunReason`, `unrunEntries` or `unrunCrashed`, so a **duplicate declaration placed after the real one is invisible to every test that lifts them**: the source-lift's non-global `.match` takes the *first* declaration, while JS hoisting makes the *last* one win at runtime.

## Evidence (from the review, run against the snapshot)

Inserting a second `function unrunReason(review) { return null; }` immediately after the real declaration in a scratch copy of `workflows/review-pr.js`, then running

```
node --test skills/fleet/scripts/review-pr-unrun.test.mjs skills/fleet/scripts/review-pr-reads.test.mjs skills/fleet/scripts/review-pr-testcmd.test.mjs
-> tests 48 / pass 48 / fail 0
```

all green — the lift exercised the real declaration. Evaluating the same source the way the engine loads it (hoisting, last declaration wins) showed the **stub** executing: `unrunReason({...tests:0...})` returned `null` instead of a reason string.

This is the bug class this repo already documented from PR #536, where a duplicate-after-real declaration for a different function passed 22/22 while being dead code at runtime — and PR #647 substantially rewrites `unrunReason`, the exact function the blind spot endangers.

## Remedy (settled)

Add `"unrunReason"`, `"unrunEntries"`, `"unrunCrashed"` to the name list in `review-pr-reads.test.mjs`'s `test("each function is declared exactly once at top level")`.

Mutation-verify it: insert a duplicate declaration of each, confirm the guard reds, restore.

feigi:
Closes #653

## What was wrong

`review-pr-reads.test.mjs`'s `test("each function is declared exactly once at top level")` looped a hardcoded list of three names. That list had frozen at exactly the names lifted by the two files that wrote it — this one and `review-pr-snapshot-path.test.mjs`. Every function a LATER lift file came to depend on was outside it, so a duplicate declaration placed AFTER the real one was invisible to the very file lifting it: the lift's non-global `.match` takes the first declaration while JS hoisting makes the last one win at runtime.

Re-derived against `origin/main` at `919269b` before implementing; the ticket's line citation for the guard was stale.

## What changed

The names are derived from the source instead of listed. Listing is the failure mode itself — a function is protected only if someone remembered to add its name, and twice nobody did (#140 shipped without `snapshotMissing`; this ticket found the unrun, testcmd and select-dimensions seams still absent).

The derivation matches every declaration form `review-pr.js` can legally contain — plain, `async`, a generator star, and any spacing around the name:

```js
[...code.matchAll(/^(?:async[ \t]+)?function[ \t*]+(\w+)[ \t]*\(/gm)].map((m) => m[1])
```

`export function` is left out on purpose rather than by oversight: `review-pr.js` compiles as a function body inside the harness VM, so a second `export` is a syntax error, not a silent rebind — measured, `export function ghostExport(x) { return x; }` reds the file's own parse test alone. Putting it in the regex would be dead code and claiming it in prose would be a false claim.

`const`-form values other files lift (`DEFAULT_DIMENSIONS`, `verifiersFor`) stay out for the same reason: a duplicate `const` in one scope is a SyntaxError, so it cannot hoist past the real declaration, and that same parse test catches it.

A second test pins the inverse — the shapes the guard must ACCEPT: a method on an object literal, a nested indented declaration, a shadowing local, and the name inside a template literal.

## Coverage, before and after

| Function | Lifted by | Before | After |
|---|---|---|---|
| `usableDiff` | reads | covered | covered |
| `readRules` | reads | covered | covered |
| `snapshotMissing` | snapshot-path | covered | covered |
| `resolveTestCmd` | testcmd | **not covered** | covered |
| `selectDimensions` | select-dimensions | **not covered** | covered |
| `resolveDimensions` | select-dimensions | **not covered** | covered |
| `unrunReason` | unrun | **not covered** | covered |
| `unrunEntries` | unrun | **not covered** | covered |
| `unrunCrashed` | unrun | **not covered** | covered |
| `decodeArgs` | not lifted | **not covered** | covered |

The ticket named three of the gaps. `resolveTestCmd`, `selectDimensions` and `resolveDimensions` were equally blind, and `decodeArgs` had no lift backstopping it at all; the same change fixes all of them.

## The non-emptiness floor

`assert.ok(names.length > 0)` is a backstop, not this file's first line of defence, and the comment now says so. This file lifts `usableDiff` and `readRules` at **module scope**, so a derivation gutted today reds there first and louder — measured, indenting every declaration but `usableDiff` gives `AssertionError: review-pr.js no longer declares readRules(...)` at `tests 1 / pass 0 / fail 1`, and the floor's own line never executes.

A pinned total was rejected deliberately. The count has only ever grown — 2 → 10 across `workflows/review-pr.js`'s whole history (2026-08-05 to 2026-08-23), non-decreasing, never once falling — so a `>= N` never reds on an addition; it quietly stops covering the new half. That is the #140 hardcoded-list failure wearing a different operator, and it would have re-introduced what this PR removes.

## Verification

Mutants were evaluated the way the ENGINE loads the file, not the way a lift reads it: the whole source compiled as an `AsyncFunction` body with `return <name>;` as the first statement, so hoisting yields the binding `review-pr.js` actually executes while none of the file runs. A source-lift read would have taken the FIRST declaration and missed the rebind entirely.

| mutant | engine binds | whole suite |
|---|---|---|
| `async function unrunReason` duplicate | `AsyncFunction`, returns `null` | `pass 1174 / fail 1` — guard alone |
| `function unrunReason (review)` duplicate | `Function`, returns `null` | `pass 1174 / fail 1` — guard alone |
| `function* unrunReason` duplicate | `GeneratorFunction` | `pass 1174 / fail 1` — guard alone |
| new `async function brandNewHelper`, declared once | — | `pass 1175 / fail 0` (must not fire) |
| nested indented `function unrunReason` inside `unrunCrashed` | real top-level `unrunReason` | `pass 1175 / fail 0` (must not fire) |
| **control:** `async` duplicate under the **pre-fix** regex | `AsyncFunction`, returns `null` | `pass 1175 / fail 0` — **green** |

The last row is the discrimination proof: that mutant was invisible to the first commit's `/^function (\w+)\(/gm` and reds under the broadened one, with nothing else in the diff responsible.

Also measured: a duplicate-after for each newly covered name reds the guard, every lift-based test staying green alongside it — which is the blind spot; and the broadened regex re-verified to yield the identical ten names in the identical order as the plain-form one, so widening the accepted forms did not start collecting anything else.

Whole suite on the final tree: **1175 pass / 0 fail**.

## Review

Six findings, all on this diff's own new prose. One `survived`-band finding applied (the regex breadth above), one `suggestion` survived its refuter and was applied (the two-file undercount, now stated as a property), three refuted with evidence, one refuted but rewritten anyway because broadening the regex invalidated its clause independently. Record ticket with the measurements: #934.

The review ran `correctness`, `tests` and `comments` only — the `tests-only` profile drops `silent-failure-hunter` and `dimensionsUnrun` was empty, so nothing announced the trim (#739).


feigi:
Review findings applied in `d2191818515ca6579735f0618632b1608b11a5b4`. Deferred/refuted findings are recorded in #934 (`wontfix`, closed — a record ticket, same shape as #483, #542 and #571), so the refutations are not re-derived and the refuted remedies are not re-applied. PR body reconciled against the final diff.

Issue #938, opened 2026-08-25
ledger.mjs: the tracker payload is a hand-rolled discriminated union with no static enforcement — every read site orders the .ok check by hand


feigi:
Deferred from PR #931 review (ticket #638). **dimension: types.** `skills/fleet/scripts/ledger.mjs:468` and its construction sites.

**Unchecked — no refuter ran on this one.** Deferred as out of #638's scope, which is what the finding itself concludes: *"Out of scope for a #638-scoped fix — not requesting a change here."* Filed so it is not lost, not because a change is being asked for.

## The finding

`tracker` is a hand-rolled discriminated union — `{ok: true, query, hits, truncated}` versus `{ok: false, query, error}` — with nothing enforcing that the field sets are mutually exclusive. Every read site depends on checking `tracker.ok` first, by convention and by comment, not by any type.

Construction sites (current head): `:475`, `:541`, `:596`, `:650`. Read sites: the `!tracker.ok` warning branch, the `verdict` ternary, and the `JSON.stringify` payload. All three order the `!tracker.ok` check before touching `.hits`/`.truncated` by hand, with no destructuring or type guard doing it for them.

The invariant is load-bearing and is currently carried by a comment:

    // `!tracker.ok` short-circuits before `tracker.hits` is read, which is
    // required: hits is absent, not `[]`, on that branch.

`hits` being **absent rather than `[]`** on the failure arm is deliberate and test-pinned (a consumer testing `.length` must fail loudly rather than read an unread tracker as clean) — which is exactly what makes an out-of-order read a silent wrong answer rather than a crash.

## Why nothing was applied

PR #931 neither touches nor worsens this shape. And the file has no type layer to hang a fix on: no `tsconfig.json` or `jsconfig.json` anywhere in the repo, and `ledger.mjs` carries zero `@type` annotations in 700 lines, so a JSDoc union here would be a stylistic departure rather than a local fix.

## What would make it actionable

If this file (or `skills/fleet/scripts/` generally) later adopts JSDoc typing with `checkJs`, a `@typedef` union on the tracker shape would let an editor flag a future `tracker.hits` read added before the `!tracker.ok` guard. Until then there is nothing to enforce it with, and the comment plus the absent-not-empty pin are what the invariant rests on.

Same weakness class as the `cause()` candidate-shape finding, which was refuted 2-0 during the same review and is not filed: an invariant living in comments and discipline rather than in something the language enforces.

feigi:
[...]

Issue #929, opened 2026-08-25
PR #914 review: the suggestion band, checked — 2 survived but declined (the mutually-exclusive enumeration rewrites), 1 refuted, 3 confirmations


feigi:
`Deferred from PR #914 review` (ticket #634). The `unverified` band arrives with **0 refuters by policy**, so the fix-applier checked each in-scope one itself. This is the record for the ones that produced no diff; the two out-of-scope ones went to #915 (already filed) and #925 (filed from this review).

Every measurement below was run against `d04c742` (the review snapshot) or `e24cfa9` (post-fix), in scratch, never in the checkout. Baseline suite at both: **`tests 1157, pass 1157, fail 0`**.

---

## Survived refutation, declined on cost — `dimension: simplify`

These two are **mutually exclusive**; the finding text says so itself. Both were dispatched to a refuter biased to refuse, and **both came back `refuted=false`**. Neither was applied. This is the part that wants a human decision.

### A. `%(refname:lstrip=2)` instead of the awk strip — `reap.sh:141-142`

```sh
# current
for b in $(git for-each-ref --format='%(refname) %(upstream:track)' refs/heads |
           awk '$2=="[gone]"{sub(/^refs\/heads\//,"",$1); print $1}'); do
# proposed
for b in $(git for-each-ref --format='%(refname:lstrip=2) %(upstream:track)' refs/heads |
           awk '$2=="[gone]"{print $1}'); do
```

**Measured behaviour-identical.** `:lstrip=2` is ambiguity-unaware exactly like `%(refname)` — tags shadowing both a flat and a nested branch, `:short` decorates to `heads/<name>` in every case, `:lstrip=2` in none:

```
short=[heads/a/b/c/deep]   full=[refs/heads/a/b/c/deep]   lstrip2=[a/b/c/deep]
short=[heads/fix/634-slug] full=[refs/heads/fix/634-slug] lstrip2=[fix/634-slug]
short=[heads/merged]       full=[refs/heads/merged]       lstrip2=[merged]
short=[has"quote]          full=[refs/heads/has"quote]    lstrip2=[has"quote]
short=[ümlaut]             full=[refs/heads/ümlaut]       lstrip2=[ümlaut]
```

[...]

Issue #925, opened 2026-08-25
reap.sh:141 — `git for-each-ref` drops an unreadable ref at exit 0, so a branch missing from the sweep leaves no trace in reaped[] or kept[]


feigi:
`Deferred from PR #914 review` (ticket #634). **dimension: silent-failure** · severity: suggestion · 0 refuters in the review's own budget; re-measured by the fix-applier before filing.

## The defect

`reap.sh` enumerates branches with

```sh
for b in $(git for-each-ref --format='%(refname) %(upstream:track)' refs/heads |
           awk '$2=="[gone]"{sub(/^refs\/heads\//,"",$1); print $1}'); do
```

`git for-each-ref` **omits a ref it cannot parse and still exits 0**, announcing it only on stderr. So the branch list can be short, and neither the exit status nor the JSON payload says so.

## Measured (fixture, git 2.50.1 Apple Git-155, against PR #914's post-fix tree)

Fixture: origin + work, one merged `main`, one unmerged `[gone]` branch `feature/unmerged`. Then `echo "deadbeef-not-a-sha" > .git/refs/heads/zzz-corrupt`.

```
$ git for-each-ref --format='%(refname) %(upstream:track)' refs/heads
warning: ignoring broken ref refs/heads/zzz-corrupt
refs/heads/feature/unmerged [gone]
refs/heads/main
git rc=0

$ git for-each-ref ... | awk '$2=="[gone]"{sub(/^refs\/heads\//,"",$1); print $1}'
warning: ignoring broken ref refs/heads/zzz-corrupt
feature/unmerged
pipeline rc=0

$ git for-each-ref ... 2>/dev/null      # the operator-facing channel removed
refs/heads/feature/unmerged [gone]
refs/heads/main
rc=0
```

The omitted ref appears in neither `reaped[]` nor `kept[]`, and rc is 0 on git and on the pipeline alike.

## Why an exit-status guard does not close this

This is **not** the swallow that PR #914's review also surfaced (pipeline takes `awk`'s status, never `git`'s — commented onto #622, where the control-flow ruling #391 reserved still stands). That one is about git *dying*. This one is git **succeeding** while returning less than it was asked for, so a guard on the captured status passes it through unchanged. Both halves live on the same statement; only one is addressed by a status check.

The in-repo precedent shows the same gap: `inflight.sh:489` guards the identical call —

```sh
if ! refs=$(git for-each-ref --format='%(refname:short)' refs/heads); then
[...]

Issue #915, opened 2026-08-25
inflight.sh enumerates branches with %(refname:short), so an ambiguous branch is reported to the operator as heads/<name>


feigi:
Declared and deliberately left by `impl-634` while implementing #634 (PR #914), and verified by the fleet controller before filing. It is a **sibling of #634 in mechanism and a different class in consequence**, which is why it is its own ticket rather than folded in.

## The defect

`skills/fleet/scripts/inflight.sh` enumerates local branches with `%(refname:short)`, the same spelling #634 fixed in `reap.sh`. That format is ambiguity-aware: when a tag shares a branch's name, git does not shorten to the bare name — it emits `heads/<name>`.

## Why the consequence is different from #634's, and much smaller

`reap.sh` **deletes**, and #634 established (measured, PR #914) that the ambiguous spelling there could authorize destroying an unmerged branch. `inflight.sh` deletes nothing — it reports.

`impl-634` verified rather than assumed the detection half:

- inflight's filter regex `(^|[/-])n([-/]|$)` **still matches** `heads/fix/95-slug`, so an in-flight ticket is still detected and the hit is still recorded.
- Only the operator-facing `local branches:` line names the prefixed form.

So this is a **diagnostic accuracy** defect: the probe reaches the right verdict and prints a branch name in a spelling that does not exist as a ref. Nobody is misled into a destructive action; an operator copying that name into a `git` command gets `branch 'heads/<name>' not found`.

## Fix

The same one that landed in `reap.sh` at PR #914 — ask for the unambiguous full refname and strip the prefix, rather than asking git to shorten:

```sh
git for-each-ref --format='%(refname) …' refs/heads | awk '{sub(/^refs\/heads\//,"",$1); …}'
```

`%(refname)` is always `refs/heads/<name>`, so the strip yields the bare name for every branch, ambiguous or not.

**Read PR #914 before implementing.** Its whole finding is that the enumeration fix is *not* safe alone in a script that also consumes `$b` as a **rev** — git resolves an ambiguous rev by preferring `refs/tags/` over `refs/heads/`. Enumerate every `$b` consumption site in `inflight.sh` and classify each as rev / string-key / namespace before changing the spelling, exactly as #914 did. If no site consumes it as a rev, the enumeration fix stands alone here; establish that rather than inheriting it.

## Also checked and deliberately not filed

[...]

PR #939, opened 2026-08-27
chore(fleet): this run's five tier rows, and the filing case step 5 never named

feigi:
Close-out for `/fleet:run-team 2 5`, 2026-08-25 → 08-27. **Controller-authored chore PR — not ticket work**, so it carries no `Closes`.

## The five tier rows

One row per implementer PR ruled this run: **#912** (#606), **#913** (#288), **#914** (#634), **#933** (#653), **#931** (#638). All five merged, all five tickets closed, all `tier=opus`, all `closed_own_ticket=yes`.

`minted_false_claim` splits 3 yes / 2 no — every one caught before merge, none shipped.

The rows carry the run's substance rather than a summary of it. Three worth reading:

- **#914** — the ticket's own prescribed remedy would have **destroyed an unmerged branch**. Restoring the bare refname makes it ambiguous *as a rev*, git prefers `refs/tags/` over `refs/heads/`, so the merge probe answers about the tag while `-D` deletes the branch. The ticket's "`git cherry` is unaffected" was verified against the pre-fix code and is false of the post-fix code.
- **#913** — a **coverage regression introduced by the item the ticket itself prescribed**. Collapsing two hand-rolled tests onto a shared helper removed the only fixture in the tree running `ci-state.mjs` outside a git repo. The simplification was right about the duplication and silent about what the duplication carried.
- **#931** — one defect, three dimensions, and nobody had it right. The `survived` finding proved the behaviour; the `refuted` finding attacked the comment and its dissenter correctly showed it refuted a sentence the file doesn't contain; an unrefuted `suggestion` held the fact that settled both — and *that* was incomplete too, corrected only by running `hasOwnProperty`.

## The step 5 gap

Step 5 ranged over `ready-for-agent` and `needs-triage`, the defect-confirmed axis. It named no arm for **a finding a refuter checked and killed** — not a confirmed defect, and not "unsure anything is broken" when you measured that nothing is.

Both fix-appliers fell through to `needs-triage`. The controller then ruled the wrong way on it, arguing separate open issues help the dedupe guard — doubly wrong, since `gh issue list --search` is full-text and the repo already had a settled convention in #483, #542, #571. A member following that convention rather than the brief, and flagging the deviation instead of complying quietly, is what surfaced the error. Four records were relabelled `wontfix` and closed afterwards.

The **split clause** carries as much weight as the rule. Of six records from one wave, two were not refutations at all — a residual a refuter surfaced, and a finding refused as out of scope and never checked. Both are real deferred work and stay open. A blanket close would have buried them, which is why the correction read all six before touching four.

## Verification

`node --test skills/fleet/scripts/*.test.mjs` → **tests 1181, pass 1181, fail 0**.

The prose edit is inside `review-and-fix.md` step 5, which no prose test slices; the TSV append is data. Neither is behaviour.

## Run outcome

Five tickets closed, five PRs merged, zero reverted, zero red merges. Every branch and worktree reaped; `main` fast-forwarded. Deferrals filed as #915–#925 and #929–#938, plus comments on #288, #603, #615, #622 and #915.


feigi:
> **Held for review in the NEXT `/fleet:run-team` run — maintainer's decision, recorded here so the next controller finds it.**

Not merged, deliberately, and **not** labelled `ready-to-merge` (that label is a reviewer's; a controller never self-applies it). CI is green: `check` and `rebase-check` both success, `behind: 0` at the time of writing — expect that behind-count to be stale by the time anyone reads this.

## What the next controller has to do, because nothing will do it automatically

Phase 3 enqueues a PR for review when **its own implementer** reports one. This PR predates that run, so **no edge will fire for it.** It has to be picked up explicitly:

```
Workflow({name: "review-pr", args: {
  pr: 939,
  branch: "chore/fleet-run-close-out-2026-08-27",
  worktree: "<abs path to a worktree on that branch>",
  testCmd: "node --test skills/fleet/scripts/*.test.mjs",
  scratch: "<per-PR scratch root, NOT shared>"
}})
```

The branch's worktree from the authoring run is at `.worktrees/closeout`; if it has been reaped, re-create one rather than pointing the review at the main checkout.

**Expect a thin review.** The diff is one TSV append plus one prose sentence, so `diff-stats` will likely profile it small or docs-adjacent and trim the dimension set — and `dimensionsUnrun` will be empty regardless, so nothing in the payload announces the trim. That is #739. Read the trim from `dimensionsRun`, not from the absence of findings.

## Two costs of holding it, both measured, both landing on the next run

**1. It pins the refill gate.** `fleet-tick` counts every open PR without `ready-to-merge` as review backlog. This PR is one, permanently, until it merges. One more open PR puts backlog at 2, and **implementers go to HOLD for the rest of that run** even with pool remaining. If the next run wants an implementer wave, either review-and-merge this early or state the backlog explicitly to `fleet-tick` net of it.

**2. The rule it carries cannot help the run that is reviewing it.** The step 5 fix lives on this branch only — `main`'s `review-and-fix.md` has no `wontfix` guidance at all (verified: zero matches). Fix-appliers read that file from the **main checkout**, not from this branch. So until this merges, every fix-applier will keep filing a review's refuted band as open `needs-triage`, which is exactly the error this PR exists to stop.

**Mitigation that does not require merging first:** the controller can state the rule inline in each fix-applier's dispatch prompt, which is what the authoring controller did for the second half of its run after discovering the gap. The rule is: *a finding a refuter checked and did not survive goes as one issue per review, titled `PR #<pr> review: the suggestion band, checked`, labelled `wontfix`, and **closed** — and a residual the refuter surfaced, or a finding refused as out of scope and never checked, is real deferred work that keeps its own open issue.*

## Provenance

`/fleet:run-team 2 5`, 2026-08-25 → 08-27. Five tickets closed (#606, #288, #634, #653, #638), five PRs merged, none reverted. This PR carries that run's five `tier-outcomes.tsv` rows and the step 5 fix.

Issue #472, opened 2026-08-14
tier-outcomes.tsv header has no test pinning it against SKILL.md's guard text


feigi:
Deferred from the PR #465 review (out of scope for that PR — it is a coverage gap, not a defect in the diff). PR #465 remains open.

**Location:** `docs/metrics/tier-outcomes.tsv` (whole file — no single line)

## Finding (verbatim, `suggestion`, dimension `tests`)

> No test in the suite checks that `docs/metrics/tier-outcomes.tsv`'s header/columns match what SKILL.md's guard text promises ("that file's header carries the column meanings"), even though the guard's entire accumulate-across-runs mechanism now depends on this file's shape.

## Evidence recorded by the reviewer

`grep -rln "tier-outcomes" skills/ docs/` returns only `skills/fleet/skills/run-team/SKILL.md` — no `*.test.mjs` references the file. The 4 seed rows and the header comment are not exercised by `node --test skills/fleet/scripts/*.test.mjs` (638 pass, none touch this path).

## Why deferred rather than applied

The review's own `suggested_fix` scopes it out: "Low priority for a docs-only PR — a seed data file, not logic — but if a future PR adds a script that reads this TSV, that script's tests should also assert the header stays column-for-column in sync with this file."

## What would close this

A pin that fails when the TSV header and SKILL.md's guard text disagree — most naturally added alongside the first script that actually reads the TSV. Adding a pin now would assert a shape nothing consumes.

Note for whoever picks this up: `skills/fleet/scripts/implementer-model-tier.test.mjs` already pins the guard prose (`at least three \`class=routine\` PRs`, `revert **\`class=routine\`**`), so the SKILL.md side of the contract has coverage; the TSV side does not.

feigi:
Evidence from the PR #523 review — the drift this issue predicts has now happened once, so this is no longer hypothetical.

PR #523 reverted `class=routine` → `sonnet` and rewrote SKILL.md's guard section to past tense, but left `docs/metrics/tier-outcomes.tsv`'s header asserting the reverted binding as **the current rule**: "class and tier are perfectly confounded by the current rule: routine always dispatches at sonnet, correction always at top". Two review dimensions found it independently (`correctness` and `comments`), and the whole suite stayed green throughout — `grep -rln "confounded\|tier-outcomes" skills/fleet/scripts/` still returns nothing.

[...]

PR #943, opened 2026-08-27
Record model and effort per fleet member, so tiering can be decided on evidence (#942)

### LINKED ISSUE #942: Record model and effort per fleet member, so tiering can be decided on evidence
Spec: docs/specs/2026-08-27-fleet-member-outcomes-instrumentation-design.md

The fleet dispatches six kinds of member and records the tier of none of them.
implementer-model-tier.test.mjs:19-21 states the gap: board.mjs parses
message.usage off the subagent JSONL and drops message.model on the same line.

Part 1 (this issue): the scraper + backfill.
Part 2 (separate): declared per-role tiers and the within-run control rule.

feigi:
Closes #942. Part 1 of 2 — the scraper and backfill. Part 2 (declared per-role tiers, the within-run control rule) is a separate plan on the same spec.

Spec: `docs/specs/2026-08-27-fleet-member-outcomes-instrumentation-design.md`
Plan: `docs/specs/2026-08-27-fleet-member-outcomes-plan-1-scraper.md`

## What this adds

`skills/fleet/scripts/member-outcomes.mjs` derives one row per dispatched fleet member from the transcripts the harness already writes, and `docs/metrics/member-outcomes.tsv` holds 5,393 backfilled rows. Phase 3 of run-team now runs it after ruling, pinned by a prose test.

The fleet dispatched six kinds of member and recorded the tier of none of them — `board.mjs` parsed `message.usage` off the subagent JSONL and dropped `message.model` on the same line. It no longer does.

Two files, opposite maintenance rules: `tier-outcomes.tsv` stays hand-ruled and append-only; `member-outcomes.tsv` is machine-derived and regenerable, keyed so a re-scrape replaces rows rather than duplicating them. No verdict column, ever — that split is what lets the facts file be regenerated when the classifier improves.

## Three defects that only running it could find

Every test passed throughout each of these. Each was caught by running the real pipeline, never by reading.

- **The idempotency key destroyed 44.6% of the corpus.** `session + member` collapsed every unnamed dispatch onto one key — 1,202 of 2,694 members lost across 85 sessions — because `member` falls back to `agentType`. Now keyed on the transcript id, carried as an `agent` column appended last so the positional awk read-outs keep pointing at the same fields. Every fixture gave its members distinct names, so nothing caught it.
- **The documented CLI form never worked.** `member-outcomes.mjs <session-dir>` exited 2 for every input: with `--file` absent, `indexOf` returns -1 and the filter dropped argument zero. All 32 tests passed `--file`.
- **The phase-3 instruction was unrunnable and failed silently.** `\$SESSION_DIR` was defined nowhere, and a wrong path exited 0 having written nothing. It now derives the path via `board.mjs`'s exported `findSubagentsDir`, accepts either that directory or its parent, prints its row count, and refuses a path with no `subagents/` inside.

## What the data can and cannot say

It can rank models. It **cannot** answer the effort question. The spec pre-registers the read-out and gates it; the header repeats the rule.

The raw pair query returns 123 within-run pairs, of which 60 are `memory-proxy` vs `memory-housekeeper` — two definitions whose tiers are pinned in frontmatter and never varied. Same-shape pairs are 63; specialists are 40 and implementers, the only role Part 2 varies, are 16.

There is no `errored` column — see the review round below for why it was removed.

## Verification

1233/1233 (`node --test "skills/fleet/scripts/*.test.mjs"`). The header's own awk queries run verbatim against the committed data. Re-running the backfill from the shipped code reproduces the file to the row — the only diff is this session's own still-growing transcripts, replaced in place.

**A second review round (2026-08-27) found three more Criticals of the same class and they are fixed in `29c4bc8`** — token columns counted jsonl lines rather than turns and biased the model comparison; the walk missed the 2,894 nested workflow transcripts, which are the half where model actually varies; and `parseTsv` padded malformed rows into unreplaceable phantoms. The corpus was regenerated (2,702 → 5,393) and the `errored` column dropped. Details in the review-round comment below.

Built with superpowers:subagent-driven-development: ten tasks, each implemented by a fresh subagent and reviewed against its brief, then a whole-branch review on Opus. 15 controller rulings are recorded in the commit messages where they changed the spec or plan.


feigi:
## Known deferrals and controller rulings

The PR body says rulings live in the commit messages. That covers the ones that changed the spec or plan, but not the findings we knowingly did **not** fix, and not the reasoning behind the judgement calls. Both belong here rather than in a session scratchpad, so here they are.

### Deferred findings — real, triaged, not fixed

Each was raised by a task review or the whole-branch review and deliberately left. None touches row-producing logic.

| # | Finding | Why deferred |
|---|---|---|
| 1 | Test named *"a member with no .jsonl is skipped without losing its siblings"* pins the `.jsonl` filter, not per-member recovery — the orphan `.meta.json` never reaches the try/catch | Recovery is genuinely covered by the malformed-meta test next to it. Rename the test, no coverage gap |
| 2 | Missing `meta.json` (ENOENT) has no dedicated fixture; only malformed meta does | Same catch block, structurally identical |
| 3 | `row()`'s default `agent` derives from `member`, so the "REPLACES" test shares an id incidentally | **Closed as a non-issue** by the final review — overriding `member` yields distinct agents, so the test fails loudly rather than passing wrongly |
| 4 | `--file` given twice refuses with the generic usage message rather than naming the problem | Fails safe: exit 2, no partial write |
| 5 | `--file` under a missing parent dir throws a raw ENOENT stack (exit 1) instead of the clean `NAME: msg` exit 2 | Inconsistent with the two written guards, but not a data risk |
| 6 | Default `--file` is CWD-relative | Deliberate — every sibling script in `skills/fleet/scripts` trusts cwd; repo-root resolution adds worktree/nesting failure modes for an unobserved problem. Callers needing cwd-independence pass `--file "$(git rev-parse --show-toplevel)/..."` |
| 7 | Header pin's column-line lookup uses `includes("run_date")`, first match top-down | Unambiguous today (one such line). A future header paragraph mentioning `run_date` above it would silently pin prose. Remedy: anchor on `startsWith("# session\t")` |
| 8 | The header pin's four `assert.match` calls scan the whole file, so a pinned rule moved into a data row would still pass | They do fail on deletion, which is the realistic drift case for #472. Remedy: match against the header slice, not the file |

**Parked residual:** SKILL.md's derivation one-liner does not special-case `findSubagentsDir`'s `{error: "..."}` return, so on a machine with no transcript dir `$SESSION_DIR` becomes a stringified error object. It fails safe — the CLI's new guard refuses it with exit 2 and writes nothing — so the cost is a garbled message, not a wrong write. That loud-and-harmless failure is precisely the property the phase-3 command lacked before this PR.

### Judgement calls worth a second opinion

- **The `agent` column is a schema change I made mid-implementation.** It is a harness-internal id now permanent in a metrics file. It earns its place by being the only per-member identifier that is unique, stable across re-scrapes, and already on disk — but the objection that it leaks an implementation detail into a published schema is fair, and the mitigation was documentation (`opaque join key`), not removal.
- **`errored` reads `no` on all 2,702 rows.** Honestly labelled now, but arguably useless. Dropping the column is a reasonable call.
- **The spec's role table was left stale** with a note, rather than recomputed, because a dated design doc is a record — but it now carries a behavioural claim the shipped classifier contradicts, annotated in place.

### Reviewing this efficiently

The three defects in the PR body were each found by running the real pipeline against real transcripts, never by reading the diff. If you want to spend review effort where it has paid off before, that is where — particularly anything where a test and its fixture share an assumption the real callers do not.

feigi:
## Review round 2 — three more Criticals, same class as the first three

Fixed in `29c4bc8`. Each passed 1,223 green tests for the same reason as the three in the PR body: every fixture made a distinction the real callers do not.

### 1. Token columns counted jsonl LINES, not turns

`board.mjs:221-232` already documents this bug and fixes it. This scraper reintroduced it — in a file that imports `classifyRole` from the same neighbourhood *"so the two files cannot fork on what a reviewer is."* They forked on tokens instead, 3.11× apart on the same member.

```
all 2,723 flat transcripts   affected 2,704 (99.3%)
cache_creation  +176.0%      turns  +140.9%      output_tokens  +1.5%
```

Not a rescale — model-dependent, because blocks-per-turn tracks how tool-heavy a turn is:

| model | members | inflation |
|---|---|---|
| claude-opus-5 | 1810 | 2.81x |
| claude-haiku-4-5 | 586 | 2.49x |
| claude-sonnet-5 | 286 | 2.39x |

opus-5 2.81× against sonnet-5 2.39× is a 17.6% spurious cost advantage handed to sonnet by the counting rule alone. Cross-checked after the fix: `board.mjs` and the scraper now both read 1,466,156 on `fix-pr-480`, turns 213 → 90.

**Why green:** no fixture line carried a `message.id`, so every fixture line was its own turn by construction — under a test named *"usage sums across turns"* that summed lines. Real transcripts: 0 of 127,102 assistant lines lack one.

### 2. The walk saw 48% of the corpus

```
2723  subagents/agent-*.jsonl                    <- scraped
2894  subagents/workflows/wf_*/agent-*.jsonl     <- never read   (37 sessions)
```

The severity isn't the volume. The nested half is `workflows/review-pr.js`'s specialists, and that file pins `model: "sonnet"` on three of its six dimensions — so **the half where model is deliberately varied was the half being dropped** (17.0% sonnet against the flat half's 10.5%). Any model ranking off the old file came from the non-varied half.

`agent` is now the path-relative stem. Flat stems are byte-identical under a recursive walk, so this widening **replaced** existing rows rather than duplicating them: 0 of the 2,702 committed keys lost. Within-run pairs 93 → 123; specialist pairs 10 → 40.

### 3. `parseTsv` padded malformed rows into unreplaceable phantoms

`agent` is the last column, so a short row parsed to `agent: ""` — a key `rowsForSession` can never produce, so `mergeRows` can never replace it. Three measured routes, all exit 0:

| trigger | outcome |
|---|---|
| torn last line | phantom row **widened to a valid 13-column row**; re-scraping that same session could not heal it |
| git conflict markers | `<<<<<<< HEAD` / `=======` / `>>>>>>> theirs` became three data rows |
| one column added ahead of `agent` | 2,702 rows → 156 |

The phantom passed a structural check, so a field-count histogram over the file could not catch it. Write is now `.tmp` + `renameSync` — `writeFileSync` truncates first, and an interrupt left a corpus the next scrape would read, merge and rewrite, losing every other session for good.

## Also fixed

- **Phase 3 scraped the wrong session.** `findSubagentsDir` answers *"newest transcript for this cwd"*, not *"this run"*. 86 sessions share `~/.claude`, and **17 of the 21 days with any subagent activity had two or more writing**. run-team now loops every session dir for the cwd — cannot pick wrong, idempotent by construction.
- **A zero-yield scrape was byte-identical to a success.** `merged.length` is the whole corpus, so a session where every member dropped printed the same line as a healthy one. Now `scraped N of M members (D dropped)` alongside the file total.
- **The `existsSync` guard closed only the ENOENT half** — EACCES and ENOTDIR walked past it back into the silent exit 0 its own comment claimed to have fixed.
- **Header preservation** hoisted below-data comments to the top and dropped blank lines, while promising "verbatim".

## `errored` removed

It read `no` on all 2,702 rows, including all 34 transcripts carrying a real terminal API failure, because what it detects is a torn read rather than a stall. Demonstrated mechanically:

```
scrape 1 (mid-write):  errored=yes  tokens_out=10
scrape 2 (finished):   errored=no   tokens_out=30
```

It describes *when the scraper ran*, not the member, and erases itself on re-scrape. It needed four header lines to stop readers taking it for a reliability signal — that caveat was the tell. Detection kept; it goes to stderr.

`agent` **stays**, and not narrowly. No alternative key is both unique and stable:

| key | distinct of 2,715 |
|---|---|
| `session+member` | 1,497 |
| `session+member+model` | 1,509 |
| `session+member+tokens` | 2,714, and unstable — counts grow while live |
| `session+agent` | 2,715 |

Defect 1 doubly disqualifies the token key. The header now also says `agent` is unique *within a session* — 5 ids appear in 2 sessions each, because a named member's stem is derived from its name.

## Pins — every one mutation-tested

Two existing pins were proven **vacuous** first: the prose pin stayed green when the entire bash block was deleted and replaced with prose, and header test 3 stayed green when the pair query was repointed from `$1/$3/$5` to `$2/$4/$6`. Both rewritten. All 8 mutations now go red:

per-line usage · non-recursive walk · padded short rows · `existsSync`-only guard · file-size-only report · misindexed pair query · unrunnable phase-3 instruction · pinned header rule moved into the data

New pins also cover `role` (it had no assertion anywhere, and it is the pair query's grouping column) and the phase-3 instruction's *runnability* — fenced block present, every variable it uses assigned in the same block.

## One recommendation deliberately not applied

A reviewer asked `parseMemberName` to strip a numeric retry suffix so `impl-137-2` books `ticket=137`. Its description reads *"Implement 137+138+139 set"* — a multi-ticket batch no single `ticket` value represents. Blank is the honest answer; the regex is unchanged with a comment saying why.

## Deferred list

The 8 deferred findings stand as reasoned, and the parked residual's *"fails safe"* claim is confirmed by execution — the stringified `{error}` stays one shell word under quoting, the guard fires, exit 2.

One disagreement, **#6** (default `--file` is CWD-relative): not asking for repo-root resolution, but this PR adds a caller that runs the scraper for its side effect on a tracked, committed, wholesale-rewritten file, which the sibling scripts do not do. There is no commit instruction for `member-outcomes.tsv` *or* `tier-outcomes.tsv` anywhere in SKILL.md — a pre-existing hole, but the diff is now 5,000 lines rather than one row. Added a sentence to the phase-3 block naming the duty; a real fix belongs with whatever owns run-artifact commits.

Two small extras: the spec's backfill loop enumerated `*/*/` (253 dirs, only 155 with `subagents/`), so it hit the new exit-2 guard ~98 times — narrowed to `*/*/subagents/`. And `normalizeModel`'s comment claimed a count-splitting hazard its own code path forecloses; reworded to say the strip is defensive and that bare aliases are a real, unhandled gap in `meta.model`.

## Verification

1233/1233. Header awk read-outs run verbatim against the regenerated file. Full regeneration from shipped code reproduces it; 0 of the 2,702 prior keys lost. Corpus 2,702 → 5,393.

Specs annotated in place rather than rewritten — the recount at `design.md:240` pinned the wrong denominator (2,687 against a real 5,617), because `ls -d .../subagents/` is exactly what hides the nested half, and plan-1 Task 9's `specialist ~878` sanity check was calibrated against the defect it was meant to catch.

No finding here concludes anything about **effort**. The gate at `design.md:401` stands, and defect 2 means the corpus is not a census either — sample composition, not an effort claim.

PR #947, opened 2026-08-27
fix(ci): provision gojq so the engine-parity gates run instead of skipping

### LINKED ISSUE #337: ci.yml: nothing provisions gojq, so the dependency-scan engine gate never runs on CI
**Deferred from PR #331 review** (issue #58). Dimension: `silent-failure`. Severity: suggestion (unrefuted — the review budgets that band 0 refuters).

`.github/workflows/ci.yml:173-176` runs `node --test skills/fleet/scripts/*.test.mjs` with no `setup-go` and no `go install`, so `findGojq()` in `skills/fleet/scripts/candidates.test.mjs:304` always takes the skip branch on CI. The engine-parity test — the only executable check that the `--jq` program holds under gojq/RE2, which is what `gh` actually applies — is green on CI whether or not the regex holds.

Measured during the review: with `GOPATH=/nonexistent-probe` the suite reported `pass 40 / skipped 1` and exited 0, fully green with the engine check never executed.

**Why deferred, not applied:** out of issue #58's scope — it edits `.github/workflows/ci.yml`, a file that ticket has no business in.

**Sequencing constraint:** #163 is a separate staged change to the *same* workflow file (node-version pin -> version file). Whoever takes this must sequence against it rather than land both blind.

**Why needs-triage:** the defect is verified, the approach is not settled — `actions/setup-go` + `go install …@v0.12.19`, a prebuilt gojq release binary, or leaving the skip in place and accepting that the engine gate is a local-only check, are all defensible and have different CI-time costs.

Note PR #331 hardened the surrounding behaviour: a missing `go` no longer crashes the whole file, and the gate now verifies the binary is actually gojq. So today CI skips loudly rather than passing on the wrong engine — this ticket is about making it *run*.

feigi:
`candidates.test.mjs` gates its engine-parity tests on a resolvable `gojq` binary — the only executable checks that the `--jq` programs hold under the engine `gh` actually applies. The `check` job provisioned none, so those gates took the skip branch on every run, and a skipped gate reports under `fail 0` exactly like one that ran.

## What changed

- `actions/setup-go@v6` plus `go install github.com/itchyny/gojq/cmd/gojq@v0.12.19` — pinned, per the Agent Brief's ruling, so a gojq release cannot change what the gate means without a commit here.
- The Tests step names the installed binary in `GOJQ_BIN`. That is the half that makes a failed provision **loud**: `findGojq()` already hard-fails on a `GOJQ_BIN` that is not gojq (PR #331's contract) instead of falling back to a skip, so a provisioning step that silently produced nothing reds the job. No new mechanism — the existing escape hatch, wired up.
- A pin in `ci-vacuous-green.test.mjs`, whose subject is already "a CI step that goes green having checked nothing": the wiring can otherwise be deleted with the whole suite still green.

`setup-go` runs after the syntax/shellcheck steps rather than beside `setup-node`, so a lint failure still fails the job without paying for the toolchain. `cache: false` because setup-go's cache keys off a `go.sum` and this repo has no Go module.

## Measurements

All against this branch, `agent-test`, darwin.

| Run | Result |
|---|---|
| `candidates.test.mjs`, `GOPATH=/nonexistent-probe`, no gojq on `PATH`, no `go` on `PATH` | `fail 0` / `skipped 2` |
| `candidates.test.mjs`, `GOJQ_BIN` set (CI's resolution path) | `fail 0` / `skipped 0` |
| whole suite, `GOJQ_BIN` set | `fail 0` / `skipped 0` |

**Controls.** A deliberately broken engine assertion must fail, and it does — each gated test's U+00A0 discriminator was flipped to the answer system jq gives, one at a time, and run with `GOJQ_BIN` pointing at real gojq: `fail 1`, `skipped 0`, the gate naming itself in the failure. Both gates were mutated, not just one.

**Guard behaviour preserved.** With no `go`, no gojq and `GOPATH` neutralised the file still skips loudly and every other test in it runs — the accept-side input PR #331's hardening exists for. With `GOJQ_BIN` naming something that is not gojq (`/bin/echo`, and a path that does not exist) the run is `fail 1`, not a skip.

**Pin.** Deleting the install, floating it to `@latest`, and deleting the `GOJQ_BIN` env each red the new test, with a message naming what was lost.

## Added CI time

**28s**, measured on this PR's own first run (`ubuntu-latest`, run 33112971037): `actions/setup-go` 12s, `Install gojq` 16s. The Tests step itself did not move — 118s here against 106s/118s/122s on the three most recent successful `main` runs, so the two gates that now actually execute cost less than the run-to-run spread.

## On CI

Run 33112971037, this branch's head:

```
ℹ tests 1234  ℹ pass 1233  ℹ fail 0  ℹ skipped 1
✔ dependency forms hold under gojq, the engine gh actually applies — not only system jq
✔ the spec predicate holds under gojq — depth widened, split heading not spanned, CRLF caught
﹣ build: --spend-since's refusal survives a gh child that has already pushed past the pipe buffer, unread # async pipe-write loss (#363) is darwin-only — a green run here is not coverage
```

Both gates executed. **One acceptance criterion cannot be met literally:** the Brief asks for `skipped 0` for the whole suite, and `board-cli.test.mjs`'s `#363` gate skips on any non-darwin platform, so a `ubuntu-latest` run reports a non-zero suite `skipped` whatever gojq does — the line above is that gate, and nothing here touches it. The criterion's substance is met and is readable per-file: `candidates.test.mjs` goes from `skipped 2` to `skipped 0`.

The control composes across the two: CI shows the gated tests executing rather than skipping, and locally — resolving gojq the same way CI does, through `GOJQ_BIN` — a broken engine assertion in either of them reds the run.

Closes #337

## Review follow-up (second commit)

Review of this PR found three defects in what the first commit added; all three are fixed here, so the diff now also touches `candidates.test.mjs`.

- **The pin could not tell a live command from a commented-out one.** All three assertions in `ci.yml provisions gojq for the engine gates, at a pinned version` matched an unfiltered read of `ci.yml`, and a `#`-commented line still contains the literal they look for. Commenting out the `GOJQ_BIN` env line, or replacing the install `run:` body with a comment plus `true`, both left the test green while the engine-parity gates went back to skipping — the same silent green the test exists to refuse. Comments are now dropped line-wise before `flat()`, in whole-line and trailing form, so all three assertions are fixed at the read.
- **A failure message named a job that does not exist.** The first assertion reported on "the Tests job". `ci.yml` has no such job — `check` is the job, `Tests` is a step in it, and the install that assertion pins lives in the separate `Install gojq` step preceding it. The message now names that step.
- **`candidates.test.mjs` still recommended `@latest`.** Its skip message and the comment above the gated test both told a developer to `go install …gojq@latest` to reproduce locally — the exact float this PR pins `ci.yml` against, so a local run could measure a different engine than CI's `v0.12.19` with nothing saying so. Both sites now name the pinned version.

**Mutation verification.** Pre-fix and post-fix versions of the test were run over the same `ci.yml` mutants in a private copy. Commenting out `GOJQ_BIN`, commenting out the install with a `true` body, and the trailing-comment form of the same neutralisation were all GREEN before the fix and are all RED after it. Two controls stay GREEN in both: reflowing whitespace inside `${{ runner.temp }}` together with an added unrelated comment line, and — post-fix — `@latest` appearing only inside a comment, which the raw read would have reddened. The negative `@latest` assertion still reds on a live `@latest` alongside the pin, and deleting the `Install gojq` step still reds.

`node --test skills/fleet/scripts/*.test.mjs` from the worktree: `tests 1234`, `pass 1234`, `fail 0`, `skipped 0`.

Four review findings are deferred rather than applied and are filed as their own issues; the numbers are posted as a comment on this PR.


feigi:
Review deferrals filed, all `needs-triage`:

- **#950** — `ci-vacuous-green.test.mjs`: the `phrase()` pins lift GitHub Actions text verbatim, so removing the spaces inside `${{ }}` reds a functionally identical `ci.yml`. *tests*. Came back 1-1; settled on method as a style question rather than a defect this PR introduces, since no formatter mechanism exists in this repo and the text lift is the file's convention at every `phrase()` site. Orthogonal to the comment-stripping fix that shipped in `0e70a41`; the body records both the reproduction and the refutation.
- **#951** — `board-cli.test.mjs`'s `#363` gate is darwin-only and the test job is `ubuntu-latest`, so it never executes on CI — and #363 is CLOSED, so it had no open home. *correctness*. No change here: fixing it in this PR would be scope creep. Filed so this PR's "the engine checks never ran on CI" framing is not read as covering every remaining CI skip.
- **#952** — REFUTED: the test does not pin `GOBIN`, but dropping `GOBIN` reds the job loudly through `findGojq()`'s hard-fail rather than reverting to a skip. *correctness*. The proposed assertion is non-vacuous but adds attribution, not detection — it fires only in a world the suite already reds. Measurements in the body.
- **#953** — REFUTED: `ci.yml` has no blank-line-between-comment-blocks convention to break — a mechanical enumeration of every comment run in the file found no site where a blank line separates two comment runs, and applying the suggested fix would create the file's only instance of the pattern it claims to restore. *comments*. The body records the adjacent observation that is real: the pre-existing failglob/Tests rationale is now detached from the step it documents, which a blank line would not fix.

Applied instead, in `0e70a41`: the comment-stripping fix covering every assertion in the gojq pin, the `Tests job` → `Install gojq` step wording correction, and the `@latest` → `@v0.12.19` update at both `candidates.test.mjs` sites. Mutation results are in the PR body.

Issue #346, opened 2026-08-11
inflight.sh probe 2: an https origin is still unbounded — http.lowSpeed* only bounds a transfer already under way


feigi:
`skills/fleet/scripts/inflight.sh:182` (comment) / `:188` (the `git ls-remote` call) — re-derive at head, the line numbers moved in PR #332.

PR #332 bounded probe 2's ssh path (`ConnectTimeout` + the `ServerAlive` pair) and set `http.lowSpeedLimit=1000` / `http.lowSpeedTime=10` for the http path. The ssh half holds. **The http half does not hold for `https`, and does not bound a connect stall on either scheme.** So the exact hazard #92 names — an unattended probe holding a fleet slot indefinitely — survives for an https origin.

## Measured (twice, by two independent refuters, against PR #332's head)

Silent TCP listener (accept, never write), `GIT_TERMINAL_PROMPT=0`, `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM=/dev/null`, `git -c http.lowSpeedLimit=1000 -c http.lowSpeedTime=10 ls-remote --heads origin`:

| origin | result |
| --- | --- |
| `http://127.0.0.1:$P/x/y.git` | rc 128 at **10.0s** — `Operation too slow. Less than 1000 bytes/sec transferred the last 10 seconds`. Bound works. |
| `https://127.0.0.1:$P/x/y.git` | **still running at 90s, 120s** — killed by the harness cap. Unbounded. |
| `http://192.0.2.1/x/y.git` (blackholed SYN) | rc 128 at **75.0s** — curl's own default, 7.5x the ssh `ConnectTimeout=10` budget. Adding `-c http.connectTimeout=5000` changed nothing. |

Cause: curl's low-speed timer never arms because no transfer ever begins — the TLS handshake does not complete. `git help config | grep -iE 'http\.[a-zA-Z]*(timeout|Speed)'` lists only `http.lowSpeedLimit` and `http.lowSpeedTime`; **git exposes no connect/handshake knob at all**, so this cannot be closed with another `-c` flag.

## Why deferred rather than fixed in PR #332

[...]

Issue #872, opened 2026-08-24
REFUTED, filed as record: stripHashGutter is duplicated in two prose tests, but hoisting it into prose-pin.mjs adds an unpinned export to the contract module — 4 of its 6 mutations already survive green


feigi:
Deferred from PR #866 review (issue #800). Dimension: **simplify**.

`skills/fleet/scripts/inflight-citation-prose.test.mjs:37` and `skills/fleet/scripts/reaping-prose.test.mjs:65` — the `stripHashGutter` helper, defined identically in both.

**This finding was REFUTED by a measured adversarial check, on the seam's value rather than on its facts.** Filed as a record, plus one thing the refuter found that is worth keeping.

## The finding

`stripHashGutter` and its explanatory comment are duplicated byte-for-byte between the two prose test files; the one-line helper belongs in `prose-pin.mjs`, which both files already import.

**Every factual premise checks out.** Byte-identity at the PR head: `diff` of the two blocks is empty, both md5 `4ebf69c005f5516204d31a6b60efb80b`. Both comment blocks are generic (shell-comment wrapping mechanics), so no file-specific wording would be flattened. The hoist applied in a scratch copy runs `tests 1127 / pass 1127 / fail 0`, identical to baseline, with behaviour equivalence confirmed (the same gutter-strip mutation reds the same single test in both trees) and a zero-behaviour reflow staying green. No sweep test objects. No doc anywhere states that prose helpers are deliberately local, so the copy is not documented as intentional — and the finding's own escape hatch (merge-conflict risk from other in-flight PRs importing `prose-pin.mjs`) is empirically nil.

## Why it was refuted

`prose-pin.mjs` is not a utility bucket — it is this repo's **contract module for prose pins, and every export it has is itself pinned** by `prose-pin.test.mjs`: `between` (3 tests) and `phrase` (2 tests), 2 of 2 exports covered.

The suggested fix adds a **third export with zero pins**. Measured mutation survival of `stripHashGutter` against all three consumer files (12 tests):

```
M1  drop \s?        (leading space kept)     pass 12  fail 0   SURVIVES
M2  drop ^ anchor   (strips mid-line #)      pass 12  fail 0   SURVIVES
M3  join("")                                 pass 11  fail 1
M4  drop \s*        (indented # kept)        pass 12  fail 0   SURVIVES
M5  add /g flag                              pass 12  fail 0   SURVIVES
M6  return text     (total gut)              pass 11  fail 1
```

[...]

PR #949, opened 2026-08-27
docs(inflight): probe 2's ssh options are defaults the user overrides, not an override

### LINKED ISSUE #350: inflight.sh comment: 'options land on top of' reads as override, but ssh -o is first-wins — the user's own value always wins
`skills/fleet/scripts/inflight.sh`, the comment sentence "A user's own ssh command is honoured, not replaced — options land on top of whatever GIT_SSH_COMMAND, core.sshCommand or GIT_SSH already says" (~line 178 at PR #332's head — re-derive, PR #332 rewrote this block).

**Filed as a deferral. The mechanism is measured and real; the consequence drawn from it was refuted.**

## The mechanism (measured, agreed by every specialist that looked)

OpenSSH is **first-wins** for a repeated `-o`. `inflight.sh` appends its bound options **last**, after the user's own command:

```
$ ssh -o BatchMode=no  -o BatchMode=yes -G example.invalid | grep ^batchmode   ->  batchmode no
$ ssh -o BatchMode=yes -o BatchMode=no  -G example.invalid | grep ^batchmode   ->  batchmode yes
$ ssh -o ConnectTimeout=45 -o ConnectTimeout=10 -G example.invalid             ->  connecttimeout 45
```

Live, against a silent listener: with the user having set `ConnectTimeout=45`, the script's 10s never applied — the run took 45.0s.

## Why the `important` version of this was refuted

The `silent-failure` finding drew the conclusion "the hardening no-ops and nothing warns". A refuter ran the real end-to-end case through the PR's own fixture and found the script still terminates on its own, at the user's chosen value: `user CT=3 → status 2 at 3.7s`, `user CT=25 → status 2 at 25.1s`. Never exit 0, never a hang. So the bound **degrades to whatever bound the user explicitly asked for**, which is the documented design ("a user's own ssh command is honoured, not replaced … so a configured identity file or proxy command still runs"). Reversing the order to ours-first would silently override a user's deliberate proxy/timeout config — a real regression traded for a hypothetical one. `grep -rn "GIT_SSH_COMMAND\|sshCommand"` over the repo finds nothing outside `inflight.sh` itself, so no caller in this repo sets either.

## What is still open

A `correctness` **`suggestion`** (budgeted 0 refuters — nothing verified it) makes a narrower, prose-only claim the refutation above does not settle:

> the comment's "options land on top of" implies the script's values override the user's, when first-wins means the opposite — a user who has already set `BatchMode=no` or a long/zero `ConnectTimeout` keeps the pre-fix prompting/unbounded behaviour.

Two sub-questions for whoever picks this up:

1. **Prose**: does "land on top of" need rewording to say the script's options are *defaults the user's own command wins over*, rather than an override? (This is the live part.)
2. **`BatchMode=no` specifically**: the refuted finding noted that `GIT_TERMINAL_PROMPT=0` does not cover ssh's host-key or passphrase prompt, so a user who has explicitly set `BatchMode=no` re-opens the exact unattended hang #92 exists to stop. That is a deliberate user opt-out rather than a defect — but it may deserve either a warning or a documented "we honour it and here is what you lose".

**Nothing was applied**, per controller ruling: an unverified copy of a claim refuted elsewhere is not applied from the fix-applier's seat.

`dimension`: `correctness` (unverified `suggestion`), with the refuted `silent-failure` finding's measurements folded in

Deferred from PR #332 review (closes #92).

feigi:
Probe 2's ssh-hardening comment said its options *"land on top of whatever `GIT_SSH_COMMAND`, `core.sshCommand` or `GIT_SSH` already says"*. That reads as override and states the reverse of the mechanism: the options are appended **last**, and ssh takes the **first** value of a repeated `-o`, so a user who has set `ConnectTimeout` keeps theirs and the 10s set here is discarded. A reader trusting the old sentence believed this probe's timeout bounded every run.

**`inflight.sh` is comment-only** — verified: with comments stripped, the script is byte-identical to `origin/main`, and no added or removed line in it is anything but a comment. The option order is deliberate and untouched; reversing it would silently override a user's proxy or timeout config, a real regression traded for a hypothetical one. The test files do change: this PR adds the prose pin and, after review, the executable pin on the append order that the whole comment rests on.

## Every restated claim re-measured against this tree and this host's ssh

`OpenSSH_10.2p1, LibreSSL 3.3.6`.

| Claim now in the comment | How it was settled |
|---|---|
| ssh takes the FIRST value of a repeated `-o` | `ssh -o BatchMode=no -o BatchMode=yes -G` → `batchmode no`; reversed → `batchmode yes` |
| `-o ConnectTimeout=45 -o ConnectTimeout=10 -G` reports `connecttimeout 45` | run verbatim |
| a user `ConnectTimeout` of 3 cuts at 3.0s, of 25 at 25.0s; the 10s applies only where the user set none | live against an accept-then-silent listener, options composed exactly as this probe composes them → 3.0s / 25.0s / 10.0s, `rc=255` every time |
| the bound degrades to the user's value, **and terminates only where that value does** | same listener, same composition: no user option → exited at 10.2s; user `ConnectTimeout=3` → 3.1s; user `ConnectTimeout=0` → **still connecting at 40s, killed**. `ssh -o ConnectTimeout=0 -o ConnectTimeout=10 -G` → `connecttimeout 0`, so 0 is accepted and wins by the same rule |
| `GIT_TERMINAL_PROMPT=0` does not reach ssh's own prompts | `GIT_TERMINAL_PROMPT=0` set throughout: `BatchMode=yes` alone fails an unknown host key at once (`Host key verification failed`); a user's `BatchMode=no` ahead of it sat on `Are you sure you want to continue connecting` until killed |
| "Reachable is the operative word" | same case with no terminal available aborts instead of waiting — the qualifier is measured, not hedged |

The ticket's own numbers were re-run rather than copied; the terminal-reachability qualifier is a distinction the ticket did not carry.

### Corrected in review: the comment's own absolute was false

The first draft of this comment closed the paragraph with *"ssh still terminates on its own, at their value, and no case here hangs."* The first half holds; the absolute does not. A user `ConnectTimeout=0` wins by the very first-wins rule the same paragraph documents and leaves the probe with no bound at all — the row above measures it still connecting at 40s where the 10s set here cuts at 10.2s. That is #92's unattended hang, reached through the user's own config, and closing it needs the same bound outside git that the https origin waits on, #346. The sentence also contradicted the paragraph on `BatchMode=no`, which describes a user sitting on the host-key prompt until killed — also a hang, also inside the scope of "no case here". The comment now states the limit instead of denying it.

## The pins

**Prose** — `inflight-ssh-precedence-prose.test.mjs`: four pins over probe 2's comment slice, bounded at both ends. Three positive (the first-wins rule, its consequence, the `BatchMode=no` opt-out plus what `GIT_TERMINAL_PROMPT=0` leaves alone) and one negative on the framing that rotted.

Mutation-tested on a copy, worktree untouched. Five mutants, each killed by **exactly one** pin: the rule flipped to LAST; the consequence sentence deleted; `-o BatchMode=no` unnamed; the `GIT_TERMINAL_PROMPT=0` clause reworded; the override framing reintroduced. Three negative controls stay green — reflow, a new sentence spliced beside a pinned clause, and an out-of-slice edit that itself contains the forbidden phrase.

The negative pin was widened in review: it could not match `landing on top of` — the exact wording of its own test name — nor any capitalised form, so the override framing could have returned in the participle with CI green. Now `/land(s|ing)?\s+on\s+top\s+of/i`, re-mutated in both directions: `landing on top of`, `lands on top of` and `LAND ON TOP OF` each redden it and nothing else, and the file's documented ceiling is unchanged — bare `"on top of"` without `"land"` still passes, so probe 2 keeps the phrase for the non-precedence claims it already makes. One false refusal measured and recorded in the file — a future sentence in probe 2 *recounting* the old wording would redden it; a word test cannot separate the claim from a mention of it, and that history belongs in the test file and on the commit.

**Behaviour** — the comment's central claim is that the bound options are appended **after** the user's own, and until review nothing executable observed it: `assertBoundOptions` in `inflight.test.mjs` asserted set membership, so inserting the options ahead of the user's command left the whole suite green while making the comment a lie. It now asserts position, and a new case gives the user a colliding `-o ConnectTimeout=45` and requires the first `ConnectTimeout=` on the invoked command line to be theirs. Under that inversion mutant exactly three tests redden — the two tier tests and the new colliding case — and shuffling the four bound options among themselves, still appended last, stays green.

## Class swept, and what is deliberately left

`git grep` over all 148 tracked files for `on top`, `land(s|ing) on top`, `layered on top`, `on top of whatever`, `takes precedence`, `override(s|n)`, `wins`, `first-wins`, `last-wins`, `clobber`, `supersed`:

- **Covered** — the probe 2 comment block, the only site making a precedence claim about repeated `-o`.
- **Left, not a member** — `inflight.test.mjs`'s three ssh test titles say "with the bound options added on top", but their claim is that the options reach the invoked command line, not which value wins. Both clauses true as written; renaming them is diff this ticket did not ask for.
- **Left, true as written** — `prove-merge.test.mjs`'s "a later merge in the same wave lands on top" (merge ordering, where last really does win) and `no-undo-audit.sh`'s `--autostash` "stores its own on top" (stash stack). Remaining hits are layout or composition, not precedence.

No file cites this block by line number, and the one cross-referenced phrase, "honoured, not replaced", is preserved verbatim.

Reversing the option order, making the script's options win, a runtime warning on `BatchMode=no` — all three measured and refused upstream. Bounding the probe from outside git — which is what both the unbounded https origin and the `ConnectTimeout=0` case need — stays #346.

Full suite: **1238/1238, 0 failures.**

Closes #350


feigi:
## Review actions — applied / deferred

Pushed `f714c26`. Full suite from the worktree: `node --test skills/fleet/scripts/*.test.mjs` → **tests 1238, pass 1238, fail 0** (baseline was 1237; +1 is the new colliding-`-o` case).

### Applied

| # | Finding | What changed |
|---|---|---|
| 1 | The new comment's *"ssh still terminates on its own, at their value, and no case here hangs"* is false, and contradicts the `BatchMode=no` paragraph in the same comment | Sentence rewritten to what was re-measured. Two dimensions found this independently; it is one defect and was fixed once. |
| 2 | The negative pin cannot match `landing on top of` — the wording of its own test name — nor any capitalised form | `/lands?\s+on\s+top\s+of/` → `/land(s\|ing)?\s+on\s+top\s+of/i` |
| 3 | The append order the whole comment rests on was pinned only as prose | `assertBoundOptions` now asserts position, not membership; new case gives the user a colliding `-o ConnectTimeout=45` |

**Re-measured before writing, not copied from the review** — `OpenSSH_10.2p1, LibreSSL 3.3.6`, local accept-then-silent listener, this probe's options appended exactly as it appends them: no user option → exited at **10.2s**; user `ConnectTimeout=3` → **3.1s**; user `ConnectTimeout=0` → **still connecting at 40s**, killed. `ssh -o ConnectTimeout=0 -o ConnectTimeout=10 -G` → `connecttimeout 0`. The `ssh -G` form is hermetic; the listener case is a few lines of python.

Mutation-verified on a scratch copy, worktree untouched, both directions:

- negative pin **reddens** on `landing on top of`, `lands on top of`, `LAND ON TOP OF` — each `4/3/1`, and each time the failing test is the negative pin and nothing else;
- negative pin **stays green** on bare `"on top of"` without `"land"` — `4/4/0`, the file's documented, deliberately-accepted ceiling, intact;
- order mutant (bound options inserted after the program word, ahead of the user's own) **reddens exactly three** tests — the two tier tests and the new colliding case — `82/79/3`;
- shuffling the four bound options among themselves, still appended last, **stays green** — `82/82/0`;
- the positive pin on the rewritten consequence sentence still bites.

`inflight.sh` remains **comment-only**: comment-stripped, it is byte-identical to `origin/main`. PR body reconciled against the final diff.

### Deferred and filed

| Finding | Where it went |
|---|---|
| `stripHashGutter` now duplicated in **three** prose test files | **Commented on #872**, not re-filed — that issue already refuted this hoist on measured grounds (`prose-pin.mjs` is the pinned contract module; the hoist adds an unpinned export). The comment records the third site and that the hoist still measures green (16/16 on the four affected files, 1237/1237 full suite) so the work is not redone. |
| Both user-config hang paths emit no diagnostic; the ssh path is unbounded outside git | **Commented on #346**, which is open and is the outside-git bound. Its body said *"The ssh half holds"* — that is now qualified with both measured cases. |
| The test header re-records the OpenSSH measurement from `inflight.sh` | **#954**, filed as a record and closed `wontfix`. Refuted by an adversarial check: the proposed pointer targets prose no test pins — deleting `inflight.sh`'s measurement paragraph leaves the suite `1237/1237` — and the negative pin structurally forbids probe 2 from hosting the other half of the pairing. |

### Not filed, with reason

- The `BatchMode=no` / host-key claims: a specialist could not execute them (macOS sshd refused to start in its sandbox) and reported nothing contradicting them. Already measured and recorded by the implementer, including the no-terminal qualifier. Sandbox limitation, not a tree defect.
- A `types` specialist noting its own dimension does not apply to a repo with no static type system.

PR #957, opened 2026-08-28
fix(agent-test): refuse an argv of flags alone instead of node's own discovery

### LINKED ISSUE #352: agent-test: a bare invocation carrying only flags still falls through to node's own discovery
Deferred from PR #333 review.

**Dimension:** correctness · **Location:** `skills/fleet/scripts/claim-ticket.sh:160`

## The finding

`[ $# -gt 0 ] || set -- .` — the #97 fix — guards a *literally empty* argv only. An invocation carrying flags but no path operand still has `$# -gt 0`, so the `for arg do` loop finds no directory to expand, argv reaches `exec node --test "$@"` holding only flags, and node falls back to its own default discovery. That produces exactly the vacuous green #97 exists to refuse.

Measured against the PR head with the fix present:

```
apply({ "t/a.spec.mjs": PASSES }).run("--test-concurrency=1")
  -> status=0   ℹ tests 0   ℹ pass 0

control, same fixture, bare:
apply({ "t/a.spec.mjs": PASSES }).run()
  -> status=0   ℹ tests 1   ℹ pass 1
```

Same fixture PR #333's own `.spec.` test uses.

## Why deferred

Outside #97's literal scope — the ticket says *bare*, and `run-team/SKILL.md:168` briefs members with `./agent-test <file>`. Pre-existing: the flag-only form behaved this way before #97 too.

## Why the approach is open

The obvious fix is to default when no *non-flag operand* is present rather than when argv is empty. But prepending `.` ahead of node's own flags reorders argv, which needs its own measurement before shipping. Whether a flag-only invocation should be defaulted at all — versus refused outright, the way a zero-match expansion already is — is itself undecided.

feigi:
Closes #352

## The defect, re-derived live

The bare-form default guards a **literally empty** argv, but a flag counts toward `$#`. `agent-test --test-concurrency=1` therefore cleared the default, gave the expansion loop no directory to expand, and reached `exec node --test` holding only flags — node's own discovery, which does not recognise the `.spec.` form this runner exists to cover.

Measured on the issue's own fixture (`t/a.spec.mjs`), against this branch's parent:

```
agent-test --test-concurrency=1   → status=0   ℹ tests 0   ℹ pass 0
agent-test  (bare, control)       → status=0   ℹ tests 1   ℹ pass 1
```

A zero-test run at exit 0 is the vacuous pass this runner exists to refuse.

## The ruling honoured: refuse, not default

Per the Agent Brief. Prepending the default ahead of node's own flags reorders argv and would need its own measurement; the runner already refuses rather than guesses where a path that does not exist or an underivable test command is what it is handed; and no briefed workflow passes flags alone.

**One shape stays exempt, and this PR does not change that.** A quoted glob that matches nothing still exits 0 having run nothing — `agent-test 'nope-*.spec.mjs'` measures status 0 with a zero-test summary and an empty stderr on the parent and on this head alike. The shell is not expanding that argument (`set -f`), node is, and the runner never learns the match count because `exec` has no return path. That is the escape hatch the file branch already documents as its ceiling; closing it would mean running node instead of `exec`ing it so the summary could be read back, which is a separate change with its own behavioural cost. The Agent Brief lists a zero-match expansion among the shapes this runner refuses — measurement does not support that, and the second commit here strikes the claim from the tree rather than restating it.

## The change

The emitted runner records whether any argument named something to run, and refuses when none did. The classification is judged on the argument's shape rather than on what that shape resolves to — a directory expands, an existing file passes through escaped, a quoted glob is node's to expand whether or not it matches anything, and a typo is refused outright — so it is one `case` beside the loop's own dispatch, and the refusal sits between the loop and the `exec`. A flag is the one shape that can name nothing at all.

The `-*)` pass-through is untouched. Only node can judge its own flags.

## Measured, on an emitted runner

| invocation | before | after |
|---|---|---|
| `--test-concurrency=1` | status 0, `tests 0` | status 1, names the missing operand |
| `-v` | status 0, prints `v26.7.0`, runs nothing | status 1, same refusal |
| `--` alone | status 0, `tests 0` | status 1, same refusal |
| `--test-only --test-reporter=tap` | status 0, `tests 0` | status 1, same refusal |
| bare | status 0, `pass 2` | unchanged |
| `--test-concurrency=1 t` | status 0, `pass 1` | unchanged |
| `'t/*.spec.mjs'` (quoted glob) | status 0, `pass 1` | unchanged |
| `'nope-*.spec.mjs'` (zero-match glob) | status 0, `tests 0` | unchanged — the documented exemption |
| `-- t` | status 0, `pass 1` | unchanged |
| `nope.test.mjs` | `does not exist`, status 1 | unchanged |
| `--test-nmae=1 t` | node's own `bad option`, status 9 | unchanged |

Re-measured under `dash` as well as macOS `sh`, since CI runs the emitted runner under the former.

## The `[ -e "$arg" ]` term, and why it stays

The classification asks existence inside the `-*` arm, so a dash-spelled path that exists is read as an operand. Measured on an emitted runner with a file named `-dash.spec.mjs` present and named as the sole argument:

```
with the term      → status 9   node: bad option: -dash.spec.mjs
without the term   → status 1   agent-test: no test file or directory in the arguments …
```

Both refuse loudly, so there is no functional bug either way and the term stays — dropping it would change the exit status for that input, and this ticket's scope is a refusal, not a re-spelling of one. But the two answers **are** observably different, so the term settles which refusal a caller sees. The first commit's comment claimed no measured input separated them; the second commit corrects that.

## Pinning

One test drives the flag-only form against the issue's own fixture and asserts the refusal, and rides the same flag alongside a real operand as the accept-direction control. Mutation-tested against the whole `skills/fleet/scripts` suite:

- refusal deleted (pre-fix behaviour restored) → the new test alone goes red
- classification always marks an operand → the new test alone goes red
- refusal predicate inverted (over-broad) → every test that hands the runner a real operand goes red, the new test's accept half among them
- classification pattern narrowed from `-*` to `--*` → the new test alone goes red, on the `-v` entry, with `v26.7.0` as the assertion's actual value
- sentinel value renamed (`operand=1` → `operand=yes`), predicate unchanged → stays green
- refusal message tail reworded, exit status unchanged → stays green

The `-v` entry is what buys the fourth of those. A corpus spelled entirely with double dashes cannot discriminate `-*` from `--*`: under that one-token mutation, with `-v` removed from the corpus, `claim-ticket.test.mjs` stays entirely green.

Full suite green.


feigi:
Review deferrals filed:

- **#960** — the emitted runner's two wrong-cause refusals: `find`'s own option parsing reported as unreadability for a dash-named directory, and an unsearchable parent reported as a typo while the directory arm gets the same input right. Both pre-existing, both unchanged by this diff, both re-measured against `bf41072`.
- **#961** — leading-dash + `[ -e ]` derived twice, by the new `operand` classification and by the loop's own dispatch, kept in agreement by a comment rather than by shared code or a failing test. Deferred deliberately: computing it once restructures the branch dispatch this ticket exists to pin. The suggested shape is recorded there.
- **#965** — record of the two in-scope `suggestion`-band findings that were checked and **refuted**, with the measurements: adding a `-dash.spec.mjs` pin test (declined — both answers refuse loudly, only the message differs) and deleting the ACCEPT tail (declined — it catches a mutant the pre-existing test misses, and its `pass 1` count assertion is the sole discriminator).

Applied in `bf41072` instead: the two false claims this diff had introduced, the ordering claim in the same comment block, and a single-dash entry in the argv corpus. The PR body has been reconciled against the final diff — it had restated the zero-match-expansion claim twice.

PR #958, opened 2026-08-28
fix(board): give the spend panel's tri-state routing a test seam, regression case first

### LINKED ISSUE #371: board.html's renderSpend has zero test coverage, and the tri-state spend contract it duck-types is prose only
`dimension: types` · `skills/fleet/scripts/board.mjs:205`, `skills/fleet/scripts/board.html:94-113`

Deferred from PR #361 review. Deferred on **scope**: #153 is scoped to `ledger.mjs check` term selection, and the fix here is a cross-file contract reshape plus new test infrastructure — too much to ride in on a ticket that has no business in it. The finding itself is not disputed.

## The finding

PR #361 introduces a tri-state `spend` contract:

- `null` — nothing yet, normal at run start; the UI hides the panel
- `{ error }` — a bug the operator must act on; the UI shows it
- a data object — success

That contract is expressed **only as prose**, duplicated across three independent, unsynchronised sites: `board.mjs`'s `gatherSpend`/`findSubagentsDir` comments, `board.html`'s `renderSpend`, and the docs/specs markdown. The discriminant is presence-of-an-error-key duck-typing (`dir && dir.error` at `board.mjs:208`, `sp.error` at `board.html:101`) rather than a tag — unlike the `tracker = { ok: true|false, ... }` shape that the *same* PR leaves in `ledger.mjs`, which discriminates on an explicit boolean checked before any success-only field is read.

The success object `gatherSpend` returns (`totals, roles, top, reviewPct, tools, attributedPct, skipped, since`) never contains a literal `error` key **today**, so the duck-type is safe by coincidence of the current field list, not by any enforced guarantee.

## Why this is worth a ticket rather than a shrug

`board.html` has **no test file at all**. `board.test.mjs:46` writes a placeholder `board.html` stub for a server test and never invokes `renderSpend`. CI's only static check on `.mjs` is `node --check` (syntax only) — no eslint, no tsc, no checkJs.

That gap is not hypothetical. PR #361's review found a real defect in exactly that untested half — `renderSpend` returned early on `!sp.totals.cacheWrite` and so hid a non-zero `skipped` count, rendering "3 of 4 transcripts unreadable" identically to "nothing yet". **The fix for it shipped in `234e668` untested**, because there is nothing to test it with.

## What is actually open

The approach, which is why this is `needs-triage` and not `ready-for-agent`:

1. **Reshape to `{ ok: boolean }`** as the original finding suggests, matching `tracker` — touches `board.mjs`, `board.html` and the spec doc together.
2. **Or keep the duck-typing** and instead extract `renderSpend`'s branch logic as a small pure function that a test can call without a DOM — smaller, and it closes the coverage gap that actually bit.

(2) is the cheaper one and addresses the measured harm; (1) addresses the stated inconsistency. They are not exclusive. Note this repo has no `package.json` and no dependencies, so anything needing jsdom is a bigger decision than it looks.

Deferred from PR #361 review.

feigi:
`board.html`'s `renderSpend` had no test coverage of any kind. The suite writes a
placeholder `board.html` stub for a server test and never reaches the renderer,
and CI's only static check on `.mjs` is `node --check`. That gap already cost
something: `renderSpend` returned early on a falsy `cacheWrite` and swallowed a
non-zero `skipped` count with it, so a run whose transcripts were unreadable
rendered identically to a run that had not started — and the fix for it shipped
untested, because there was nothing to call.

Per the ticket's ruling this takes option 2 — extract the branch logic — and
leaves the `{ ok: boolean }` contract reshape alone.

## What changed

`spendView(sp)` returns the panel's outcome as a value: a `kind` tag plus the
strings that outcome renders. `renderSpend` switches on it and does the DOM.
Pure, DOM-free, no new dependency.

It stays in `board.html`'s inline script rather than moving to a module.
`board.mjs` stages that page with a single `copyFileSync` and `createBoardServer`
routes only `/`, `/board.html` and `/board.json`; a module would need a second
staged file and a fourth route, and would blank the panel whenever staging the
second file failed. The test lifts the function out of the source instead — the
technique `select-dimensions.test.mjs` already uses against `workflows/review-pr.js`.

## Rendering is unchanged

Verified rather than asserted: both versions of `renderSpend` were rendered
against a stub DOM across every outcome — hidden, error, the skipped-only note,
the funded panel, the tool cap, an absent `attributedPct` — and the resulting
trees compared. All identical. The harness was checked against a known-different
pair first, so the agreement is a measurement rather than an artifact of a
comparison that cannot fail.

The two text-only outcomes already rendered through one identical `spend-wrap`
div, so they now share one `append`.

## What the tests pin

Every outcome of the contract, plus the sub-cases inside "success" that can
render wrongly. Leading the file is the case this ticket exists for: a success
object with a falsy `cacheWrite` and a non-zero `skipped` must report the skip,
not hide. Both directions are covered — inputs the panel must SHOW as well as
inputs it must hide, so neither direction can drift unnoticed.

Asserting on what `spendView` RETURNS is only half the claim, though, and review
found the other half missing: nothing pinned how `renderSpend` *consumes* the
value it gets back. Single-token breaks in the renderer left every
return-value case green — delete `if (v.kind === "hidden") return;` and a hidden
decision falls into the `!== "panel"` branch and appends an empty `spend-wrap`
every tick at run start, which is a box where the panel must render nothing and
is this ticket's own behaviour reintroduced; or swap the two column lists, or the
two header strings, and nothing goes red either. Each field of the decision now
has a pin saying which slot it lands in.

Because a lift reads a copy, further guards tie it back to the page: a
declared-exactly-once check (a duplicate declaration placed after the real one
wins at runtime by hoisting and is otherwise invisible to a lift), a call-site
pin (the decision can otherwise be disconnected in one token with every case
still green), a check that the raw argument is not branched on again anywhere in
the renderer, and a parse check on the inline script.

Every regex that matches a declaration is whitespace-tolerant around the name and
the paren, and matches parameter names as `\w+`. Neither a rename nor a reformat
is a defect, and the two must be loosened together: loosening the
declared-exactly-once count without loosening the lift lets a reformatted
declaration pass the count and then die in the lift. The wiring pins match a
whitespace-collapsed body for the same reason — where a line breaks is not
wiring.

Mutation-verified both ways, every pin in the file: each semantic mutation reds
exactly one test and it is the right one, and every zero-behaviour control stays
green. The controls exercise a reworded comment, a renamed local, a renamed
parameter in either function, an unrelated CSS change, a reformatted lambda,
reformatted declarations, and a long call wrapped across lines at each of the
places a formatter would break it. Two of the reformat controls were red before
the pins were loosened and are green now, so the loosening removed real false
alarms rather than only widening what the pins accept.

## Docs

`docs/specs/2026-07-24-fleet-cockpit-design.md` asserted the page is "not
unit-tested". True before this branch and false after it, so it now says what is
true: the spend panel's branch decision is extracted and unit-tested, and the DOM
rendering around it is not — which is what the wiring pins above are about.

## Deliberately left

`{ error: "" }` is routed to the success path and hidden, because the
discriminant is truthiness while the contract is presence-of-a-key. That
divergence is what the test measures and pins. Whether any producer actually
reaches it is a separate question and is not settled here: `gatherSpend`'s outer
catch returns `{ error: e.message }` and `e.message` is empty for any error
thrown without one, so the contract admits the shape, but no live message-less
throw into that catch has been established. Fixing the routing would change
rendering, which this ticket rules out, so the current routing is pinned instead
and the case is left for #959.

## Deferred from review

Filed rather than folded in: #962 (four regexes pin the same two declarations; a
test-consolidation refactor, not a defect), #963 (collapsing the `error`/`note`
tags — refuted by measurement, filed as the record), #964 (the parse check and
the lift both read only the first `<script>` block, so a second block parses
green while the lift takes a decoy — confirmed and measured).

Closes #371


feigi:
Review actions applied. Pushed `74e514d`.

**Applied**

- **`renderSpend`'s consumption of the view was unpinned** (survived, 2-0; and the `silent-failure` finding that nothing pins a non-panel decision reaching the DOM, which is the same defect from the other side). Seven wiring pins over the renderer's body, one per field of the decision. Confirmed here that all three demonstrated breaks were green before: deleting `if (v.kind === "hidden") return;`, swapping the two `spendRows` lists, and swapping `lede` for `note`.
- **The duplicate-declaration guard was whitespace-evadable** (survived, 2-0). `/^function spendView\(/gm` missed `function spendView (sp)` — one space, still valid JS, still binds the name, and a duplicate placed after the real declaration wins by hoisting. Now `\s`-tolerant.
- **The re-derivation guard was too narrow** (survived, 2-0; same finding as the `simplify` dimension's). It required the raw argument to be followed by `.` or `)`, so `if (sp && sp.error)`, `if (sp?.error)` and `const { error } = sp` all walked through. It now asserts the property directly: the raw argument reaches exactly one place, the `spendView` call.
- **A bad commit citation, removed.** The regression case cited `234e668` as where the fix shipped. That is not a valid object in this repository — `git cat-file -t 234e668` gives `fatal: Not a valid object name 234e668`, exit 128. It came from #371's own body and restating it minted the citation into the tree. The historical point stands without it.
- **The empty-message case asserted reachability**, which is not established; it now asserts the contract/code divergence, which is what it measures. See #959.
- **`docs/specs/2026-07-24-fleet-cockpit-design.md`** asserted the page is "not unit-tested" — true before this PR, false after it. One line, saying what is now true and that the DOM rendering still is not.

**Found while verifying, and fixed in the same push**

Mutation-testing the pins against zero-behaviour controls turned up three false reds, all one class — a pin anchored on literal whitespace. Loosening the count guard without the lift left a pure reformat passing the count and then killing the lift (11 red). The re-derivation guard stripped the signature with a second literal-spaced regex, so a reformatted signature survived the strip and the argument was "found" in it. The wiring pins required their line unbroken. All three fixed; the two reformat controls are green now and were red before.

**Refuted, not applied** — each dispatched to its own refuter, biased to refuse, settled by running the mutation rather than by reasoning:

- *The `k` row of the duplicate guard is dead weight, since a duplicate `const k` is a SyntaxError the parse check already catches.* Refuted. A `const k` in a **second `<script>` block** reds the `k` row and nothing else — the parse check stays green, because its regex is non-global and reads only the first block. Reproduced on the post-fix tree. Dropping the row would strictly reduce coverage. The refutation exposed a real defect, filed as **#964**.
- *The `kind` discriminant carries four values but the consumer distinguishes three.* Refuted. The premise that `renderSpend` is the only consumer is false — `spend-view.test.mjs` asserts the tags as data, and collapsing them reds two tests with tag-only diffs. Filed as **#963**.

**Deferred and filed**

- **#962** — four regexes across two tests pin the same two declarations. A test-consolidation refactor, not a defect, and it would have entangled a cleanup with two fixes whose mutation evidence needed to stay readable.
- **#963** — the refuted `kind`-collapse suggestion, filed as the record.
- **#964** — the parse check and the lift both read only the first `<script>` block. Confirmed defect, remedy open.

**No action** — the review could not confirm from its inputs that #371 rules the `{ ok }` reshape out of scope; it does, in both the Brief and the ruling, so that comment is accurate and unchanged.

**Tests**: `node --test skills/fleet/scripts/*.test.mjs` from the worktree — `tests 1263`, `pass 1263`, `fail 0`. Baseline before this work was 1263 minus the seven new wiring pins. Mutation matrix run separately against both this tree and `65bbd1f`, on copies under scratch: every semantic mutant reds exactly one pin and the right one, every zero-behaviour control green.

PR #969, opened 2026-08-28
docs(fleet): make the finisher's member name a written convention

### LINKED ISSUE #326: run-team/SKILL.md:34-35 — the finisher has no documented name, but compute-spend now classifies on one
**dimension:** correctness
**Deferred from PR #321 review** (which closes #318).

## The gap

PR #321 taught `classifyRole` to recognise finishers by their agentType:

```js
// skills/fleet/scripts/compute-spend.mjs:46
if (/^finish-|finish pr|finisher/.test(hay)) return "finisher";
```

That `^finish-` keys on a naming convention that is **stated nowhere**:

- `skills/fleet/skills/run-team/SKILL.md:34-35` enumerates the member names — `impl-<issue#>`, `fix-pr-<pr#>`, `review-pr-<pr#>`, `merge-bot-<wave#>` — and **omits the finisher entirely**.
- `compute-spend.mjs:16`'s own comment lists the names that are "controller-authored and stable because run-team fixes them", with the same four and no finisher.

So the classifier depends on a convention the controller is never told to write.

## Why this is `needs-triage` and not a one-word doc fix

**At least three spellings are live simultaneously**, which is why the fix is a decision rather than a gap-fill:

- `finish-<pr#>` — e.g. `finish-315`, `finish-316`, `finish-319`, `finish-320`
- `finisher-pr-<pr#>` — e.g. `finisher-pr-224` … `finisher-pr-293`
- `finisher-*` is recorded as the convention in at least one project memory

Classification itself is currently fine — `^finish-` catches the first, the unanchored `finisher` alternative catches the other two. The open question is **which spelling is canonical**. Writing one into `SKILL.md:34-35` standardises on it, which in turn decides whether the other alternatives in `:46` are permanent compatibility or drift to be removed. That is a convention call for the maintainer, not something to guess.

## Suggested remedy (pending that call)

Once the canonical finisher name is decided, add it to both enumerations so the convention the classifier reads is the convention the controller is told to write:

- `skills/fleet/skills/run-team/SKILL.md:34-35` (the member naming list)
- `skills/fleet/scripts/compute-spend.mjs:16` (the "run-team fixes them" comment)

## Why deferred rather than applied in #321

`compute-spend.mjs:16` is that PR's own file, but `run-team/SKILL.md` is **not in its diff**, and the two edits are coupled: adding `finish-<pr#>` to the `:16` list alone would make that comment's own claim — "stable because run-team fixes them" — false, since run-team does not fix it. PR #321 instead documents the dependency at `:16` honestly and points here. Related: #318.

*Deferred from PR #321 review*

feigi:
Closes #326

`classifyRole` recognises finishers by name, but no naming list mentioned a finisher — so the classifier depended on a convention the controller was never told to write.

## The ruling, verified before standardising

Canonical name is **`finisher-pr-<pr#>`**: the house shape `<role>-<subject#>`, with `-pr-` where the subject is a PR.

The brief required this be checked against what the controller actually dispatches. `docs/metrics/member-outcomes.tsv` is derived from the harness's own subagent transcripts, so it records dispatched names. Grouped by run date, `finisher-pr-<n>` is the predominant spelling overall, appears on more distinct run dates than any other, and is the spelling of the most recent recorded run. `member-outcomes.mjs`'s own dated measurement across every `meta.json` ranks it first independently. The ruling holds; the check is recorded on the ticket.

## What changed

- **Both naming lists** name the finisher — `run-team/SKILL.md` and `references/member-lifecycle.md`, which carries the same list and is where `SKILL.md`'s list points. Fixing only the first leaves a reader who follows the pointer on the incomplete one.
- **`compute-spend.mjs`'s comment** adds `finisher-pr-<n>` to the names it calls *"stable because run-team fixes them"* — true only because the naming-list edit lands with it, which is the coupling PR #321 deferred rather than half-do. The surviving alternatives are labelled deliberate compatibility, with the consequence of removing one, so a later reader does not clear them as drift. Its two `:46` line references are gone; this edit would have made them wrong.
- **`member-outcomes.mjs`** described #326 as still open; that clause now states the canonical name.

## A fourth spelling the ticket does not name

The ticket enumerates three. The record contains **`finish-pr-<pr#>`** as well, across several run dates. `classifyRole` already classified all four correctly, so this was a documentation gap, not a live misclassification — but AC-3's "other live spellings" is a set of three, and all four are now named. No matching logic changed (per Out of scope).

## Tests

The suite could not see either naming list: adding the finisher to both moved no test. Three new pins in `finisher-name-prose.test.mjs`, each mutation-verified to red alone and to survive a reflow of the paragraph it reads:

- both naming lists name every member, compared as a **set** so a reorder stays green
- the names the classifier comment cites are **derived from that comment** and checked against run-team's list — so the half-fix (name added to the comment alone) reds
- `classifyRole` still **accepts** all four dispatched spellings plus a retry suffix, the accept side of the documented compatibility

Full suite green, 1266 passing, 0 failing.

## Review applied (846ba78)

The review found the first pin's slice ended at `doc.indexOf(".", …)` — the first period anywhere after the lead-in, not a sentence boundary — so an abbreviation, or the `references/member-lifecycle.md` pointer the same paragraph already carries, truncated the slice, red the pin, and reported every name as removed while none was. It now slices the markdown **paragraph**, through `between` from `./prose-pin.mjs` — the shared two-ended slicer this directory already uses — rather than a local reimplementation. The docs are read unflattened, since `flat()` eats the blank line the slice ends on.

Four smaller changes landed with it:

- the second pin's per-role loop matched the whole flattened `SKILL.md`, not the naming list its own comment claims it checks. Narrowed. This is a tightening, not a false-green fix — the first pin already catches a name dropped from a list it restates. What the narrowed haystack newly catches is a name cited in `compute-spend.mjs`'s comment that reaches `SKILL.md`'s narrative but never its naming list: the half-fix #326 was deferred to avoid, and the case the first pin is structurally blind to.
- the retry case `finisher-pr-958-b` folded into the loop above it, gaining the per-type failure message. It is not redundant — it is the only one of the five that reds when the finisher match is anchored to a trailing number.
- this file's header comment cited `review-path-default.test.mjs` "at the site that dispatches it" for the fix-applier pin. Measured wrong: deleting the name from the naming paragraph reds `the member-naming rule still names the fix-applier` (:516) and leaves the dispatch-site pin (:89) green. Cited by test name now.
- **`compute-spend.test.mjs`** — a fifth file, not in the list above — stated the pre-#326 convention as present-tense fact ("run-team dispatches finishers as agentType `finish-<n>` … the controller's actual naming convention"), contradicting this PR's own module comment ten lines away. `finish-<n>` is 18 of 289 recorded dispatches and unused since 2026-08-11. Reworded to what the assertion pins: a historical spelling that must stay classifiable. No assertion moved; 21/21 before and after.

Every changed pin was mutation-verified both ways after the fix: the two probes that produced the false red now stay green, a reorder-plus-reflow of the naming paragraph stays green, and real deletions from either list still red alone.

Deferred and filed: #974, #975, #976, #977 (`needs-triage`). Refutations recorded closed in #978.


feigi:
Review applied and pushed as `846ba78`. `node --test skills/fleet/scripts/*.test.mjs` → **tests 1266, pass 1266, fail 0**, exit 0.

**Applied (5):** the naming pin's slice moved from the first period to the markdown paragraph, through `between` from `./prose-pin.mjs`; the second pin's haystack narrowed from the whole file to the naming list; the `finisher-pr-958-b` retry case folded into the loop; the header comment's citation of the fix-applier pin corrected; `compute-spend.test.mjs`'s present-tense statement of the pre-#326 convention reworded.

Two review findings asked for `between(doc, "Names follow", ".", label)`. Their literal end anchor is the defect the `important` finding had just proved, so the paragraph anchor was used instead and their point — one implementation of "slice between two anchors, fail loudly if either moved", not two — is served by the same edit.

**Deferred and filed (4, `needs-triage`):** #974 (plan-1-scraper's stale quoted list), #975 (member-outcomes.mjs's dated spelling counts vs the as-of anchor), #976 (no shared spelling list across two docs and two modules), #977 (review-path-default.test.mjs:516 possibly subsumed — verify by mutation before deleting coverage).

**Refuted by measurement, recorded closed in #978 (`wontfix`):** the classifyRole reviewer/finisher precedence claim (289 recorded finisher dispatches, zero misbooked), the "dead branch" claim on the parse-shape guard (deleting it swaps a true diagnostic for a false one, and removing both makes the loop pass vacuously), the "retry suffix adds no discriminating power" claim (it is the only one of the five that catches a number-anchored match), and the two findings the review itself returned refuted.

PR #970, opened 2026-08-28
docs(agents): issue-tracker carries the issue-read block whole, and a pin holds it there

### LINKED ISSUE #374: docs/agents/issue-tracker.md still lacks the read-before-touching-code sentence after #79 restores the other two
Surfaced by `impl-79` while implementing #79, and re-verified independently by the controller against `origin/main` before filing.

## What

#79 is about `docs/agents/issue-tracker.md` reproducing the implementer issue-read block from `run-team/SKILL.md` and dropping the `## Agent Brief` and `Respec` sentences. PR #373 restores those two. **A third sentence is still missing**, and it is not the one #79 names.

The block grew since #79 was filed. #79's body describes a **four**-sentence block at `:149-154`; on `origin/main` it is **five** sentences at `:238-245`:

```
$ git show "origin/main:skills/fleet/skills/run-team/SKILL.md" | sed -n '238,245p'
> Read the issue with `gh issue view <N> --json title,body,comments --jq '...'`.
> Not bare `gh issue view <N> --comments` — non-interactively that prints only
> the comments, and nothing at all when there are none, dropping the title and
> body either way, exit 0, so the loss is silent. The `## Agent Brief` comment
> is authoritative over the issue body. Honor its `Respec` block — it may
> explicitly rule out hypotheses the body raises. Read the issue **before
> touching code**: with the repo in front of you, still undecided or needing
> human hands you do not have → bail, name the cause, do not implement.
```

The tracker doc's copy ends at `the loss is silent.`:

```
$ git show "origin/main:docs/agents/issue-tracker.md" | sed -n '34,37p'
Read the issue with `gh issue view <N> --json title,body,comments --jq '...'`.
Not bare `gh issue view <N> --comments` — non-interactively that prints only the
comments, and nothing at all when there are none, dropping the title and body
either way, exit 0, so the loss is silent.
```

So after #373 lands, the copy carries sentences 1–4 and still not sentence 5.

## Why it was not folded into #373

Deliberately out of scope, and correctly so. The fifth sentence states a *different* rule from the two #79 names — bail before implementing when the ticket is undecided or needs human hands — and correction tickets are capped at the ticket's stated size precisely because added prose is where new wrong claims enter. Widening would have made #373 a bigger diff than the ticket authorised.

## Open question (why `needs-triage`)

Whether the tracker doc's section should be a **complete** copy of the block or deliberately a partial one. That question is upstream of this fix and is not answered anywhere:

- If complete, this is a one-sentence restore and the section needs something that keeps it complete as the block grows again — it has now drifted twice.
- If partial, the boundary needs stating, because "carries the command and its caveat, omits the judgement rules" is currently accidental rather than chosen. The same accident is what #79 exists to fix.

Related: #23 owns which `--comments` caveat wording the three *skill* files should share, and is sequenced after #79 for the same reason this is — #373 changes what the authoritative text is.

## Scope searched

`gh issue list --state all --search "issue-tracker.md block sentence"` and a phrase search for `"before touching code"` — no existing issue covers this. `ledger.mjs check` returned `verdict: clean` with the tracker reached.

feigi:
`docs/agents/issue-tracker.md` reproduces the implementer issue-read block from `run-team/SKILL.md`. The copy stopped at the `Respec` sentence; the block goes on to say to read the issue before touching code and to bail naming the cause when the ticket is undecided or needs human hands the reader does not have.

Re-derived against `origin/main` before implementing. The ticket's line numbers had drifted — the block is at `skills/fleet/skills/run-team/SKILL.md:611-618`, the copy at `docs/agents/issue-tracker.md:34-39`, not the `:238-245` / `:34-37` the body cites. `grep -rn "before touching code" --include="*.md"` matched only a spec doc, confirming the sentence had not landed. #23 is CLOSED with `caveat-unification-prose.test.mjs` on `main`, and #373 is MERGED, so the sequencing precondition the body raises is discharged and the authoritative text is settled.

## What changed

- The copy carries the block whole. The added text is verbatim from the source — this PR mints no prose of its own in the doc.
- `skills/fleet/scripts/tracker-block-copy-prose.test.mjs` compares the two paragraphs **whole**, so the copy is pinned complete rather than pinned to contain some sentence. A presence pin on the restored sentence would leave the next one added upstream to drift exactly as it did.

The far end of each slice is anchored past the rule, never on the block's own last sentence. A closing-phrase anchor stops the slice at today's last sentence, so a sentence appended after it falls outside both slices and the pin passes with the copy already behind.

This PR's first draft used the paragraph break as that anchor, and review measured it to be the same mistake one size up: a whole new `>` paragraph appended beside the block in `run-team/SKILL.md` alone fell outside both slices, and all three tests — and the whole 160-test prose suite — stayed green while the tracker was genuinely behind. So each slice now runs from the opener to the first paragraph that is **not** this rule: the next heading in the tracker, where the block owns a `##` section, and the rule that follows the block in run-team's prompt, where it is one blockquote among siblings. Growth of any size, sentence or paragraph, now lands inside the slice. Documenting the hole as a ceiling was the alternative and is the weaker answer — the file already refuses that trade for the sentence-sized version of the same hole.

Both slices are unquoted and whitespace-normalized before the opener is matched, because the two files hard-wrap the same prose at different widths today.

## Mutation matrix (measured, not asserted)

Each mutation applied to a committed tree, run, then restored with `cp`; `git status --porcelain` verified empty after every restore.

| # | Mutation | Intended | Measured |
|---|---|---|---|
| 1 | Drop a sentence from the copy | red | **red** |
| 2 | Append a new sentence upstream in the source | red | **red** |
| 3 | Reword one side only | red | **red** |
| 4 | Same reword applied to both sides | green | **green** |
| 5 | Pure rewrap of the copy, no text change | green | **green** |
| 6 | Upstream growth arriving as a **new sibling paragraph** | red | **red** |
| 7 | Extractor reverted to the paragraph-break slice | only test 4 red | **only test 4 red** |

(2) is the one that matters: it is the drift this section has produced twice, and it is what a presence pin cannot see. (6) is the case review found the first draft missing — it was **green** under the paragraph-break slice — and (7) is its mutant-kill: reverting the extractor reddens the sibling-paragraph test and nothing else, so that test pins the anchor rather than restating (2).

(3) reds **by design**, and the test's header comment says so where the next person to trip it will read it: one wording, two files, so the file you did not edit is the copy that just went stale. The fix is to mirror the edit, never to loosen the comparison. (4) and (5) are what keep that from meaning "reds on everything" — a rewrap is not an edit, and (5) is pinned by a live accept-case test rather than left to this table.

## Refusal surface

The extractor asserts **exactly one** paragraph opens with the block's clause, so a second copy is refused loudly rather than silently compared. A positive control asserts the slice is longer than its opener, so a stale anchor cannot make both slices empty and pass the equality vacuously. The end anchors are guarded the same way: a paragraph matching each must exist after the block, so an anchor that goes stale reddens naming itself rather than silently widening the slice to EOF.

## Class enumeration

Covered: the restored sentence; future upstream growth, both inside the block's paragraph and as a new paragraph beside it (measured, mutations 2 and 6); and the tracker's `--comments` caveat, which whole-paragraph equality now pins as a side effect — `caveat-unification-prose.test.mjs` had deliberately left it unpinned, and compares a disjoint pair (`next-ticket` / `sizing-a-ticket`), so nothing collides.

Deliberately left: `issue-tracker-prose.test.mjs`'s presence pins from #79 stay, because they pin that the *rule is stated* independent of the source's wording — if the source ever legitimately drops one, this pin goes green and that one stays red. Different property, both worth having. `docs/agents/domain.md` and `docs/agents/triage-labels.md` carry no copy of this block (grepped for the command, `Agent Brief`, and `Respec` — no matches), so the sibling class is empty. The `next-ticket` / `sizing-a-ticket` paraphrases are #23's, closed and already pinned, and paraphrase makes whole-paragraph equality the wrong instrument there.

## Acceptance criteria

- [x] The copy carries the block whole.
- [x] A pin reds when the copy falls behind (mutations 1, 2 and 6).
- [x] Both directions demonstrated (table above), with the design intent for a one-sided reword stated rather than left implicit.
- [x] **The authoritative block is unchanged** — `git diff origin/main --stat` touches two files, and `run-team/SKILL.md` is not one of them. The mutation runs that edited it temporarily were restored with `cp` and the tree verified clean.

## Tests

`node --test skills/fleet/scripts/*.test.mjs` — **tests 1267, pass 1267, fail 0, cancelled 0**. The four new tests appear in the glob expansion, so the suite exercises the mode this PR adds.

Closes #374


feigi:
Review deferrals filed: **#979** (`needs-triage`, open — the issue-read section's implementer-only "do not implement" rule, out of scope for #374 because it would edit the authoritative block) and **#980** (`wontfix`, closed — the suggestion band, checked: five entries, all measured and none a defect).

Applied on this PR: the critical `tests` finding and the `correctness` suggestion restating it, one defect fixed once in `6a44a9b` — the slice now ends past the rule instead of at the paragraph break, with the sibling-paragraph case added as a fourth test.

Issue #347, opened 2026-08-11
Every unattended git network call in the fleet outside inflight.sh probe 2 still carries #92's prompt/hang hazard


feigi:
`skills/fleet/scripts/release-ticket.sh:306` was the specialist's citation — **re-derive it at head, PR #330 just moved line numbers in that file.** The finding is not about that one line; it is about the class.

## The finding

Every sibling unattended network call in the fleet still carries the exact hazard #92 named for `inflight.sh` — none sets `GIT_TERMINAL_PROMPT`, `BatchMode`, or any transport bound. PR #332 fixed `inflight.sh`'s probe 2 alone. The other callers — `release-ticket.sh`'s push/fetch paths and every other `git` or `gh` call the fleet makes without a human present — can still prompt on a credential or host key, or stall on a transport that connects and then goes quiet, holding a fleet slot indefinitely.

## Why this is not a #92 follow-on but its own ticket

- It generalises #92's single-call fix to a fleet-wide policy. That is a materially bigger change than #92 asked for, touching several scripts.
- `release-ticket.sh` merged via PR #330 during this same fleet run; editing it from PR #332's branch would have put two live PRs in conflict over the same lines.
- The right shape is a decision, not a patch: a shared helper that every unattended network call routes through, versus repeating the knobs at each call site, versus a watchdog wrapper. That interacts directly with #346 (the https/connect-stall bound `inflight.sh` still lacks) — if #346 lands a watchdog, this ticket's answer is "route everything through it" rather than "copy four `-o` flags around". **Sequence this after #346, or decide both together.**

## Starting point for whoever takes it

1. Enumerate the unattended network calls: `git ls-files '*.sh' | xargs grep -nE 'git (ls-remote|push|fetch|clone)|gh (issue|pr|api)'`.
2. For each, record whether it can prompt (no `GIT_TERMINAL_PROMPT=0` / no `BatchMode=yes`) and whether it is bounded (no `ConnectTimeout`, no `lowSpeed*`, no watchdog).
3. Note that `gh` is separately bounded — measured during the PR #332 review, `gh issue view` against a silent listener failed at 10.0s with `net/http: TLS handshake timeout`, so the `gh` calls need no knob. The gap is the raw `git` transport calls.

`dimension`: `correctness`

Deferred from PR #332 review (closes #92). Controller ruling: out of scope for #92.

feigi:
> *This was generated by AI during triage.*

**Moving to `ready-for-agent`.**

[...]

PR #972, opened 2026-08-28
fix(inflight): bound probe 2's ls-remote outside git with a watchdog

### LINKED ISSUE #346: inflight.sh probe 2: an https origin is still unbounded — http.lowSpeed* only bounds a transfer already under way

feigi:
Closes #346

`http.lowSpeed*` aborts a transfer already under way and nothing before one, so
an https origin whose TLS handshake never completes was never bounded. Re-derived
against `origin/main` before implementing: with an accept-then-silent listener and
the probe's own knobs set, `inflight.sh` outlived a 30s `spawnSync` cap and had to
be killed from outside — and it outlived the cap because a surviving
`git remote-https` held the caller's stderr, so the caller hung on a script that
had already answered and exited.

git exposes no connect knob to close that: `git help config` lists `lowSpeedLimit`
and `lowSpeedTime` as its whole http timing vocabulary, and `http.connectTimeout`
is not a key git will read.

## What the watchdog does

The fetch is backgrounded and a sleeper kills it at a budget. `wait` returns the
moment the fetch does, so a reachable origin pays ~18ms — measured against this
same script at `origin/main`, not estimated. A polling loop was the other option
and was rejected: at 1s granularity it costs every probe a second.

The unit killed is the process **subtree**. `set -m` would say that in one signal
but is not available where this runs — measured, dash with no controlling
terminal answers `can't access tty; job control turned off` and the group kill
then fails, and dash is `/bin/sh` on the Linux CI runs on. So the subtree is
walked from a `ps` snapshot taken *before* anything is signalled, and the same
walk reaps the sleeper on the path where the fetch returned normally.

A killed fetch gets its own `add_unknown` reason, per the second maintainer
comment: the only failure surface here used to be reached when `ls-remote`
RETURNED, so a stall reached no label at all and was indistinguishable from a
slow link.

The watchdog **says** it fired rather than being inferred from a signal number.
Review found the first cut read "our watchdog fired" off `ls_rc -eq 143`, which
this script does not own: measured, a fetch SIGTERM'd from outside at 4s under
the default budget was reported as `did not finish within 30s`. It now records
the fact in a temp file the sleeper writes before it kills, and the wording reads
that. Probe 2 remains the one probe that still answers when `mktemp` is broken
(#185's contract, pinned) — there the old exit-status test is kept as a fallback,
costing the sharper wording and never the verdict.

`kill_tree`'s documented `ps`-absent ceiling stays a ceiling but stops being
silent. Measured with `ps` shimmed to exit 127: byte-identical stderr to a
healthy run while a `git remote-https` survived and held the caller for its whole
cap. The fallback now records itself and the reason says the process table could
not be read. And the kill escalates to SIGKILL with no pause — a descendant that
traps TERM otherwise holds the fetch's inherited stderr and the caller hangs past
the budget anyway (measured against a `GIT_SSH_COMMAND` wrapper doing
`trap '' TERM`). No pause, because the watchdog subshell is killed through this
same function and would be sitting inside it.

| case | covered | how |
|---|---|---|
| https connect/handshake stall | yes | silent listener, bounded at the budget |
| http transfer stall | yes | `lowSpeed*` still fires first; watchdog backs it |
| connect phase, either scheme | yes | bounded whatever curl's own default is |
| ssh user `ConnectTimeout=0` | yes | silent listener + first-wins `-o` |
| ssh user `BatchMode=no` | by mechanism, **untested** | needs a reachable terminal; the harness has none, and ssh aborts rather than asking there. Recorded in the script's comment. |
| credential helper that never answers | yes | 401 responder + a helper that never returns — served from a **separate process**, see below |
| other `git` network calls repo-wide | **out of scope — #347** | not touched |

That credential-helper row was vacuous until review caught it. `spawnSync` blocks
this process's event loop for the whole child run, so the in-process 401 server
accepted at the kernel level and never served: measured, the handler ran zero
times and `mute-helper.sh` was never executed, leaving a case that only re-tested
"an http origin that never answers" — which `http.lowSpeedTime=10` bounds on its
own, so it stayed green even with `kill_tree`'s subtree walk neutered. The server
now runs in a process that is still scheduled while `spawnSync` holds this one,
and the case reddens on that mutant as it always should have.

`kill_tree`'s awk closure also gained a case at its own level. Collapsing the
`do { … } while (grew)` fixpoint to a single pass survives **every** end-to-end
test, because `ps -A` prints parents before children; only a canned table with
each child ahead of its parent discriminates, and that is what the new case
feeds. It runs in 1.3s against a shimmed `ps`, where the five end-to-end cases
need real stalls.

I could not re-derive the issue's 75s blackholed-SYN figure on this host:
`192.0.2.1` is refused here in ~4s, so the connect-stall is covered by the https
listener instead. The comment states the property rather than a number I could
not settle.

## The other half — what this could wrongly REFUSE

A watchdog that turns a working slow fetch into a false exit 2 is worse than the
hang, since exit 2 is a verdict the fleet acts on. So AC-3 gets a real control: a
transport that genuinely works — serving the fixture's own bare repo through
`git upload-pack` — delayed on the way in, which must come back with the ordinary
`taken` verdict. It is paired with its own sensitivity control (same transport,
budget under the delay), because an accept-only case passes just as well against
a watchdog that never fires.

The test-only budget override can only ever **shorten**. A knob that could lengthen
it would be one more way for configuration to remove the bound — which is the
defect the ssh half of this issue reports.

That property, and the 30s default itself, went in argued-in-prose and unpinned:
review measured that raising the default tenfold **and** flipping the clamp to
`-gt 0` left the whole inflight suite green, because all six override sites set
1, 5 or 20 — every value below the default — so neither branch was ever the
operative path. One case now sets the knob **above** the default against the
silent listener and asserts the reported budget is still 30s; both mutants redden
it, and an equivalent rewrite of the clamp leaves it green.

The overflow input was likewise claimed here and not exercised. It was also
wrong: `*[!0-9]*` catches `5s` and `-1` but passed any all-digit string into `[`,
which printed a raw diagnostic naming a line number — the guard's own comment
("an invalid value is not an error") contradicted on the surface the guard writes
to. A digit-count arm closes it, with a case refusing both shell wordings, since
`sh` on darwin and the dash on CI word it differently and a pin against either
alone is vacuous on the other host.

## What the restructuring itself could break

Backgrounding changes what carries the exit status and what `set -e` covers.
`wait` is captured through an explicit `|| ls_status=$?` for that reason: a bare
`wait` on a killed child aborts the subshell at that line, after the fetch is
dealt with but *before* the sleeper is — and the sleeper would be the leak. The
status is carried out by an explicit `exit` rather than by whatever the block
happens to end with. Every new case asserts `r.error` is unset beside its exit
code, which is the only assertion that can see a descendant holding the pipes.

## AC-6 — the per-transport knobs

**They stay**, as defence in depth, and the reason is written into the comment and
pinned: each fails earlier than the watchdog and in git's own words, and dropping
them would make every ssh stall wait out the full budget where `ConnectTimeout`
ends it far sooner. The pin is mutation-tested in both directions.

The negative half of that pin was scoped too narrowly to do its job. It ran over
the http paragraph alone, while the stale sentence this PR deletes lived in the
**ssh** paragraph above it — measured, reinstating that exact sentence left both
`doesNotMatch` assertions green. The negative pins now span the whole probe-2
comment block; the positive ones keep the narrow slice, where a bound at both
ends is what stops an unrelated block from satisfying them.

## Verification

- `node --test skills/fleet/scripts/*.test.mjs` — **1273/1273 pass**
- `.github/scripts/check-tracked.sh '*.sh' shellcheck -x -S warning` — clean
- `sh -n` and `dash -n` on `inflight.sh` — clean
- three prose-pin mutants, each applied to the real file and reverted, each
  reddening the intended assertion and only that one
- and six more from the review pass, each applied to a scratch copy at a pinned
  ref, never to the checkout: `ls_budget=30`→`300` **red**; the clamp
  `-lt "$ls_budget"`→`-gt 0` **red**; the clamp `-gt 0`→`-ge 1`, an equivalent,
  **green**; the awk fixpoint collapsed to one pass **reds the new canned-table
  case and leaves all three end-to-end watchdog cases green**; the digit-count
  arm dropped **red**; the same arm widened to seven digits, still catching a
  twenty-digit value, **green**


feigi:
Review findings applied — deferrals filed as #981 (open, `ready-for-agent`: two test-helper copies left behind, out of #346's scope) and #982 (closed, `wontfix`: the suggestion band and the 2-0 refutations, checked and recorded).

PR #983, opened 2026-08-28
fix(fleet): release-ticket.sh's die keeps the receipt after a block

### LINKED ISSUE #387: release-ticket.sh: a `die` after any `block` discards every accumulated blocker — exit 2 with no JSON receipt on stdout
Deferred from PR #380 review (closes #86). Specialist dimension: **`silent-failure`**, severity `suggestion` — the band the review workflow budgets **0 refuters**, so this arrives unverified and was scope-split rather than refuted.

## Finding

`skills/fleet/scripts/release-ticket.sh:448` (PR head `eb94b2d`, the `.git` linkage `die`), and every other `die` that fires after a `block` has run.

`block()` accumulates into `$blockers` and echoes prose to stderr. If any later `die` fires, the run exits 2 and **stdout carries no JSON receipt at all** — so every blocker computed up to that point is discarded whole. That is the same stdout-silence the `halt` helper was introduced to fix.

Pre-existing ordering; PR #380 does not introduce it. What #380 changes is the *exposure*: its two new guards are the first blockers that can accumulate **before** the dirty check, so there is now more computed-and-then-discarded state on the path to a `die` than there was.

Adjacent but distinct from #84 (`wt_list`'s `git worktree list` aborting under `set -e` without a receipt) — that one is a single unguarded command; this is the general shape of `die` after `block`.

## Proposed fix — two candidate approaches, and this is the open decision

1. Emit the blocked receipt as soon as `$blockers` is non-empty, before anything that can `die`.
2. Have `die` print the same receipt shape with the accumulated blockers appended, so no computed finding is lost to a later unknown.

The two approaches produce materially different receipts and different orderings of the whole precondition phase — approach 1 changes when the payload is written, approach 2 changes what `die` means. The review verified the defect but settled neither, so a maintainer picks before this costs a claim.

feigi:
## Summary

`release-ticket.sh`'s `die()` used to discard every accumulated blocker whole when it fired after a `block` already ran: exit 2, prose on stderr, and no JSON receipt on stdout at all — so a caller that already had real findings computed got none of them.

`die()` now prints the same receipt shape `halt()` and the blocked checkpoint already use whenever `$blockers` is non-empty, with its own reason appended last exactly as `block()` would have recorded it — approach 2 from the brief, ruled in over emitting the receipt early. A die with nothing accumulated is untouched: the `[ -n ]` guard means the JSON branch is never reached, so every existing die-only regression test stays green unchanged.

Sizing: light

Closes #387

## Enumeration

Every `die` call site in the script funnels through the single shared `die()`, so the fix is one guard in that shared function rather than a patch per call site. Enumerated the sites that can actually fire *after* a `block` has run:

- `git rev-list --count` failure (ahead count)
- `git cherry` failure
- `git ls-remote` failure
- "cannot tell whether $wt exists" (unstatable worktree)
- "$wt has no .git file"
- "cannot resolve $wt" (`wt_canon`)
- "cannot read the git linkage of $wt" (`toplevel`)
- "$wt's .git does not point at $wt" (linkage mismatch)
- "cannot read the status of $wt"

All nine are covered by the same `die()` fix — no per-site patching needed. Every other `die` in the script (usage/argv checks, the worktree-registry counting section, the `wt`/`stray`/`main_branch` lookups, the JSON-field escaping) fires before anything has been accumulated, so the guard is false there and those paths are byte-identical to before.

Deliberately left out (per the brief's own scope line): #84 (`wt_list`'s `git worktree list` under `set -e`) and the ordering of the precondition phase itself — approach 1 would have touched that, which is part of why it wasn't the ruling.

## Review fixes

The review of this PR found three things in the new `die` and they are fixed here.

**1. The guard read an *inherited* `blockers`.** `set -u` is satisfied by an ambient environment variable just as well as by one the run computed, so an inherited `blockers` took `[ -n "${blockers:-}" ]` true on an early die — before `$issue` exists — and the receipt printf aborted on `issue: unbound variable`, losing the very diagnostic the die is there to print. A regression this PR introduced: restoring the pre-PR one-line `die` on the same copy gave the usage prose back. Fixed by assigning `blockers=""` above the `die` definition, which overrides whatever was inherited. Any non-empty value triggered it — `[ -n ]` tests non-emptiness, not JSON validity, so `[]` was as fatal as `[x]` — and the exit code it produced is platform-asymmetric (1 under bash-as-`/bin/sh`, 2 under `/bin/dash`), which is why the regression test pins the prose and not a number.

**2. The receipt grew a format string only this caller used.** `[%s"%s"]` fed raw `$blockers` plus a separate escaped variable, where the blocked checkpoint already had `[%s]` fed the spliced accumulator. Doing the splice in the *argument* makes the two format strings byte-identical — four distinct receipt formats in this file become three. The temporary is also renamed `block_j` → `die_j`, because `block()` owns `block_j` and `block()` is itself a caller of `die`, so the callee was clobbering the caller's variable mid-call (unobservable today, since `die` always exits, but it is what made the shared name unsafe to reuse).

**3. A bare `&&` chain left both of its failures mute.** `|| die` — what `block` and `halt` use — would recurse here, so the chain had no voice. The review found the failed-`jstr` case. Measuring it turned up the worse one it had missed: with the chain as the last command before `fi`, a receipt printf that *cannot write* takes `set -e` and kills the function before its own stderr prose, so the die reason vanishes and the script exits **1** — this script's `NOT released` verdict, fabricated out of a write error. Measured on `/bin/sh` (bash 3.2.57), `/bin/dash` and bash 5.3.15; only zsh survived it. A single `||` arm ends the list so nothing is left to trip. It names neither failure specifically, because `a && b || c` fires `c` for both and this file does not report a cause it did not measure. That write-failure defect is *separate* from #387's and the arm closes it only incidentally, so it has its own regression test rather than riding along — an incidental fix without a pin is how a defect returns under a later refactor that looks unrelated.

## The other half — false-positive check

The brief's own control criterion: "a run that dies with no blockers accumulated behaves exactly as it does today." The `[ -n ]` guard short-circuits to the untouched path when the accumulator is empty. The existing suite already exercises this directly — e.g. "an unreachable remote is an unknown answer, never a 'not pushed'" and the two `.git`-linkage die tests all assert `json === null` for a die with zero blockers computed — and all of them still pass unmodified, confirming no false-positive JSON emission on that path.

## New edit's own risk

Checked "exactly one receipt per run on every path": `die()` emits at most one receipt `printf` before its one `exit 2`, and it only runs at all on a path that would otherwise have printed nothing — no path in the script can now emit two receipts. Checked the exit code stays 2 (not the blocked checkpoint's 1) on every path through the new branch, including the two failure paths of the `&&`/`||` chain.

Differential against the pre-review-fix commit, four shells (`/bin/sh` = bash 3.2.57, `/bin/dash`, bash 5.3.15, zsh 5.9), `die` extracted with the real `json.sh` sourced:

- blocked die, cause containing a quote and a backslash — stdout and stderr byte-identical, rc 2 both sides, receipt parses
- early die, accumulator empty — byte-identical, rc 2 both sides
- `jstr` stubbed to fail — the arm speaks on all four shells, then the die reason, rc 2
- receipt write failing on a healthy `jstr` — before, the die reason was lost entirely on sh, dash and bash 5.3; after, both lines print on all three

`shellcheck -x -S warning` clean.

## Test

"a die after a block still emits the accumulated blockers, not a bare exit 2 (#387)" drives exactly the named shape: the `ahead` check blocks first (one real finding), then a `git cherry` shim fails, hitting the die this fix reaches. Asserts the receipt is non-null, still exits 2, `label` is `null`, and `blockers` is `[ahead-finding, cherry-die-cause]` in that order.

The review found that test did not discriminate three of the fields it appeared to cover — `die` has its own inlined printf, shared with none of the other three receipt sites, so nothing else in the file pins them for *this* emitter. Added:

- `branch` and `worktree` assertions on the die receipt. Hardcoding literals for both left the whole 1290-test suite green. `git worktree list --porcelain` reports the realpath, so the expectation is `realpathSync(c.wt)`, not the `/var` symlink the fixture built.
- "the die receipt's applied field is the flag, not a constant (#387)" — the same path with `--apply` withheld. This path is reached exactly once in the file and the harness defaults the flag on, so a die printf ignoring `$apply` entirely stayed green across all 98 cases.
- three "an inherited blockers=… does not silence an early die (#387)" cases, pinning fix 1 on the prose rather than the exit code.
- "a receipt that cannot be written still leaves the die reason and exit 2 (#387)", pinning fix 3's write-failure half directly. Closing fd 1 is the portable way to make the write fail with a healthy `jstr` — `stdio: "ignore"` opens /dev/null, whose writes succeed — so the wrapper is a shell that execs the script after `exec >&-`. Pinned on the prose *and* the exit code, unlike the inherited-`blockers` cases: there the code is platform-asymmetric, here the code IS the defect, since 1 is a verdict this script means something by.

Every new assertion was mutation-tested in both directions on a scratch copy:

| mutation | result |
|---|---|
| `applied` hardcoded `true` | 1 red — only the new `applied` case |
| `branch` hardcoded | 1 red — only the `#387` case |
| `worktree` hardcoded | 1 red — only the `#387` case |
| `blockers=""` moved back below `die` | 3 red — exactly the three inherited-`blockers` cases |
| the `\|\|` arm deleted | 1 red — only the write-failure case |
| `applied` hardcoded, against the write-failure pin | that pin stays **green** — it is scoped to its own property |
| `branchRewritten` hardcoded (must NOT be caught) | **0 red** — the pins are scoped, not blanket |

## Test plan

- [x] `node --test skills/fleet/scripts/release-ticket.test.mjs` — 103/103 pass (was 98)
- [x] `node --test skills/fleet/scripts/*.test.mjs` — 1295/1295 pass (was 1290)
- [x] Original test verified red on the unfixed script, green on the fixed one
- [x] All five new pins mutation-tested red-and-green, per the table above


feigi:
Review adjudication, PR #983 (10 findings: 2 survived, 2 refuted, 6 unverified).

**Applied** (pushed as `3d6ca81`, `151c436`, `de4f866`):

| finding | dimension | outcome |
|---|---|---|
| survived[0] + refuted[0] | silent-failure / correctness | the guard read an *inherited* `blockers`. The same defect found independently by two dimensions; three of four refuters reproduced the mechanism. Fixed. |
| survived[1] | tests | the die receipt's `applied` was not discriminated from a hardcoded `true`. Fixed. |
| unverified[0] | silent-failure | give the failed escape a voice. Survived a refuter briefed against its mutually-exclusive sibling; landed with corrected wording (see below). |
| unverified[1] | tests | `branch`/`worktree` unasserted in the die receipt. Refuter could not refute; the mutation survives the **full** suite, not just the targeted test. Fixed. |
| unverified[4] + unverified[5] | simplify | one change, not two — collapse the die format string onto the blocked checkpoint's, rename `block_j` → `die_j`. Refuter verified inert. Applied **after** the guard fix and re-verified, because the replacement keeps the `[ -n ]` guard and would otherwise have reinstated the defect. |

Two corrections to findings' own text, both measured: the guard failure is **not** value-dependent (`blockers=[]` fails too — `[ -n ]` tests non-emptiness, not JSON validity), and its exit code is **platform-asymmetric** (1 under bash-as-`/bin/sh`, 2 under `/bin/dash`), so the regression test pins the diagnostic prose and not a number.

`unverified[0]`'s own suggested wording was also wrong and was not copied: it named the escape as the cause, but `a && b || c` fires `c` on a write failure too. Landed as *"no JSON receipt for #%s — the escape or the write failed"* — true of both arms.

**Deferred and filed:**

- #987 — `unverified[2]` (types): one shared "give me the blockers as a JSON array body" accessor across the four receipt printf sites. Out of scope for #387, and the reviewer measured that the current duplication is not broken. `needs-triage`.
- #988 — the refutation record, filed closed as `wontfix`: `refuted[1]` (partial-scan marker — the confirming probe never read stderr, which carries the discriminating verdict line) and `unverified[3]` (the "safe deliberate degrade" reading of the same lines `unverified[0]` covers).

**The mutually-exclusive pair.** `unverified[0]` and `unverified[3]` addressed the same lines with opposite remedies — one adds output, the other documents its absence — so **one** refuter was briefed with both and made to settle them against each other, rather than two independent refuters each approving their own side. `unverified[0]` won. What killed `unverified[3]`: its "safe degrade" was measured over only one of the two ways die's AND-OR list can fail. With stdout closed and a **healthy** `jstr`, the shipped shape loses the die reason entirely and exits **1** — this script's `NOT released` verdict, fabricated out of a write error — on `/bin/sh`, `/bin/dash` and bash 5.3.15; only zsh survived. Full evidence in #988.

Suite 1290 → 1294, all green. All four new pins mutation-tested in both directions, including one mutation they must **not** catch. PR body reconciled against the final diff.

feigi:
Follow-up to the adjudication above: the **write-failure** defect is now pinned directly rather than riding on the `||` arm as an incidental fix.

`bd13598` adds "a receipt that cannot be written still leaves the die reason and exit 2 (#387)". It is a separate defect from #387's — a receipt `printf` that cannot write took `set -e` and killed `die` before the function reached its own `printf … >&2`, so the die reason vanished and the script exited **1**, a code this script defines as `NOT released — nothing was touched`, fabricated out of a write error.

Drivable from the existing harness, so it was pinned rather than filed. Closing fd 1 is the portable way to make the write fail with a healthy `jstr` — `stdio: "ignore"` opens `/dev/null`, whose writes succeed — so the wrapper is a shell that execs the script after `exec >&-`, wrapped around the same die-after-block fixture the #387 case already builds.

Pinned on the prose **and** the exit code, unlike the inherited-`blockers` cases where only the prose is portable. Here the code *is* the defect: 1 is a verdict this script means something by.

Mutation-checked three ways on a scratch copy:

| mutation | result |
|---|---|
| the `\|\|` arm deleted | 1 red — only the new write-failure case |
| `applied` hardcoded (caught by its own pin) | the write-failure pin stays **green** — scoped to its own property, not a blanket failure |
| `branchRewritten` hardcoded | **0 red** |

Suite 1294 → 1295, all green. `release-ticket.test.mjs` 98 → 103.

One thing the first mutation shows and is worth stating: deleting the arm reds *only* the write-failure case, so the arm's other half — the failed-`jstr` path — remains unpinned by any test. One pin holds the arm as a whole against deletion, which is what matters for keeping it alive; driving a failed `jstr` would need a `sed`/`tr` PATH shim and is not in scope here.

Issue #144, opened 2026-07-31
run-team: the finisher's pinned SHA races the reviewer that is still absorbing relays — 6 of 8 PRs this run


feigi:
Measured across one full `/fleet:run-team` session, 8 PRs, ~40 specialist reports.

## The loop

`run-team/SKILL.md` tells the controller to dispatch a finisher pinned to the reviewed SHA, and tells the finisher that a moved head halts it. Both rules are right. What neither anticipates is that **the reviewer is usually still alive and still receiving specialist relays when the finisher is dispatched.**

Specialist reports arrive 5–16 minutes apart (measured: fastest 4m, slowest 15m30s). A reviewer composes its verdict when it has enough, sends it, and then keeps working as late reports land. Every late relay it acts on moves the head off the SHA the finisher was pinned to.

Observed sequence, four times:

1. reviewer sends verdict at SHA `A`
2. controller dispatches finisher pinned to `A`
3. late relay reaches the reviewer; it applies the finding and pushes `B`
4. finisher halts on the moved head — **correctly**, and labels nothing
5. controller re-pins or re-dispatches

## Frequency

**6 of 8 PRs had a head that moved after the reviewer's first verdict**: #107 (`ee81d14`→`fdac65d`→`89c8310`), #108 (`1ca997d`→`3c7f05b`), #109 (`308b776`→`2af7d65`→`2c36ead`), #110 (`a707462`→`4a669d0`), #115 (`4d21525`→`02f64bf`), #130 (`1e158b7`→`b38e651`).

Two produced finisher halts (#109, #130). One produced a *second* halt after a re-pin (#108). The halts were all correct — one caught a file mid-write with monotonic diffstat growth across four samples — but each cost a round trip.

## Two distinct causes that read identically to a finisher

A finisher sees only "head ≠ pin". This run produced both causes, and the distinguishing evidence turned out to be cheap:

- **live editor** — worktree dirty, mtimes inside the check window, diffstat grows between two reads
- **rebase under the reviewer** — worktree clean, `git reflog` shows `reset: moving to origin/...`, diff across the move is byte-identical

#109 was a rebase (the branch was replayed onto `dc22aa9` when #108 merged; the reviewer had not pushed at all). #130 was a live editor. Once finishers were told to produce mtimes and a growth check, a halt became a one-message resolution instead of a re-review.

## Contributing factor worth naming separately

[...]

Issue #654, opened 2026-08-18
review-pr-cost-and-apply-policy spec :285 cites review-pr.js:284-286 for the 'tests 0' convention; that range is the BOUND EVERY READ paragraph


feigi:
**Deferred from PR #647 review** (issue #143, head `b393dff`). Dimension: **comments**. Location: `docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md:284-286`.

**Why deferred:** pre-existing drift in a design doc PR #647 does not touch — out of #143's ticket scope. Flagged because it sits in the exact subject area the PR modifies.

## The claim, verified from the citing text (not from the cited line)

The design doc says:

> A bad run reuses the convention already in the specialist prompt (`review-pr.js:284-286`) rather than inventing one: **`tests 0` is a FAILED run, not a pass.**

The **claim itself is true** — the specialist prompt does carry that convention. The **citation is wrong**: `workflows/review-pr.js:282-290` is the `BOUND EVERY READ` read-bounding paragraph, ending in a closing template-literal brace. The rule the doc points at lives at `workflows/review-pr.js:845` today (`grep -n "FAILED run, not a pass" workflows/review-pr.js` -> single hit, `845`).

## Remedy is OPEN

Two defensible options, and picking between them is a judgement about what these specs are for:

1. **Drop the line citation**, keeping the quoted convention — a dated design doc is a historical record, and any `:NNN` in it rots by construction.
2. **Update it to `:845`** — accurate today, stale again on the next edit to that prompt.

Label is `ready-for-agent` because the broken citation is confirmed, but say which of the two the repo wants before editing; the same choice presumably applies to every other bare `:NNN` in `docs/specs/`.

feigi:
**Second instance in the same file, from the PR #996 review** (dimension: `comments`).

`docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md:123` cites `run-team/SKILL.md:382` for the finisher's duty summary:

> The finisher is "a fresh small agent, not the fix-applier resumed" (`run-team/SKILL.md:382`) whose duties are worktree audit, confirm deferrals filed, add `ready-to-merge`.

The quoted string now lives at `SKILL.md:1262`. Line 382 holds unrelated text — the tail of the test-path expansion paragraph, immediately above **A reused worktree may lack the runner**. Drift of ~880 lines.

[...]

PR #996, opened 2026-08-28
fix(run-team): give the finisher's mutation gate its own throwaway tree (#389)

### LINKED ISSUE #389: run-team: the finisher is a WRITER — its mutation gate edits the same two files the fix-applier owns, with no lock between them
Measured live during a `/fleet:run-team` run, 2026-08-11, on PR #380. No work was lost, and the reason it was not is luck of timing plus one member's judgement — not a property of the design.

## The shape

`run-team/SKILL.md` treats the finisher as an auditor: it verifies, then labels. But a finisher's gate **mutates source files** — the controller routinely asks it to independently re-run the ticket's acceptance mutation (loosen a regex, delete a guard clause, flip a per-entry reset), run the suite, and restore. That makes it a **writer** in the same worktree the fix-applier owns, and nothing sequences the two.

They overlapped here because the controller dispatched the finisher while three relays to the fix-applier were still unacknowledged — a controller error, and a variant of #144. But **#144 is about a stale SHA; this is about concurrent writes to the same paths**, and it would occur on any overlap however caused.

## What was actually observed

At the moment the fix-applier announced one more clause to land:

```
$ git -C .worktrees/86-release-dry-apply-parity status --porcelain
 M skills/fleet/scripts/release-ticket.sh
```

The finisher was inside that worktree. Two orderings were available, and neither is detectable after the fact from the diff alone:

1. The finisher's restore (`git checkout -- <path>`) lands on top of the fix-applier's uncommitted clause → **the clause is destroyed silently**.
2. The finisher's mutation is still in the tree when the fix-applier commits → **a deliberately broken guard is committed as though it were the fix**, in a commit nothing re-reviews. (`review-and-fix.md`'s fix commits are not re-reviewed; the review snapshot predates them.)

Outcome 2 is the dangerous one: a mutated guard reads as a plausible correction.

## Why nothing broke

Three independent things, none of them structural:

- The controller's halt instruction said "restore the mutated paths" — **wrong, and it would have caused outcome 1.** The finisher refused, because it could not prove the diff was its own, and reported it instead. The never-force rule beat an explicit controller instruction.
- The finisher halts at step 1 on any dirty worktree, so it never reached its own mutations.
- The fix-applier's commit was `1 file changed, 5 insertions(+), 3 deletions(-)` — a diffstat small enough that a stowaway hunk would have been obvious, and it checked.

- **Sequence them explicitly.** The finisher does not start until the fix-applier has reported *and* the controller's outbox to it is empty. Cheapest, and it is the rule #144 already implies — but it is a controller-discipline fix for what is now a data-race, and controller discipline is what failed here.
- **Give the finisher its own tree.** Its mutation gate needs a worktree, not *the* worktree: a throwaway checkout at the PR head, mutated and discarded. Removes the shared-path hazard entirely and costs one `git worktree add`/`remove` per finish. A merge bot in this same run already used a throwaway detached worktree to measure `origin/main`'s suite, so the pattern exists.
- **Make the finisher non-mutating.** Have the fix-applier record its mutation results and have the finisher check the record rather than re-run it. Cheapest at runtime, but it gives up the independent verification that has caught two unpinned fixes today (#377's acceptance criterion, #380's sibling-lock fixture), so it trades away the thing that makes the gate worth having.

The middle option looks strongest, but the trade — an extra worktree per finish against a silent-corruption class — is a maintainer call.

## Related

- **#144** — the finisher's pinned SHA races a reviewer still absorbing relays. Same overlap, different consequence (stale SHA vs concurrent writes). A fix for #144's ordering would prevent this too, which is an argument for treating them together.
- **#381** — filed this run; also about a worktree the fleet leaves in a state its tooling does not expect.

## Scope searched

`ledger.mjs check` → `verdict: clean`, tracker reached (nearest #151 at 0.20, unrelated). Plus `gh issue list --state all --search "finisher fix-applier worktree"` and a read of #144. Note the check's own reliability is filed as #388.

feigi:
Closes #389

Sizing: light

## The defect

`run-team/SKILL.md` calls the finisher an auditor, but its gate **mutates
source**: the controller has it independently re-run the ticket's acceptance
mutation, run the suite, and discard. That made it a writer in the worktree the
fix-applier owns, with nothing sequencing the two. Measured on #380 — a
finisher was inside that worktree at the instant the fix-applier announced one
more clause to land. Neither ordering is detectable afterwards from the diff: a
restore eats the uncommitted clause, or the mutant is still there when the
fix-applier commits and a deliberately broken guard ships as the fix, in a
commit nothing re-reviews.

## The change

The Agent Brief's ruling, taken as ruled — the middle option, its own tree. The
gate is documented where the finisher already reads (duty 2, which is where it
verifies a claim by reading a diff; running a mutation is the half reading
cannot do). It adds a detached worktree at **its dispatch pin**, under a scratch
path carrying its own member name, mutates there, and `--force`-removes it on
every path including failure.

Explicitly **not** the other two options: sequencing stays as the complement it
already is (the `#144` outbox rule is untouched), and the gate still mutates and
runs the suite itself rather than reading a recorded result.

The gate mandates `<testCmd>` three times, and nothing told the controller to
hand a finisher one — the runbook's pass-list named only the workflow and the
fix-applier, and no `fleet-finisher` agent definition exists to carry a default,
so the finisher holds only what the controller's prompt contains. Measured this
run: four finishers dispatched, none given a `testCmd`. The gate would have
shipped with the same hole the runbook already names for the fix-applier —
"substituting `<testCmd>` with nothing leaves it no gate at all". So the
pass-list now names the finisher, and duty 1's dispatch instruction says to hand
one over alongside the two-cause block.

## Facts settled against the tree, not the ticket

- `git worktree remove` on a dirty tree: rc 128, `contains modified or untracked
  files`, directory left on disk; `--force` removes it at rc 0 (git 2.50.1).
- `reap.sh`'s branchless sweep is bounded to `*/.worktrees/*`, so a tree under
  `<scratch>` is a `kept` entry it reports and never collects — the leak is not
  swept for the finisher.
- `claim-ticket.sh` writes `./agent-test` only into a worktree it freshly
  claims, so an added tree has none.

## Verification

`skills/fleet/scripts/finisher-own-tree-prose.test.mjs`, slice-scoped to the
duty-1 and duty-2 blocks. Every pin mutation-tested both ways: each of six
semantic mutants (pin swapped for `headRefOid`, `--force` dropped, `--force`
scope dropped, every-path removal weakened, owned-worktree ban dropped, gate
made non-mutating, neighbour paragraph clipped) reds exactly one test, and two
rewordings that should not be caught stay green.

Three assertions were then rewritten under review, each re-measured against the
mutants that motivated it:

- **The path is read off the `worktree add` and required back on the
  `worktree remove`.** The scoping token appears twice, so the previous bare
  match was satisfied by either occurrence alone — descoping exactly one command
  left add and remove targeting different paths and stayed 5/5 green in both
  directions. Now descoping the add only reds the path assertion, descoping the
  remove only reds the linkage assertion, and a reword elsewhere in the slice
  stays green.
- **The duty-1 halt phrase is matched through the file's own flattener**, and
  against a duty-1 slice rather than whole-file. It previously ran on raw
  `RUN_TEAM`, so a pure rewrap splitting the pinned phrase — the source line is
  exactly 80 chars, on this file's wrap boundary — reddened it with a message
  claiming the halt had been weakened. Measured: rewrapping inside the span is
  now green, weakening the words is red, and relocating the phrase out of duty 1
  behind a verbatim decoy is red (the whole-file form was green on that).
- **The controller's `<testCmd>` handoff is pinned**, contiguous with the reason
  it exists. Deleting it reds; rewrapping it stays green; severing the handoff
  from "duty 2's mutation gate runs it" across a sentence boundary reds.

Includes the accept side: duty 1's dirty-owned-worktree halt — the guard that
caught this defect live — and duty 2's own negative-claim scoping caveat are
pinned unchanged, so isolating the write cannot quietly weaken the detection or
swallow the paragraph it is appended after.

Full suite: 1339 pass, 0 fail.

## Review

Three findings applied (above). One suggestion — that duty 2's removal
paragraph never says what the finisher does when the suite will not start — was
checked by a refuter and **refuted**: "your own halt at any duty" in that same
sentence already covers it, every duty states halt-before-label, and the file
already adjudicates a missing suite result as "halt here, same as a dirty
worktree". Recorded in a closed `wontfix` issue. One pre-existing citation drift
in `docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md` was
deferred to the issue that already tracks that file's positional citations.

https://claude.ai/code/session_018PuZ79KKWm4WtxKUpo3raA


feigi:
**Review applied** at `89f6fcb`. Suite: `node --test skills/fleet/scripts/*.test.mjs` → **1339 pass / 0 fail**.

**Applied (3)**
- `correctness` — nothing told the controller to hand a finisher a `<testCmd>`, which duty 2's new gate mandates three times. Pass-list and duty-1 dispatch instruction now name the finisher; pinned.
- `tests` — the path-scoping assertion was satisfied by either of the path's two occurrences, so descoping one command silently split add from remove. Path is now read off the add and required back on the remove.
- `tests` — the duty-1 halt pin ran on raw `RUN_TEAM`, so a pure rewrap reddened it. Now flattened and sliced to duty 1.

**Filed (2)**
- #998 — `PR #996 review: the suggestion band, checked`, `wontfix`, closed. One suggestion refuted with measurement: "a suite that will not start" is already adjudicated by "your own halt at any duty" in the same sentence, by halt-before-label at every duty, and by the no-CI block's "halt here, same as a dirty worktree".
- #654 (comment) — pre-existing citation drift at `docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md:123` (`SKILL.md:382` → now `:1262`). Same file and same class as that issue; commented rather than duplicated.

**No action (1)** — `review-and-fix.md:8`'s `tests 937` figure. The sentence disclaims its own drift by design ("compare against a full run rather than against zero"); the finding's own verdict was no action needed.

PR #1000, opened 2026-08-29
fix(claim-ticket): pin the file guard's logical pwd against a symlinked node_modules entry

### LINKED ISSUE #401: claim-ticket.sh: the vendored file guard's logical `pwd` is unpinned — `pwd -P` survives all 50 tests and reintroduces #100 on a workspace-symlinked node_modules
## What

`claim-ticket.sh`'s emitted `agent-test` decides whether a file argument is vendored by comparing the argument's directory, resolved with the shell's **logical** `cd`/`pwd`, against `"$PWD"/node_modules/`:

```sh
case "$arg" in */*) argdir="${arg%/*}" ;; *) argdir="." ;; esac
case "$(cd "$argdir" 2>/dev/null && pwd)/" in
  "$PWD"/node_modules/*) ... refuse ... ;;
esac
```

The omission of `-P` is **deliberate and correct** — logical `pwd` collapses `..` textually while leaving symlinks unresolved, which is what node's own `path.normalize` does. Nothing in the suite pins it.

## Evidence

Found re-gating PR #394 at head `b134407`, on a freshly materialized runner (script `cksum 3556071636`, emitted template stamp matching):

- Mutating `cd`/`pwd` → `cd -P`/`pwd -P` leaves **all 50 `claim-ticket.test.mjs` tests green.**

It is **not an equivalent mutant.** With the ordinary npm/pnpm workspace layout where `node_modules/pkg` is a symlink to `../packages/pkg`:

| | `node_modules/pkg/w.test.mjs` |
|---|---|
| node | **DROPS** it (silently — pass 1, only the anchor ran) |
| shipped code (logical `pwd`) | **REFUSES** — correct |
| `pwd -P` mutant | **exit 0, ran nothing extra** — the silent drop #100 describes, reintroduced |

So a future "hardening" pass that adds `-P` for apparent robustness would reintroduce this ticket's own defect and the suite would stay green.

## Not already owned

- **#186** is the *directory* branch's `find` filter, with the symlink pointing **into** `node_modules`. This is the *file* branch's `pwd` resolution mode, with the symlink **inside** `node_modules` pointing out.
- **#230** is the file-vs-directory anchoring incoherence, a different axis.
- **#229** is the `-type d` prune clause on the directory branch.

## Suggested fix

One test that lays down `node_modules/pkg` as a symlink to a sibling real directory, and asserts the guard refuses a file under it. That reddens under `-P` and stays green under the shipped form.

Per this repo's standing constraint: mutation-test the new pin **both ways** — confirm it goes red under `-P`, and confirm a harmless nearby edit leaves it green.

## Provenance

Not a defect in PR #394 — the shipped behaviour is correct. This is a missing pin on correct behaviour, the class filed as #379 and #345.

feigi:
## Summary

The file branch's vendored-check in the `agent-test` template `claim-ticket.sh` resolves an argument's directory with logical `cd`/`pwd` (no `-P`) before comparing it against `node_modules`. That is deliberate and correct — it mirrors node's own textual path normalization, which never resolves symlinks either — but nothing in the suite pinned the choice. Mutating the guard to `cd -P`/`pwd -P` left the whole suite green, and would silently re-admit the class #100 closed: in the ordinary npm/pnpm workspace shape where a `node_modules` entry is itself a symlink to a sibling real directory, physical resolution walks the symlink out of `node_modules`, this guard stops matching and falls through unrefused, and node then drops the file anyway on its own unresolved spelling — silently, at exit 0.

- Adds `runner: a vendored file behind a symlinked node_modules entry refuses`, laying down `node_modules/pkg` as a symlink to a sibling real directory and asserting the guard refuses a file under it.
- States the `-P` hazard directly in the guard's own comment, so the next reader meets the reason before touching the pin.

**Not in scope** (per the ticket): #186 (the *directory* branch's `find` filter, symlink pointing *into* `node_modules` — a different mechanism) and #230 (file-vs-directory anchoring incoherence, a different axis). The resolution mode itself is unchanged; the section below records the one behavioural change review added on top of the pin.

**The directory branch does not have this gap.** It already resolves with `pwd -P` (physical), deliberately, for the opposite reason (#186 needs to follow the argument's own symlink to its real target to catch a symlink pointing *into* a vendored tree) — and that choice is already pinned by the existing `runner: a symlink whose target is inside a vendored tree refuses` test, which requires `-P` to pass. Only the file branch's *logical* form was unpinned.

## Added in review: the same guard's missing `CDPATH=`

Review of this PR found that the comment above declares this guard refuses vendored arguments, and under an inherited `CDPATH` it does not — so the diff as first pushed shipped a claim its own code did not hold. The file branch resolved `$argdir` with a bare `cd`, while the sibling *directory* branch has carried `CDPATH= cd --` since it was hardened, with a comment naming exactly this hazard. `cd` consults `CDPATH` before the cwd, and the divergence lands the same way twice:

- a decoy on `CDPATH` holding a same-named `node_modules/…` resolves the guard into the **decoy**, so `"$PWD"/node_modules/*` misses — on both `/bin/sh` and `/bin/dash`;
- a **non**-matching `CDPATH` fails the `cd` outright under macOS `/bin/sh` (bash 3.2, which does not fall back to the cwd), `2>/dev/null` eats the error, and the case subject is a bare `/`.

Either way the guard falls through without refusing, the vendored argument reaches node, and node discards it silently at exit 0 — #100's silent drop with the loud refusal removed. The runner is spawned with `{...process.env}`, so a developer's `CDPATH` reaches it.

- Adds `CDPATH= ` and `--` to the file branch's `cd`, the same spelling the directory branch already uses. The resolution mode stays **logical** — no `-P` — which is orthogonal and is what the rest of this PR is about.
- Adds `runner: a vendored file argument refuses under an inherited CDPATH`, using the decoy shape rather than a bare `CDPATH=/tmp`: a non-matching entry only diverts bash 3.2 and would pin this on the dev platform alone, while a matching one diverts both shells.

So the resolution mode is unchanged, but this PR is no longer pin-only: it also closes the `CDPATH` hole in the guard it documents. Mutation-tested both ways — stripping `CDPATH= ` reds the new row and only it; adding `-P` reds the #401 row and leaves the new one green. Full suite: `tests 1341, pass 1341, fail 0`.

Sizing: light

Closes #401


feigi:
Review deferrals filed: #1005 (argdir indirection + unreachable arm — `unverified[0]`/`unverified[3]`, one question, `ready-for-agent`), #1006 (permission-vs-typo mislabel — `unverified[1]`, `ready-for-agent`), #1007 (the suggestion band, checked — the `unverified[2]`/`unverified[4]` comment-duplication pair, both refuted by one refuter briefed with both, `wontfix`, closed).

Applied here instead of deferred: the two `survived` findings, which are one defect at two mechanisms — the file guard's missing `CDPATH=` — fixed once in `41d7743`.

PR #1001, opened 2026-08-29
fix(ledger): check's verdict reads the scores it already computes (#388)

### LINKED ISSUE #388: ledger.mjs check: the near-miss rows carry the signal the verdict throws away — clean alongside a 0.45 near-miss, tracker-hit on two score-0 issues

feigi:
Closes #388

Sizing: heavy

`check`'s verdict was derived from the presence of tracker rows alone, so it printed `tracker-hit` over rows its own overlap rates 0.00, and `clean` directly above near-miss rows that named the right issue. Both numbers were already computed, already in the payload and already on stderr. The verdict now reads them.

## What changed

- **`soft-hit`**, a new verdict at **exit 0**, for signal that is not an established duplicate: a tracker set with no scoring row, or — once the tracker read succeeds — a filed near-miss at or above `NEAR_SOFT_HIT`. A read that failed stays `unverified`, which outranks both.
- **`NEAR_SOFT_HIT = 0.2`** in `ledger.mjs`, beside the near-miss display cap, with the measured basis in its comment: the near-misses #388 records a reader going on to confirm as the genuinely adjacent issue scored from 0.20 up to 0.45, and #388 also records an adjacent row at 0.09 that this floor does not reach.
- **`tracker-hit` and exit 3 now require a row that scores.** That is the one shape whose exit code moves: 3 → 0. Every other code is untouched, and the exit code stays a pure function of `verdict`.
- **stderr agrees with the verdict.** The score reaches each tracker row (`TRACKER HIT — #388 (OPEN, score 0.58) …`), a set with no scoring row is headed `TRACKER ROW` instead of the blocking `TRACKER HIT`, and a soft hit says on stderr that its exit 0 is not permission.
- **Every document that states the contract** is updated: `review-and-fix.md` step 5, run-team's **Run ledger** section, and the fleet-plugin-design script-surface row, which is the third place these exit codes are written down.

## Exit codes after this change

| code | verdict |
|---|---|
| 0 | `clean`, `unverified`, **`soft-hit`** |
| 1 | `already-filed` |
| 2 | usage error (no verdict) |
| 3 | `tracker-hit`, and only with a scoring row |

Minting a code for `soft-hit` was ruled against by the same argument #152 used for `unverified`: it would break `check "$s" && gh issue create` on runs where the answer is advisory, and #145/#152 already ruled near-misses advisory rather than blocking. So the payload and stderr carry the distinction and the callers are taught to read it.

## Tests

`node --test skills/fleet/scripts/*.test.mjs` — 1346 pass, 0 fail.

New cases drive all four acceptance shapes plus the two measured failures from the issue body: the wontfix-vocabulary hit set scoring 0.00, and the score-0 hits sitting beside a filed row well above the floor. The control that stops every check becoming a soft hit has a near row **below** the floor rather than none at all, so it pins the floor from underneath.

Mutation-checked, each red only the intended test: `NEAR_SOFT_HIT` 0.2 → 0.9 (near-above-floor case), 0.2 → 0.1 (the clean control), and the score gate back to `tracker.hits.length` (both zero-score cases). The two new prose pins were mutated on their own text as well.

Review fixes add three pins the suite was missing, each mutation-checked with a negative control: the `SOFT HIT` stderr line (deleting the block reds exactly the two tests now asserting it, and no others), and the floor's value and inclusivity (a fixture scoring exactly 0.20 reds under both `>=` → `>` and 0.2 → 0.35, while staying green under an equivalent rewrite of each).

Four existing tests are re-pointed, not weakened. Their fixtures scored 0.00 against their own subjects and asserted exit 3 — that is the defect shape, so the ones pinning truncation reporting and the unread-ledger arm now carry a title that scores, and the #232 title-less row keeps everything #232 asked for (read, describable, never `unverified`) at the advisory exit code its zero score now earns.

## Deliberately left

- The tracker query and its term selection, per the brief's ruling — #153's residue, its own ticket.
- The already-filed subset match at exit 1: unscored by design, and the behaviour callers gate on.
- `unverified` (#231) and the display caps (#154).


feigi:
Review fixes applied in `e396577`. Full suite `node --test skills/fleet/scripts/*.test.mjs` — **1346 pass, 0 fail** (1345 baseline + 1 new test).

**Applied (4):**
- The `SOFT HIT` stderr line, unpinned while every neighbouring verdict's was pinned — asserted in both #388 tests that produce a soft hit.
- The near-miss floor's **value and inclusivity**, both free (`>=`→`>` and `0.2`→`0.35` were each green across the whole suite) — closed by a fixture scoring exactly 0.20.
- `SKILL.md`:1913 claimed a filed row above the floor moves the verdict to `soft-hit` unqualified; a failed tracker read stays `unverified`, which outranks it. Qualified.
- `ledger.mjs`:669 said both best-scores are "used by both the report and the verdict" — true of `bestHit` only. Route restated; conclusion was sound and is kept.

Each added test was mutation-checked with a negative control: deleting the `SOFT HIT` block reds exactly the two tests now asserting it and no others; the floor test reds under both `>=`→`>` and `0.2`→`0.35`, and stays green under an equivalent rewrite of each.

**Deferred (3 issues):**
- #1009 — `PR #1001 review: the suggestion band, checked` (`wontfix`, closed). Seven entries, each checked by a refuter that ran something and did not survive. Includes the batch-level `TRACKER HIT` heading (mechanism real, headline false — cured by this diff's own per-row `score 0.00`), the "vacuous" combined test at `ledger.test.mjs`:676 (it is the suite's **only** witness of the both-operands-true case, and the sole killer of an escalate-when-both mutation that re-mints exit 3 — kept), and the `NEAR_SOFT_HIT` comment's cited figures, which were checked against #388 with `gh` and are **accurate**.
- #1010 — the `verdict` string domain has no single source of truth (`needs-triage`, out of scope, unchecked).
- #1011 — three style deferrals on the `bestHit`/`bestNear`/ternary region (`needs-triage`, out of scope, unchecked).

One proposed remedy was rejected on measurement rather than on scope: raising the floor toward 0.45 would discard the confirmed-adjacent rows #388 records at 0.20, 0.22 and 0.30 — three of its four — sending the two `clean`-over-the-right-issue failures the ticket exists to fix back to `clean`.

Issue #1002, opened 2026-08-29
Presence pins on the verbatim dispatch blocks cannot see an exception appended inside a block — needs a mechanism other than a regex


feigi:
## What

The pins #400 adds to phase 2's verbatim dispatch blockquotes are presence-and-adjacency pins. They cover the **location** half — a block deleted, gutted, or reordered reds. They do not cover a sentence **appended inside** a block that carves an exception out of a rule already pinned there: every pinned fragment survives untouched, so the suite stays green while the rule the block exists to carry no longer governs.

## Evidence

Measured against the pins in `skills/fleet/scripts/dispatch-block-pins-prose.test.mjs`, on the prose-pin subset of the suite with a per-mutant control confirming each mutation reached disk. Three carve-outs, each appended inside a block whose rule it contradicts, each leaving the block's pinned spans byte-identical:

- Commit-incrementally block, after `effectively unrecoverable.` → `A single commit at the end is fine when the diff is small.` — **green**.
- Enumerate-the-class block, after `A green suite is evidence only about the paths it exercises.` → `Skip the enumeration entirely when the ticket names only one case.` — **green**.
- Re-derive block, after `say which, and stop.` → `If the criterion looks close enough, implement it anyway.` — **green**.

The same ceiling is already recorded for `member-prompt-prose.test.mjs`, measured on the identity block and on `Never apply ready-to-merge, never merge`. So this is a property of the pin **form**, not of any one file's craft.

## Why a tighter regex is not the fix

`assert.doesNotMatch(/\bunless\b|\bexcept\b/i)` was tried on this exact question and failed all three controls: scoped to the pinned sentence it does not kill its own mutant; broadened enough to kill it, rewordings preserve the defect and stay green — including one carrying no conditional word at all; and it reddens on a legitimate edit, since a standing rule containing the word *unless* is exactly the kind of text that belongs in a verbatim block. Confirmed here: a mutant adding `Do not call the AgentTool unless requested.` to a block stays green under the pins as written, and would not under a blacklist.

Asserting the absence of arbitrary natural-language negation is unbounded. The mechanism has to change.

## Candidate mechanisms

- **LLM judge over the instruction file.** Ask whether each block's rules are still unconditional, rather than whether its words are still present. Costs a model call in the suite and needs a stable verdict; the open question is how to make it deterministic enough to gate on.
[...]

Issue #998, opened 2026-08-28
PR #996 review: the suggestion band, checked


feigi:
One `suggestion`-band finding from the PR #996 review was checked by a refuter and did not survive. Recorded here rather than left open: it is not a confirmed defect, and nothing is queued.

## Refuted — "a suite that will not start" has no stated verdict

**Dimension:** correctness · **Severity:** suggestion
**File:** `skills/fleet/skills/run-team/SKILL.md:1336`

**Claim.** Duty 2's removal paragraph lists three triggers — "a mutant that comes back green, a suite that will not start, your own halt at any duty" — and the middle one gets no verdict: the block never says whether the finisher halts like duty 1 or falls through to duty 3 and labels. A `git worktree add` tree is a bare checkout with no installed dependencies, which is the ordinary way a suite fails to start.

**Refuted, with evidence.**

1. **The sentence already contains the verdict.** The third item, "your own halt at any duty", is not a peer failure mode alongside the first two — it is the general clause covering a halt inside duty 2. The sentence's job is "the `worktree remove --force` runs regardless", not enumerating gate outcomes.
2. **Halt-before-label is stated at every duty.** Duty 1: "Dirty or diverged halts the finisher *here*, before the label … labels nothing." Duty 2: "Not filed → file it or halt, never label over it." Duty 3: "Zero or more than one halts the finisher before the label, naming which it found." Duty 4: report "anything it halted on — cause and evidence".
3. **The exact proposed clause is already written for the analogous absence.** SKILL.md's no-CI block: "Either missing → **halt here, same as a dirty worktree**: report `no CI configured, no verified suite run on record` and add no label", and "**Absence never reads as pass.**"
4. **The named non-start cause is pre-empted with a remedy, not merely omitted.** "`claim-ticket.sh` writes `./agent-test` only into a worktree it freshly claims, so the tree you added has none — run `<testCmd>` in it, per **A reused worktree may lack the runner** above."
5. **The stated mechanism is empirically false in this repo.** The refuter extracted the PR SHA into a bare, never-installed directory (`git archive | tar -x`; no `npm install`, no `node_modules`, **no `package.json` anywhere in the repo**) and ran the suite: `tests 1339 / pass 1339 / fail 0`, exit 0. A tree lacking installed dependencies cannot be the ordinary non-start cause here — there are none to lack.

[...]

Issue #988, opened 2026-08-28
PR #983 review: the suggestion band, checked


feigi:
Refutation records from the PR #983 review (ticket #387, `skills/fleet/scripts/release-ticket.sh`). Filed **closed**: these were checked and did not survive, so they are neither confirmed defects nor open questions. Kept as a durable record so a future run re-deriving either lands on the evidence instead of re-litigating it.

---

## 1. "The blocked-checkpoint receipt cannot be told from an aborted scan" — REFUTED, refutation upheld

Dimension: **silent-failure**.

Claim: a caller cannot distinguish a receipt from a *completed* precondition scan from one whose scan aborted partway, because the payload carries no partial-scan marker.

**Refuted, and the refutation was upheld on review.** The confirming probe captured only the exit code and `blockers` — it never read **stderr**, which is the channel carrying the discriminating verdict line.

Three channels discriminate the two states, not one:

1. **exit code** — 1 for a completed scan that found blockers, 2 for a die
2. **the `#<n> NOT released — nothing was touched` stderr line**, which the blocked checkpoint writes and `die` never does
3. **the appended cause sentence** in the die's own blockers array

A caller reading only stdout and the exit code sees less than the script actually tells it; that is a property of the probe, not of the script.

Not applied, and not filed as a defect.

---

## 2. "die's silent degrade when `jstr` fails is safe and deliberate — just document it" — REFUTED

Dimension: **types**.

Claim: when `jstr` fails inside `die`, the receipt is dropped but `set -e` does not abort, the original die reason still prints, and the script still exits 2 — a safe degrade matching pre-PR behaviour, needing only a comment (or an explicit `|| true`) to say so.

This was **mutually exclusive** with a sibling finding from the `silent-failure` dimension, which wanted the failed escape given a voice instead. One refuter was briefed with **both** claims and made to settle them against each other, rather than two independent refuters each approving their own side.

**Refuted on its central claim.** The "safe degrade" was measured over only one of the two ways `die`'s AND-OR list can fail. The other — the receipt `printf` itself failing on a healthy `jstr` (closed fd, EIO, a full filesystem on a redirect) — was measured with stdout closed:

| shell | shipped | `\|\| true` | the voice arm |
|---|---|---|---|
| /bin/sh (bash 3.2.57) | **exit 1, no prose** | exit 2, prose | exit 2, prose |
| /bin/dash | **exit 1, no prose** | exit 2, prose | exit 2, prose |
[...]

Issue #982, opened 2026-08-28
PR #972 review: the suggestion band, checked


feigi:
Refutation record for the PR #972 review (ticket #346). **Nothing here is work.** Filed closed so a future run re-deriving one of these lands on the measurement instead of re-litigating it — `ledger.mjs check` reads the tracker at `--state all`.

Every entry below was *checked* and did not survive. That is different from "torn": it is not that nobody is sure whether something is broken, it is that someone measured that nothing is.

---

## 1. `assert.notEqual(r.signal, "SIGTERM", …)` missing from the credential-helper test

**Dimensions: `types` (`unverified[6]`) and `comments` (`refuted[4]`) — the same observation from two specialists.** `refuted[4]` came back 1-1 and was ruled here.

The textual claim is true: that line is present at `inflight.test.mjs:2038` and `:2118` and absent from the credential-helper test. The implied gap is not.

Measured on darwin / node 26.7.0, `spawnSync` with a `timeout`:

| child command | `error` | `status` | `signal` |
|---|---|---|---|
| `sleep 30` | ETIMEDOUT | null | SIGTERM |
| `sh -c 'sleep 30' & exec sleep 1` | ETIMEDOUT | 0 | **null** |
| `sh -c 'sleep 30' &` | ETIMEDOUT | 0 | **null** |
| `sh -c 'sleep 30' & sleep 30` | ETIMEDOUT | null | SIGTERM |
| `sh -c 'sleep 30 >&1 2>&2' & exec sleep 0.5` | ETIMEDOUT | 0 | **null** |

`error` is ETIMEDOUT in **every** case; `signal` is null in three of five. So `assert.equal(r.error, undefined, …)` — which the credential-helper test *does* carry — catches every shape the missing line would, and the missing line would pass vacuously on the descendant-holds-the-pipes shape that is the actual failure mode here. The one case `error` misses (an external SIGTERM with no timeout expiry) is caught by `assert.equal(r.status, 2)`, present in all three tests.

Control: a child SIGTERM'd *not* by the timeout returns `{error: undefined, status: null, signal: "SIGTERM"}`.

**Ruling: leave all three tests alone.** Adding the line adds a provably vacuous assertion. Removing it from the two siblings is cosmetic churn on a green PR. If anyone touches this later, removal is the defensible direction, not addition.

---

## 2. The budget comment "never preempts a bound that would have produced git's own diagnostic"

**Dimension: `correctness`. Refuted 2-0 by the review.**

[...]

PR #1003, opened 2026-08-29
test(run-team): pin the verbatim dispatch blockquotes phase 2 carries to every member

### LINKED ISSUE #400: Every verbatim dispatch blockquote in run-team/SKILL.md is unpinned — deleting one leaves the suite fully green

feigi:
Closes #400

Sizing: light

## What

Phase 2 hands the controller a run of `>` blocks and says to carry each one verbatim into every implementer prompt. Those blocks are the only text a member ever sees, so a block that goes missing does not misinform the member — it silently stops governing it. This adds `skills/fleet/scripts/dispatch-block-pins-prose.test.mjs`, which anchors them.

## Partial-fix finding, re-derived against `origin/main`

Of the three blocks the ticket names:

- **Enumerate-the-class** — unpinned. No test file mentioned `enumerate every member of that class`.
- **Commit-incrementally** — unpinned. No test file mentioned `Commit incrementally`.
- **Issue-read** — pinned since the ticket was filed, and better than a location pin: `tracker-block-copy-prose.test.mjs` compares the block **whole** against `docs/agents/issue-tracker.md`, so deleting or gutting it on either side alone reds. One case that comparison cannot see: an identical gutting of **both** copies agrees with itself and passes. Measured, and this PR's pins catch it.

  Review found the first pass had left the block's **last rule** — read the issue *before touching code*; still undecided or needing human hands you do not have → bail, name the cause, do not implement — outside all four content pins, so it was still open to exactly that escape: deleted from both files, the suite's only red was `tracker-block-copy-prose.test.mjs`'s own `TAIL` fixture going stale, and doing what that red's message instructs took the tree to **1345/1345 with the rule gone from both files**. The sibling `Respec` sentence in the same paragraph *is* caught by a content pin, so the harness discriminates — this sentence was specifically the one left open. Now pinned to its timing and its trigger rather than to the word *bail*, so the inversion that matters (`→ implement anyway`) reds too. Replaying that escape against the final tree reds 5 tests, one of them this pin naming itself.

The ticket's "498/498" figure predates the tree; the suite runs 1345 tests here, 1339 before this PR.

## Shape

A positive regex over the whole member-prompt region is a vacuous pin — every word of every block is somewhere in it. Two things do the work instead:

1. **Each block gets its own slice**, bounded at both ends by the neighbouring block's opening words. A block deleted outright reds on its own anchor and names itself. Anchoring on a closing phrase instead would leave anything appended after it outside the slice.
2. **Each pinned fragment is bound to the content beside it.** Independent presence checks pin facts and never the relation between them — the measured case being `**step 6**` and `**step 7**` asserted separately, which stayed green through a swap of the two numbers. So a rule is pinned together with the command that carries it out, and an instruction together with the alternative it excludes.

No word blacklist: `doesNotMatch(/\bunless\b|\bexcept\b/i)` was tried on this question and failed all three controls, including reddening on a legitimate edit this repo is likely to make.

## Enumerate-and-declare

The class is *verbatim `>` blocks in phase 2's member prompt that no test anchors*. Built from the ticket's prose first, then from the region itself. **Covered** — the three named blocks, plus two siblings the ticket does not name and that were equally unpinned: the **worktree** block (`Do NOT create another worktree` and the `rev-parse` check that settles it) and the **re-derive** block (the very rule that produced the partial-fix finding above). Pinning three of five in one prompt would be the missed-sibling shape the enumerate-the-class block exists to prevent.

**Deliberately left** — the identity block and the two `next-ticket` step handoffs, already pinned in `member-prompt-prose.test.mjs`; the fix-applier dispatch prompt in the Reviewers section and the refuter block nested inside it, already pinned in `review-path-default.test.mjs` and outside this ticket.

## What this change can wrongly refuse

A pin is a constraint on future edits, so its false-positive class is the legitimate edits it now blocks: rewording a pinned span, reversing the order of the three halves of the enumerate block, renaming a block's opening words (which reds on the slice anchor with an "update this test" message rather than a bare mismatch), or moving a block out of phase 2. Rewording the unpinned connective prose between pinned spans stays green, and so does a reflow — the slices are `>`-stripped and whitespace-collapsed, and the accept-direction test derives its rewrap from the live block rather than quoting a line.

**One false positive of that class shipped in the first pass and review caught it.** `phrase()` escapes `*` as a regex metacharacter, so wherever a `**` seam sat inside a pinned span its exact position was literal test surface. Narrowing the bold on the re-derive sentence to `**Re-derive the ticket's claims** against \`origin/main\` before implementing —` — same words, same order, same meaning — reddened with a message claiming `origin/main` was no longer pinned, which was false. `flatten()` now strips `**` in the same pass as the `>` gutter, so a pin tolerates where emphasis sits the way it already tolerated where a line wraps. Only `**`: a single `*` and `_` are live characters in this region (`*.test.mjs` globs, `SNAKE_CASE` identifiers inside code spans), so stripping those would corrupt the text rather than normalize it.

## Mutation evidence

Every pin tested both ways, on a scratch copy of the tree, mutated by exact string replacement in node rather than `sed` — with a per-mutant control that fails loudly when the replacement does not change the file. That control fired once, on a mutant whose literal ignored a hard wrap; without it that mutant would have read green and been indistinguishable from an unpinned block.

- **Delete each of the five blocks** → red, each naming its own test.
- **Hollow each of the five** (opening kept, content gutted) → red.
- **Hollow the issue-read block in both files at once** → red on this PR's pin; the whole-block comparison stays green, which is the gap this pin closes.
- **Swaps**, every token still present: enumeration order (mechanism first, prose second); `is a bail, not a thing to implement` reversed; the Agent Brief precedence reversed; `origin/main` and the working tree exchanged → all red.
- **Benign rewords stay green**: unpinned connectives in the enumerate, commit and re-derive blocks; a rewrap at a different width; an unpinned clause in the issue-read block reworded in both copies; and a standing rule containing the word *unless* added to a block.

Re-run after review, on a `git archive` copy of the final commit, same per-mutant did-it-land control — which fired twice, on two mutants whose tracker-side literals ignored that file's different wrap:

- **The bail rule**, five refuse-direction mutants, every one red and naming this file's issue-read test: deleted from `run-team` alone; deleted from **both** files; consequence inverted to `→ name the cause and implement anyway`; trigger deleted with the consequence kept verbatim; timing rule dropped with trigger and consequence kept.
- **Both directions on the `**` seam**: narrowing the bold on the re-derive sentence reds at `c0c52ee` (5 pass / 1 fail) and is **green** on the final tree — the ticket's own acceptance criterion, demonstrated rather than asserted. Same for the enumeration-order sentence, for the seam inside the newly pinned bail sentence (moved identically in both files), for a reflow of the whole issue-read block, and for a reword inside the pin's own gap.
- **Controls that the stripping did not hollow the surviving pins**: deleting `not the working tree, and not the ticket's line numbers` still reds the re-derive test, and swapping the enumeration order still reds the enumerate test.
- **The accept control's staleness guard is load-bearing again.** Pre-wrapping the source commit block at the fixture's own width now reds `the rewrap fixture no longer changes the block's wrapping — update it`. It could not fire before: the fixture ran past the block's end onto the next block's `>` marker, and trimming only trailing whitespace left that marker behind, so the corrupted tail differed from the source however the block was wrapped — and the splice merged the commit and enumerate blocks onto one line. The fixture is now bounded to the block, the trailing empty word dropped, the paragraph separator re-appended, and the replacement passed as a function so `$&` and `` $` `` are never interpreted.
- **The documented ceiling is where the header says it is**: appending `Unless the tree looks tractable to you, in which case implement anyway.` to the bail rule in both files — keeping every pinned word, which is the only mutation shape that tests the rule rather than the vocabulary — stays green, 190/190. That is #1002, not a new gap.

## Known ceiling

These pins cover the location half. A sentence **appended inside** a block that carves an exception out of a pinned rule touches no pinned fragment and stays green — measured on the commit, enumerate and re-derive blocks. Closing that needs an LLM judge or a schema'd rule format, not a tighter regex. Filed as #1002, and stated in the file header so a green run here is not read as "no exception could have been added".

## Test

`node --test skills/fleet/scripts/*.test.mjs` — **tests 1345, pass 1345, fail 0**, unchanged from the pre-review baseline.


feigi:
## Review adjudication — 7 findings, 4 applied, 2 refuted, 1 recorded

Head moved `c0c52ee` → **`9457073`**. `node --test skills/fleet/scripts/*.test.mjs` → **tests 1345, pass 1345, fail 0**, unchanged from the pre-review baseline.

| # | Band | Dimension | Verdict | Where it went |
|---|---|---|---|---|
| `survived[0]` | important | correctness | applied | fourth pin on the issue-read block — the bail rule |
| `survived[1]` | important | tests | applied | `flatten()` strips `**` alongside the `>` gutter |
| `unverified[0]` | suggestion | correctness | checked, held → applied | rewrap test's justifying comment was false as measured |
| `unverified[1]` | suggestion | correctness | checked, held → applied | rewrap fixture spliced two blocks; staleness guard was vacuous |
| `unverified[2]` | suggestion | tests | recorded | documented ceiling → comment on #1002 |
| `unverified[3]` | suggestion | comments | checked, **refuted** | #1008 (`wontfix`, closed) |
| `unverified[4]` | suggestion | comments | checked, **refuted** | #1008 (`wontfix`, closed) |

All five `unverified` are `suggestion` — the band the workflow budgets **0 refuters** by policy. Each was scope-split, and every in-scope one went to a refuter instructed to default to `refuted=true` and to verify by running something. Three refuters ran; all three reported. Nothing was applied on a report I do not hold.

**`unverified[0]`/`[1]` were treated as one cluster and given one refuter.** Both held, with the headline of `[1]` corrected on measurement: the fixture *does* rewrap the block, but it also destroys the block boundary, merging the commit and enumerate blocks onto one line — and it made the guard `the rewrap fixture no longer changes the block's wrapping — update it` unable to ever fire, since the corrupted tail differed from the source however the block was wrapped. Remedy run before commit: file alone 6/6, full suite 1345/1345, and the guard now reds when the source block is already at the fixture's width.

**Both refutations are measurements, not judgements.** `[3]`: the header's "the blocks between those" is literally true — "those" includes the `next-ticket` step handoffs, which are the region's **last** block, so prior coverage is `{1, 3, 7}` and the four new blocks `{2, 4, 5, 6}` all sit strictly between two pinned ones. `[4]`: `member-prompt-prose.test.mjs`'s "two other test files carry their own copy" counts importable `section(source, …)` helpers and is still exactly 2 at this head; this PR adds none, and `git log -S` on the introducing commit `0d9c225` settles that the count was correct when written. Full evidence in #1008.

**Mutation evidence, both directions, on a `git archive` copy of `9457073`** — per-mutant did-it-land control kept, and it fired twice on tracker-side literals that ignored that file's different wrap:

- Bail rule, five refuse mutants (deleted from `run-team` alone; deleted from **both** files; consequence inverted; trigger deleted, consequence kept; timing rule dropped) → all red, each naming this file's issue-read test.
- The `**` seam, both directions: the emphasis move that reddened at `c0c52ee` (5 pass / 1 fail) is **green** on `9457073`. That is this ticket's own acceptance criterion — *a benign reword inside a block that preserves its meaning stays green* — demonstrated rather than asserted. Same for the enumeration-order sentence, a seam move inside the bail sentence in both files, a reflow of the whole issue-read block, and a reword inside the pin's own gap.
- Controls that the stripping did not hollow anything: deleting `not the working tree, and not the ticket's line numbers` still reds the re-derive test; swapping the enumeration order still reds the enumerate test.
- The documented ceiling is exactly where the header says: appending a carve-out that keeps every pinned word stays green, 190/190. Recorded on #1002.

**#1004 — this narrows it, does not close it.** The G1b escape (delete the bail rule from both files, repoint `TAIL` as its own red instructs) went **1345/1345 green** at `c0c52ee` and now reds 5 tests, one of them the new pin. But `tracker-block-copy-prose.test.mjs` is untouched and `TAIL` is still hard-coded, so the misleading-message defect #1004 names is intact. A derived `TAIL` would **not** have been a better fix for this finding: it fixes a message, and cannot pin content, since that test compares the two files against each other and an identical gutting of both passes there however `TAIL` is derived. Reasoning and measurements posted to #1004.

`ledger.mjs check "PR #1003 review: the suggestion band, checked"` → exit 0, `verdict: "tracker-hit"`, `found: false`; the hits are the same-shape records for other PRs (#1007, #998, #988, #982, #980), read before filing, none about #1003.

PR body reconciled against the final diff.

PR #1012, opened 2026-08-29
docs(fleet): a claim in an immutable body needs its settling command inline (#402)

### LINKED ISSUE #402: Commit bodies carry `Verified:` measurement claims with no settling command — `86/86 across both test files` was wrong on arrival and shipped twice
## What

Fleet commit bodies routinely close with a `Verified:` line asserting measured counts. Nothing requires the number to have been re-measured at the commit that ships it, and nothing makes it re-runnable afterwards. A commit body is immutable, so a wrong number there is permanent.

## Evidence

Found gating PR #399. Two of its commits (`75906a8`, `0bfdd2d` — post-rebase `f961cf2`, `acce6ee`) both close:

> Verified: 498/498 repo-wide, **86/86 across both test files that read** run-team/SKILL.md

Measured at HEAD:

| file | reads `run-team/SKILL.md`? | how | tests |
|---|---|---|---|
| `candidates.test.mjs` | yes | `join(import.meta.dirname, "..", "skills", "run-team", "SKILL.md")` | 50 |
| `fleet-tick-prose.test.mjs` | yes | `join(REPO, "skills", "fleet", "skills", "run-team", "SKILL.md")` | 9 |
| `implementer-model-tier.test.mjs` | yes | same form | 4 |
| `review-path-default.test.mjs` | yes | same form | 11 |
| `inflight.test.mjs` | **no** | one comment at `:701`, no read | — |
| | | **four readers** | **74** |

**Four files, not two. 74, not 86.** No grouping of these files yields 86. The `498/498` in the same sentence is correct and was actually run — so a measured number and a remembered number shipped in the same clause, under the same `Verified:` header, twice.

## Why it is worth a rule and not just a fix

`Verified:` is the highest-assurance construct available in a commit body. It is the thing a later reader trusts *instead of* re-deriving. A wrong number there is worse than no number, and it cannot be corrected in place — only retracted in a later commit, which is itself a fresh unverified historical claim.

This sits in a measured pattern on that PR: seven false claims across four commits and the PR body, and the generator is the same each time —

> **a claim whose settling command was never re-run at the commit that ships it.**

The converse held four for four on the same PR: every claim a settling command *was* run against was correct. This is not a care problem, it is a missing-mechanism problem.

## Suggested shapes (triage to pick)

1. **Require the settling command inline**, the way issue #48's own body does — every assertion accompanied by the command that produces it, so a reader can re-run rather than trust. Cheapest, no tooling.
2. **Ban bare counts in `Verified:` lines**, allowing only a suite total that CI independently re-derives (`498/498` would survive; `86/86 across both test files` would not).
3. **A pre-commit check** that re-runs any `N/M` claim it can parse. Most expensive, and probably over-built for the failure rate.

Option 1 also fixes the class of "N files that do X" claims, which is what actually broke here — the count was wrong *because the file set was wrong*, and no command was run to establish either.

## Related, not duplicate

- **#48** — three stale assertions in `docs/specs/2026-07-23-fleet-plugin-design.md`. Same family (a prose assertion that no longer holds) but a **mutable** artifact: it can be re-checked and fixed in place. #48's body is, incidentally, a good model for shape 1 above — it pairs every assertion with a runnable settling command.
- **#400** — verbatim dispatch blockquotes in `run-team/SKILL.md` are unpinned. Adjacent gap, different artifact.

## A trap for whoever implements this

Establishing "which test files read `run-team/SKILL.md`" by grep is unreliable and produced two wrong answers while writing this issue:

- `grep -ln "run-team/SKILL.md" *.test.mjs` → **2 files**. Three of the four build the path via `join(...)` and are invisible to a literal-string grep; both apparent hits were *comments*, not reads.
- Refining to the exact `join(REPO, "skills", "fleet", "skills", "run-team", "SKILL.md")` string → **3 files**, still wrong, because `candidates.test.mjs` uses a third spelling.

Only reading the four files settled it. Any check written for this must not be a literal-path grep.

feigi:
Sizing: light

Closes #402

## What

The correction-ticket discipline in `run-team/SKILL.md` scoped every rule to
prose **the diff restates**. A commit body is not in the diff, cannot be edited
after push, and carries the `Verified:` header a later reader trusts instead of
re-deriving — so none of the existing rules reached it.

Adds the rule at the widened scope the Agent Brief rules for (counts, line
numbers, SHAs, attributions, positional references), pinned to the **shipping**
commit rather than the motivating one, with the inline form and the reason
(immutability: a retraction is itself a fresh unverified historical claim). Adds
the prohibition on a literal-path grep as the settling command for "which files
read X", with the mutation method that replaces it.

## Where the guidance went, and how that was settled

`skills/fleet/skills/run-team/SKILL.md`, in the correction-ticket discipline
block — the paragraph run that already carries the settling-command rule, the
positional-reference ban and the anti-count rule, and already says to give them
to every implementer regardless of class.

Settled rather than assumed:

- `git grep -n 'Verified:' origin/main` matches only `pathVerified` identifiers
  in test and workflow files plus one prose *usage* in
  `run-team/references/ci-and-staleness.md`. No file described the construct.
- `grep -rn 'commit body\|commit message\|commit bodies' skills/ docs/` reaches
  `review-and-fix.md` (reviewer-side, PR body only) and
  `references/correction-tickets.md` (evidence, listing a commit-body defect as
  one of its classes). Neither states a rule an implementer writing a body reads.
- `next-ticket/SKILL.md` step 7 was the dispatch brief's candidate. Read: it is
  PR mechanics (rebase, push, create, label) and says nothing about claims. The
  discipline block is where the sibling rules live and where phase 2 relays them.
- `references/correction-tickets.md` holds the evidence the block cites, not the
  rules the implementer is handed. Left unchanged; the rule reaches nobody there.

## Applied from the review

**The rule now reaches the hand-over, not just the block.** Delivery here is a
controller copying an enumeration, not the block — nothing in this repo renders
the block into a member prompt verbatim, settled with
`grep -rn "correction-ticket discipline" skills/ agents/`, whose every hit
selects, annotates or pins the discipline and none of which renders it. Both
enumerations now name this rule: phase 0's parenthetical describing what phase 2
hands the implementer, and the give-them-to-every-implementer sentence in the
block. Phase 0's also regains the anti-count rule, which was absent from it
before this PR —
`git show 38d9d15:skills/fleet/skills/run-team/SKILL.md | sed -n '303,310p' | grep -in "count\|tally"`
exits 1.

**The unsettleable-claim sentence gained the ending the cross-repo citation
convention already sanctions.** As written it was a settle-or-drop binary, and
read literally it banned what that convention (pinned by
`cross-repo-citation-prose.test.mjs`) directs: asserting foreign evidence as
prose saying it cannot be settled from this repo. A foreign claim's settling
command cannot be written *here*, so the binary made the sanctioned form
unreachable. The sentence now carries all of the endings — settle inline, mark
unsettleable per the convention, or drop — and the convention text itself is
untouched, since it is pinned and is a separate blast radius.

**The `Verified:` rationale sentence is no longer unpinned.** It is the ticket's
own "why a rule and not just a fix", and it could be deleted from `SKILL.md`
with the entire suite green: the ranking span stopped before it and the
worked-instance span started after it, leaving it in a gap no assertion reached.
Now bound as an ordered span like its siblings. This also makes the `flatten()`
decision to leave single `*` unstripped load-bearing rather than merely
motivated, since the assertion matches on `*instead of*`.

## Verification

`node --test skills/fleet/scripts/*.test.mjs` from the worktree →
`tests 1363 / pass 1363 / fail 0`. Baseline before the review fixes was
`1361 / 1361 / 0`; the two added tests are the `Verified:` pin and the
hand-over-enumeration pin.

Worked instance in the guidance, re-settled at this commit rather than restated
from the issue:

    git log -1 --format=%b f961cf26b3ac2fb88f9a9fbeea6f62e6171c284f | grep -A2 'Verified:'
    git log -1 --format=%b acce6eec2d7ca2d94b47cd43756a261c26fa0d6b | grep -A2 'Verified:'
    gh pr view 399 --json commits --jq '.commits[].oid'

Both bodies carry `86/86 across both test files`; both commits are listed on
#399 and are ancestors of `origin/main`.

The file-set half was settled by the method the guidance now prescribes:
replacing `skills/fleet/skills/run-team/SKILL.md` with `MUTATED` and running the
suite reds a set of suites that a grep over the path spellings does not name,
including files that reach it through a local `read` helper.

## Pin

`skills/fleet/scripts/immutable-body-claim-prose.test.mjs`, house style from
`dispatch-block-pins-prose.test.mjs`: a slice bounded at both ends, fragments
bound to the content beside them rather than asserted alone, a `flatten()` that
strips the `>` gutter and `**` in one pass, and a rewrap accept control derived
from the live text. The hand-over pin takes a second bounded slice, since phase
0's enumeration is a delivery surface that drifts independently of the block.

Mutation-tested both ways in a copy of the tracked tree under a scratch
directory — never the worktree — with a did-it-land assertion on every literal
and the mutant diff printed. Every must-red mutant reds naming its own
assertion, and every benign control (reflow, bold-seam move) stays green. The
must-red set covers, for each pin added or changed here: deleting the pinned
sentence, inverting it, and removing exactly one element of an enumeration while
leaving the rest intact.

**Ceiling, measured not assumed.** Two shapes, and only one of them was already
known:

- *Appended clause* — the known half, disclosed in the file header from the
  first commit and tracked by #1002. Appending an explicit carve-out that keeps
  every pinned word — "a claim already settled once earlier in the branch needs
  no re-run at the shipping commit, and a `Verified:` line may restate it bare"
  — stays **green**. A presence pin cannot see an appended exception.
- *Mid-gap clause* — recorded nowhere before this PR. A meaning-changing clause
  inserted strictly between the two anchors of one ordered span is absorbed by
  that span's `.{0,N}?` wherever the gap still has headroom, and stays green.
  Measured on three spans; a fourth reds only because its gap has no headroom
  left, so the property is per-gap headroom rather than one every span has.

Tightening the gaps is not the fix, and not for the obvious reason. Reflow is
not what it costs — `flatten` collapses whitespace before matching, so the
endings gap narrowed to its exact content with zero headroom stayed **green**
through a hard re-wrap of the whole rule at 42 columns. What it costs is
word-level copy-editing: that same zero-headroom gap **red** on inserting the
single word `it` into "not one to assert bare". Narrowing the gaps therefore
trades a silent hole for false reds on meaning-preserving rewording, and a pin
that reds on one gets weakened or deleted by the next person to touch the prose.
Recorded in the file header, separating the known half from the new one, and
filed as a comment on #1002, which owns the class.


feigi:
**Review fixes applied at `f3c7626`.** `node --test skills/fleet/scripts/*.test.mjs`
from the worktree → `tests 1363 / pass 1363 / fail 0` (baseline `1361/1361/0`).

Applied:

- the `Verified:` rationale sentence, which was deletable with the whole suite
  green, is now bound as an ordered span;
- both hand-over enumerations name this rule, and phase 0's regains the
  anti-count rule that was absent from it before this PR;
- the unsettleable-claim sentence gained the marker ending the cross-repo
  citation convention sanctions, resolving a literal collision. The convention
  text is untouched.

Filed:

- #1014 — `PR #1012 review: the suggestion band, checked` (`wontfix`, closed):
  the PR-body-mutability finding refuted by measurement, the pin-discrimination
  finding recorded as a positive result, and the `#1002` citation confirmed
  accurate by `gh issue view 1002`.
- A comment on #1002 recording the **mid-gap** half of the pin ceiling. The
  appended-clause half was already known and already tracked there; only the
  mid-gap half is new.

PR #1013, opened 2026-08-29
fix(claim-ticket): judge the file branch's argument by where it resolves (#424)

### LINKED ISSUE #424: agent-test: a symlink to a FILE resolving inside a vendored tree is not refused — the file-branch sibling of #186
Deferred from **#186** / PR #422, declared out of scope by the implementer rather than folded in.

## The gap

#186 fixed the runner's **directory** branch: `agent-test <arg>` now judges the argument's *resolved* directory (`cd "$arg" && pwd -P`) instead of its spelling, so a symlink whose target resolves inside a vendored tree is refused.

The **file** branch got no equivalent. A symlink to a *file* whose target resolves inside a vendored tree does not reach `[ -d ]` at all — it falls into the else/file arm, which has separate validation machinery and no resolved-path check. So the same class of input, one `-d` away, still runs vendored code.

## Why it was left rather than fixed

The implementer's stated reasoning, recorded here so triage does not have to re-derive it:

- #186's Current/Desired-behavior narrative and all three of its named routes are about the directory branch's find-based filter.
- #186's own `Out of scope` points at "Arguments node discards silently — #100", which owns the file branch's validation machinery.
- Scope was explicitly bounded to the directory branch, so widening it in that PR would have been an unrequested scope increase.

That reasoning is sound for the PR. It leaves the class open, which is what this issue is for.

## Reachability

Same profile as the case #186 fixed: this repo has **zero tracked symlinks** (`git ls-tree -r HEAD | awk '$1=="120000"'` → 0), so it takes a deliberately constructed symlink to reach. It was not judged unreachable in #186 either — it was judged out of that PR's scope.

## The open question for triage

Whether this belongs to the file branch's own validation (i.e. folded into **#100**), or is a second application of #186's resolved-path rule to the file arm. That disposition is the decision this ticket needs; the defect itself is a direct read of the shipped code, not a measurement.

Filed by the `/fleet:run-team` controller from `impl-186`'s enumerate-and-declare list. Related: #186 (parent), #100, #123, #125 (closed).

feigi:
The runner's file branch judges a vendored argument by its own spelling and by
the logical directory that spelling names — which is what node's silent-discard
rule reads. A symlink whose name carries no `node_modules` but whose target is
vendored defeats every term there, and node does not discard that argument: it
runs it, so the suite's result comes to hang on third-party code passing.

That is #186's hazard rather than #100's, and it takes #186's remedy — judge
where the argument RESOLVES — applied to the file arm.

## Shape

Added **alongside** the existing check. #401's logical `pwd` and its `CDPATH=`
are untouched; the one edit to that block is the routing pattern above it,
bounded to a path segment (see below), which moves no verdict there.

The new check reads only arguments whose spelling does not name a
`node_modules` **path segment**. Those are exactly the ones the existing check
declines, so the two never see the same argument — which is what lets this one
resolve physically while #401's stays logical, without either overturning the
other.

Resolution is `realpath`, not `cd`+`pwd -P`: `cd` resolves symlinked *directory*
components, but a symlink whose own target is a *file* is this ticket's input and
`cd` never reaches it.

The match is anchored at the divergence from the runner's own location, reusing
the directory branch's `$shared` walk: a `node_modules` above that point is an
ancestor of the runner too and says nothing about the argument.

## What the review changed

Three ways the first shape of the guard reached a verdict nobody chose:

- **One input, two verdicts.** BSD `realpath` fails on a nonexistent final
  component and GNU's succeeds on it, so a quoted glob spelled through a
  symlinked vendored directory ran green on macOS and was refused on
  `ubuntu-latest`; a typo under such a directory reported *"resolves inside
  node_modules — not missing"* on CI, asserting a missing file was not missing.
  The file passed 77/77 under **both** semantics — nothing discriminated them.
  The question is now asked only of an argument that **exists**: nothing absent
  can be vendored, and both resolvers agree on everything that is there.
- **Absence is not a fallback.** With no `realpath` on PATH the guard disarmed
  whole — vendored file ran, exit 0, `pass 2`, zero bytes on stderr. It now
  refuses, as `release-ticket.sh` already does for the same reason (*"none is
  guaranteed to exist"*), and dropping `2>/dev/null` puts `realpath`'s own
  reason on stderr on the way out.
- **The skip was unbounded.** `*node_modules/*` also swallowed a directory
  whose name merely *ends* in the word, handing `vendor_node_modules/link.test.mjs`
  to the guard above as already-judged — whose own test is anchored at
  `$PWD/node_modules/` and never fired. A symlink under such a name to a
  vendored file was judged by neither and node ran it. Both arms now carry the
  segment bound, which keeps them disjoint.

The refusal also names the resolved target, so it states what was applied to
*this* path rather than a criterion the runner does not apply to every spelling
(a literal `t/node_modules/pkg/x.test.mjs` still runs — #401's ruling,
deliberately).

## Acceptance criteria

- [x] A symlink at the worktree root pointing at a file inside `node_modules` is refused.
- [x] A symlink pointing at a non-vendored file is still accepted and runs.
- [x] A worktree that itself lives under a `node_modules` path still works — file legs added to the existing ancestry row.
- [x] The existing file-branch checks are unchanged in behaviour: normalized-relative discard, absolute-spelling exemption, bracket-escaping, typo refusal.
- [x] #401's logical `pwd` is untouched.
- [x] Tests cover the symlink-to-vendored-file case and the non-vendored symlink control.
- [x] The verdict is the same on macOS and on the platform CI runs.
- [x] Every new pin mutation-tested both ways.
- [x] Full suite green.

## Verification — every figure re-run at `212c3c9`, the commit that ships it

Each command below was run at `212c3c9` (not at the commit that motivated it) and its output is quoted beside it. `<ship>` is a clean tree, `git archive 212c3c9 | tar -x -C <ship>`.

**Full suite, macOS 26.4, native BSD `realpath`:**
```
$ node --test skills/fleet/scripts/*.test.mjs        # from the worktree, HEAD=212c3c9, clean
ℹ tests 1358   ℹ pass 1358   ℹ fail 0   ℹ skipped 0   ℹ todo 0
```

**`claim-ticket.test.mjs`, macOS, native BSD `realpath`:**
```
$ cd <ship> && node --test skills/fleet/scripts/claim-ticket.test.mjs
ℹ tests 79   ℹ pass 79   ℹ fail 0
```

**`claim-ticket.test.mjs`, real Linux, real GNU coreutils 9.1 — the platform CI runs on:**
```
$ docker run --rm -v <ship>:/w -w /w node:22 node --test skills/fleet/scripts/claim-ticket.test.mjs
# tests 79   # pass 74   # fail 0   # cancelled 0   # skipped 5   # todo 0
```
**Read that count before trusting it:** `pass 74`, not 79. The five are *pre-existing* root-guarded permission tests — `# SKIP root reads every directory` / `# SKIP root searches every directory` — which skip because the container runs as root, and none of them touches `realpath`. Zero failures. All three new pins **ran** and passed there (`ok 30`, `ok 32`, and the `unresolvable argument` pin at `pass 1 / skipped 0`).

**The platform split, settled in both directions on real Linux** — the pins red at `973581b` and green at `212c3c9`:
```
$ docker run --rm -v <973581b tree>:/w -w /w node:22 \
    node --test --test-name-pattern="reported missing" skills/fleet/scripts/claim-ticket.test.mjs
not ok 1 - runner: a missing file under a symlinked vendored directory is reported missing
    'agent-test: dirlink/nope.test.mjs resolves inside node_modules — excluded from the run, not missing'
# tests 1   # pass 0   # fail 1

$ docker run --rm -v <973581b tree>:/w -w /w node:22 \
    node --test --test-name-pattern="unresolvable argument" skills/fleet/scripts/claim-ticket.test.mjs
not ok 1 - runner: an unresolvable argument refuses rather than running unchecked
    # pass 2
# tests 1   # pass 0   # fail 1
```
The `# pass 2` is the defect itself: the guard disarmed and the vendored file counted toward a green run.

**Why nothing in the suite caught this before.** At `973581b` the file is `77 tests / 77 pass / 0 fail` on macOS and `77 tests / 72 pass / 5 skipped / 0 fail` on real Linux (same five root skips). Zero failures under either `realpath`, so no row discriminated them — which is what let one input draw opposite verdicts unnoticed.

**Mutation matrix** — mutants cut from a tree byte-identical to `212c3c9` (verified: `git archive 212c3c9 | tar -x` then `diff -q` against the mutation source, both files identical), each run as `node --test skills/fleet/scripts/claim-ticket.test.mjs`, macOS:

| mutation | result |
|---|---|
| revert the loud `realpath` failure to `2>/dev/null` | `79 / pass 78 / fail 1` — ✖ *an unresolvable argument refuses rather than running unchecked* |
| unbound the skip back to `*node_modules/*)` | `79 / pass 78 / fail 1` — ✖ *a symlink to a vendored file refuses however it is spelled* |
| drop the resolved target from the refusal message | `79 / pass 78 / fail 1` — ✖ *a symlink to a vendored file refuses however it is spelled* |
| **negative control:** reword an unrelated guard's message | `79 / pass 75 / fail 4` — reds four *existing* tests, all three new pins stay green |

Each mutation applied by exact string replacement with an occurrence-count assertion and its diff printed, never `sed -i`.

**Shim honesty.** A GNU-semantics `realpath` shim on macOS was used during development and agrees with the container (`79/79` at `212c3c9`); it was validated by confirming a deliberately-wrong version reds this PR's own test (`tests 3 / pass 2 / fail 1`). Every figure above that names Linux comes from the real container, not the shim.

## Left open, deliberately

- A **quoted glob** whose matches resolve into a vendored tree. The runner never expands globs; that is the file branch's already-stated ceiling. GNU `realpath` refused one such spelling by accident, reading `*` as an ordinary character, and charged for it with the "not missing" misreport above — so this is now the same on both platforms rather than a platform split.
- A **hard link** to a vendored file. It has no resolution to follow, so no path-based guard can see it.

Sizing: light

Closes #424


feigi:
Review deferrals filed: #1016 (dedupe of the anchor and divergence walk across the directory and file branches, `ready-for-agent`, out of #424's scope) and #1017 (`PR #1013 review: the suggestion band, checked` — `wontfix`, closed, the durable record of the six entries a refuter checked and did not confirm, plus the three positive tests-dimension results and the two re-adjudicated `refuted` rulings).

feigi:
Correction to this PR's commit body, which is immutable and cannot be edited.

`212c3c9`'s message says *"The whole file passed 77/77 under both semantics"*. That figure is macOS-only. Re-run at the platform CI actually uses:

```
$ docker run --rm -v <973581b tree>:/w -w /w node:22 node --test skills/fleet/scripts/claim-ticket.test.mjs
# tests 77   # pass 72   # fail 0   # skipped 5
```

The five skips are pre-existing root-guarded permission tests (`# SKIP root reads every directory`); the container runs as root and none of them touches `realpath`. The substantive claim the sentence was making — that **nothing in the suite discriminated the two `realpath` semantics** — holds: zero failures under either. Only the literal `77/77`, read as a cross-platform figure, is wrong.

Every other figure in that commit body has been re-run at `212c3c9` and holds; the commands and their output are now inline in the PR description under **Verification**, including the real-Linux runs that settle the platform split in both directions.

Issue #730, opened 2026-08-20
reap.sh + 3 siblings: the dirty gate reads an rc-0 EMPTY `git status` as clean — with status.showUntrackedFiles=no it deletes untracked work and exits 0


feigi:
Found by the **PR #726** review (issue #470), `silent-failure` dimension, severity **critical**, **2-0 could-not-refute**. Both refuters reproduced it end-to-end with matched controls. Deferred from that PR rather than applied — see *Why not fixed in #726* below.

## The defect — this destroys uncommitted work and reports success

`reap.sh`'s dirty-worktree guard treats an **rc-0 empty** answer as "clean":

```sh
if ! status_out=$(git -C "$wt" status --porcelain 2>/dev/null); then
```

That fails closed only on a **non-zero** exit. But `git status --porcelain` honours `status.showUntrackedFiles`, so with `showUntrackedFiles = no` in the operator's gitconfig it returns **exit 0 with empty output** on a worktree holding untracked work. The follow-up `[ -n "$status_out" ]` dirty test then reads "clean".

Measured, unmodified script, non-fleet linked worktree on a merged-then-`[gone]` branch:

```
GIT_CONFIG_GLOBAL = "[status]\n showUntrackedFiles = no"

$ printf 'DRAFT\n' > $R/outside/notes.txt
$ git -C $R/outside status --porcelain
[]                                     <- rc 0, empty, no stderr

$ sh skills/fleet/scripts/reap.sh --apply
    REAPED feature/merged
{"applied":true,"reaped":["feature/merged"],"kept":[]}
rc=0
notes.txt: DESTROYED
branch:    DELETED
```

**Control, identical fixture with the setting absent:** `KEEP feature/merged — dirty worktree …`, file alive. Only the config differs.

git 2.50.1 (Apple Git-155).

## It also falsifies the script's own comment

`reap.sh:265` states: *"No `--force`, ever. It refuses on modified and untracked files."* In the run above `git worktree remove` did **not** refuse — it deleted the untracked file and cleared the registration. `git worktree remove` uses the same `wt_status` machinery the config silences, so the comment's guarantee is conditional on a setting nothing checks.

## Fleet worktrees are exposed identically

The `--ignored` probe is the only downstream guard, and it is skipped entirely for `*/.worktrees/*` paths — so the fleet exemption removes the one thing that might otherwise have looked. It runs the same silenced machinery in any case.

## This is the `git cherry` swallow, in a second place

[...]

PR #1015, opened 2026-08-29
fix(release-ticket): unset ambient GIT_DIR/GIT_WORK_TREE

### LINKED ISSUE #427: release-ticket.sh: an ambient GIT_WORK_TREE makes the new linkage die blame a healthy .git, and the suite is blind to it by construction
Deferred from PR #423 review (issue #135), where it was filed as `refuted[1]`. **Dimension: `silent-failure`.**

`needs-triage` because the **defect is verified but the approach is open** — two candidate fixes are named below with different blast radii, and closing it properly also requires a decision about the test harness.

## Read this first: the finding's FRAMING was refuted, the residue was not

The finding as submitted claimed the new guard "hard-fails a **HEALTHY** worktree". Two refuters split 1-1; the tie was ruled by the controller after reading both. **The framing is refuted, and this ticket is only about the residue.** Do not reopen the framing — it was measured, twice, in both directions:

Refuter 2 removed the coincidence refuter 1's probe depended on. In the original probe the blocker fired only because `.worktrees/` happened to be untracked in the main tree. With `.worktrees/` gitignored so the main tree reads clean:

- **BASELINE (pre-PR)** believed the worktree clean and **went ahead and MUTATED** — `release-ticket: #9 HALTED mid-release — nothing landed: git worktree remove refused …: fatal: '…/9-release-ticket' contains modified or untracked files, use --force to delete it`; worktree removed: false, branch deleted: false, **ticket left reading as taken.** It was saved only by `git worktree remove`'s own refusal.
- **PR HEAD** exited 2 having **touched nothing**; `precious.txt` present.

So this PR makes the env-poisoned case **strictly safer**. It is not a false refusal: the tree is healthy, but the *reading* of it is not, and **neither build can answer about `$wt` while `GIT_WORK_TREE` points elsewhere.** The finding's own author conceded HEAD is the safer build.

## The residue that is real

The die blames the wrong thing:

```
release-ticket: …/9-release-ticket's .git does not point at …/9-release-ticket — it resolves to …/w — so whether it holds uncommitted work is unknown
```

That `.git` was written by `git worktree add` in the same probe and is **intact**. The real cause is an inherited `GIT_WORK_TREE` outranking `git -C`. The message sends an operator to inspect a healthy file.

Scope, stated honestly: `release-ticket.sh` **never clears `GIT_DIR`/`GIT_WORK_TREE`** (confirmed: `grep -n 'GIT_DIR\|GIT_WORK_TREE' skills/fleet/scripts/release-ticket.sh` returns nothing). That is true of baseline too, so the exposure is **pre-existing in kind** — the pre-existing `git -C "$wt" status --porcelain` leaks on the same env var. Only the *misleading message* is new.

**Negative control, measured:** `GIT_DIR` **alone does not reproduce** — both builds rc=1 with the correct blocker, because the work tree then defaults to cwd. It takes `GIT_WORK_TREE`.

## The house already handles this by hand, three times

- `skills/fleet/scripts/ledger.mjs:355-356` — `delete gitEnv.GIT_DIR;` / `delete gitEnv.GIT_WORK_TREE;`, under the comment at :335 *"ambient GIT_DIR/GIT_WORK_TREE retargets BOTH the probe below and gh's …"*
- `skills/fleet/scripts/inflight.test.mjs:221-222` — same two deletes, under :217 *"Inherited git vars outrank `cwd`, so an ambient GIT_DIR silently retargets …"*; see also its :750-751.

So the hazard is known and handled everywhere except the script that acts on the answer.

## Why no test caught it, and why a test costs more than it looks

**The suite cannot see this by construction.** `skills/fleet/scripts/release-ticket.test.mjs:32-33` sets

```js
  GIT_DIR: undefined,
  GIT_WORK_TREE: undefined,
```

for **every** fixture. That is correct hygiene — it is what stops a suite run from a poisoned shell from going vacuous — but it means no fixture can currently exercise this path. **Adding a test here requires changing the fixture harness** to let one fixture opt back in, and that change needs care not to reintroduce the vacuity the scrub prevents. That is a real part of this ticket's cost, not an afterthought.

## The two candidate fixes (pick one at triage)

1. **Root cause:** `unset GIT_DIR GIT_WORK_TREE` near `set -eu` (`release-ticket.sh:26`) — what `ledger.mjs:355` does. This also hardens the **pre-existing** `status --porcelain` leak, not just the new guard. Wider blast radius: it changes behaviour for any caller that deliberately sets those vars, so it wants a moment's thought about whether any does.
2. **Message only:** widen the die to name the other suspect — `"…does not point at $wt — it resolves to $toplevel (an ambient GIT_DIR/GIT_WORK_TREE will do this too)"`. Stops the message misdirecting an operator; leaves the underlying exposure, including the pre-existing one.

These are not equivalent: 1 fixes the class, 2 fixes the sentence.

## Provenance

Deferred from PR #423 review — dimension `silent-failure`, severity `important`, submitted as refuted after a 1-1 refuter split that the controller ruled. Not applied in PR #423 because the framing it was filed under is wrong, the residue is pre-existing in kind, and the root-cause fix reaches well outside #135's scope.

feigi:
Closes #427

## What

`release-ticket.sh` never cleared `GIT_DIR`/`GIT_WORK_TREE`. An ambient
`GIT_WORK_TREE` outranks `-C`, so the linkage guard's
`git -C "$wt" rev-parse --show-toplevel` answered about the ambient target
instead of `$wt` and blamed a healthy `.git` for it — the message the issue
was filed over. The pre-existing `git -C "$wt" status --porcelain` reads the
same poisoned environment.

Fix: `unset GIT_DIR GIT_WORK_TREE` in the preamble.

Note on framing: the issue's own history records that an earlier version of
this finding (guard hard-fails a *healthy* worktree) was refuted at triage —
this PR does not reopen that, and does not touch the guard's logic, only the
environment it reads.

## Settling commands (re-run against this PR's head, self-contained)

Which variable reproduces the misdirection — a healthy worktree `wt`, no `-C`
needed to see it:
```
git init -q repo && cd repo && git commit -q --allow-empty -m i
git worktree add -q -b b wt
GIT_WORK_TREE=/tmp git -C wt rev-parse --show-toplevel   # -> /tmp, not wt: reproduces
GIT_DIR="$(pwd)/../other/.git" git -C wt rev-parse --show-toplevel  # (../other a real, unrelated repo) -> wt's own path: does NOT reproduce
```
`GIT_WORK_TREE` alone is what reproduces the *linkage guard's* misdirection;
`GIT_DIR` alone does not (the work tree falls back to discovery). That is why
the two fixtures are shaped differently: the `GIT_WORK_TREE` one drives the
guard, and the `GIT_DIR` one cannot, so it measures the damage instead. See
**Test** below.

Placement below `export LC_ALL=C` rather than immediately after `set -eu`:
```
grep -n 'const PROLOGUE' skills/fleet/scripts/locale-pin-prose.test.mjs
```
`PROLOGUE = /^(#.*|set -[eux]+|\s*)$/` — an `unset` line placed above the
locale pin fails that test on principle (any non-comment, non-blank,
non-`set -[eux]+` line there is treated as work the pin must sit above),
independent of what the line does.

## Bug class — what's covered

- `git -C "$wt" rev-parse --show-toplevel` (the guard #427 names) — fixed.
- `git -C "$wt" status --porcelain` (the pre-existing sibling call, same `-C
  "$wt"` shape) — fixed by the same unset; both read the same environment.
- Every other git call in the script has no `-C` at all (`git rev-parse
  --git-dir`, `--git-common-dir`, `worktree list --porcelain`, `rev-parse
  --verify`, `rev-list --count`, `git cherry`, `git ls-remote`, `worktree
  remove`, `branch -d`, `worktree prune`) — with no `-C` to even attempt
  insulating them, an ambient `GIT_DIR` retargets these more directly than it
  does the `-C` calls. The preamble unset is what reaches all of them in one
  place, which is the reason this PR takes the root-cause fix over a
  message-only one.

Settle the `-C` split at this head:
```
grep -nE '(^|\$\(|\| |\|\| |&& |! |^\s*)git ' skills/fleet/scripts/release-ticket.sh | grep -vE ':\s*#|die "|block "|printf|echo '
```
Only the two linkage-guard lines carry `-C`; every other executed git call
runs without one.

**Out of scope, filed rather than fixed:** the same class is still open in
every sibling fleet script — `worktree-audit.sh` run inside repo A under an
ambient `GIT_DIR` silently audits repo B at rc 0, and
`GIT_WORK_TREE=<repo root> git -C <worktree> status --porcelain` returns rc 0
with empty output, a silent "clean". #427 is scoped to `release-ticket.sh`, so
that is **#1020**, not this PR.

## Could this newly refuse anything?

Behavioral check (stronger than a text search — an unconditional `unset` is
exactly the kind of change a literal grep can't fully vouch for, since a
setter could be assembled indirectly): ran the full fleet scripts suite
against this change. Before this PR's placement fix, one test failed
(`locale-pin-prose.test.mjs`, addressed above and now green); no other
regression surfaced across the suite, meaning nothing test-covered in this
repo currently depends on an inherited `GIT_DIR`/`GIT_WORK_TREE` reaching this
script.

Supporting text search (necessarily incomplete for an indirectly-assembled
setter, so treated as corroboration, not proof) — production files only, since
the test files legitimately mention both vars:
```
grep -rn 'GIT_DIR\|GIT_WORK_TREE' skills/fleet --include='*.sh' --include='*.mjs' | grep -v '\.test\.mjs:'
```
At this head that returns only `ledger.mjs`, which *deletes* both vars from a
child's env rather than setting them, and `release-ticket.sh` itself (this
comment block and the `unset`). No production caller sets either var before
invoking `release-ticket.sh`.

## Test

`release-ticket.test.mjs`'s `ENV` scrubs `GIT_DIR`/`GIT_WORK_TREE` for every
fixture (correct hygiene — otherwise a poisoned suite run would go vacuous on
this and every other case). Both new fixtures opt back in through `release()`'s
own per-call `env` override, so only those two tests see an ambient var; the
shared `ENV` object is untouched and every other fixture stays scrubbed.

Two fixtures, because the two variables cannot be pinned the same way:

1. **`an ambient GIT_WORK_TREE does not make the linkage guard blame a healthy
   worktree (#427)`** — drives the guard's two `git -C "$wt"` calls directly.
2. **`an ambient GIT_DIR does not aim the release at another repository
   (#427)`** — `GIT_DIR` alone does not misdirect `-C "$wt"` at all, so this
   one cannot use the guard as its detector. It clones the fixture origin a
   second time, gives that clone the claim's branch name, and asserts both
   that this claim's artefacts are gone **and** that the other clone's branch
   survives. Under `unset GIT_WORK_TREE` alone the run reports
   `released: true` after deleting the branch name in the *other* repository
   while leaving this one's directory, worktree and branch all standing —
   cross-repository destruction reported as success. Measured: each of those
   two assertions reds on its own under that mutation; the pair is what says
   the damage is cross-repository rather than merely a missed release.

Fixture 1 also had to be fixed to fail by *assertion*. Pre-fix it failed with
`TypeError: Cannot read properties of null (reading 'blockers')` — `release()`
returns `json: null` when the script dies before printing a receipt, and the
fixture dropped `stderr` — so it could not tell the linkage-guard misdirection
from any other `die`. It now asserts the exit code first with `stderr` as the
message, which makes the same pre-fix run fail with an AssertionError carrying
the script's own refusal. That is the file's stated standard ("asserting the
blocker it emits, not merely that the script exited non-zero") applied to a
`die` that prints no receipt to assert against.

Mutation-tested both ways in the worktree after committing, restored with `cp`
(never `git checkout --`, which would have discarded uncommitted sibling work
in this shared worktree). Each half of the `unset` reds exactly its own
fixture and nothing else, and the behavior-preserving reorder stays green:
- `unset GIT_DIR` (dropping `GIT_WORK_TREE`) → `pass 104 / fail 1`, the single
  failure being fixture 1.
- `unset GIT_WORK_TREE` (dropping `GIT_DIR`) → fixture 2 reds by assertion,
  fixture 1 stays green.
- `unset GIT_WORK_TREE GIT_DIR` (tokens reordered, behavior-preserving) → both
  green.

A read-only variant of fixture 2 was also considered — assert the script
answers about the *cwd* repo with rc 2 while `GIT_DIR` points at the repo
holding the claim. It was built and measured against a family of mutants
(drop `GIT_DIR`; move the `unset` below the `git rev-parse --git-dir` guard;
move it below `git worktree list --porcelain`) and its kill set was
**identical** to the damage-shape fixture's in every case. It is not included:
a second test for the same mutant is one defect turned into two near-duplicate
fixtures.

## Suite

`node --test skills/fleet/scripts/*.test.mjs` at this head: `tests 1365 /
pass 1365 / fail 0`.

Sizing: light


feigi:
## Review deferrals filed

- **#1020** (`ready-for-agent`) — the ambient `GIT_DIR`/`GIT_WORK_TREE` class is still open in every sibling fleet script. Dimensions `correctness` + `silent-failure`; two findings filed as one issue because they are one class with two instances. Deferred as out of scope: #427 is scoped to `release-ticket.sh` and its brief says so. Cross-linked from #730, which shares the symptom but not the cause.
- **#1021** (`wontfix`, closed) — the suggestion-band findings that were checked and did **not** survive: both halves of the proposal to dedupe the `unset` comment in `release-ticket.sh` against the fixture comment in `release-ticket.test.mjs`. Closed on filing; it is a durable record, not queued work.

Applied in `1619d25` rather than deferred: the four `survived` findings (one defect seen from three dimensions, plus the comment-scope one) and the checked-and-survived pair about the fixture's pre-fix failure mode. Details in the reconciled PR body.

Issue #77, opened 2026-07-30
2026-07-23-fleet-plugin-design.md:367 understates the citation rot by ~88 lines — all nine run-team SKILL.md:NNN citations are dead, and :365's figures are wrong too


feigi:
## Problem

`docs/specs/2026-07-23-fleet-plugin-design.md` has a migration-debt table whose two rows about `run-team` carry figures that do not measure. Distinct from #48, which enumerates three stale assertions in the *same file* at `:208`, `:225`, `:428` — a different section and a different set of claims.

### `:367` — the stated offset is wrong by ~88 lines, and every citation is dead

The row reads:

> Every `run-team.md:NNN` citation in **this** spec (6 sites) | … stale twice over: wrong path, and **+2 lines** off. The command's 4-line frontmatter became a 6-line skill frontmatter, so a body line at N is now at N+2. The one that bites is the Plan 2 instruction citing `run-team` SKILL.md:423 for the settings.json prohibition — it resolves to nothing.

Take the row's own worked example. `:286` cites `run-team` SKILL.md:423 for the `settings.json` prohibition. That prohibition is not at `423 + 2`:

```
$ git show origin/main:skills/fleet/skills/run-team/SKILL.md | grep -n 'settings.json'
513:*this run produced*. No speculative polish. Never `settings.json`, permissions, or

$ git show origin/main:skills/fleet/skills/run-team/SKILL.md | sed -n '423p'
edge-triggered loop that only refills on completion stalls silently the moment the
```

**423 → 513 is +90**, on `origin/main`, independent of any open PR. So "a body line at N is now at N+2" understates the rot by roughly a factor of forty-five, and the row's stated *cause* — 4-line frontmatter becoming 6-line — does not explain it. Line `:423` holds unrelated text about the edge-triggered refill loop.

Two specialists independently resolved **all nine** `run-team SKILL.md:NNN` citations in this spec against both refs. **None of them pointed at their claimed content, at either ref.** The offsets are not uniform, so no single delta repairs them — each citation has to be resolved individually.

### `:365` — four figures, all wrong

The row reads:

> `docs/specs/2026-07-22-run-team-agent-fleet-design.md` | 4 dead `~/.claude/commands/…` paths and 7 bare `/run-team` references. **Not a dead document** — `run-team` SKILL.md:13 sends the reader to it as "Rationale" …

| claim | actual |
|---|---|
| 4 dead `~/.claude/commands/…` paths | **0** |
[...]

PR #1018, opened 2026-08-29
fix(inflight): state the `set -e` exemption at every `probe_X || :` site, and pin a probe body against it (#415)

### LINKED ISSUE #415: inflight.sh: `probe_X || :` exempts each probe's WHOLE body from `set -e`, so a future unguarded command fails silently — no live defect today
Deferred from PR #406 review (closes #96, head `865129e`; fix applied as `cc60e73`).

**dimension:** raised three times across three dimensions — `types` (`refuted[6]`), `correctness` (`unverified[1]`), `silent-failure` (`unverified[3]`). Filed once, deliberately: three reviewers independently re-derived the same latent hazard, and both halves of the answer are recorded below so a fourth does not have to.

## The mechanism (measured, real)

`skills/fleet/scripts/inflight.sh` sets `set -eu` at the top, then calls its three probes as `probe_pr || :` (`:196`), `probe_remote || :` (`:311`), `probe_local || :` (`:531`). POSIX exempts a function's execution from `-e` for as long as its own status is what an `||`/`&&`/`if` is testing — so `-e` is inert for the **entire body** of all three probes, not merely at the failure sites #96 converted to `add_unknown …; return 1`.

Reproduced by injecting a bare `false` plus a marker echo as the first two lines of `probe_local`'s body in a copy: the marker printed, the rest of `probe_local` ran to completion, `no local branch or worktree for #8` was reported as a definite absence, and the injected failure left no entry in `hits`, none in `unknown`, and no other trace. Independently confirmed as shell semantics — not a bash quirk — in `/bin/sh`, `bash`, `dash` and `zsh`.

## The direction is the unsafe one — measured, and this is why the hazard is not mere tidiness

A missing guard inside a probe body does **not** produce a safe abort. It produces a **wrong free** — the one answer this script exists to prevent, because a free verdict puts two agents on the same ticket while a taken verdict merely skips one.

Measured at `cc60e73`. Fixture: reachable bare origin with no matching branch, `gh` stubbed to return no PRs, **local branch `fix/77-slug` genuinely on disk**, and `awk` shimmed to fail only the local-branch filter program (selector `$0 ~`, so the worktree filter and counter still run). Two legs, differing only in whether that one filter carries its guard:

```
guarded (as shipped):
  EXIT=2  {"issue":77,"taken":false,"hits":[],"unknown":["local"],…}
  stderr: inflight: could not filter the local branches for #77

unguarded (guard deleted):
  EXIT=0  {"issue":77,"taken":false,"hits":[],"unknown":[],…}
  stderr:     no local branch or worktree for #77
```

`EXIT=0`, empty `unknown`, and a positive report of absence — for a ticket whose branch is sitting in `git branch --list`. The failure is silent *and* fail-open: nothing records an unknown, and the fall-through `else echo "no … for #$n"` branch renders the crash as a definite negative.

Note this is specific to **value-producing** commands. The cosmetic result count at `:181` is the counter-example and is correctly `|| raw="?"` rather than a guard — its failure cannot change the verdict, so it must not produce an unknown. The hazard is any *new* command inside a probe body that feeds a verdict field.

## Why nothing fails open today (also measured)

The bodies were audited comment-stripped, all three, at the reviewed head. Every command capable of a non-zero status carries its own explicit guard — `x=$(…) || { add_unknown …; return 1; }`, `if ! x=$(…); then … return 1; fi`, or `|| true`. The only unguarded statements are `echo … >&2`, `rm -f /tmp/.inflight.$$`, plain assignments, `[ … ]` used as a condition, and the `add_hit`/`add_unknown` string appends — none of which changes a verdict.

The #96 conversion was also 1:1: every line removed in the probe regions was already a `die` guard, so no command lost protection that it previously had. **This is a maintenance hazard for the next edit inside a probe, not a behaviour change in #96** — which is why the reviewer's `refuted=true` vote is recorded as accepted for the *defect* claim. But the severity is higher than "latent tidiness", per the measurement above: the failure mode a future missing guard opens is a wrong free, not a loud abort.

A calibration mutation confirms the fault-injection point is simply untested: unguarding `git for-each-ref` in a copy still passed 57/57.

## Why this is `needs-triage` and not `ready-for-agent`

Three materially different remedies were proposed and the review settled none of them:

1. **Comment only** — note at the `|| :` sites that `set -e` no longer applies inside a probe, so every new command needs its own guard, and that an unguarded one reads as a free ticket. (`cc60e73` partially did this: the verdict-printf comment now states the exemption explicitly, and the `probe_pr` count comment names it and records the measured silent-failure behaviour. The three `|| :` sites themselves still do not.)
2. **A completion sentinel** — initialise `ran_pr=""` etc. beside the evidence defaults, end each probe body with `ran_pr=1`, and after `probe_pr || :` add `[ -n "$ran_pr" ] || add_unknown "pr" "probe 1 stopped before it finished"`. One line per probe; converts the whole latent class into a recorded unknown rather than a silent free. **Interacts with `cc60e73`'s per-signal change** — a probe that stops early may now legitimately have already recorded a hit, so the sentinel must not retract it.
3. **Restructure away from `probe_X || :`** — capture status by another mechanism that is not `-e`-exempt. Largest change; re-opens the shape #96 deliberately chose.
4. **A per-probe mutation test** that injects a genuinely unguarded failure and asserts the run does not complete silently — i.e. asserts against the `EXIT=0`/`unknown:[]` payload measured above.

(2) adds a new seam and (3) changes architecture, so two implementers reading only this ticket would build materially different things. Pick one at triage.

**Sequencing:** remedies (2) and (3) both change what happens at the three `probe_X || :` wrappers, which is also where #416 proposes moving the probe-name recording. Decide #415 and #416 together, or sequence them; do not hand them to two implementers in parallel.

## Files

- `skills/fleet/scripts/inflight.sh:196`, `:311`, `:531` — the three wrappers
- probe bodies: `probe_pr` `:74`-`:195`, `probe_remote` `:199`-`:310`, `probe_local` `:334`-`:530`

feigi:
Closes #415

Sizing: light

Remedy 1 + remedy 4, per the ticket's ruling. Remedies 2 (completion sentinel)
and 3 (restructure away from `probe_X || :`) are not here.

## What the ticket asked for, and what it did not

`inflight.sh` sets `set -eu`, then runs each probe as `probe_X || :`. POSIX
exempts a function's whole execution from `-e` while its status is what an `||`
is testing, so `-e` does not reach a command run **directly** in a probe body.
Re-measured at this commit rather than taken from the ticket:

    $ for sh in /bin/sh /bin/dash /bin/bash; do printf '%s -> ' "$sh"; \
        $sh -c 'set -eu; p() { false; printf "body ran past the failure; "; }; p || :; printf "rc=%s\n" "$?"'; done
    /bin/sh -> body ran past the failure; rc=0
    /bin/dash -> body ran past the failure; rc=0
    /bin/bash -> body ran past the failure; rc=0

**That result does not carry over to a command substitution nested in the
body, and the wrappers no longer say it does.** A substitution is a subshell,
and only bash extends the exemption into one. Measured at this commit, in the
`v=$( … ) || rc=$?` shape `probe_remote`'s own body is built around:

    $ for sh in /bin/sh /bin/dash /bin/bash; do printf '%s -> ' "$sh"; \
        $sh -c 'set -eu; p() { rc=0; v=$( false; printf "sub ran past the failure"; exit 7 ) || rc=$?; printf "v=[%s] rc=%s\n" "$v" "$rc"; }; p || :'; done
    /bin/sh -> v=[] rc=1
    /bin/dash -> v=[] rc=1
    /bin/bash -> v=[sub ran past the failure] rc=7

Under `/bin/sh` and `/bin/dash` the added command is skipped, the capture
truncates, and the deliberate `exit 7` is replaced by the failing command's
status — so it never reaches the guard the rule tells the editor to write. The
advice the wrappers give is unchanged and was always right; only the stated
reason moved, and it is now scoped to the shape that was actually measured.

`-u` is exempted at neither level and still aborts — but that abort is not the
loud backstop the bare clause reads as, so the wrappers now say where it lands.
Measured on this script, `echo "$UNSET_VAR_415" >/dev/null` injected as
`probe_local`'s first statement, run against a repo carrying `fix/77-slug`:

    unmutated   rc=1 with the full payload, under all three shells
    /bin/sh     rc=0, stdout empty
    /bin/bash   rc=0, stdout empty
    /bin/dash   rc=2, stdout empty

The EXIT trap's last command succeeds and resets the shell's status, so
wherever `/bin/sh` is bash the abort lands on rc 0 — this script's **free**
verdict, with no payload — which is the wrong free #415 exists to prevent,
advertised as a safety net. `/bin/dash` differs and reaches rc 2.

**No live defect, as the ticket states.** Nothing here adds a guard to a
command, changes a verdict, or touches a probe's behaviour. The diff is a
comment at three sites plus two test files.

## Re-derived against `origin/main` (d4a6276)

The ticket's line numbers are stale — the wrappers are not at `:196`/`:311`/`:531`.
The ticket's calibration figure is stale too. Both re-measured here instead.

The premise re-derived: **the suite was blind to exactly the mode this ticket
is about.** Removing `git for-each-ref`'s guard from probe 3 — what an edit
that simply forgets a guard looks like — left the entire inflight file green
before this PR. It reds now.

## Enumerate and declare

Value-producing commands inside a probe body, and whether removing the guard
is already caught. Measured by applying each mutant to a scratch copy and
running the file:

| probe | command | guard-removal caught? |
|---|---|---|
| 1 | `mktemp` (gh stderr capture) | yes, already |
| 1 | `gh issue view` | yes, already |
| 1 | `gh pr list` | yes, already |
| 1 | python3 PR filter | **yes — measured, mutant reds** |
| 1 | python3 tally `\|\| raw="?"` | cosmetic by design; must NOT record an unknown |
| 2 | `git ls-remote` | yes, already |
| 2 | awk remote filter | **yes — measured, mutant reds** |
| 3 | `git rev-parse --git-common-dir` | **no — measured, mutant survives** |
| 3 | refs dir `-r`/`-x` pair | yes, already |
| 3 | `find` over refs subdirectories | **no — measured, mutant survives** |
| 3 | `git for-each-ref` | **was NO — measured; this PR closes it** |
| 3 | awk local-branch filter | yes, already |
| 3 | `mktemp` (worktree listing) | yes, already |
| 3 | `git worktree list --porcelain -z` | **no — measured, mutant survives** |
| 3 | registry count / recount, worktree filter, worktree counter | yes, already |

**Covered here:** the comment at all three wrapper sites — no sibling missed,
which is the shape the ticket calls out — and the behavioural pin at probe 3's
`git for-each-ref`.

**Deliberately left:** probe 3's `git rev-parse --git-common-dir`, its `find`
guard, and `git worktree list`. Each measured as surviving, not assumed. They
are per-BRANCH coverage, which `inflight.test.mjs`'s own comment scopes to a
separate issue, and the ruling here is a per-PROBE mutation test: every probe
now has at least one guard-removal-sensitive pin. Sweeping them in would take
that other ticket's scope, which this ticket's Out of scope forbids for its
named neighbours.

The new case asserts `hits: []` alongside exit 2. That is the answer the run
legitimately produces — nothing is established when the shim fires, because
probe 3's branch half commits its hit *after* this lookup. So it does **not**
forbid a probe stopping early on a hit already recorded, which is the case the
ticket names against remedy 2.

It would red on a legitimate change in two places, both loudly: if probe 3
stopped reading branches through `for-each-ref`, and if the guard's message
were reworded.

**The injection point's assumption.** The shim selects on the git subcommand
as `$1`, not on an invocation count — this suite's own helpers shell out to git
through the same PATH, so a counted shim would fire on whichever call happened
to be nth. That assumes `for-each-ref` stays `$1`: a future `git -c x=y
for-each-ref` moves it, the shim stops firing, and the run goes free — which
this case then reds on as exit 0, not silently. A reordering that moved probe
3's branch half after its worktree half would make `hits` legitimately
non-empty and red the case too, again loudly.

Applied by python string replacement asserting the literal occurs exactly once,
on scratch copies; the worktree was never mutated and each mutant diff was
printed and read.

    -if ! refs=$(git for-each-ref --format='%(refname:short)' refs/heads); then
    -  add_unknown "local" "git for-each-ref failed, so whether #$n has a local branch is unknown"
    -  return 1
    -fi
    +refs=$(git for-each-ref --format='%(refname:short)' refs/heads)

    GUARDED (as shipped) -> passes
      EXIT=2  {"issue":77,"taken":false,"hits":[],"unknown":["local"],…}
      stderr: inflight: git for-each-ref failed, so whether #77 has a local branch is unknown

    UNGUARDED (guard deleted) -> fails, `0 !== 2`
      EXIT=0  {"issue":77,"taken":false,"hits":[],"unknown":[],…}
      stderr:     no local branch or worktree for #77

which is the payload the ticket measured, reproduced rather than cited.

The prose pin was mutation-tested three ways: the rule deleted at
`probe_remote` only reds that probe's cases and no others; `probe_local`'s
block keeping `probe_pr`'s key reds exactly the key case; and a negative
control re-wrapping `probe_pr`'s block at a narrower width stays green, so the
pin holds the rule and not the line breaks.

## Shell semantics

The new case spawns `sh`. CI's ubuntu-latest resolves that to dash while this
worktree resolves it to Apple's bash in sh mode, so a pin measured on one shell
can be vacuous on the other. Both legs re-measured with `sh` resolving to
`/bin/dash`: byte-identical payloads, same pass and same fail. `sh -n` and
`dash -n` both accept the edited script.

That same asymmetry is why the `-u` landing spot above is stated per shell
binary rather than per shebang: `#!/bin/sh` gets rc 0 on this host and rc 2 on
a host where `/bin/sh` is dash. Writing it as one number would have been the
vacuous-pin shape one level up, in prose.

## Review fixes applied on top

Two review findings survived adversarial verification and are fixed here, in
the same commit as the pin that was locking the overclaim in:

- the three wrapper blocks asserted the `-e` exemption unconditionally and
  certified it "measured under /bin/sh, /bin/dash and /bin/bash". The
  measurement was real but taken on the adjacent shape. Narrowed at every site,
  with the substitution carve-out and the `-u` landing spot each given their own
  assertion so the sentence cannot be widened back.
- the pin covered only the `add_unknown "<key>"` half of the guard shape; the
  `; return 1; }` half is now pinned in the same span (rewriting all three
  wrappers to `|| add_unknown "K" "…"` previously left the prose suite fully
  green). And the unsafe-direction test used three independent `phrase()`
  matches over one sentence, so an exception clause spliced between two of them
  stayed green — now one contiguous span, the form
  `finisher-pin-race-prose.test.mjs` already uses for this class.

Each was mutation-tested both directions before applying: red on the defect it
names, green on a pristine tree, on a 66-column reflow, and on a benign reword
of unpinned prose in the same block. One-site-only discrimination re-confirmed
after the change — the rule deleted at `probe_remote` alone reds only
`probe_remote`'s cases.

The remaining `simplify` suggestions on the new test file were checked by
measurement and refuted; they are recorded in the review's `wontfix` record
issue, not applied.

## Tests

    ℹ tests 1377
    ℹ pass 1377
    ℹ fail 0


feigi:
Review adjudication filed: **#1022** — `PR #1018 review: the suggestion band, checked` (`wontfix`, closed on filing).

Applied here in `eb26d31`, not deferred: the two `survived` correctness findings (the `set -e` overclaim at all three wrapper sites, and the `-u` backstop parenthetical) plus the two `suggestion` findings that survived adversarial verification (the unpinned `; return 1` half of the guard shape, and the three-independent-`phrase()` splice gap). The pin moved in the same commit, since it was locking the overclaim in.

Recorded in #1022, checked and not applied: `refuted[0]` (`types` — prose pins do not behaviourally enforce the rule; a self-disclosed ceiling, 1-1 refuter split, refutation accepted, with the demonstration preserved) and the three `simplify` suggestions on the new test file, each refuted by measurement.

Related: **#1019** already tracks the three probe-3 guards `impl-415` deliberately left.

PR #1023, opened 2026-08-29
fix(no-undo-audit): render worktree/branch as JSON null, never retract an established audit (#431)

### LINKED ISSUE #431: no-undo-audit.sh: the escape chain's `|| die` retracts an already-established audit over a formatter failure — #120's class, unfixed here
Deferred from PR #425 review (which fixed the same class in `inflight.sh`).

**dimension:** correctness · **severity:** suggestion · **file:line:** `skills/fleet/scripts/no-undo-audit.sh:421`

## The finding

`no-undo-audit.sh` still retracts an established verdict over a formatter failure — verbatim the shape #120 rules out for `inflight.sh`:

```sh
wt_j=$(jstr "$wt") && wt_rw=$(jrewritten "$wt") \
  && branch_j=$(jstr "$branch") && branch_rw=$(jrewritten "$branch") \
  || die "could not escape the audit fields for $branch"
printf '{"worktree":"%s",...,"clean":%s,"stash":%s,"conflicts":[%s],"atRisk":[%s]}\n' ...
```

`$clean`, `$stash`, `$conflicts` and `$atRisk` are all established by the time this runs. A `sed`/`tr` outage there converts a finished audit into "unanswerable" — the rationale `inflight.sh:25-29` now spells out for the same situation.

Confirmed against the tree while applying #425 (read directly, not inferred): the block is exactly as quoted.

`release-ticket.sh`'s copy is **not** the same shape and correctly needs no change — its escape at `:219-221` runs *before* any mutation (`git worktree remove` at `:550`, `gh issue edit` at `:571`), so its `die` aborts before anything is established.

Out of #120's scope as filed. #120 is `inflight.sh`; widening PR #425 into a second script would edit a file that ticket has no business in.

The defect is verified, but the **approach is open**, and two competent implementers would build materially different things:

1. **Payload schema.** `inflight.sh`'s answer was to emit JSON `null` for the unrenderable field. Here `"worktree"` and `"branch"` are currently always strings, and the consumer is a runbook step that hands the path to `git diff -- <path>`. Whether `null` is right — versus omitting the field, versus keeping `die` — is a schema decision this review did not make.
2. **It collides with open #190.** #190 (`ready-for-agent`, still open) targets this same printf block, and the shape quoted above *is* #190's proposed fix — assign-first-then-`||`-die. So #190's remedy is what introduced the retraction. Someone has to rule whether this supersedes #190, folds into it, or is a follow-on.

Not a duplicate of #119 (unescaped JSON + `jstr` duplication; it explicitly routes `no-undo-audit.sh` elsewhere) and not of #190 (whose mechanism is `set -e` blindness in a `printf` argument list, a different failure).

## Also noted, unfiled

`jstr()` now has three copies — `inflight.sh:595`, `release-ticket.sh:186`, `no-undo-audit.sh:46` — and only `inflight.sh`'s carries the empty-input short-circuit added by #120, so the three have silently diverged. That divergence is #119's territory; recorded here so it is not rediscovered.

Deferred from PR #425 review.

feigi:
Closes #431

Sizing: light

## What changed

`no-undo-audit.sh`'s escape chain for `worktree` and `branch` ended in `|| die`. Every field the payload carries is established by the time it runs, so a `tr` that had gone missing there converted a finished audit into exit 2 with no payload — discarding `clean`, `stash`, `conflicts[]` and `atRisk[]` over the formatting of two values the caller typed. Each field renders independently now and reports JSON `null` when it cannot, the answer #120 shipped for this same class in `inflight.sh`.

## Enumerating the class, and what is deliberately left

The class is *established findings retracted by a downstream formatting failure*. This script has three escape sites that `|| die`, not the one the ticket names, and all three run after `clean` and `stash` are measured:

    $ grep -n 'could not escape' skills/fleet/scripts/no-undo-audit.sh
    485:  || die "could not escape the conflicting paths for $branch"
    538:  || die "could not escape the at-risk commits for $branch"

**Covered:** the `worktree`/`branch` escape. Both are echoes of argv — the caller supplied them and still holds them — so `null` costs the operator nothing it cannot recover, and every finding still reaches it.

**Deliberately left:** the conflicting-path and at-risk escapes, which keep their `die`. They render FINDINGS obtainable nowhere else, and the runbook hands `conflicts[]` to `git diff -- <path>`, where an absent list reads as "nothing to prove" rather than "unproven". That reading is already pinned deliberately, and this PR leaves those pins untouched:

    $ node --test skills/fleet/scripts/no-undo-audit.test.mjs 2>&1 | grep 'jarr that\|jarr_rewritten that'
    ✔ a jarr that cannot escape the conflicts is exit 2 with a cause, never the refusal that means dirty (593.234875ms)
    ✔ a jarr that cannot escape the at-risk commits is exit 2 with its own cause (712.898792ms)
    ✔ a jarr_rewritten that cannot answer is exit 2 too — the `&&` chain covers both operands (585.412417ms)

**No field in this payload is legitimately unestablished** at the escape chain: `clean` and `stash` are measured above it, both arrays are already rendered, and `worktree`/`branch` are argv. The ticket's claim holds as written.

`release-ticket.sh`'s copy needs no change, confirmed rather than inherited. The ticket's line numbers have drifted, so the ordering is settled by position in the current file. Fixed-string matching, because `$wt` and `$issue` in a BRE make `$` an end-of-line anchor and the pattern then matches nothing while still exiting cleanly:

    $ grep -nF -e 'could not escape the receipt fields' -e 'git worktree remove "$wt"' -e 'gh issue edit "$issue" --remove-label' skills/fleet/scripts/release-ticket.sh
    387:  || die "could not escape the receipt fields for #$issue"
    948:    if ! err=$(git worktree remove "$wt" 2>&1); then
    969:    gh issue edit "$issue" --remove-label in-progress >/dev/null ||

The escape is at top level, above the first function definition that follows it:

    $ grep -n '^[a-z_]*() {' skills/fleet/scripts/release-ticket.sh | head -3
    134:die() {
    418:halt() {
    456:block() {

So it runs before either mutation, and its `die` aborts before anything is established — the ticket's reasoning holds even though its citations no longer do.

Replacing `die` with a `null` emission turns an abort into a completed run: a caller that previously saw exit 2 now sees the verdict's own code — 0 on a clean worktree, 1 on a dirty one. The one caller that gates on the status is the no-undo runbook, which reads 2 as "ask a human". That reading is now narrower, and both carriers of the contract say so; the audit's actual gate (`clean`) and its findings are unaffected, so proceeding on 0 is the correct verdict rather than a relaxed one.

Because that is the false-reassurance direction, the runbook and the design spec both rule out the dangerous reading by name — a `null` path is "this run could not render the path you passed in", never "there is no worktree", and never a path to paste into a pathspec.

## The suite could not see this mode before

No test drove an escape failure through `jstr`/`jrewritten` for these two fields; the existing `withBrokenEscaper` cases all target the arrays. Two cases now drive it, and a third pins the prose.

The injection targets `tr`, never `sed`, and the pipeline shape was measured rather than inherited from the ticket. `jstr`'s last command is a pipeline closed by `tr`, so the function's status is that tool's:

    $ sed -n '107,113p' skills/fleet/scripts/json.sh
      json_esc=$(printf '%s' "$1" \
        | LC_ALL=C sed -e ':a' -e '$!N' -e '$!ba' \
            -e 's/\\/\\\\/g' -e 's/"/\\"/g' \
            -e "s/$(printf '\010')/\\\\b/g" -e 's/\t/\\t/g' -e 's/\n/\\n/g' \
            -e "s/$(printf '\014')/\\\\f/g" -e 's/\r/\\r/g') || return 1
      printf '%s' "$json_esc" | LC_ALL=C tr '\001-\007\013\016-\037' ' '

Worth recording for whoever reads the ticket next: the premise that a broken `sed` is *masked* was true before #119 and is not true now — that `|| return 1` reads `sed`'s own status. `tr` is still the right target because it closes the function under either regime.

## Shell semantics, measured

`add_field` reads a failing escaper through `if ! ev=$(jstr "$2")`. Measured in that exact shape — `set -eu`, a function ending in a pipeline closed by a failing `tr`, and `jrewritten`'s always-zero AND-OR last line — under macOS `/bin/sh` (bash 3.2.57), `/bin/dash` and bash 5.3.15. All three agree on all three arms: a failing `jstr` yields `null`/`null` naming `jstr`, a failing `jrewritten` yields `null`/`null` naming `jrewritten`, a healthy run yields the quoted value with `false`, and every shell runs on to the payload at exit 0 rather than aborting under `set -e`.

Mutants applied to a `git archive` copy of this head under a scratch directory, never to the worktree; each replacement asserts its literal was found **exactly once** before writing, and that assertion fired once during this run on a badly-escaped literal of mine, which is the control working.

    RED    revert-to-die                          (only the two new escape cases)
    RED    null-becomes-empty-string              (only the two new escape cases)
    RED    rewritten-flag-keeps-its-stale-value   (only the two new escape cases)
    RED    escaper-name-dropped-from-message      (only the two new escape cases)
    RED    runbook-drops-the-not-an-absence-rule  (only the prose pin)
    RED    spec-drops-what-null-means             (only the prose pin)

No mutant survived, and each redded only the cases meant to catch it. The semantic revert:

    @@ -601,8 +601,7 @@
         ev="\"$ev\""
       fi
       if [ -n "$why" ]; then
    -    ev=null
    -    rw=null
    +    die "could not escape the audit fields for $branch"
         printf '%s: could not render the %s for %s as JSON (%s) — reported as null\n' \
           "$NAME" "$1" "$branch" "$why" >&2
       fi

The control direction — a healthy run's payload is unchanged — is measured, not asserted. This head's script and `origin/main`'s run against one fixture carrying a `"` in the branch, a space in the conflicting path, a stash entry, conflicts and atRisk:

    $ cmp old.h.out new.h.out && echo "stdout BYTE-IDENTICAL"
    stdout BYTE-IDENTICAL
    $ cmp old.h.err new.h.err && echo "stderr BYTE-IDENTICAL"
    stderr BYTE-IDENTICAL
    $ shasum -a 256 old.h.out new.h.out
    7d257d9769b1c218a272e63c1073d86a409187d9b11848f713629b4fbce8e189  old.h.out
    7d257d9769b1c218a272e63c1073d86a409187d9b11848f713629b4fbce8e189  new.h.out

Byte-identical on stdout and stderr for the clean branch (exit 0) and the refusal branch (exit 1) alike.

## Suite

    $ ./agent-test 2>&1 | grep -E '^. (tests|pass|fail)'
    ℹ tests 1372
    ℹ pass 1372
    ℹ fail 0
    $ git ls-files '*.sh' | xargs shellcheck -x -S warning && echo clean
    clean


feigi:
**Review adjudication — nothing applied, HEAD unchanged at `6fcf73c`.**

All 6 dimensions ran (`dimensionsUnrun: []`): 0 survived, 1 refuted, 7 unverified. The 7 `suggestion`-band findings arrive unchecked (the workflow budgets that band 0 refuters), so each in-scope one was checked here with a refuter briefed to refute and to verify by running something. Four refuters, clustered by question:

| finding(s) | dimension | outcome |
|---|---|---|
| `refuted[0]` | silent-failure | refuted 2-0 by the review; residual already filed as #1024 |
| `unverified[0]` + `[2]` | correctness, tests | refuted — the two existing tests plus six healthy-path exit-1 pins already cover it; the proposed test's only unique kill is a mutant nobody would write |
| `unverified[1]` | correctness | refuted + out of scope — the obvious shared-helper shape reinstates #431's own defect, and the full suite is green with that regression in it |
| `unverified[4]` | types | refuted — the invariant is structurally enforced and test-pinned; the class is a standing measured refusal (#307) |
| `unverified[5]` + `[6]` | simplify | `[5]` behaviour-identical but falsifies a comment this PR just shipped; `[6]` survived and was held — aesthetic only, and marginally less forgiving than what ships |
| `unverified[3]` | comments | not checked → deferred and filed |

**Filed:** #1025 (`ready-for-agent`, open) — the null-field docs call a null `branch` "the path you passed in"; `branch` is a ref name, and the phrase is pinned verbatim in three places. #1026 (`wontfix`, closed) — the durable per-PR record with every measurement.

No commit was manufactured to avoid the no-op. The PR body still matches the diff, which is unchanged.

Issue #339, opened 2026-08-11
candidates.mjs: a bold label alone on its line, with refs in the list below it, yields no dependencies


feigi:
**Deferred from PR #331 review** (issue #58). Dimension: `correctness`. Severity: suggestion.

**Finding:** the two collection passes are asymmetric — a *heading* gets multi-line section scope, a bold/plain *label* does not. Measured, identical under jq 1.7.1 and gojq 0.12.19:

```
"**Blocked by:**\n\n- #12\n- #13\n"  -> []
"Blocked by:\n\n- #12\n"             -> []
```

while the handled neighbours give `"## Blocked by\n\n- #12\n- #13\n" -> [12,13]` and `"**Blocked by:** #12\n" -> [12]`.

**Partly refuted, and not applied.** The finding's stated worry is that `candidates.mjs`'s comment claims `d` covers *"every body form to-tickets publishes"*. That claim was checked directly against `~/.claude/skills/to-tickets/SKILL.md` (gitignored, so invisible to the review's snapshot): the local-file template writes `**Blocked by:** <numbers/titles>` on ONE line (`:75`) and the issue template writes `## Blocked by` + a list (`:99-101`). Both forms are covered and tested. The comment is accurate; this shape is not a form `to-tickets` publishes.

**Still worth a decision** because a human writing an issue by hand plausibly writes exactly this. It is under-fire (a blocked ticket is admitted), not the over-fire class PR #331 just closed.

**Why needs-triage:** the approach is open, and the obvious fix is the one PR #331 deliberately rejected. Carrying a "label seen, refs pending" flag into the following lines re-opens the unclosed-section sweep that invented phantom blockers — the exact defect that PR fixed by restricting in-section collection to list-item lines. A narrower rule (a bare label line arms only the immediately-following list block, closing on the first non-list line) might work, but nobody has measured it.

feigi:
> *This was generated by AI during triage.*

**Closing as `wontfix`.** The measurement holds — a bold label alone on its line with refs in a list below yields nothing, identically under both engines — and it is still not work to take.

[...]

PR #1028, opened 2026-08-29
fix(fleet): a bolded ref and a noun-form Dependencies heading both scan to no dependency (#439)

### LINKED ISSUE #439: candidates.mjs: a bolded ref after the keyword (`Blocked by **#179**`) scans to no dependency — #208 reached phase 0 with an open blocker
Found by the `/fleet:run-team` controller during phase 0 today, not by a review. It produced a live wrong admission, which is why it is filed rather than noted.

## Measured

`depnums`'s inline scan is

```
(?i)(?:depends on|blocked by|requires|after)\*{0,2}:?\*{0,2}\s*(#\d+(?:\s*(?:,|and)?\s*#\d+)*)
```

The `\*{0,2}` alternatives sit **before** `\s*`, so they only match asterisks that are flush against the keyword. Markdown puts them after the space, on the ref:

```
$ PAT='(?i)(?:depends on|blocked by|requires|after)\*{0,2}:?\*{0,2}\s*(#\d+(?:\s*(?:,|and)?\s*#\d+)*)'
Blocked by **#179** (decides gone shape)      -> []
Blocked by #179 (decides gone shape)          -> ["#179"]
blocked by **#178**                           -> []
Depends on **#5**                             -> []
```

Bolding the ref is enough. The keyword, the colon and the ref shape are all unchanged.

## The live consequence

**#208**'s Agent Brief ends with:

> - Blocked by **#179** (decides `gone()`'s shape, which `Indeterminate` needs), which is itself blocked by **#178**.
> Recording those as native dependency edges so the fleet does not claim this early.

`candidates.mjs` reported `#208` with `d:[203]` — #203 only, which is MERGED. Both real blockers (#179, #178) are open and both were dropped. The ticket therefore survived the dependency scan and reached the survivor read as claimable. It was caught only because the controller read the brief in full and saw the prose; nothing in the tooling flagged it.

Note the author explicitly believed they were recording machine-readable edges. That is the failure mode — the syntax looks right and is silently not.

## Second, independent miss on the same ticket

The section is headed `## Dependencies`. The section-mode branch gates on

```
(?i)^#{1,6}\s+(depends on|blocked by|requires)\b
```

which does not match the noun form `Dependencies`, so the list-item branch never engages either. Both routes into that section fail, for two different reasons.

## Scope — what this is NOT

- **Not #339**, which is a bold *label* alone on its line with refs in a list below. Here the label is inline and unbolded; the **ref** is bolded. Different position, same regex.
- **Not #58** (CLOSED), which was about to-tickets' blocking edges.
- **Not #205**, which is the missing test coverage for `d:` — though a fix here plainly wants a pin, and #205 is the ticket that owns that gap.

## Suggested direction, unverified

Moving `\s*` ahead of the asterisk groups, or accepting `[\s*]*` between keyword and ref, would cover the bold case. Neither is measured — and the noun-form heading is a separate decision (widen the heading regex, or accept that only the inline form works). Marked `needs-triage` for that reason: there are two defects here and the second one has an open choice.

> *Filed by the `/fleet:run-team` controller, phase 0, from a live wrong admission.*

feigi:
Closes #439

Sizing: light

Two independent misses on one section of `candidates.mjs`'s `depnums` jq program, both live against `origin/main` at cbbc6fa, plus a residual of the first miss found reviewing the fix — all three fixed here.

## What changed

**Inline separator, label side** — `\*{0,2}:?\*{0,2}\s*` becomes `[\s*]*:?[\s*]*`. The asterisk groups sat *ahead* of the separating whitespace, so they matched only asterisks flush against the phrase; markdown puts them on the other side of the space, on the ref.

**Inline separator, between refs** — the continuation `\s*(?:,|and)?\s*` becomes `[\s*]*(?:,|and)?[\s*]*`. Widening only the label side left the run breaking at the *second* ref's leading `**`, so `Blocked by **#12** and **#13**` answered `[12]` — one open blocker silently dropped, exit 0, nothing on stderr. That is #439's own failure one ref to the right, and it survived the first fix.

**Heading gate** — gains the noun form, anchored to the whole heading with an optional trailing `:`: `((?:depends on|blocked by|requires)\b|dependenc(?:y|ies):?\s*$)`. The noun form cannot take the verb forms' `\b` tolerance — `## Dependency injection` is an ordinary section title in a code repo, and arming on it would invent blockers out of its bullets.

## Measured, both engines

Standalone harness applying the shipped program under `/usr/bin/jq` (jq-1.7.1-apple) and gojq 0.12.19 — the version `.github/workflows/ci.yml:226` installs. Every row agreed between the two engines. `before` is cbbc6fa, `after` is this branch's head.

| fixture | before | after |
|---|---|---|
| `Blocked by **#179** (decides gone shape)` | `[]` | `[179]` |
| `blocked by **#178**` | `[]` | `[178]` |
| `Depends on **#5**` | `[]` | `[5]` |
| `Blocked by: **#12**` | `[]` | `[12]` |
| `**Blocked by** : #12` | `[]` | `[12]` |
| `Blocked by **#12** and **#13**` | `[]` | `[12, 13]` |
| `Blocked by: **#12**, **#13**` | `[]` | `[12, 13]` |
| `Blocked by #12 and **#13**` | `[12]` | `[12, 13]` |
| `Depends on **#5**, #6` | `[]` | `[5, 6]` |
| `Blocked by #179 (decides gone shape)` — control | `[179]` | `[179]` |
| `**Blocked by:** #12` / `**Blocked by**: #12` — controls | `[12]` | `[12]` |
| `Blocked by: #12, #13` — control | `[12, 13]` | `[12, 13]` |
| `## Dependencies` + list | `[]` | `[12, 13]` |
| `## Dependency injection` + bullet naming a ref | `[]` | `[]` |
| #208's `## Dependencies` section verbatim | `[]` | `[178, 179, 390]` |

The four rows that move from a single number to a pair are the between-refs residual; `Blocked by #12 and **#13**` is the one that shows it is the *later* ref's asterisks that broke the run, since its first ref was never bolded.

The `PR #390` bullet in #208's section is collected by construction — in-section collection reads every `#N` on a list-item line — and is inert at the consumer, which drops a ticket only for a blocker still open.

## The other direction

The phantom-blocker sweep PR #331 closed stays closed, unchanged on both engines and re-measured at this branch's head after the between-refs widening: `See #99 for context` under a heading → `[]`; a later paragraph's `#42` → not collected; a fenced block's `fixes #77` → `[]`; a continued list line → first ref only. All four are carried through the newly-armed noun heading as well. A separator loosened for bold is exactly the change that would reopen them, which is why they are re-derived here rather than inherited from the earlier commit.

Every guard pin was mutation-tested rather than left to pass vacuously, each mutant applied to a scratch copy of this branch's head and the suite run with `GOJQ_BIN` set:

- widening the noun alternative from `:?\s*$` to `\b` → *the noun form arms only when it is the WHOLE heading* reds, and so does the gojq parity test.
- dropping the list-item restriction (`$nextsec and $item` → `$nextsec`) → *the phantom-blocker sweep stays closed* reds, alongside the two #331 pins it sits with.
- reverting the whole inline separator to its pre-#439 form → both bolded-ref tests and the gojq parity test red.
- reverting the heading gate to the verb forms only → *#208's own brief text* reds, while the new between-refs test stays green.

Those last two are also the discrimination check on the new test: it reds under its own mutant and stays green under one it has no business catching.

The new pins were confirmed RED against a401dca's parent before the first fix landed: exactly the bolded-ref, noun-heading, #208-brief, phantom-sweep and gojq-parity tests failed, and no other.

## A comment corrected

The #208-brief fixture's comment claimed it "reds if either half regresses". Measured against this branch's head, it does not: its bolded refs sit on list-item lines inside the armed section, so the list-item branch reads them whatever the label separator does, and a full separator revert leaves that fixture green. Only the heading half is pinned there. The comment now says that and names where the separator half is pinned instead — the two bolded-ref tests, which the same mutant does red.

## Engine semantics

The widened separator keeps `\s` rather than spelling an ASCII class, so it inherits the existing Oniguruma/RE2 split rather than pre-empting #383, which owns that question. The between-refs widening uses the same token and inherits it identically. The gojq parity test gains discriminator rows for the bolded and multi-ref forms: `Blocked by: **#12**` is `[12]` under jq and `[]` under gojq, exactly as the bare form already was.

## Deliberately left

- `## Blockers`, `## Prerequisites`, `## Requirements`, `## Blocked` as heading forms. `## Requirements` in particular is ordinary spec prose whose bullets routinely name refs — the same argument that keeps `after` out of the heading alternation.
- A ref inside backticks (`` Blocked by `#12` ``) — unchanged, and not a form the ticket names.
- A bold split landing *mid-phrase* (`**Blocked** by #12`) — the label alternation still needs one unbroken literal. Deferred rather than fixed here; it is not a shape #439's acceptance criteria name.
- A bolded heading (`## **Dependencies**`) still arms nothing. Same reason: out of this ticket's scope, and it interacts with the `## Dependency injection` end-anchor this PR deliberately added. Deferred.
- #339's shape (a bold *label alone on its line* with refs in a list below) is **not** made free by this change: measured at 9ecd630, `**Depends on**\n\n- #12\n- #13` and `**Blocked by:**\n\n- #12` both still answer `[]` on both engines. It needs section scope, which this PR does not touch.

## Test result

`node --test skills/fleet/scripts/*.test.mjs` at this branch's head, `GOJQ_BIN` set: pass equals tests, fail 0, skipped 0. `skipped 0` is what says the gojq gate executed rather than skipping past the engine `gh --jq` actually applies.


feigi:
Review adjudicated. Twelve findings across all six dimensions; no coverage gap.

**Applied here** (`1cb24e8`):

- The two `survived` findings — `correctness` and `silent-failure` at `candidates.mjs:183` — are one defect and took one edit: the between-refs continuation was widened with the same token the label side already used. Both refuters on both findings confirmed it 2-0, and it was re-measured here on `/usr/bin/jq` and `/Users/chris/go/bin/gojq` before commit.
- A `tests` suggestion that the label-side widening's pre-colon half had no discriminator. Dispatched a refuter biased to refuse; it could not refute, and found the discriminating shape is bare whitespace before the colon, not an asterisk: `**Blocked by** : #12` is `[12]` under the merged form and `[]` under the pre-#439 asterisk-only group, on both engines. Added as a fixture in both engine paths.
- Two `simplify`/`tests` findings that the #208-brief fixture's comment overclaimed its coverage. Re-derived against the post-fix tree rather than copied from the findings, since widening the separator could have made the claim false — it did not. A full separator revert leaves that fixture green; only the heading half is pinned there. Comment corrected.

**Deferred and filed** — #1031, #1032, #1033, #1034, #1035, #1036, each with its dimension, its `file:line`, the measurement behind it and why it was not applied here. Three are real defects outside #439's acceptance criteria (bold-blind heading gate, no signal for an unparsed declaration, mid-phrase bold split). Two are coverage deletions declined on a PR that had just proved this suite under-discriminating. One is a cosmetic regex group that predates this PR and needs a two-line sweep rather than a rider.

The `types` dimension returned a no-op — a jq-program-as-string diff has no types to review — and is recorded as ruled rather than filed.

`node --test skills/fleet/scripts/*.test.mjs` at `1cb24e8` with `GOJQ_BIN` set: pass equals tests, fail 0, skipped 0.

PR #1037, opened 2026-08-29
fix(run-merge-bot): answer the staleness probe's `+` on the primary path, and pin its window (#447)

### LINKED ISSUE #447: run-merge-bot step 1: the `git cherry` staleness probe on the server-side path has no stated outcome for `+` — the worktree-diverged STOP now lives only in the fallback
Deferred from PR #441 review (closes #149). **dimension: correctness.**

## The residue

`skills/fleet/commands/run-merge-bot.md:60` tells the merge bot to probe a post-rebase stale worktree:

> To prove it holds nothing unique, `git cherry origin/<branch> HEAD` from the worktree — **not** `git cherry origin/main HEAD` …

and then stops. **Neither outcome is given an instruction.** `-` for every commit presumably means "stale but safe, proceed", and `+` presumably means stop — but the doc says neither, on the path that is now the default.

Meanwhile the `worktree-diverged-#<pr>` STOP lives only inside the **Fallback** block (`run-merge-bot.md:74`, under the `**Fallback (server-side rebase unavailable).**` heading), and `run-team/SKILL.md:1036` still carries the row

> `| Merge bot finds the worktree ahead of the PR head | `worktree-diverged-#<pr>`, PR stays queued …`

So the primary path runs a staleness probe whose result routes nowhere.

## Why this is only the residue, and what was refuted

The original finding claimed more: that the check is **vacuous without a `git fetch`**, because a stale `origin/<branch>` still equals HEAD. That half was **refuted by measurement** and must not be re-filed:

- **Vote 1 (refuted=false)** reproduced the benign case only: stale worktree, no local commit, no fetch → `git cherry origin/feat HEAD` prints zero lines, and concluded the check was vacuous.
- **Vote 2 (refuted=true)** ran the case vote 1 never did — a stale worktree **with an unpushed stray commit**, still without a fetch:

  ```
  == case A: stale worktree, clean (no local commit), NO fetch ==
  exit=0            # zero lines = "holds nothing unique" — the correct answer
  == case B: stale worktree + unpushed local commit, NO fetch ==
  + 6943bf94cb16cb1df73878f7c8b17ce8931f8ecc
  exit=0            # the dangerous case IS detected, with no fetch at all
  ```

  The stale `origin/<branch>` is the *pre-rebase* head, so the stray commit is genuinely not in it. **The un-fetched check catches the only case that matters.** Vote 1 generalised "vacuous" from the benign case alone.

Ruled by the controller: **UPHELD REFUTED — do not add the fetch, do not re-file the vacuity claim.** Only the missing `+` handling survives, and it is filed here.

The defect is verified but the **approach is genuinely open**, because vote 2's second argument cuts against the obvious fix:

1. **Add the outcome branches to the primary path** — `-` → proceed, `+` → STOP and report `worktree-diverged-#<pr>`. Restores the `run-team/SKILL.md:1036` contract on the default path. But it may be dead weight: the server-side path runs no local rebase and pushes nothing (step 4 merges the remote head via `gh pr merge`), so an unpushed local commit is *structurally unable* to enter the merge. The STOP's own stated rationale — "your rebase would carry into the merge" — is a hazard that exists only in the fallback, which is exactly where the STOP now lives.
2. **State explicitly that `+` cannot matter here**, and say why — one sentence, no new STOP. Cheapest, and it closes the reader's question ("I ran the probe, it printed `+`, now what?") without minting an outcome the path cannot act on.
3. **Move the probe into the fallback too**, if it proves nothing actionable on the primary path.

(2) looks smallest and matches the mechanism, but whether a `+` on the primary path should still halt a merge is a judgement about how much the runbook trusts "structurally unable" — a maintainer's call, not an implementer's.

PR #441's applied fixes were the three ruled defects (unbounded poll loop, the false `no autoMode.allow entry` cause, the fallback's un-mergeable dead end). Adding an outcome branch here is a fourth behavioural change to the same step with an unsettled shape — and the review's own evidence on it split two ways.

> *Deferred from PR #441 review. Finding at `skills/fleet/commands/run-merge-bot.md:60`, dimension `correctness`.*

feigi:
Closes #447.

Sizing: light

Ruling followed: **option 2**. `+` is stated to not halt the primary path, with the structural reason. No outcome branch, no new STOP, and **no fetch** — the vacuity claim is UPHELD REFUTED and is not re-filed here.

A review of the first draft found it had minted a false claim while fixing the missing outcome, and that its fallback-STOP pin was vacuous. Both are corrected here; the sections below describe the tree as it now stands, not the first draft.

## What changed

`run-merge-bot.md` step 1, `run-team/SKILL.md`'s failure table, and the pins.

1. **The probe's window, in the same sentence as the command.** `run \`git cherry origin/<branch> HEAD\` from the worktree **before the merge, never after**`. GitHub deletes the remote branch when the PR lands, so `origin/<branch>` outlives the merge only until the next pruning fetch — and step 4's own proof runs one seconds later, under this machine's global prune. Past that fetch the probe dies on a missing ref, an error easy to misread as the divergence it was meant to rule out. The reason is stated conditionally on purpose: the local tracking ref does **not** vanish the instant GitHub deletes the branch, and the first draft said it did.

2. **Which probe output is clean.** The clean answer is the absence of a `+`, not the absence of output. `git cherry` emits a line per commit in `origin/<branch>..HEAD` and marks a patch-equivalent one `-` rather than dropping it, so a worktree that has fetched past the server-side rebase reads all-`-` and one that has not — the state this step leaves it in, since nothing here fetches — reads nothing. Both are clean. Only a `+` is the divergence signal.

3. **`+` answered, and routed.** A `+` names a commit the worktree holds and the rebased branch does not: it is reported alongside the merge, since it is the only signal that something local is unpushed, and it does **not** stop the merge — this step rebases through the API and pushes nothing, step 4 merges the remote head, so an unpushed local commit is structurally unable to enter it. The scoping stays explicit: the STOP's own rationale names a hazard belonging to the fallback, where it is stated and still applies.

4. **`run-team/SKILL.md` — the failure row reconciled.** It promised `worktree-diverged-#<pr>` for a worktree ahead of the PR head with no path qualifier while the halt lives only in the fallback; the row now reads `**on the local-rebase fallback**`. That scoping is what left the primary path's `+` routing nowhere, which point 3 closes.

## The `git cherry` measurement

Ran at the head this PR ships. Positive control first: `git rev-parse --show-toplevel` in the fresh scratch dir → `fatal: not a git repository`, rc 128, before any `git init`. Fixture: bare remote, branch with two commits, `main` advanced, branch rebased onto it and force-pushed from a second clone (the server-side rebase), and a clone left at the pre-rebase head (the stale worktree).

| Worktree state | `git cherry origin/feat HEAD` |
|---|---|
| Stale, not fetched since the rebase — what this step's own flow leaves | nothing, rc 0 |
| Stale, after `git fetch origin` | a `-` per rebased commit, rc 0 |
| Either of the above, plus an unpushed commit on top | the same lines, plus a `+` |
| Reset to the rebased head (not stale at all) | nothing, rc 0 |

So silence is a real clean reading — it is just not the only one, and it is not what the paragraph's own premise (a worktree gone stale *after* the rebase, having seen it) produces.

## Settling commands, re-run at the head this PR ships (`55b7bfe`)

| Claim | Command | Result |
|---|---|---|
| The primary path pushes nothing | `grep -c "That's an API call, not a push — no local git command runs" skills/fleet/commands/run-merge-bot.md` | `1` |
| The merge takes the remote head | ``grep -c '^4\. `gh pr merge <pr> --merge`' skills/fleet/commands/run-merge-bot.md`` | `1` |
| GitHub deletes the branch on merge | `grep -c 'GitHub removes the remote branch anyway' skills/fleet/commands/run-merge-bot.md` | `1` |
| Step 4's proof fetches, plain — no `--prune` flag | `grep -n 'git fetch --quiet origin' skills/fleet/scripts/prove-merge.sh` | line 65, unflagged |
| …and that fetch prunes here | `git config --show-origin --get fetch.prune` | `file:/Users/chris/.gitconfig	true` |
| The fallback pushes nothing either | `grep -c 'never needs to reach the remote' skills/fleet/commands/run-merge-bot.md` | `1` |
| The STOP sits inside the fallback block | `grep -n 'your rebase would carry into the merge' …` vs `grep -n 'Fallback (server-side rebase unavailable)' …` | STOP at 101, heading at 91 — the STOP resolves after it; the citation this PR adds resolves at 87, before it |
| **AC-4: the fallback STOP is untouched** | `git diff origin/main -- skills/fleet/commands/run-merge-bot.md \| grep -c 'Worktree ahead'` | `0` — the bullet is byte-identical to `origin/main` |

## Prose pins

`skills/fleet/scripts/run-merge-bot-prose.test.mjs`. Five pins, each mutation-tested both ways at this head, mutations applied to a copy outside the worktree and confirmed applied before each run.

| Mutation | Reds |
|---|---|
| Split the ordering constraint out of the command's sentence, keeping every pinned word and parking it in a later sentence | only the ordering pin |
| Keep the `+` verdict, delete the structural reason | only the outcome pin |
| Revert the clean-answer rule to "No output is the clean answer" | only the clean-answer pin; 19 of 20 still pass |
| Delete the rationale from the fallback's `**Worktree ahead**` bullet, leaving the citation of the same phrase byte-identical | only the fallback-survives pin |
| Revert run-team's row qualifier | only the cross-file pin |
| **Control** — an unrelated reword in the same paragraph | nothing; 20 of 20 green |

Two are worth naming. The ordering mutation changes no vocabulary at all, so it proves that pin anchors the constraint's *placement* beside the command rather than its presence somewhere in step 1. The `**Worktree ahead**` mutation is the one the first draft failed: the rationale phrase occurs twice in the file (`grep -o -c 'your rebase would carry into the merge' …` → `2`, the STOP's own and this PR's italic citation), and a step-1-wide assertion was satisfied by the citation alone, so deleting the rationale stayed green. The pin now slices the fallback block and asserts the bullet's whole rationale clause inside it.

## The comment that asserted too much

The pin's comment said that on the fallback an unpushed commit "genuinely does ride into the merge". The file's own untouched step-4 sentence denies it — the fallback "never needs to reach the remote", and step 4 merges "the pre-rebase head, since nothing here was pushed". Settled against a fixture running the fallback's documented sequence: the unpushed commit appears in neither the merge commit's ancestry nor its tree, while a control that pushes the rebased head does land it. The comment now names the real hazard — it corrupts what you verify — and still names the STOP's own wording as the thing it pins. The STOP's wording itself is pre-existing and out of this ticket's scope; it is unchanged.

## Why the ordering constraint earns its line

The runbook not carrying it has a live cost: the controller had to hand "run the probe **before** the merge, never after" to both merge bots dispatched in today's run by writing it into their dispatch prompts, because the doc does not say it. That is a controller observation from this run, not a measurement taken here. A review finding argued the constraint is outside #447's residue; the ticket's own acceptance criteria require it in the same sentence as the command and give the reason, so it stays.

## What I deliberately left

Enumerated from the ticket's prose first, then from the mechanism:

- `git cherry origin/main HEAD` — named in passing as the wrong upstream. Already has a stated outcome ("proves nothing"). Not defective.
- `git ls-remote origin "$ref"` (step 1's poll) — reads the same branch ref, but runs before step 4 by construction and already has all four outcomes enumerated.
- `git merge-base --is-ancestor <pr_head> origin/<branch>` — same ref, but reachable only on the desync path, where the merge has *not* happened.
- `git rev-list --count origin/<branch>..origin/main` — already carries "before **each** merge" in its own sentence.
- The fallback STOP itself — AC-4, unchanged, and now pinned as unchanged, rationale included.
- No new report token for the primary path's `+`, and no run-team failure row for it. The bot says what it saw with the merge; minting a token would need a matching table row and that is a wider change than this residue.

Suite at this head: `node --test skills/fleet/scripts/*.test.mjs` → tests 1397, pass 1397, fail 0.


feigi:
Cross-reference, added by the fleet controller after the finisher noted this PR carries no pointer to it.

**#1038** was filed from this PR's review and deliberately **not** fixed here: #447's AC-4 requires the fallback `**Worktree ahead**` STOP bullet byte-identical, and the defect is in that bullet's rationale.

The fallback block states the local rebase is verification-only and never reaches the remote, while the STOP bullet gives "your rebase would carry into the merge" as the hazard. Those cannot both hold. A refuter measured the unpushed commit reaching neither the merge commit's ancestry nor its tree. The STOP is correct; its stated reason is not — the real hazard is that you verify a tree that is not the reviewed head.

Verified byte-identical against `origin/main` by this PR's finisher (`cmp` on the extracted bullet), so AC-4 holds and #1038 remains open for the wording.

PR #1043, opened 2026-08-29
fix(fleet): rule how a refuter observes a test run, at all three refuter dispatch sites

### LINKED ISSUE #503: fleet refuter briefs: 'until grep <marker>' polling on a node --test log hard-deadlocks — 3 of 5 refuters reached for it independently
Measured during `/fleet:run-team`, PR #498's fix-apply pass, 2026-08-16.

## What happened

A refuter dispatched by `fix-pr-498` wedged permanently on its own polling loop:

```sh
until grep -q "^# fail" .../pr498/f0/mut_a.txt 2>/dev/null; do sleep 20
```

Its `mut_a.txt` ended:

```
ℹ tests 520
ℹ pass 517
ℹ fail 0
```

`grep -c "^# fail" mut_a.txt` → **0**. The agent waited on the `#` pipe-reporter prefix while `node --test` wrote the `ℹ` terminal prefix. The marker can never appear, so the loop is infinite. Transcript frozen 15+ minutes; the agent had to be killed by the controller and its finding went **unchecked**.

## Why this is a pattern, not one agent's slip

**3 of 5 refuters in one dispatch independently reached for the same idiom.** A second (`f8`) used `until grep -q 'MUT_RC=' mut2.txt` and survived only because its writer does append that marker.

## The `ℹ` vs `#` split is already known here — as a *reading* bug

`review-and-fix.md` warns about the same prefix split for the zero-test rule (`tests 0` is a FAILED run). This is that same split resurfacing as a **liveness** bug: not "the agent misread the result" but "the agent can never observe a result at all". The existing warning does not cover it, because it is phrased as a rule about interpreting output that has already arrived.

## Suggested fix

In the refuter brief — `workflows/review-pr.js`'s refuter prompt, and the fix-applier's refuter brief in `skills/fleet/skills/run-team/SKILL.md` — state that a refuter runs its verification command **synchronously and reads the exit code**, and must never poll a log file for a completion marker. If a marker check is genuinely wanted, it must be prefix-agnostic: `grep -E "^(ℹ|#) fail"`.

## Also worth pinning

A frozen transcript is **not** a dead agent — it can be a blocked tool call. The two are distinguishable only from the agent's scratch dir: here the record counts were unchanged across two polls while `pin.log` / `mut2.txt` / `base.log` all carried fresh mtimes. Any guidance about detecting a stalled member should say that.

## Not fixed in this run, deliberately

`run-team/SKILL.md` is edited by two open PRs (#498, #499) right now, and `review-pr.js`'s refuter prompt is in active use by #499's running review — changing it mid-review would alter behavior under a review already in flight.

Related: #496 (same refuter prompt, scratch-dir and cd-chaining rules).

feigi:
Closes #503

Sizing: heavy

## What this does

The refuter briefs told an agent to verify by RUNNING something and never said how to **observe** the result. Measured consequence: refuters reach for `until grep -q "<marker>" <log>; do sleep N`, and one waited on the tap `#` prefix while node wrote the spec one — the marker could never appear, the loop was infinite, and the finding went unchecked.

Per the ticket's scope ruling, this states **the rule about matching node's reporter output**, in the refuter briefs. It does not sweep the pins (#514 owns those).

## The measurements, reproduced rather than copied

All on node v26.7.0, `node --test` over a three-test file, piped through `cat -v`:

| condition | summary line |
|---|---|
| `env -u FORCE_COLOR`, piped | `ℹ pass 3` |
| `FORCE_COLOR=1`, piped | `^[[34mℹ pass 3^[[39m` |
| `env FORCE_COLOR=`, piped | `^[[34mℹ pass 3^[[39m` |
| TTY via `script -q /dev/null`, no `FORCE_COLOR` | `^[[34mℹ pass 3^[[39m` |

Three things fall out, and the third is why the narrow fix would have shipped a broken rule:

- **The colored line begins with ESC**, so an anchor on *either* prefix matches nothing — empty output at exit 0, indistinguishable from a hung run and from a zero-test run. Accepting both prefixes is necessary but **not sufficient**; SGR must be stripped too.
- **`FORCE_COLOR=` empty colors identically to `FORCE_COLOR=1`** — the claimed control is confirmed *not* a control. `env -u FORCE_COLOR` is, and the rule says so.
- **A TTY colors on its own**, with `FORCE_COLOR` unset entirely. So the SGR hazard is not reachable only by forcing color.

**One axis I could not reproduce, stated rather than asserted:** only node v26.7.0 is installed here and it emits the spec prefix *even when piped*, so the tap `#` form appeared only under an explicit `--test-reporter=tap`, never as a version default. The rule is therefore written to be **safe** under a prefix this machine cannot produce — demanding SGR stripping rather than resting on a prefix alternation — instead of asserting a version boundary I did not observe.

## Three dispatch sites, not two

The brief named two. `grep -rn 'Try to REFUTE this finding'` at the shipping commit reports the instruction quoted **verbatim** in a third: `review-and-fix.md` step 2, which `/fleet:review-and-fix` hands to a refuter word for word.

Widening is not my invention — `refuter-scratch-prose.test.mjs`'s own header already ruled it: *"A fix added to only one is the one that does not reach whichever path a given caller takes, so both slices are pinned here."* The brief's stated blocker for widening was a collision with PR #939, which is **merged**; no open PR touches any of the three files.

`review-and-fix.md`'s zero-test rule — the part the brief put out of scope — is untouched.

## Wording

Follows the existing CI precedent in `run-team/SKILL.md` (*"prefer ONE blocking `gh run watch` to a poll loop … permission to look, never as the verdict"*) rather than inventing a second idiom. The clause also states what a refuter **may** still do — reading a log the run has already finished writing is fine — so the ban cannot be read as forbidding a legitimate synchronous check.

## Verification

`skills/fleet/scripts/reporter-output-rule-prose.test.mjs`, five pins × three sites. Each keeps the two halves of a rule in **one span**, because for each one a single half was the measured defect.

Mutation-tested both ways, 10 mutants, every one confirmed applied before its run:

- 7 destructive mutants (splice into the poll ban; drop the SGR half; drop the reason `FORCE_COLOR=` is not a control; drop the pattern-would-have-missed half; spell the polling idiom; and the SGR/poll-ban halves on the other two sites) each red **exactly** their intended pin, zero collateral.
- 3 must-stay-green mutants (a reflow across the blockquote gutter; an edit after the slice; an edit to a clause inside the slice these pins do not own) stay green.

Settling commands re-run at `d56612a`:

```
$ for f in workflows/review-pr.js skills/fleet/skills/run-team/SKILL.md skills/fleet/commands/review-and-fix.md; do ... done
workflows/review-pr.js:                     poll-ban=1 force_color=1 prefixes=1
skills/fleet/skills/run-team/SKILL.md:      poll-ban=1 force_color=1 prefixes=1
skills/fleet/commands/review-and-fix.md:    poll-ban=1 force_color=1 prefixes=1

$ grep -rn 'until grep\|do sleep' <the three briefs>
rc=1   # no match — the acceptance criterion's pass
```

`node --test skills/fleet/scripts/*.test.mjs` → **1412 pass, 0 fail, 0 skipped, 0 cancelled**.


feigi:
## Review applied — 12 findings, split

Head `47fe1ad229fb23e2a90bc150af737678f77f68f5`.

**Applied (6 findings, 1 defect + 1 guard):**

Five findings — `correctness`, `types`, `tests`, and two `unverified` entries — were one defect: the presence pins carried `.{0,N}` gaps far wider than the text they bridged, so a sentence reinstating polling spliced *inside* a pinned clause and stayed green on all three sites. Gaps measured against their budgets at the parent commit:

| gap | real | budget |
|---|---|---|
| `read its exit code.` → `Never poll` | 1 char | `.{0,80}` |
| `completion marker` → `Reading a log…` | 107 chars | `.{0,200}` |
| `missed.` → `grep over one ref` | 3 chars | `.{0,160}` |
| `about history` → `token a diff removed` | 27 chars | `.{0,160}` |

Both halves of the remedy shipped, because neither alone holds: every join is now `\s+` and no pin regex carries a `.{0,N}`, so a splice has nowhere to land; and the negative pin, which banned three literal shell idioms the defeating sentence does not contain, now bans the wait vocabulary instead.

Discrimination proved with a **splice** mutant rather than a deletion mutant — the ten deletion mutants run during implementation all missed this, because deleting pinned vocabulary tests vocabulary-removal and not slack. The splice keeps every pinned word byte-identical and inserts, after `never as the answer.`, the sentence `For a long run, poll the log until the marker appears instead.` Applied to all three sites and confirmed per site with `grep -c` on the inserted text — an `old not in new` check is wrong for an insertion whose replacement contains the search string:

| tree | result |
|---|---|
| parent commit's pins, spliced | tests 15 / pass 15 / fail 0, exit 0 — **green** |
| this commit's pins, unmutated | tests 18 / pass 18 / fail 0, exit 0 — control |
| this commit's pins, spliced | tests 18 / pass 12 / **fail 6**, exit 1 |

The six reds are two per site across all three sites. Counts read from the reporter's `fail` summary line, not by counting `✖` — node prints a `✖ failing tests:` header that a line count scores as an extra failure.

The sixth applied finding pins that the shared slice anchor occurs exactly once per source file. `between()` takes the first match by contract, so the guard could not go in `prose-pin.mjs`: adding it there was measured to take the suite from 1412 pass to 1410 pass / 2 fail (`prose-pin.test.mjs:20`'s own fixture, and `release-label-split-prose.test.mjs:44` where `gh pr create` genuinely repeats).

Two comments that asserted the opposite of what was true are corrected, not just reworded.

**Deferred and filed (3):** #1045 (#503's "Also worth pinning" item, outside this ticket's acceptance criteria), #1046 (fold the `SITES` closures into data rows), #1047 (host `stripQuoteGutter` in `prose-pin.mjs`). #1046 and #1047 say in their bodies that they land only on top of this commit — a dedupe of the harness in the same breath as the pin fix is how a refactor reinstates the defect a survived finding just proved.

**Checked and refuted (1), plus 2 refuted 2-0 by the review's own refuters:** recorded in #1048, closed `wontfix`.

**Suite at this head:** `node --test skills/fleet/scripts/*.test.mjs` → tests 1415, pass 1415, fail 0, skipped 0, exit 0 (1412 before; +3 anchor guards).

PR #1044, opened 2026-08-29
fix(next-ticket): say why step 7's push keeps --force-with-lease unconditionally

### LINKED ISSUE #443: next-ticket step 7 pushes with `--force-with-lease` unconditionally — the first push never needs it, and it is the exact string #149 documents as classifier-denied
Declared as a deliberately-left sibling by the `impl-149` fleet member while fixing #149, then independently verified and **narrowed** by the controller. Filed with the narrowing, because the obvious stronger version is false.

## The line

`skills/fleet/skills/next-ticket/SKILL.md:84`, inside step 7's block:

```bash
git fetch origin && git rebase origin/main   # rebase, never merge main in
<test command>                               # re-run after rebasing
git push --force-with-lease -u origin HEAD
```

## What is actually wrong — and what is NOT

**NOT wrong: the lease does not fail on a fresh branch.** The intuitive claim — that `--force-with-lease` is inert or errors when no remote ref exists — is **false**. Measured, git 2.50.1, bare origin + clone, branch never pushed:

```
$ git push --force-with-lease -u origin HEAD
 * [new branch]      HEAD -> fix/9-fresh
branch 'fix/9-fresh' set up to track 'origin/fix/9-fresh'.
rc=0
```

Clean success. Recording that so nobody re-derives it or files the stronger version.

**NOT wrong: the flag is unconditionally useless.** It is not. Step 7 rebases immediately before pushing, so a member that re-enters step 7 after an earlier push genuinely needs the force. The flag is right on the second pass.

**What IS wrong: the command is unconditional, and the common case is the first push.** In fleet use, `claim-ticket.sh` creates the branch fresh and the member pushes once. On that path the lease buys nothing — and it is the exact command string #149 documents as intermittently denied by the Claude Code auto-mode classifier, before the push is attempted, so it is not a lease race and retrying does not help.

## Measured operational cost, this run

`/fleet:run-team` today, member `impl-149`. It completed #149's work, committed twice, and stopped before step 7. The controller found the commits by auditing the worktree and had to instruct a **deviation from the documented command** — a plain `git push -u`, on the grounds that the branch had never been pushed so no force was needed.

That worked. But it means the documented step was routed around by hand, on the very ticket about this flag being denied, and the reason it was safe to route around had to be re-derived by the controller rather than read off the page.

The fix has an open shape and at least three candidates, none obviously right:

1. Make it conditional — plain push when `git ls-remote --exit-code origin <branch>` finds nothing, lease otherwise. Correct, but puts a two-branch shell conditional into a prose runbook that is deliberately a flat copyable block.
2. Split step 7 into first-push and re-push forms, and say which applies.
3. Leave the command and add one sentence saying the lease matters only on a re-push, so a denial on a first push is safe to route around with a plain push. Cheapest, and it is exactly the knowledge the controller had to supply by hand today.

(3) is the smallest and directly matches the observed failure, but it documents a workaround rather than removing the trigger — a maintainer's call.

## Scope

- **Not #149.** That is the merge bot's rebase-and-merge sequence and its ancestry proof. PR #441 fixes it with server-side `gh pr update-branch --rebase`. This is PR *creation*, where no `gh pr update-branch` target exists yet and there is no currency gate or ancestry proof to protect — which is exactly why #149's brief scoped it out and why `impl-149` correctly left it.
- **Not the `autoMode.allow` permission configuration**, which #149 also rules out.
- `run-merge-bot.md:49` carries the other `--force-with-lease` in the repo; that one is #149's target and PR #441 replaces it.

> *Filed by the `/fleet:run-team` controller from a member's enumerate-the-class declaration, verified independently, stronger version refuted by measurement.*

feigi:
## Summary

Step 7's push (`git push --force-with-lease -u origin HEAD`) was documented unconditionally, with no explanation of what the lease buys. In fleet use the branch is created fresh and the member pushes once, where the lease protects nothing — and it is the exact command string #149 documents as intermittently denied by the auto-mode classifier before the push is even attempted, so a denial is not a lease race. The denial is judged per invocation and intermittently, never once for the command, so a retry is a coin flip rather than a fix.

Ruling (per the Agent Brief, option 3): keep the push command exactly as written, add one sentence explaining why the lease is there, so a denial on a first push is safe to route around with a plain push, while a denial on a re-push is not — the two REFUTED-claim guardrails from the brief (lease does not fail on a fresh branch; flag is not unconditionally useless) hold.

## Review fix — the pins were presence-only

The two pins this PR first shipped were `assert.match` over the whole step-7 section. That reddens when a pinned phrase is **deleted** and stays green when a contradicting clause is spliced in beside it, every phrase intact — including the exact wrongly-ACCEPT case the test file's own comment names. Both halves were measured passing 3/3 while the sentence asserted the opposite of the truth.

The fix narrows before banning, the shape `candidates-exit3-prose.test.mjs` already uses on one table row: `note()` slices out the sentence, and its own `;` splits the force-is-required half from the plain-push-is-safe half (sliced at the **last** `;`, so an appended clause lands inside the fallback half rather than escaping both pins). Then it pins the predicate a reversal actually rewrites — the two original pins are both subject-side — and bans the contradiction in each half.

This also drops a false clause from the test file's own header comment. It read "retrying does not help". `run-merge-bot.md:49` records the denial as "judged and intermittently denied per invocation", and #149 records the same command succeeding for another actor the same day, so a retry is a coin flip, not a no-op. Restated as the property.

## Test plan

- [x] Full suite at the shipping commit `48f1d3b`: `node --test skills/fleet/scripts/*.test.mjs` → `tests 1395`, `pass 1395`, `fail 0`, `skipped 0`
- [x] Mutation-tested with **negation/widening** mutants, not deletion mutants — each keeps every pinned phrase byte-identical and changes only the surrounding meaning; each confirmed applied before running, and reverted from a pristine copy between runs:

| mutation | pinned phrases | result |
|---|---|---|
| append `, though in practice it does not need the force there` to the reason half | untouched | **RED** (`doesNotMatch`), 2 pass / 1 fail |
| append `, and the same plain push is just as safe to use after a re-push denial.` to the fallback half | untouched | **RED** (`doesNotMatch`), 2 pass / 1 fail |
| *control* — append ` onto the updated base` to the reason half (same shape, innocuous) | untouched | GREEN, 3 pass |
| *control* — reword `is safe to route around` → `is fine to route around` | untouched | GREEN, 3 pass |
| *control* — append `, since a first push does not need the force at all.` (a **true** statement of a refuted claim) | untouched | GREEN, 3 pass |

The two controls that share a mutant's shape are what make this discrimination rather than brittleness; the third confirms neither new ban forecloses a true statement of either REFUTED claim.

Sizing: light

Closes #443


feigi:
Reachability observation from the fleet run that produced this PR (2026-08-29), recorded because it bounds this ticket's motivating symptom without refuting it.

**Every first push in this run used the documented `git push --force-with-lease -u origin HEAD`, and none was denied.** Three separate implementers on three separate branches, each pushing a branch that had never been pushed, each succeeding at rc 0 on the first attempt. The controller had pre-authorised a plain-push fallback for exactly the denial this ticket describes; no member needed it, and there is no denial text to record.

**This does not weaken the ticket, and option 3 is the right fix precisely because of it.** The denial is judged per invocation — the ticket's own measured occurrence stands, and today's runs stand alongside it. A remedy keyed on the denial happening (option 1's conditional) would be keyed on a condition that is not stable, and would have been dead weight in every push of this run. Explaining *why* the lease matters only on a re-push is useful whether or not the classifier ever fires again.

**One place the reason was load-bearing this run, in the opposite direction.** The controller read a member's committed-but-unpushed branch as the stall this ticket documents, and instructed a plain push on that inference. The inference was wrong — the member was waiting on a full-suite run, not on a denial — and the correction came from the member saying so. The sentence this PR adds is what lets a reader distinguish those two cases without guessing: a plain push is correct **only** where the branch has never been pushed, and the lease is load-bearing on any re-push after a rebase.

feigi:
Review deferrals filed: #1049 (`needs-triage`, open — the re-push route the Brief deliberately withheld, with the measured evidence and a pin-safe candidate remedy) and #1050 (`wontfix`, closed — the three-finding `suggestion` band, each checked by one biased-to-refuse refuter and all three refuted by measurement).

Two of those refuters converged independently on the same fact — the classifier denial is judged **per invocation** and non-deterministically (`run-merge-bot.md:49`: "judged and intermittently denied per invocation"; #149: "Not universal"). That falsified "retrying does not help", a clause this PR carried in both its test-file header comment and its own body. Both have been corrected to state the property; the fix for finding 1 would have spread it further onto a member-read page.

PR #1051, opened 2026-08-30
docs(fleet): name refs/stash recovery, and the audit's self-named cause (#437)

### LINKED ISSUE #437: run-merge-bot runbook: the `unknown` path names no ref-based recovery — `git show refs/stash` recovers the work in four measured states where both file reads come back empty

feigi:
## Summary

The `unknown`-path paragraph in `run-merge-bot.md` (the no-undo-audit
section) named no ref-based recovery for the states where `ls -l`/`cat`
come back empty, and didn't say that one of the four `unknown` causes
already names itself in the audit's own stderr. This PR adds both,
in one edit over the paragraph.

**AC already met before this diff (nothing to do — re-derived against
`origin/main`, not the ticket's own tables):**
- The cause count is already correct: `origin/main` reads "three of
  its four causes leave it empty at rc 0" (fixed by f3903c72, already
  an ancestor of `origin/main`). Untouched here — re-mutating it would
  mint a new false correction into a correction PR, the exact failure
  this ticket class is filed to avoid.
- `ls -l` remains the primary diagnostic and `cat`'s one-cause scoping
  is unchanged — neither sentence was touched.

**What this diff adds (the two genuinely open items):**
1. A ref-based recovery for the states where the file reads come back
   empty (unreadable / emptied / gc'd-away reflog). `git show refs/stash`
   reaches the ref directly, loose or packed, and is named for the
   `WIP on <branch>` label; `git stash show -p refs/stash` is named for
   the content, because `git show` alone prints NO diff when the stash
   was taken with the change staged (see Verified). Where the ref FILE
   itself is the broken one (unreadable, or holding non-SHA text) both
   fail, and the reflog's second-field SHA with
   `git stash show -p <that sha>` is what applies there.
2. The stash-object-missing cause is not silent: `git stash list`'s
   own stderr already ends `fatal: bad object refs/stash`, so the
   paragraph now says to read that line before the file archaeology.
   The cause is **named**, not counted to — this paragraph already
   spends "the fourth" on the absent-ref/live-reflog case, and #437's
   brief asks for naming over counting.

**Verified** (git 2.50.1, Apple Git-155, throwaway repos under scratch,
re-run at this PR's head `4b8a628`):

Which of the four `unknown` causes `git stash list` is silent about —
`git stash list`, rc / stdout lines / stderr:
```
unreadable ref file      rc=0  0 lines  stderr empty
unreadable reflog        rc=0  0 lines  stderr empty
absent ref, live reflog  rc=0  0 lines  stderr empty
missing stash object     rc=1  0 lines  stderr: fatal: bad object refs/stash
```

The recovery commands, over {staged, unstaged} x {loose, packed} x
{reflog unreadable, reflog blank, reflog removed} — twelve states:
```
git show refs/stash          WIP label in all 12; NO diff in all 6 STAGED states
                             (stash tree == index-commit parent, default --cc
                              suppresses every hunk)
git stash show -p refs/stash the hunk in all 12
```

Where the ref FILE is the broken one (`chmod 000 refs/stash`;
`printf 'not-a-sha' > refs/stash`), staged and unstaged:
```
git show refs/stash            rc=128  ambiguous argument / ignoring broken ref
git stash show -p refs/stash   rc=1    refs/stash is not a valid reference
git show <reflog 2nd-field sha>        again NO diff when staged
git stash show -p <that sha>   rc=0    the hunk, staged or not
```
So both commands the paragraph names fail exactly where it says they do,
and the fallback it names recovers the content in every state it scopes.

No mutation of the stash stack anywhere in the new text — `show` only,
never `apply`/`pop`/`drop`.

**Enumerate-and-declare** (the ticket's four `unknown` causes: unreadable
reflog, ref pointing at a missing object, unreadable ref file, malformed
ref file — plus the pre-existing fifth shape, ref-file-absent-with-live-reflog):
- Covered by the new recovery sentence: unreadable reflog,
  emptied/gc'd-away reflog.
- Covered by the (pre-existing, untouched) SHA-from-`cat` sentence:
  ref-file-absent-with-live-reflog.
- Newly covered via the SAME new sentence's fallback clause: unreadable
  ref file, malformed ref file (both fail the two `refs/stash` commands,
  both recover via the reflog's SHA).
- Deliberately left uncovered: the stash-object-itself-gone cause —
  nothing recovers it, the object is destroyed; the new text only
  points at the audit's own diagnostic for it, not a recovery.
- Reachable but **not** covered, and filed rather than fixed: `refs/stash`
  and its reflog **both** unreadable. The premise clause holds, every
  command the sentence names fails, and `cat` printed no SHA to fall back
  on. Out of #437's size — **#1055**, `needs-triage`.

**Review round (findings applied after the first push, `cbe3b97..4b8a628`):**
- The "plus the diff" promise was false for a staged stash → item 1 above
  now names `git stash show -p`. The sentence's tail carried the same
  defect and now names `git stash show -p <that sha>`.
- The new sentence bound "the fourth" to a second referent inside a
  paragraph that already spends that ordinal on the opposite-shape cause
  → the cause is named instead. Both remaining "fourth"s are the
  opposite-shape case:
  `awk 'NR==205' skills/fleet/commands/run-merge-bot.md | grep -oi 'the fourth[^:.]*'`
  → `The fourth cause is the opposite shape` / `the fourth cause`.
- The two new pins used literal-space regexes, so a same-words reflow
  inside a pinned span would false-red them → their input is wrapped in a
  `flat()` normalizer. Scoped to these two pins on purpose; the file's
  other 8 `assert.match(noUndoAudit()` pins are left alone, since
  converting them is a file-wide convention change #437 did not ask for.
- One further finding — that the fallback clause is dead prose — was
  refuted by measurement and not applied. Its residual is #1055.

## Test plan
- [x] `env -u FORCE_COLOR node --test skills/fleet/scripts/*.test.mjs` — `tests 1426 / pass 1426 / fail 0` (same count as `origin/main` + this branch's baseline; the two #437 tests were rewritten, not added to)
- [x] `env -u FORCE_COLOR node --test skills/fleet/scripts/run-merge-bot-prose.test.mjs` — 22/22
- [x] Both #437 pins mutation-tested in a copy under scratch, through a guard that throws unless the target occurs exactly once. Four kill legs — the `fatal: bad object refs/stash` string, `loose or packed`, assertion 2's span alone, `reflog's second field` — each reddened its own test and only its own test (21/1). Four control legs stayed 22/22: a pure space→newline reflow inside each of the two pinned spans, and an edit to unrelated doc text. Defanging `flat()` to the identity function and re-running the first reflow reddened it (21/1) — the false red the normalizer prevents.
- [x] Rebased on `origin/main` (already current, no-op)

Sizing: light

Closes #437


feigi:
**Review fixes applied — `cbe3b97..4b8a628`.**

Applied (3 findings, one coherent rewrite of the same paragraph — the prose edits first, then the pins re-derived from the post-fix text, never copied out of a finding):
- `git show refs/stash` promised "plus the diff" and prints none for a staged stash → the paragraph now names `git show refs/stash` for the `WIP on <branch>` label and `git stash show -p refs/stash` for the content. The sentence's tail had the same defect and now names `git stash show -p <that sha>`.
- "The fourth" bound a second referent inside a paragraph that already spends that ordinal on the opposite-shape cause → the cause is named, not counted to.
- The two #437 pins used literal-space regexes → their input is wrapped in a `flat()` normalizer, scoped to those two pins only.

Deferred and filed: **#1055** (`needs-triage`) — with `refs/stash` and its reflog **both** unreadable, the recovery sentence's premise holds while every command it names fails and `cat` printed no SHA to fall back on. This is the residual of a finding whose "dead prose" framing was refuted by measurement; the residual is a different claim, and out of #437's size. A file-wide `flat()` conversion of this test file's other 8 pins was likewise judged out of scope and not filed as a defect — it is a convention change, not a fault.

`env -u FORCE_COLOR node --test skills/fleet/scripts/*.test.mjs` → `tests 1426 / pass 1426 / fail 0`.

Issue #449, opened 2026-08-13
Fleet members and the controller read the runbook and every script from the LOCAL main checkout, which is 55 commits stale — a merge bot nearly ran the step 1 that PR #441 had just deleted


feigi:
Caught by `merge-bot-3` during the final wave of a `/fleet:run-team` run, then found by the controller to be far wider than the bot could see.

## The near-miss

PR #441 rewrote `run-merge-bot.md` step 1, replacing a local `git push --force-with-lease` with server-side `gh pr update-branch <pr> --rebase`, and merged **20 minutes before** `merge-bot-3` was dispatched. The bot read its runbook from disk and got **the pre-#441 version** — the exact step the PR existed to delete. In its own words:

> My first read of the file gave me the old step 1 — local rebase plus `git push --force-with-lease`, no mention of `update-branch`. I only caught it because your dispatch described a step 1 that did not match what I was reading.

It recovered with `git show origin/main:skills/fleet/commands/run-merge-bot.md`. **It recovered only because the controller happened to describe the new step in the dispatch prompt.** With a neutral prompt it would have taken the force-push path — and reported success, since that path still works about half the time.

**The hazard is self-concentrating: the more recently a runbook was fixed, the likelier the next member runs the version that was just proven wrong.**

## Measured extent — it is not just the runbook, and not just the bot

The controller checked its own checkout after the report:

```
$ git rev-parse HEAD          9a281521
$ git rev-parse origin/main   4f7bf2da
$ git rev-list --count HEAD..origin/main
55
$ git diff --stat HEAD origin/main -- skills/fleet/scripts/
 ... 23 files changed, 2496 insertions(+), 192 deletions(-)
   ledger.mjs             150 +++--
   reap.sh                 24 +-
   release-ticket.sh       38 +-
   no-undo-audit.sh        81 +
   inflight.test.mjs      421 ++++--
   ...
```

**55 commits behind, 23 changed script files.** So for this entire run:

- The **controller** ran `candidates.mjs`, `inflight.sh`, `claim-ticket.sh`, `ci-state.mjs`, `reap.sh`, `ledger.mjs`, `fleet-tick.mjs`, `verify-sha.sh`, `pr-overlap.mjs` and `worktree-audit.sh` from the stale tree — including `reap.sh` and `release-ticket.sh`, which **delete branches and worktrees**.
[...]

Issue #819, opened 2026-08-23
docs/specs/2026-07-23-fleet-plugin-design.md: the script surface table lists none of the scripts added since it was written


feigi:
Deferred from PR #813 review (issue #238).

**dimension:** `comments`

## The finding

`docs/specs/2026-07-23-fleet-plugin-design.md:165` introduces the Script surface table with "All eleven obey the three rules above." The review claimed PR #813 makes that sentence more false by adding `staleness.mjs` — a script on the same 0/1/2 exit contract — without a row.

**That half was ruled refuted and is not what this ticket asks for.** The sentence quantifies over the rows that follow it, those rows are untouched by PR #813, and every one of them still obeys the contract, so the diff does not change its truth value.

## What is real, and pre-existing

The table does not list every fleet script on that contract, and did not before this PR. `release-ticket.sh` and `drop-merged-label.sh` were both added after the spec was written (`git log --diff-filter=A`), `derive-testcmd.sh` likewise, and all three follow the same 0/1/2 shape with no row. `board.mjs`, `fleet-tick.mjs` and `diff-stats.mjs` are also absent.

The implementer's reasoning for leaving `staleness.mjs` out was upheld on review: a row would invent a currency the table does not have, given how many of its siblings it already omits.

## Why deferred

The question is what the table is FOR — an enumeration of every fleet script, or a curated set — and that decides whether the fix is a row per missing script, a scope sentence, or nothing. Not a defect anyone has established, so `needs-triage` rather than `ready-for-agent`. Do not expand PR #813's diff for it.

feigi:
Triage note, added by the filer: a class ruling already covers most of this. #257 (CLOSED) settled that `docs/specs/*` are dated historical records — default is to leave stale wording alone, and the one exception is a line making a **behavioural claim the code now contradicts**, corrected by annotation rather than a silent rewrite (#782 is the worked instance of the exception).

Applying that test here: a reader taking the table as current gets no wrong belief about what any script DOES — every listed row is still accurate, and the introducing sentence quantifies over those rows. What they can get is the impression that the table enumerates every fleet script, which is a completeness/phrasing gap rather than a false behavioural claim. Under the #257 ruling that reads as no-action.

Filed anyway rather than dropped because the ruling was made about quotes and line references, not about a table other code cites as live truth — `inflight.sh`'s header points at this table for its exit-2 causes. Whether that citation is enough to pull the table out of the dated-record default is the decision this ticket wants.

[...]

PR #1052, opened 2026-08-30
fix(fleet): the controller refuses to read an instrument that changed under it (#436)

### LINKED ISSUE #436: run-team: the controller runs its instruments FROM the main checkout, which no worktree isolation protects
Measured as a **near-miss** during a `/fleet:run-team` run, 2026-08-11. No wrong reading was produced, and the reason is that the member caught its own mistake — not a property of the design.

## What happened

`impl-269` was dispatched into `.worktrees/269-ci-state-shape-guards` to fix `ci-state.mjs`. It disclosed, unprompted:

> early on I edited the two files in the main checkout (`/Users/chris/.claude`, branch `main`) instead of my assigned worktree — caught it via a hash mismatch before committing, diffed it out, reapplied to the worktree, and reverted the main checkout to clean

The controller verified the cleanup was real: `git status --porcelain` clean apart from a pre-existing unrelated change, both files hashing byte-identical to `origin/main`, no stray commits, and no reflog activity on `main` since before the session started.

## Why this is a controller-correctness problem, not a tidiness one

**The controller runs its scripts from the main checkout.** During the window that member had `ci-state.mjs` modified, the controller's CI monitor was polling `node skills/fleet/scripts/ci-state.mjs` every 120s from that same path, and a live finisher ran it to decide whether to apply `ready-to-merge`.

Every instrument the controller reads the world through lives there:

- `ci-state.mjs` — every CI gate decision, and the finisher's label gate
- `worktree-audit.sh` — every dirty/diverged halt
- `verify-sha.sh` — every SHA acceptance
- `reap.sh` — branch and worktree deletion
- `ledger.mjs` — the run's entire durable state
- `fleet-tick.mjs` — dispatch decisions

A member editing any of those in the main checkout silently changes what the controller measures: no error, no diff in any PR, nothing in the ledger.

## Why worktree isolation does not cover it

The fleet's isolation story is per-member worktrees plus `./agent-test` port derivation. Both protect members **from each other**. Neither protects the **main checkout**, which every member can write to and which the controller depends on.

A dispatch prompt says "you are ALREADY in worktree `<abs-path>`, do NOT create another". That establishes where a member should work — not what it must not touch.

The failure is also silent in the direction that matters: a broken instrument does not announce itself. A modified `ci-state.mjs` still returns a verdict; it may just not be the right one. The controller has no reason to re-check a script it did not know had changed.

## Bounding it honestly

- Reachability is low — it takes a member losing track of its working directory. Observed once, self-caught, self-disclosed.
- Blast radius when it happens is high and undetectable after the fact: the readings are gone and the file is reverted.
- No wrong decision is known to have been made in this run. Readings taken during that window were consistent with readings taken before it — evidence, not proof.

## Directions, none obviously right

- **Detect rather than prevent.** The controller hashes its instrument set at run start and re-checks before each gate decision, refusing rather than reading a changed script. Cheap, and it fails closed.
- **Tell members explicitly.** Dispatch prompts say where to work; they could also say the main checkout is off-limits and why. Weakest option, since this is a slip rather than a decision.
- **Run instruments from an immutable copy.** The controller invokes scripts from a snapshot cut at run start rather than from the checkout. Strongest, and the largest change.

## Relationship to existing tickets

**#389** (the finisher is a writer in a shared tree, with no lock) is the same theme from a different surface — a role assumed read-only that mutates. This ticket is about the *controller's* dependency on a tree it does not own, and its measurement path rather than its source files.

`skills/fleet/skills/run-team/references/isolation.md` covers member-vs-member and the shared docker stack. It does not cover member-vs-main.

Filed by the `/fleet:run-team` controller. `ledger.mjs check` returned `tracker-hit` on #389; read and judged distinct.

feigi:
Closes #436

Sizing: heavy

## What this does

The controller reads every gate probe out of the main checkout, which every
member can write to and which no worktree isolation covers. `instruments.sh`
pins a digest of that set at run start and re-checks it before each gate;
`run-team/SKILL.md` carries the rule, the pin, and the one legitimate re-pin.

Per the brief's ruling, **detect rather than prevent**. Neither ruled-out
direction is built: no immutable copy, and "tell members explicitly" is not the
fix here.

**Behaviour on detection, stated:** exit 0 is the only code that lets a gate
proceed. Exit 1 (the set changed) and exit 2 (the check could not answer) both
refuse, and the rule says to report what the script printed and not re-read the
instrument. A guard that fails open on "could not look" protects nothing, which
is why exit 2 does not follow the permissive convention `staleness.mjs` uses.

Nothing of the kind existed before this branch — re-derived against `origin/main`
at `8787f4f`, the commit this branch sits on:

```
$ git show origin/main:skills/fleet/skills/run-team/SKILL.md | grep -ci instrument
0
$ git ls-tree -r --name-only origin/main -- skills/fleet/scripts/ | grep -icE 'preflight|instrument'
0
```

## Enumerate and declare

**Covered** — the digest is every *tracked* file under `skills/fleet/`, so the
class is covered by construction rather than by a list that rots on the next
commit: the six probes the ticket names, the siblings it does not (the supply
scan, the liveness probe, the claim/release/merge-proof scripts), the shared
`arg.mjs` / `json.sh` / `net.sh` libraries that one edit mutates every node
instrument through, and the runbooks — which the controller also reads out of
that same writable tree, as phase 0 step 0 already knows.

- **Refs are not in the per-gate digest.** This is the brief's *consider*, and I
  decided against it. The corroborating evidence on the ticket is a stray
  `fix/42-slug` branch, which no hash of files can see — but `claim-ticket.sh`
  creates a branch per ticket and `reap.sh` deletes them, in the ref store every
  worktree shares, so ordinary work moves refs several times a wave. A per-gate
  refusal on that is exactly the noise AC-3 forbids, and noise is the one failure
  a guard the controller must run constantly cannot survive. Sweeping for
  unexpected refs is a drain-cadence question, not a gate one. The reason is
  recorded in the script header and in the runbook, and pinned by a test, so the
  next reader does not re-derive it as an obvious improvement.
- **Untracked files under the set.** A `.DS_Store`, an editor swap file or a
  member's scratch output changes no reading the controller takes, and refusing
  on one fires the guard on ordinary runs.
- **`docs/metrics/*.tsv`.** The tier guard reads them, but the fleet also
  *writes* them during a run — they are output, and digesting them would refuse
  on the fleet's own work.
- **`.github/workflows/` and the repo-root `agent-test`.** CI runs from the PR's
  tree on a runner, not from the controller's checkout; `agent-test` is a
  member's test command, not a gate input.

## What it can wrongly refuse

The false-positive class is not the false-negative class, so the suite feeds it
input it must **accept** as well as input it must reject. Accepted: an untouched
tree, an untracked dropping under the set, a bare `touch` (content is hashed, so
the stat cache is not a channel), branch and worktree churn in the shared ref
store, any change outside the set, and a commit that does not move the bytes.

The one real false positive is the controller's **own** mid-run tooling fix,
which legitimately edits this tree. `--pin` is that path, and the runbook puts
the re-pin as the first item of "Fix the tooling mid-run" with the consequence
spelled out — skip it and the next gate refuses on your own fix, which is how a
guard gets ignored rather than fixed.

## Cost

AC-3 asks that an unchanged set cost effectively nothing. One `git ls-files` and
one pass of `shasum`; no network, no `gh`, no fetch, no subprocess per file.
Measured at this head:

```
$ for i in 1 2 3; do /usr/bin/time -p ./skills/fleet/scripts/instruments.sh 2>&1 >/dev/null | grep real; done
real 0.10
real 0.08
real 0.08
```

## Tests

```
$ ./agent-test skills/fleet/scripts
tests 1450, pass 1450, fail 0
```

Mutants, each applied with its diff printed and then restored by `cp`:

- missing-baseline refusal replaced by `exit 0` → kills only *no baseline
  refuses with exit 2*.
- `ls-files` widened to include untracked files → kills the two accept-side
  cases, and nothing on the refuse side.
- the mismatch `exit 1` replaced by `exit 0` → kills every refuse-on-change case.
- the runbook's exit-2 clause reworded into "re-run it and carry on" (the
  fail-open rewrite this exists to prevent) → kills only *the rule states that
  BOTH non-zero codes refuse*.
- the re-pin clause deleted from the tooling-fix section → kills only that pin.

Review of this PR found seven clauses and guards these suites claimed to cover
and did not. Each fix carries its own mutant, applied with the diff printed and
restored by `cp`, plus a control it must NOT catch:

- the not-a-checkout `die` replaced by a silent fallback → now kills only
  *outside a git checkout it refuses instead of certifying nothing*. It did not
  before: the run falls through to `git ls-files`, which fails on the same
  non-git path and also exits 2.
- an exists-but-unreadable baseline made fail-open → kills only the new
  unreadable-baseline case.
- the gate name `a reap` removed from the bullet, the unrelated `reap.sh`
  mention left standing → now kills only *the rule names the gates it runs
  before*. Renaming that `reap.sh` mention instead leaves the suite green, which
  is the control: before, it was the only thing holding the assertion up.
- the fast-forward command reworded away → now kills only *phase 0 step 0 pins
  the set, and only after the fast-forward*. The `indexOf < indexOf` form it
  replaces stayed green, because `-1` is less than anything.
- the re-pin governance bullet deleted, and separately softened to permit
  clearing a refusal → each kills only the new rule pin.
- every prose mention of the instrument path renamed → kills the path
  assertion; renaming the script on disk instead kills it too.
- control: a runbook sentence away from every pinned clause reworded, and the
  untracked-dropping fixture's content changed → both green.

Two negative controls, to show the pins discriminate rather than merely fire:
rewording `instruments.sh`'s `--pin` stderr note, and rewording an unpinned
sentence inside the pinned slice — both applied, both green.

Which suites actually read the script, settled by replacing it with the word
`MUTATED` and running the fleet suite rather than by a literal-path grep: only
`instruments.test.mjs`, every case in it red. `instrument-check-prose.test.mjs`
stayed green, which is correct and is its documented ceiling — it pins that the
path the prose names resolves to a real executable, not what that executable does.

## Seam for #449

#449 wants a run-start staleness check on the same checkout and the brief calls
the two one preflight. **#449 is not claimed this wave and nothing here
implements it.** The seam it would use, named rather than built: phase 0 step 0
now has an explicit "after the fast-forward, before anything reads them" slot,
and `instruments.sh --pin` is the single call sitting in it. A staleness check
belongs beside that call, in that slot, and would want its own exit-code
contract rather than sharing this one — this script answers "did the set move
under this run", not "was the set current when the run began".


feigi:
Review adjudication, deferrals filed.

**Applied** (head `fa65476`) — the seven confirmed defects, each with its own mutant and a control it must not catch; details in the commit body and the Tests section above.

**Deferred and filed:**

| Issue | Label | What |
|---|---|---|
| #1057 | `ready-for-agent` | the instrument rule names two gates its own readers never run — the finisher's label gate and the merge bot's CI gate. Does **not** block closing #436; the remedy is outside its stated scope |
| #1058 | `ready-for-agent` | an unreadable baseline is reported as `no baseline` and prescribed `--pin` — the residual of a finding refuted 0-2 on a broader claim |
| #1059 | `ready-for-agent` | a mode-only change to a tracked instrument passes the digest, and the header names every other exclusion but this one |
| #1060 | `needs-triage` | the INT/HUP/TERM trap repeats the EXIT trap's `rm`; unmeasured on CI's `dash` |
| #1061 | `needs-triage` | phase 0's "Pinning ahead of the fast-forward…" states a counterfactual with nothing marking it as one |
| #1062 | `needs-triage` | `PIN_STEP`'s slice ends on the next paragraph's heading, so renaming it reds the phase-0 pin — reproduced while running this PR's own negative controls |
| #1063 | `wontfix`, closed | the checked-and-did-not-survive record: the `--pin`-as-fail-open-remedy claim (refuted 0-2), the unreadable-*instrument* reading (already pinned), and two framings refuted whose defects were applied anyway |

**Not filed** — the spec's "All eleven obey the three rules above" is #819, already ruled `wontfix` for the same claim on PR #813. Commented there rather than filing a third instance.

Issue #455, opened 2026-08-13
release-ticket.sh:397 — the unresolved-HEAD blocker names the registry root but not which entry to open


feigi:
Deferred from PR #450 review (issue #179). Dimension: **silent-failure**. Severity: **suggestion** — 0 refuters by workflow policy.

## The finding

`skills/fleet/scripts/release-ticket.sh:397`

The blocker PR #450 adds names the registry **root** (`$wtroot`) but not which entry inside it to open, while every sibling arm hands the operator an exact command — so on a repo with several linked worktrees the operator has to scan `.git/worktrees/*/gitdir` to find the one being talked about.

Verbatim, measured:

```
BLOCKED: worktree /…/w/.worktrees/9-x is this claim's but git could not read its HEAD, so its
branch is unknown — inspect its entry's HEAD file under /…/w/.git/worktrees by hand
```

Sibling arm for comparison, same probe: `git worktree unlock <path>, then prune or remove it`.

## The "by hand" framing is correct, and that is not the gap

The reviewer ran all four candidate remedies against a corrupt-HEAD entry and none cleared it:

- `git worktree repair` → rc 0, entry unchanged
- `git worktree prune -v` → rc 0, prints nothing, entry unchanged
- `git worktree remove` and `git worktree remove --force` → rc 128, `fatal: validation failed … '.git' is not a .git file, error code 7`

So there is no command to name. The only gap is **which entry**, not what to do.

## Approach still open — hence `needs-triage`

The suggested fix has two variants and the cheap one is **measurably wrong** in exactly the case the sibling finding documents:

> Name the entry. git names the admin entry after the worktree directory's basename, so `${stray##*/}` resolves it in the common case: `… inspect $wtroot/${stray##*/}/HEAD by hand`. If that guess is unwanted (a relocated worktree keeps its original entry name — measured: entry stayed `9-x` after the directory became `9-x-renamed`), say so instead: `… find its entry by grepping $wtroot/*/gitdir for $stray`.

`${stray##*/}` would make this message assert a path the script never measured, after `git worktree move` — which is the *same defect class* #179 was filed to remove, reintroduced one arm over. The grep-the-gitdir variant is always correct but is a longer instruction, and a third option (leave it) is defensible since the operator is already being sent to inspect by hand.

Needs a call on which, before an agent picks it up.

## Related

The relocated-worktree measurement comes from the same probe as the silent-release defect filed alongside this one from the same review.

feigi:
[...]

PR #1054, opened 2026-08-30
fix(release-ticket): a moved worktree with an unresolvable HEAD blocks instead of releasing silently (#453)

### LINKED ISSUE #453: release-ticket.sh: a MOVED worktree whose HEAD git cannot resolve is released silently — branch deleted, staged work orphaned, rc 0
Deferred from PR #450 review (issue #179). Dimension: **silent-failure**. Severity: **important**. State: **survived** — two independent refuters attempted to refute it and both reproduced it.

## The defect

`skills/fleet/scripts/release-ticket.sh:396`

The blindness PR #450 is built on — git cannot resolve a worktree's HEAD, so the `branch` line vanishes and the `wt` lookup finds nothing — still produces a **silent full release at rc 0** whenever the worktree's directory no longer ends in `/$issue-$slug`, because `stray` is then empty too and the new `unresolved_head` arm is never reached.

Both empties have the same cause seen from two angles:

- `release-ticket.sh:126` — `wt=` keys on the `branch refs/heads/$branch` line, which a corrupt admin HEAD removes.
- `release-ticket.sh:140` — `stray=` matches on the directory path's exact `/$issue-$slug` suffix, which `git worktree move` changes.
- `release-ticket.sh:392-399` — the whole `if [ -z "$wt" ] && [ -n "$stray" ]` block, including the new `unresolved_head` arm, is therefore **unreachable**.
- The registry-count guard at `:123` passes (1 entry, 1 linked worktree) and `has_branch=true`, so the `die` at `:151` does not fire either.

Result: full release. Branch deleted, label removed, `"released":true`, rc 0, **staged uncommitted work orphaned in a directory that is no longer a git repository**.

## Measured — refuter 1

> Confirmed against the official snapshot, not the other agent's copy. `stray` (`release-ticket.sh:140-142`) matches only on the worktree path's exact `/$issue-$slug` suffix, and `wt` matches only on the `branch refs/heads/...` line — a corrupted admin HEAD removes the branch line AND `git worktree move` changes the path suffix, so both come back empty. `has_branch` is true, so the die at `:151` does not fire, the whole `[ -z "$wt" ] && [ -n "$stray" ]` block (including the new `unresolved_head` arm) is skipped, and the script does a full release: branch deleted, label removed, released:true, rc=0, staged uncommitted work orphaned in a directory that is no longer a git repo. Registry has 1 entry and git lists 1 linked worktree, so the completeness guard at `:123` passes. Contrast run with the identical HEAD corruption but WITHOUT the move blocks correctly with the new message and rc=1, confirming the arm exists and is simply unreachable in the moved case. The "pre-existing, not introduced by this diff" qualifier is consistent with the diff, which touches only the helper and the arm ordering, not `stray`/`wt` computation.

```
$ sh p.sh   (probe5.sh with SNAP repointed at the review snapshot)
--- porcelain ---
worktree .../w/.worktrees/9-x-renamed
HEAD 0000000000000000000000000000000000000000
--- registry entries --- 9-x
--- release-ticket.sh 9 x fix --apply ---
$ gh issue view 9 --json labels
$ git branch -d fix/9-x
$ gh issue edit 9 --remove-label in-progress
{"issue":9,"branch":"fix/9-x","worktree":"","label":true,"released":true,"applied":true,"blockers":[]}
rc=0
--- branch after --- main
--- precious.txt after --- work that exists nowhere else
--- reachable? --- fatal: not a git repository: .../w/.git/worktrees/9-x

$ sh p2.sh   (same fixture, `worktree move` line deleted)
BLOCKED: worktree .../.worktrees/9-x is this claim's but git could not read its HEAD, so its branch is unknown ...
"released":false ... rc=1
```

## Measured — refuter 2

> Reproduced against the official snapshot (snapshot script byte-identical to the probe's copy — `diff` clean). Fixture: worktree on fix/9-x with staged uncommitted precious.txt, `git worktree move .worktrees/9-x .worktrees/9-x-renamed`, then `printf 'garbage\n' > .git/worktrees/9-x/HEAD`. Result: `{"released":true,"blockers":[]}`, rc=0, branch fix/9-x deleted, precious.txt orphaned, `.worktrees/9-x-renamed` no longer a git repo.
>
> Mechanism confirmed by reading the code, not just the output: [...] The new `unresolved_head` arm lives entirely inside `if [ -z "$wt" ] && [ -n "$stray" ]`, so with stray empty it is unreachable. Nothing else in the script covers this path — not handled elsewhere.
>
> Pre-existing is also correct: the diff adds only the `unresolved_head` helper and one `elif`, touching neither the `wt` nor the `stray` computation. Claim's scope (needs BOTH the moved dir and the unresolvable HEAD) matches what the evidence shows — not weaker than stated.

```
$ sh p5.sh  (snapshot script)
--- porcelain ---
worktree .../w/.worktrees/9-x-renamed
HEAD 0000000000000000000000000000000000000000
--- release-ticket.sh 9 x fix --apply ---
{"issue":9,"branch":"fix/9-x",...,"released":true,"applied":true,"blockers":[]}
rc=0
--- branch after --- main
--- precious.txt after --- work that exists nowhere else
--- reachable? --- fatal: not a git repository
```

Both refuters also confirmed **pre-existing**: reverse-applying PR #450's diff (`patch -p1 -R`, `grep -n unresolved_head` → no match) gives a byte-identical result, same `"released":true,"blockers":[]`, rc 0, same orphaned `precious.txt`.

## Why deferred rather than fixed in PR #450

Controller ruling on the review. The finding itself states "Pre-existing, not introduced by this diff", and its remedy **adds a new blocking path** where `wt` and `stray` are both empty — that changes what the script *does*. Issue #179's brief scopes that ticket to what the message *claims*, with outcome, exit code and blocker behaviour explicitly unchanged. Real defect, wrong PR.

The finding's own sketch is one candidate, not a settled design:

> `unresolved_head` is currently asked only about `$stray`. Ask it about the listing as a whole before trusting the absence of `wt`: after `stray=` is computed (`release-ticket.sh:140`), if `wt` and `stray` are both empty but ANY linked worktree in `$wt_list` shows the null-OID-and-no-branch shape, block instead of releasing — the same reason the registry count guard at line 123 exists, since an unresolvable HEAD makes the listing's answer about this branch as untrustworthy as a dropped entry does. Roughly: drop the `P=` filter into an `any_unresolved_head()` variant over `$wt_list` (n>1 records only), and add `if [ -z "$wt" ] && [ -z "$stray" ] && any_unresolved_head; then block "a linked worktree's HEAD is unreadable, so whether #$issue has one is unknown — inspect $wtroot by hand"; fi`.

Two competent implementers would build materially different things here, which is the `decided?` test failing:

- **A new blanket blocker** keyed on ANY linked worktree's unresolvable HEAD blocks releases that have nothing to do with the broken worktree — a false refusal is the failure #37 was filed against.
- **Widening `stray`** to key on the registry entry name rather than the directory suffix (the entry stays `9-x` after the directory becomes `9-x-renamed` — measured in the probe above) reaches the same case without a new arm, and reuses the arm PR #450 already added.
- **Doing nothing** is defensible for the `wt`-and-`stray`-both-empty case in general, but is not defensible once the listing shows an unresolvable HEAD.

Needs a maintainer's call on which, and on whether the block should be a `block` or a `die`, before an agent picks it up.

## Related

- #83 — reap.sh and release-ticket.sh take opposite defaults on an unreadable worktree.
- #395 — release-ticket.sh's worktree-registry cross-check.

feigi:
Closes #453

Sizing: heavy

## The defect

A claim whose worktree was **moved** *and* whose HEAD git cannot resolve was released silently. Two lookups go empty from the same claim seen at two angles: the unresolvable HEAD drops the `branch` line, so `wt` is empty; `git worktree move` changes the directory, so the `/<issue>-<slug>` suffix key is empty. Both empty, the whole `[ -z "$wt" ] && [ -n "$stray" ]` block — the `unresolved_head` arm PR #450 added included — never runs.

Reproduced against `origin/main` before touching anything, on a fixture with staged work in the worktree:

```
$ sh skills/fleet/scripts/release-ticket.sh 9 x fix --apply
{"issue":9,"branch":"fix/9-x",...,"released":true,"applied":true,"blockers":[]}
rc=0
branch after: main                        <- fix/9-x deleted
precious.txt: work that exists nowhere else
reachable?  : fatal: not a git repository
```

## The fix

The registry **entry** is what survives a move, so it is the second key. Measured, git 2.50.1:

- `git worktree move` does not rename the entry — it rewrites the entry's `gitdir` file to follow the directory.
- The path git prints in the porcelain is that file's content with `/.git` stripped, echoed **verbatim** with no canonicalisation of its own (a hand-written `gitdir` naming a path through a symlink is printed back through the symlink), which is what lets the comparison be byte equality against the listing rather than a second opinion about the same path.

But an entry is a **name**, and a name cannot prove ownership: git derives it from the directory's basename at `worktree add` and appends a digit when that name is taken, so an unrelated worktree registered first under `<issue>-<slug>` owns that entry while this claim's is `<issue>-<slug>1`. So the entry key is a **corroborated fallback**, not a union:

1. the directory-suffix key answers first, unchanged;
2. the entry key is consulted only when it found nothing;
3. and its answer is credited only where git cannot read that worktree's HEAD.

Step 3 is the ownership test the name cannot give. A worktree that resolves cleanly to a branch which is not this claim's is *demonstrably* not this claim's — the `else` arm's own message says exactly that about it — so it must not refuse the release. One whose HEAD is unresolvable **may** genuinely be this claim's, and refusing is the safe direction there.

Where the entry key is what found the worktree, the blocker states only what the entry establishes — the entry name — and no longer asserts "is this claim's". Same fixture as the repro above, in both modes:

```
--apply   BLOCKED: worktree .../9-x-renamed holds this claim's registry entry
          name 9-x but git could not read its HEAD, so its branch is unknown —
          inspect its entry's HEAD file under <wtroot> by hand
          "released":false  rc=1
dry run   same blocker, "applied":false, "released":false, rc=1
branch after: fix/9-x, main               <- alive
precious.txt: work that exists nowhere else
```

`block`, not `die`: this lands where the unmoved case already lands.

## Why not a plain union — measured

The first version of this fix OR-ed the two keys and let porcelain listing order decide between them. That shipped a regression, on a fixture whose premise is this PR's own operation. Register a decoy under `9-release-ticket` (git suffixes this claim's entry to `9-release-ticket1`), `git worktree move` the decoy to `.worktrees/00-moved-decoy` so its path no longer carries the suffix, and the entry key reaches a worktree the suffix key never would:

| fixture | as a plain union | corroborated (this PR) |
|---|---|---|
| claim's own HEAD corrupt, decoy moved aside | rc 1, `worktree .../00-moved-decoy is this claim's but is not on fix/9-release-ticket` — wrong directory, wrong cause | rc 1, names the claim's own directory and the unreadable HEAD |
| claim removed and pruned, nothing of it left | rc 1, `released:false`, blocked on `.../00-moved-decoy` forever | rc 0, `released:true`, `blockers []` |

The second row is the **false refusal** the ticket rejected the blanket-blocker alternative over — a claim with no worktree, no registration and nothing to hand-release, refused permanently over a stranger sitting on `decoy/9`. It reaches it through the entry key instead of through a listing scan, which is why the existing `a broken worktree elsewhere does not refuse a claim that has NO worktree of its own` control did not catch it: there the sibling's entry carries a different name.

Both rows are now regression tests.

- **A healthy claim in a moved worktree still releases.** It is still on its branch, so `wt` finds it and the stray block stays gated shut regardless of what either key matched.
- **A broken worktree elsewhere does not refuse a claim that has no worktree of its own.** The entry is gone with the worktree, so the fallback's `[ -r .../gitdir ]` guard never opens.
- **A moved stranger holding this claim's entry name does not refuse it either** — new, and the corroboration is what buys it.

What remains: a worktree that took this claim's entry name, was moved, **and** whose own HEAD is unreadable. The script cannot tell that from the claim itself, so it still blocks — the safe direction — and the blocker now names the entry rather than claiming the directory is this claim's.

## Enumerated, and deliberately left

`stray` can still miss a registered worktree of this claim when the entry name is **not** `<issue>-<slug>` — git's collision suffix (`9-x1`) — **and** the directory was moved, so neither key matches. Left deliberately: reaching it means matching entry names by pattern, and this lookup is anchored on an exact `<issue>-<slug>` in both keys precisely because `<slug>` is caller-supplied. The unsuffixed half of that class is covered and pinned.

## Adjacent, not touched

**#455** is about the wording of this same arm's blocker. This change makes the arm fire in a newly-reachable shape, and the message written for it names the entry (`9-x`) alongside the worktree's current path (`.../9-x-renamed`), so *"inspect its entry's HEAD file under `<wtroot>`"* now has the name the operator needs in the same sentence. The rest of #455's wording territory is untouched. #83 and #395 untouched.

## Tests

Seven cases in `release-ticket.test.mjs` for this ticket: the defect in **both** modes (the script is dry-run by default and every other case in this family passes `--apply`), the two false-refusal controls, the moved sibling, the entry-name collision only the suffix key can reach, and the two regression cases above.

Verified at `6cfaa4d`, in the worktree:

```
$ env -u FORCE_COLOR node --test skills/fleet/scripts/release-ticket.test.mjs
tests 112  pass 112  fail 0

tests 1431  pass 1431  fail 0  skipped 0
```

Mutants, each built from the shipped script under a scratch path and never in the checkout, each applied through a guard asserting the target occurs exactly once and each diff printed before running. `--test-name-pattern='#453'`, 7 tests, at `6cfaa4d`:

| mutation | reds |
|---|---|
| the plain union this PR first shipped | both new regression cases, and only those |
| suffix-first ordering but no `unresolved_head` corroboration | the false-refusal case, and only it |
| entry fallback removed entirely | the moved-worktree defect case, and only it |
| drop the `[ -r .../gitdir ]` guard | the no-worktree-of-its-own control, and only it |
| drop `[ -z "$wt" ]` from the stray block's gate | the moved-sibling control, and only it |
| reword an unrelated blocker | nothing |


feigi:
Review findings adjudicated. Applied in `6cfaa4d`; deferrals filed.

**Applied** — the critical regression and the false prose that licensed it:

- **Finding 5** (critical, survived 2-0): the registry-entry key was OR-ed into the stray lookup and could select an unrelated *moved* worktree. Two probes, both measured on this suite's own harness, both reverted to correct behaviour by the fix. Detail is in the PR body's "Why not a plain union — measured" table.
- **Findings 1 + 6 + 10** (one defect): the comment asserting the union "adds no new ambiguity" was false in exactly the moved case this PR exists to handle. Removed, replaced with what the probes measure.
- **Findings 2 + 11** (one defect): `entry` collided with the registry-scan loop's variable. Renamed `entry_wt` / `entry_stray`.
- **Finding 7**: both new regression tests kill their own mutant and survive a change they should not catch; the full mutation matrix is in the PR body.

**Filed:**

- **#1064** (`ready-for-agent`, open) — the residual the finding-1 refuter surfaced: the pre-existing directory-**suffix** key names an unrelated same-basename worktree as "is this claim's". Measured byte-identical before and after this PR's hunk, so out of #453's scope. Remedy open, and best decided with #455.
- **#1065** (`wontfix`, closed) — the `suggestion` band, each in-scope entry checked by running something: finding 3 dissolved by the restructure, finding 4 **refuted** by measurement (deleting that test would ship a mutant green across the whole `#453` family), findings 8 and 9 confirming observations.
- **#1056** — already filed by the controller: the single full-suite failure under a `git archive` snapshot is that snapshot not being a git repo, not a regression here.

Also commented on **#455**, whose cited line and verbatim message this PR moved. Its gap stands on the suffix-key path.

**Tests at `6cfaa4d`, in a worktree:**

```
$ env -u FORCE_COLOR node --test skills/fleet/scripts/release-ticket.test.mjs
tests 112  pass 112  fail 0

tests 1431  pass 1431  fail 0  skipped 0
```

Issue #798, opened 2026-08-22
release-ticket.sh:517 — release_outcome reads an awk that could not RUN as "not registered", writing a false "Deregistered" into the JSON receipt


feigi:
Deferred from PR #788 review (issue #243). Dimension: **silent-failure**. Severity: **important**. Band: **survived** — both refuters tried and failed to refute it.

Out of #243's scope, which is the four `$(... awk ...)` assignments over `git worktree list --porcelain` at `:228`–`:250`. This is a fifth awk site with a worse failure mode: it does not abort, it asserts a wrong state.

## The finding

`skills/fleet/scripts/release-ticket.sh:517`

```sh
       P="$1" awk '/^worktree /{if (substr($0,10)==ENVIRON["P"]) f=1} END{exit !f}'; then
```

The awk runs as an `elif` **condition**, so `set -e` is exempt and a non-zero status means only "false". But awk encodes its answer in that same status (`END{exit !f}`), so "awk could not run" (rc >= 2) and "the path is not registered" (rc 1) are indistinguishable. Control falls to `elif occupied "$1"; then echo Deregistered`.

The result is a factually false post-state in the JSON receipt — "Deregistered — the registration is cleared" — while `git worktree list --porcelain` still contains the registration.

## Measured

Fixture: an ordinary claim, with `GH_DIRTY` dirtying the worktree during the `gh` round trip so `git worktree remove` refuses at rc 128 with **registration and directory both kept**. awk shimmed on PATH to exit 2 only for the program containing `) f=1}`, a string unique to this site (it does not match `:428`'s `/^locked/{f=1}`).

| | blockers[] | artefacts |
|---|---|---|
| baseline | `...is Unreleased — registration and directory both still present` | `{dir:true,worktree:true,branch:true}` |
| shimmed | `...is Deregistered — the registration is cleared, the directory is still on disk` | `{dir:true,worktree:true,branch:true}` |

Exit 2 in both arms. In the shimmed arm the stderr headline also flips to `#9 PARTIALLY RELEASED`, where the shipped baseline test ("a refused worktree removal leaves the label on the issue") asserts `#9 HALTED mid-release — nothing landed`.

So both the receipt's state claim and the headline become false on an awk that merely could not run.

## Why this is a regression of an existing fix

[...]

PR #1068, opened 2026-08-30
fix(release-ticket): capture awk's status in both worktree-state predicates (#454)

### LINKED ISSUE #454: release-ticket.sh:396 — unresolved_head in elif position cannot tell a dead helper from an honest false

feigi:
`locked` and `unresolved_head` answer THROUGH awk's exit status, so an awk that
could not run lands above the range the answer occupies and, in the condition
position every caller uses, is read as the FALSE answer — the permissive one, in
guards whose whole job is to refuse to answer permissively.

Triage ruled option 3 of the three the ticket recorded: guard BOTH predicates.

Sizing: light

## What changed

Each predicate tests awk's status itself and refuses above the range its
answer occupies. Guarded INSIDE each helper, not at the call sites: every caller
is then covered by construction, and no call site changes shape, so `set -e`
still sees at each of them exactly what it saw before. `|| die` cannot serve
here — there is no status left over to die on once the answer has consumed it —
which is what the comment PR #788 corrected already said; it stays true, and the
neighbouring sentence about assignments is now scoped so it cannot be read as
covering the predicates.

## Claims, each with the command that settles it at dfeda07

**The fail-open is live on this tree, end to end through the real script.** The
test added for it is red before the fix, and its failure is the ticket's measured
shape rather than a bare assertion:

```
$ git stash && ./agent-test skills/fleet/scripts/release-ticket.test.mjs   # at f34f93f
✖ a LOCK lookup that could not run refuses instead of releasing (#454)
  blockers: ["git worktree remove refused …: fatal: cannot remove a locked
  working tree … — worktree … is Unreleased"]
✖ an unresolved-HEAD lookup that could not run refuses instead of guessing (#454)
  1 !== 2
```

The lock blocker never fired and the run reached the mutation the precondition
exists to prevent.

**`set -e` cannot catch this** — it is ignored throughout a function invoked in a
condition, so the body runs to completion for a dead awk exactly as for an honest
false:

```
$ for sh in /bin/sh /bin/dash /bin/bash /bin/zsh; do printf '%s: ' "$sh"; "$sh" errexit2.sh | tr '\n' ' '; echo; done
/bin/sh: body ran to completion, rc=1 body ran to completion, rc=2
/bin/dash: body ran to completion, rc=1 body ran to completion, rc=2
/bin/bash: body ran to completion, rc=1 body ran to completion, rc=2
/bin/zsh: body ran to completion, rc=1 body ran to completion, rc=2
```

**`&& return 0` ahead of the status test, not a bare pipeline whose status the
next line reads.** With `&& return 0` the pipeline is a non-final AND-OR element,
where `set -e` is ignored in EVERY context, so a helper called OUTSIDE a condition
still reaches its own guard. Written the bare way that call never reaches the guard
at all: `set -e` kills the script on the pipeline itself, silently, and the process
carries awk's own status out instead of this guard's exit 2 and its `$NAME`-prefixed
diagnostic. It is **not** a fail-open — the bare form answers nothing, it dies
unattributably. Both forms, both shells, awk shim exiting 5, helper called bare:

```
/bin/sh    FORM=guarded CALL=bare exit=2    out=[release-ticket: could not read whether the worktree at /p is locked]
/bin/sh    FORM=bare    CALL=bare exit=5    out=[]
/bin/dash  FORM=guarded CALL=bare exit=2    out=[release-ticket: could not read whether the worktree at /p is locked]
/bin/dash  FORM=bare    CALL=bare exit=5    out=[]
```

The bare form's exit status is always awk's own, never the guard's: with the shim
at 2 it exits 2, at 127 it exits 127. Called in a condition both forms reach the
guard, so the difference is confined to the unconditioned call site.

An earlier revision of this body claimed the bare form would "put the fail-open
back". That was the wrong failure direction and the PR's own review caught it; the
comment in the source said the same thing and was corrected in the same commit.

**Each shim marker is unique to the awk program it selects**, so the other awks in
the run still answer and only the predicate under test fails:

```
$ grep -cF -- '/^locked/' skills/fleet/scripts/release-ticket.sh   # 1
$ grep -cF -- 'nullhead' skills/fleet/scripts/release-ticket.sh    # 3, all but one in comments, which argv cannot match
```

**Every awk left unguarded in the file is the one this PR does not own**, `release_outcome`'s
inline condition — issue #798, same file, explicitly out of scope here:

```
$ grep -vE '^[[:space:]]*#' skills/fleet/scripts/release-ticket.sh | grep -E 'awk'
...
       P="$1" awk '/^worktree /{if (substr($0,10)==ENVIRON["P"]) f=1} END{exit !f}'; then
```

That single remaining line is `release_outcome`'s registration probe. PR #1068's
review rediscovered it from three dimensions; the reproduction and two corrections
to that issue's remedy were posted to #798 rather than applied here.

**Mutation-tested at dfeda07, each mutant against the WHOLE suite.** Dropping
either guard reds that predicate's case and nothing else; widening the guard to
refuse on the FALSE answer as well reds the acceptance case and much of the file
besides; the equivalent rewrite stays green.

```
drop locked() guard            → ℹ pass 1461  ℹ fail 1   ✖ a LOCK lookup that could not run refuses instead of releasing (#454)
drop unresolved_head() guard   → ℹ pass 1461  ℹ fail 1   ✖ an unresolved-HEAD lookup that could not run refuses instead of guessing (#454)
widen  [ $? = 1 ] → [ $? = 0 ] → ℹ pass 1390  ℹ fail 72
equiv  [ $? = 1 ] → [ $? -le 1 ] → ℹ pass 1462  ℹ fail 0
```

**Full suite green at dfeda07:**

```
$ ./agent-test skills/fleet/scripts/
ℹ tests 1462
ℹ pass 1462
ℹ fail 0
ℹ skipped 0
```

**Lint gate green at dfeda07:**

```
$ shellcheck -x -S warning skills/fleet/scripts/release-ticket.sh   # the ci.yml:183 gate
$ echo $?
0
```

## Enumerate and declare

The class is *lookups over `git worktree list --porcelain` whose answer IS awk's
exit status*.

Covered — both predicates, guarded once inside each, which covers every call site
in the file by construction (`grep -nE '(^|[^_a-z])(locked|unresolved_head) "'`):
the lock precondition on the claim's own worktree, the lock arm of the stray
chain, the unresolved-HEAD `elif` of that chain, and the registry-entry corroboration.

Deliberately left — `release_outcome`'s inline `printf | awk` condition, the same
status-collapse shape in the same file. That is #798, named out of scope by this
ticket's brief; this PR's review reproduced it and posted the evidence there. The assignments over the same listing are already guarded with
`|| die`, which is correct for an assignment and is left alone.

Also left — sibling fleet scripts. `reap.sh`, `inflight.sh` and `worktree-audit.sh`
read the same porcelain, but through assignments, not through a predicate's exit
status, so the class does not extend to them.

Nothing at any call site: the `if`/`elif` chain and the `[ -n … ] &&` guards are
byte-identical, so what `set -e` covers there is unchanged. Inside each helper it
adds one new exit path — `die`, exit 2, this script's unanswerable verdict, the
same code and voice its sibling lookup guards already use. The ticket's own AC-2
is the criterion this restructuring could newly violate, which is why the
acceptance case exercises both the TRUE and the FALSE answer of both predicates
and is the mutation target above.

## Review fixes applied at dfeda07

PR review ran 6/6 dimensions. What landed on top of 4c74b0b:

- **The `locked()` comment's failure direction was wrong** and is corrected — see the
  `&& return 0` claim above. A correction for a permissive guard must not ship a
  fresh false claim of the same family.
- **`awk_rc` dropped.** It was a shell global both guards shared; `[ $? = 1 ]` reads
  the same status inline. Behaviour-identical across both shells, every awk status,
  `set -e`/`set -u` on and off, bare and conditional call sites. The mutually
  exclusive alternative (`local awk_rc=$?`) was refuted: it fails the `ci.yml:183`
  shellcheck gate with SC3043 and would be the first `local` in an sh-dialect file
  in this repo.
- **`assert.equal(code, 2, …)` in the LOCK-refusal test is annotated** to say it does
  not discriminate its own mutant — the pre-guard run reaches `git worktree remove`,
  halts, and also exits 2. The empty receipt and untouched tracker are what red.
- **A result binding renamed** from `d` (with no `c` in scope) to the fixture it releases.

Deferred, each with a durable home: #1073 (a pre-existing stale test-case count in
this file's prose, untouched by this diff), #1074 (the checked suggestion band —
refuted and refused entries, closed as `wontfix`), and a comment on #798.

Closes #454


feigi:
Review deferrals filed, one durable home each:

- **#1073** — `release-ticket.test.mjs:398` asserts a live count of this file's own test cases ("all 98 cases"); the file now runs 115. Pre-existing and untouched by this diff, so deferred rather than widened into this PR. `ready-for-agent`; the remedy is the by-shape wording, not a fresh number. Dimension: `comments`.
- **#1074** — `PR #1068 review: the suggestion band, checked`. The `local awk_rc=$?` alternative, refuted by measurement (SC3043 against the `ci.yml:183` gate); the proposal to delete the awk-healthy control, refused because that test is issue #454's acceptance criterion 2; and one positive confirmation. `wontfix`, closed — finished records, not queued work. Dimensions: `types`, `simplify`.
- **#798, [comment](https://github.com/feigi/claude-config/issues/798#issuecomment-5469012015)** — `release_outcome`'s registration probe at `:685`, rediscovered here from three dimensions (`correctness`, `silent-failure`, `types`), each 2-0. Already `ready-for-agent` and named out of scope by #454's brief, so it went as a comment rather than a duplicate issue. The comment carries two corrections to that issue's own remedy: a `die` cannot serve there because the function is called inside `$( )`, and the remedy sketch's bare pipeline yields an **empty** outcome under `set -e` (measured, `/bin/sh` and `/bin/dash`).

Fixes pushed at `dfeda07`. Suite 1462/1462, shellcheck gate exit 0, mutation-checked against the whole suite — details in the PR body.

Issue #584, opened 2026-08-18
fleet scripts: ledger.mjs absorbs an unknown --flag into the check subject, flipping the duplicate-filing guard from exit 1 to exit 0


feigi:
Deferred from PR #575 review (dimensions: **correctness**, **silent-failure**, **types** — three specialists landed on the same defect independently: `survived-0` correctness/important, `refuted-0` silent-failure/important, `unverified-4` types/suggestion).

`skills/fleet/scripts/ledger.mjs:80`

## The behaviour — real, reproduced by three independent agents

`ledger.mjs` splices `--file`/`--require-file` out of `argv` itself and then takes `const [cmd, ...rest] = argv`. Every *other* `--` token stays positional, and for the `check` subcommand `rest.join(" ")` (`ledger.mjs:205`) folds it straight into the **duplicate-filing subject**. So one mistyped flag flips the duplicate-filing gate from exit 1 to exit 0:

```
$ node ledger.mjs --file l.md filed 123 "widget guard missing"     # seed, exit 0
$ node ledger.mjs --file l.md check "widget guard missing"
  ledger: ALREADY FILED — #123 widget guard missing   ...  "verdict":"already-filed"   exit=1
$ node ledger.mjs --file l.md check --requre-file "widget guard missing"
  {"subject":"--requre-file widget guard missing","found":false,"verdict":"unverified"}  exit=0
```

`~/.claude/skills/fleet/commands/review-and-fix.md:10` instructs the model to run `ledger.mjs check "<subject>"` and gate filing on it, so a misspelt flag silently searches for a *different* subject, an existing duplicate goes unfound, and a second issue is filed. That is the exact fail-open #365 names, in the script #365 names by name.

Partial mitigation, measured, and it does not close the gap: the run still prints `ledger: near-miss 1.00 — #123 widget guard missing` and `review-and-fix.md:10` warns that exit 0 is not automatically "safe to file". A caller reading stdout would likely catch it; a caller gating on `$?` alone — which the same doc invites — files the duplicate.

Scope check, measured: of the non-test fleet scripts, `board`/`ci-state`/`diff-stats`/`pr-overlap` all call `sweep()` after #575; `candidates.mjs` has its own stricter `parseArgs` (deliberately excluded, documented in `arg.mjs`); `compute-board`, `compute-spend`, `fleet-tick`, `strip-comments` read no flags. **`ledger.mjs` is the only flag-reading fleet script left unswept.**

## Why it is unowned

[...]

PR #1069, opened 2026-08-30
fix(arg): refuse a bare or single-dash positional stray, not just an unknown --flag

### LINKED ISSUE #463: fleet scripts: board.mjs and ci-state.mjs silently ignore stray positional arguments — candidates.mjs refuses them for free via strict parseArgs

feigi:
## Summary

`makeSweep()` (arg.mjs) only ever refuses a `--`-prefixed token — deliberate, per #365's triage ruling, since it must stay ignorant of which names take a value (board.mjs's `build`/`serve` subcommands, and every valued flag's own value, both look positional to it). That left a bare word or single-dash token nothing reads to ride through in total silence: `ci-state.mjs --pr 42 basee main` computed a real, wrong verdict against the default `base` at exit 0/1 — the fleet's own merge gate — and `diff-stats.mjs`/`pr-overlap.mjs` had the same gap.

Widening the sweep itself can't fix this without reopening #365's ruling: `startsWith("-")` refuses a legitimate negative value like `--spend-since -1` (which `arg()` accepts today), and widening past `--` would refuse `board.mjs`'s own subcommands.

This adds `makeStray()` beside `makeSweep()` in `arg.mjs` — a new guard, not a change to the sweep. It takes the set of flag names that consume a value (so it can skip over one, whatever it looks like) and, for board.mjs only, the one positional slot its subcommand fills. Placed after each script's own `sweep()` call and before any `gh` query.

## The class is five scripts; the ticket names four

#463's ACs name `ci-state.mjs`, `diff-stats.mjs`, `pr-overlap.mjs` and `board.mjs`. There is a fifth `makeSweep` consumer:

```
$ grep -rln 'makeSweep\|makeStray' skills/fleet/scripts/*.mjs
  arg.mjs  arg.test.mjs  board.mjs  ci-state.mjs  diff-stats.mjs
  pr-overlap.mjs  staleness.mjs
```

`staleness.mjs` was left unguarded by the first two commits and is now wired the same way. Measured before the fix:

```
$ staleness.mjs --path README.md --present needle
$ staleness.mjs --path README.md --present needle JUNKTOKEN
  byte-identical payloads; JUNKTOKEN never mentioned
$ staleness.mjs --path README.md --present two words
  {"needle":"two", ...} — `words` silently discarded
```

The unquoted-needle shape is why this member is not the least important one: `staleness.mjs`'s verdict feeds phase-0 ticket selection, so a confident `fixed`/`live` about a truncated needle retires live supply or claims dead work. Both invocations now exit 2 naming the token.

`candidates.mjs` stays untouched — its own `parseArgs` already refuses these for free and is the reference behaviour named in the ticket. `ledger.mjs` stays untouched — its free-text tail is deliberately unswept (#584), a separate ticket with its own ruling.

## Also in this PR, from its own review

- **The refusal names the flag given wrong, not the innocent token behind it.** For a value flag read BELOW `stray()`, a missing value surfaced as `unexpected argument '<next token>'`. `board.mjs build --prev --port 9000` named `9000` (`--port`'s value) and `ci-state.mjs --pr 42 --workflow-file --base main` named `main` (`--base`'s value). Both reads are hoisted above the guard, restoring `--prev needs a value` / `--workflow-file needs a value`, each with its own pin. The ci-state read sits immediately above `sweep()` and no higher: past the boolean guards it would take the `--declare-no-ci=` refusal off the guard that words it better. Three further `board.mjs` reads — `--port`, `--interval`, `--spend-since`, all inside `serve()`/`gather()` rather than at branch top — still name the innocent token; that needs a restructure of those functions' argv fallback and is deferred to #1076.
- **One `VALUE_FLAGS` list per script**, with the sweep's larger set derived from it, instead of the same names spelled once per `sweep()` and once per `stray()` call. `board.mjs`'s accepted-flag line reorders as a result (`--open` now last); nothing asserts that order.
- **`makeStray()`'s `const name = a.slice(2).split("=")[0]` dropped** — it is read only under `!a.includes("=")`, where the split is a no-op.

## What's covered / deliberately left

**Covered** — every case the ticket names, plus the fifth script: bare positional (`ci-state.mjs --pr 42 basee main`), single-dash (`-basee`), `board.mjs build`/`serve` still working with a stray alongside either refused, `--spend-since -1` still accepted, existing `--`-flag sweep behaviour unchanged, and `makeSweep()`'s own comment updated to point at the new guard instead of leaving the class read as closed.

**Deliberately left**: `candidates.mjs` and `ledger.mjs` as above; the three `board.mjs` reads in #1076; and a per-script flag-arity table deriving both lists from one declaration, deferred to #1077.

## Testing

- `skills/fleet/scripts/arg.test.mjs`: a `STRAY_POSITIONALS` matrix mirroring the existing `#365` `STRAYS` table (exit 2, stray named, `gh` never called) across every wired script, including both the bare and single-dash form on `ci-state.mjs`, a second `board.mjs` row whose extra token is itself a declared subcommand, and a `staleness.mjs` row. Plus unit tests on `stray()` via a throwaway consumer: the declared-positional slot, a SECOND declared positional, the value-flag skip that lets `--spend-since -1` through, the drift direction (a boolean flag's name in `valueFlags` hides the next token), stray-before-flags ordering, and that a `--`-prefixed token is left to `sweep()` alone.
- `skills/fleet/scripts/board.test.mjs`: the `serve`-path regression (`serve --port 0 junk`), which needs the existing `serveArgs()`/`serveOpts()` harness rather than the generic matrix; and the `--prev` ordering pin.
- `skills/fleet/scripts/ci-state.test.mjs`: the `--workflow-file` ordering pin.
- Mutation-verified, each against a scratch copy of `arg.test.mjs` + `board.test.mjs` + `board-cli.test.mjs` + `ci-state.test.mjs` (baseline 158/158):
  - `makeStray()` reverted to a no-op → reds only stray-specific tests, nothing pre-existing.
  - the `usedPositional` gate deleted → reds exactly the two new second-positional pins (156/2). Before those existed the same mutation left everything green, while `board.mjs serve --port 0 build` went from exit 2 to exit 0 with the server actually starting.
  - the value-flag skip made unconditional → reds the drift test alone (157/1).
  - either hoisted read put back below the guard → reds exactly that flag's ordering pin (157/1, twice).
  - `staleness.mjs`'s `stray()` call deleted → reds its matrix row alone (157/1).
  - A naive widening that drops the value-skip entirely reds every accept case that passes a value, `board-cli.test.mjs`'s `--spend-since`/`--interval` suite and the `-1` unit case included — those tests already discriminate this exact failure mode.
  - Benign controls: renaming the gate's internal binding, and reordering a `VALUE_FLAGS` array, both leave 158/158 green.
- Full suite: `./agent-test skills/fleet/scripts/` — 1477/1477 pass.

Sizing: heavy

Closes #463


feigi:
Review findings applied at `6202af2`. Deferrals filed:

- **#1076** (`ready-for-agent`) — `board.mjs`'s `--port`/`--interval`/`--spend-since` reads still sit below `stray()`, so a missing value names an innocent token. Confirmed defect, remedy open: hoisting them means restructuring `serve()`/`gather()`'s argv fallback, which #463's brief does not cover.
- **#1077** (`ready-for-agent`) — the `valueFlags` arity invariant is still per-call-site prose. The within-file half is closed here (one `VALUE_FLAGS` list per script, sweep derived from it) and the untested drift direction is now pinned; what remains is whether a per-script flag-arity table is worth the seam.
- **#1078** (`wontfix`, closed) — the suggestion band, checked. `unverified[2]` (an accepted-flag list in `stray()`'s refusal) refuted by measurement: `stray()` never receives the boolean flags, so the suggested list tells a caller who typed `-quiet` that `--quiet` is not accepted. `unverified[5]` had no action.

One finding was considered and dropped without a ticket: `runStray()` duplicating `runSweep()`'s probe rig, which the finding itself calls defensible to leave.

Issue #446, opened 2026-08-13
review-pr.js:459 — the snapshot dir is `mkdir -p` + `tar -x`, so a reused scratch OVERLAYS one PR onto another; the byte-identity claim at :226 is then false


feigi:
Measured live during `/fleet:run-team` today. It produced **two false findings** on PR #441 that asserted the branch could not claim a green suite.

## Mechanism

`workflows/review-pr.js:459-460`:

```sh
mkdir -p ${scratch}/snapshot
git -C ${worktree} archive HEAD | tar -x -C ${scratch}/snapshot
```

`mkdir -p` does not clear an existing directory, and `tar -x` **overlays** — it writes the archive's entries and leaves every pre-existing file that the archive does not contain. So when two reviews are handed the same `scratch`, the second review's snapshot is the **union** of both PRs' trees, not the second PR's tree.

The script states the opposite as a guarantee, at `:226`, in the text handed to every specialist:

> The snapshot IS the source of truth: 'git archive HEAD', byte-identical to ...

Under a reused scratch that sentence is false, and it is false in the one direction that matters: specialists are explicitly instructed that anything disagreeing with the snapshot is a probe artifact (`:229`), so the contaminating files outrank a specialist's own correct observation.

## Measured, this run

Reviews of PR #440 then PR #441 were both given `scratch=<session>/scratchpad`, so both resolved to `<session>/scratchpad/snapshot`.

```
$ git ls-tree --name-only 6bf8025a -- skills/fleet/scripts/ | grep -c member-prompt-prose
0                       # NOT in PR #441's tree
$ git ls-tree --name-only origin/main -- skills/fleet/scripts/ | grep -c member-prompt-prose
1                       # only in main, after #440 merged
$ ls .../scratchpad/snapshot/skills/fleet/scripts/member-prompt-prose.test.mjs
.../member-prompt-prose.test.mjs        # present in #441's snapshot anyway
```

Provenance check — the snapshot is a genuine mixture, not simply the wrong tree:

```
6bf8025a     run-merge-bot.md: MATCHES snapshot     # PR #441's own file: correct
origin/main  run-merge-bot.md: differs
09f99147     run-merge-bot.md: differs
member-prompt-prose.test.mjs: differs from merged #440   # stale copy from #440's PRE-fix snapshot
```

So the PR-under-review's own files were right, and a foreign file from the previous review sat beside them.

## The false findings it produced

[...]

Issue #274, opened 2026-08-10
resolveDimensions accepts a repeated dimension key — two specialists, one scratch dir, double-counted coverage


feigi:
`resolveDimensions` accepts a repeated dimension key, so the fan-out dispatches the same specialist twice against one scratch dir.

**dimension:** correctness / types (raised twice by two specialists — `unverified-2` and `unverified-8` of the same review; filed once)
**file:** `workflows/review-pr.js:407` / `:419`

## Finding

`resolveDimensions(["correctness","correctness"], DEFAULT_DIMENSIONS)` returns a 2-element array whose entries are the *same* `DEFAULT_DIMENSIONS` object reference — no throw, no warning. The fan-out then derives the specialist scratch dir, the agent label and `dimensionsRun` from `d.key`, so:

- `scratch dirs: ['/tmp/review-pr-271/correctness/', '/tmp/review-pr-271/correctness/']` — 1 distinct of 2
- `labels: ['review:correctness', 'review:correctness']`
- `dimensionsRun: ['correctness', 'correctness']` — the coverage record the caller reads back double-counts one dimension (`skills/fleet/skills/run-team/SKILL.md:403` reads it as coverage)

Suggested fix at the time: dedupe by key before the empty check, e.g. `const deduped = [...new Map(resolved.map((d) => [d.key, d])).values()];`.

## Why deferred — read this before re-arguing it

**The adversarial pass rejected this substance unanimously, 2-0**, as `refuted-2` of the same review. Do not re-run that argument from scratch; the two refutations were:

1. **Not opened by this PR.** Pre-PR, `explicitDimensions || selectDimensions(...)` passed a caller-supplied `[obj, obj]` straight to the fan-out with the identical result. The PR added *string-key* resolution and required-field checks on objects; duplication was always reachable and is unchanged. A duplicate object override behaves the same before and after the diff.
2. **The cited design comment does not cover it.** `workflows/review-pr.js:425-430` is the Snapshot section header, about *read* collisions on the shared immutable tree (mutation probes, mid-review worktree edits, racing specialists). It says nothing about scratch-dir *write* uniqueness, so the "sibling-collision class the design exists to prevent" framing borrows that comment's authority for a different concern.
[...]

PR #1082, opened 2026-08-30
fix(review-pr): give every refuter its own scratch dir, and the cd/toplevel rules (#496)

### LINKED ISSUE #496: review-pr.js:664-676 — the workflow's refuter prompt shares one scratch dir per dimension and carries no cd-chaining/toplevel rule; #258's exclusion rationale covers only half of it
Deferred from PR #495 review (issue #258, `fix/258-refuter-scratch-isolation`).

Two halves of one gap in the refuter prompt `workflows/review-pr.js` dictates. This is the fleet's **default** review path — `run-team/SKILL.md`'s Reviewers section dispatches `Workflow({name: "review-pr"})` — so it is the path most refuters actually take.

## Half 1 — one-level scratch dir, shared by every refuter of a dimension

`workflows/review-pr.js:676`

```js
Scratch: ${scratch}/verify-${d.key}/`,
```

That string is built inside `Array.from({ length: n }, (_, i) => () => agent(...))` at `:662`, so **all `n` refuters of one dimension are handed the identical path**. Only the `Lens ${i + 1}` line at `:675` varies per refuter. This is exactly the collision shape #258 measured and fixed on the hand-dispatch path: a refuter's mutation matrix landing on a sibling's generic filenames in a shared scratchpad.

The fix #258 shipped is a **two-level** namespace, `<scratch>/pr<N>/<finding>/`. The workflow's is one level and keyed on the dimension, not the finding — and a dimension is precisely the unit that fans out.

**dimension:** `correctness`

## Half 2 — no `cd` chaining rule, no toplevel assertion

`workflows/review-pr.js:664-671`

```
Try to REFUTE this finding from PR #${pr}. Default to refuted=true if uncertain.
...
Verify against the snapshot ${snap.path} by RUNNING something — compile it, run
the test, apply the mutation. Do not reason your way to agreement.
```

Nothing here says `cd "$D" && git …` rather than `cd "$D"; git …`, and nothing asks for a `git rev-parse --show-toplevel` check around a fixture's own `git init`/`git commit`. PR #495 added both to the two hand-dispatch copies (`run-team/SKILL.md`'s fix-applier block and `review-and-fix.md` step 2). This third live site was left as it was.

**dimension:** `comments`

## Why this was not fixed in PR #495, and why the stated reason is incomplete

#258's brief puts this path explicitly out of scope, and the PR is correct not to have touched it. The brief's line 32:

> The `review-pr` workflow path is **not** affected in the same way — it provisions `scratch` per run and passes it down — which is part of why this only surfaces on the fix-applier's own hand-dispatched refuters.

That is true, and it does neuter the **cross-PR** collision — two fix-appliers on different PRs cannot land in each other's directories when each run gets its own `scratch` root. It does not cover either half above:

- **Half 1** is a collision *within* one run, between the `n` refuters of a single dimension. Per-run provisioning is upstream of it; they share the sub-path regardless.
- **Half 2 has nothing to do with scratch provisioning at all.** A `cd` that silently fails leaves the following `git` running in whatever directory the agent was already in — the checkout. Where scratch was provisioned is irrelevant to that; the hazard is orthogonal and live on the workflow path. It is the failure that produced commit `020d6ea` during PR #488's fix-applier run, sweeping up an uncommitted `settings.json` change in the maintainer's live checkout.

So the exclusion is right on scope and incomplete on rationale. Filing rather than widening PR #495, which is deliberately refuter-shaped and hand-dispatch-shaped.

## Not a duplicate

`ledger.mjs check` returns `verdict: "clean"`, `found: false`, nearest row 0.09 — but that is an ANDed-query near-match result and is **not** a certification that the tracker has nothing. Checked by hand against the closest existing issues:

- **#274** — `resolveDimensions` accepts a repeated dimension key, giving two specialists one scratch dir. Different cause (duplicate keys in the dimension list); Half 1 happens with a perfectly well-formed dimension list.
- **#446** — `review-pr.js:459`, a reused snapshot dir overlaying one PR onto another via `mkdir -p` + `tar -x`. That is the snapshot, not the refuters' write area.

Neither covers either half.

The defect is verified but the approach is open. Half 1 needs a decision on what the second level keys on — the workflow has no stable per-finding id at that point in the loop (`f` carries `claim`/`file`/`line`, and the `unv<N>` ids are assigned later, at report assembly), so it is a choice between the lens index `i`, a hash of the claim, or hoisting the id assignment earlier. Half 2 needs a ruling on whether the workflow prompt should inline the same prose the two hand-dispatch copies now carry, or whether all three should read from one shared source — the duplication that made PR #495 patch two documents in lockstep is itself the thing worth deciding on.

feigi:
Closes #496

Sizing: light

The `review-pr` workflow handed every refuter of a dimension the same scratch
directory, and its refuter prompt carried neither the cd-chaining rule nor the
toplevel assertion the two hand-dispatch briefs gained in PR #495 (#258). This is
the fleet's default review path — `run-team/SKILL.md` dispatches
`Workflow({name: "review-pr"})` — so it is the prompt most refuters take.

## Half 1 — the class is two axes, not one

The brief rules the second level onto the lens index. Implementing exactly that
leaves half the collision standing, so the fix keys both axes.

The fan-out under a dimension nests: `parallel` over the dimension's findings,
and `Array.from({ length: n }, ...)` over each finding's lenses inside that. A
path keyed on the dimension alone is therefore shared by every refuter under the
dimension — across findings as much as across lenses. AC-1 asks that "two
refuters of the same dimension" differ, and the same lens number on two findings
is such a pair.

Measured rather than argued. Narrowing the path to the lens index alone — the
ruling as literally worded — leaves exactly one pin red, the one asserting two
findings of a dimension differ:

    L=$(mktemp -d) && git archive HEAD | tar -x -C "$L" \
      && node -e 'const f=process.argv[1],fs=require("fs");const s=fs.readFileSync(f,"utf8");const o=s.replace("f${fi + 1}-l${i + 1}","l${i + 1}");if(o===s)throw new Error("mutant applied nothing");fs.writeFileSync(f,o)' "$L/workflows/review-pr.js" \
      && cd "$L" && node --test skills/fleet/scripts/review-pr-refuter-scratch.test.mjs
    # -> only "the same lens on two findings of one dimension ..." fails

`fi` comes from the findings `.map` that was already there. No `unv<N>` id is
hoisted and no claim is hashed, per the ruling.

## Half 2 — the rules, inlined

The cd-chaining rule and the toplevel assertion are inlined in the terms the two
hand-dispatch copies use, rather than factored into a shared source: one site is
a `.js` template and the others are Markdown. The three-copy drift hazard is real
and is its own ticket.

    grep -Fc 'cd "$D" && git …' skills/fleet/skills/run-team/SKILL.md \
      skills/fleet/commands/review-and-fix.md workflows/review-pr.js

`-F` is load-bearing: this repo's grep is ugrep, which reads the bare pattern as
a regex and reports every file as a non-match.

A `cd` that silently fails leaves the following `git` running wherever the agent
already was — the checkout. That is what produced `020d6ea`, whose whole diff is
a `settings.json` edit swept up from a maintainer's tree.

    git show --stat --format='' 020d6ea

## Enumerate and declare

Every prompt `review-pr.js` builds, its fan-out, and whether siblings share an
interpolated path:

| prompt | fan-out | shared path? | this PR |
|---|---|---|---|
| snapshot | one per run | no sibling exists | left |
| specialist | one per dimension, keyed on the dimension | no sibling exists | left, per the ruling |
| refuter | findings × lenses under each dimension | yes, on both axes | **covered, both axes** |

Deliberately left, each with its reason:

- **The specialist scratch path.** Ruled out of scope: specialists are
  one-per-dimension, so there is no sibling to collide with. A repeated dimension
  key would manufacture one, and that is #274 — a different cause.
- **De-duplicating this prose across the three sites.** Ruled out of scope.
- **The snapshot directory's cross-run reuse.** #446, a different write area.

- **Adds a level; does not relocate.** AC-3 is pinned by a test that renders
  several finding/lens cells and asserts each stays under the run's provisioned
  `scratch` root — and, separately, that none climbs back out through a `..`
  segment, which a prefix match alone carries.
- **Confines refuter writes in one unbroken clause.** The write ban is pinned by
  a single contiguous regex spanning "nowhere else", "never write targets" and
  the `git show` exemption — the discipline the cd rule already used, and the one
  the sibling prose pin for the hand-dispatch copies already applied. Asserted in
  halves it was not pinned at all: deleting either half, or splicing an exception
  between them, left the file green.
- **Pins what the path segments mean, not only that they differ.** Every
  difference assertion is satisfied just as well by swapping the finding and lens
  indices at the interpolation. The paths stay distinct, so the collision this PR
  fixes is still prevented; what breaks is the prompt's own sentence about `f<N>`
  and `l<N>`. One literal equality keeps that sentence true.
- **Changes the prompt every refuter reads.** A prompt gaining write rules can
  gain one that stops a legitimate verification, so a pin runs the other way:
  the prompt must still order a real run, and must still permit
  `git show`/`git archive` at a pinned ref. Cutting either reddens it — a
  refuter that can neither run nor read the PR's objects refutes nothing while
  defaulting every finding to refuted.

## Tests

The pins render the refuter prompt rather than grepping it, because the criterion
is a property of two rendered prompts: they must differ in the scratch path and
not only in the lens line, which no regex over the source can see. The template
literal is lifted out of the comment-stripped source and evaluated, and the
assertions read its output.

Rendering binds the two indices from the test file, so a render-only pin stays
green if the call site stops varying them — a lift tests a copy. A further pin
closes that by asserting the two fan-out expressions bind the names the template
interpolates.

Every pin was mutated alone in a `git archive` copy carrying no `.git`, and each
killed its own mutant and only its own: reverting the path level, narrowing it to
the lens, cutting the cd-chaining sentence, cutting the toplevel clause, cutting
the pinned-ref permission, and unbinding the findings index. Reflowing the added
prose across a different line break and rewording the lens question left every
pin green, so the pins discriminate rather than merely detecting an edit.

The pins added under review were mutated the same way, against a `git archive`
copy of the shipping commit. Deleting "and nowhere else", deleting the checkout
sentence, and splicing `, unless a fixture is easier to build in the checkout`
between them each redden the write-ban pin alone, while the same words rewrapped
across different line breaks stay green. Dropping `fi` from the findings fan-out
and `i` from the lens fan-out redden the fan-out pin both plainly and with the
removed form repeated in a trailing comment on that same line — `stripComments`
blanks whole-line comments only, so the unanchored pattern read the regression
back out of the comment — while respacing either argument list, and wrapping the
`.map` onto its own line, stay green. Swapping the two interpolated indices, and
separately reordering this file's own `SCOPE` list, each redden the pin on the
`f<N>`/`l<N>` mapping; both were green before that pin existed. Splicing `../`
into the path template reddens the escape assertion — and the mapping pin with
it, since that one asserts the whole rendered line.

Against the pre-fix prompt every pin goes red except the provisioned-root one,
which held before this ticket and had to:

    B=$(git merge-base origin/main HEAD) && L=$(mktemp -d) \
      && git archive HEAD | tar -x -C "$L" \
      && git show "$B:workflows/review-pr.js" > "$L/workflows/review-pr.js" \
      && cd "$L" && node --test skills/fleet/scripts/review-pr-refuter-scratch.test.mjs

The merge-base, not `origin/main`: the commit bodies spell this with `origin/main`,
which is pre-fix only until this merges. The merge-base keeps resolving afterwards.

Full suite green.

    node --test skills/fleet/scripts/*.test.mjs


feigi:
Review findings applied at `099e0c4`; deferrals filed.

**Applied** — the write ban pinned contiguously (deleting either half, or splicing an exception between them, left the file green before this); the two fan-out pins anchored so a trailing comment cannot smuggle the removed form back past `stripComments`, with whitespace tolerance for a benign reformat; the `f<N>`/`l<N>` mapping pinned by literal equality rather than by difference alone; the provisioned-root assertion extended to reject a `..` escape; `render`'s single-valued `dimension` parameter dropped.

**Filed**
- #1083 `ready-for-agent` — `verify-*` refuter scratch dirs are never cleaned and the default root is per-PR, so a re-review reads the previous run's mutants. Raised against this PR and correctly refuted as a defect *of* this PR; the mechanical facts are pre-existing and the refutation is recorded in the body.
- #1084 `ready-for-agent` — the specialist prompt at `workflows/review-pr.js:897` carries neither rule this PR added to the refuter prompt. Remedy open: the shared-const route is ruled out by #496's brief.
- #1085 `needs-triage` — every refuter of a dimension is dispatched under one identical `verify:<dim>` label. Not confirmable from inside this repo; the host's label semantics decide it.
- #1086 `wontfix`, closed — the suggestion band, checked: the `SCOPE`/`render` restructure refuted by a dispatched refuter (it measured the positional form catching the same rename just as loudly, and the restructure adding a mostly-silent reverse direction), and the request to delete the "not merely the lens line" test refused on #496's AC-1.

Suite green from the worktree: `./agent-test skills/fleet/scripts/` — 1490 tests, 1490 pass, 0 fail (1488 before, plus the two pins added here).

Issue #1076, opened 2026-08-30
board.mjs still names an innocent token when --port/--interval/--spend-since is given no value


feigi:
Deferred from PR #1069 review (issue #463).

**dimension:** `correctness` (finding `unverified[1]`, extended by its refuter)

## The defect

`makeStray()` refuses a leftover positional token. For a value flag whose `arg()` read sits BELOW the `stray()` call, a missing value now surfaces as `unexpected argument '<the next, innocent token>'` instead of `--<flag> needs a value` — the refusal names a token the caller got right.

PR #1069 fixed the two reads that are plain statements: `board.mjs`'s `arg("prev")` and `ci-state.mjs`'s `arg("workflow-file")`, both hoisted above the guard with their own pins.

Three reads in `board.mjs` are left, all measured by the review's refuter against `cb78fca4` and still true at `6202af2`:

```
board.mjs serve --port --open x       → unexpected argument 'x'   (want: --port needs a value)
board.mjs serve --interval --open x   → unexpected argument 'x'   (want: --interval needs a value)
board.mjs build --spend-since --open x → unexpected argument 'x'  (want: --spend-since needs a value)
```

`file:line` — `skills/fleet/scripts/board.mjs:53` (`argPort()`), `:67` (`argInterval()`), `:481` (the `--spend-since` read inside `gather()`).

## Why it was deferred rather than fixed

These three are not statements at the top of a branch; they are `arg()` calls inside `argPort()`/`argInterval()` (called from `serve()`) and inside `gather()`. Hoisting them means `main()` reading the values and passing them in — a change to the argv fallback of an exported function that in-process callers also use, which is a restructure #463's brief ("a new guard alongside the sweep, not a change to it") does not cover.

## Remedy is open

Two candidates, neither settled:

1. Hoist all three reads into `main()` above `stray()` and pass them down, keeping `serve({...})`'s in-process contract intact.
2. Make `stray()` itself refuse when a value flag's next token starts with `--`. The refuter measured this: it fixes `build --prev --port 9000` but NOT `serve --port --open x` (there `x` is a genuine trailing token, so `stray()` fires regardless of the skip rule), and a differential run showed it changes `stray()` in 165/78642 argv combinations no test observes. Less complete and less pinned than (1).

## Pinning note

[...]

PR #1090, opened 2026-08-30
fix(board): evaluate --port/--open guards at CLI entry, not only in serve()

### LINKED ISSUE #468: board.mjs: the build subcommand never evaluates --port or --open, so both are silently ignored at exit 0
Deferred from PR #464 review (dimension: **silent-failure**). Finding state: **verified and reproduced, behaviour half deferred; the prose half was fixed in PR #464.**

## The defect

`argPort()` and `has("open")` are only ever called inside `serve()` (`board.mjs:439` and `:442`). The `build` path never reaches them, so `--port` and `--open` are accepted and dropped in silence — including the two malformed spellings the shared guards in `arg.mjs` exist to refuse.

Measured on PR #464 head:

```
$ node board.mjs build --port          -> exit 0   (flag ignored, full board JSON on stdout)
$ node board.mjs build --port abc      -> exit 0   (garbage value ignored; #366's guard never runs)
$ node board.mjs build --open=1        -> exit 0   (the --flag=value spelling #364 exists to refuse)

$ node board.mjs build --interval      -> exit 2
$ node board.mjs build --spend-since   -> exit 2
$ node board.mjs build --ledger        -> exit 2
```

The contrast is the point: `ledger`/`prev`/`spend-since`/`interval` are read on the build path too (`gather()` evaluates `argInterval()`, and `:389` evaluates `has("spend-since")`), so they refuse there. `port` and `open` are read in `serve()` alone.

**Not a regression from #367.** Identical old-vs-new in the review's differential matrix — both trees exit 0.

## Already done in PR #464 (do not redo)

`board.mjs`'s comment listed `port` among the flags whose silent-default it had closed, which is false on the build path. PR #464 corrected that comment only — it now says outright that a guard fires only where the flag is actually read, and names `build --port` / `build --port abc` / `build --open=1` as ignored at exit 0. **No behaviour changed.**

## Why the behaviour half is deferred, and why `needs-triage`

The approach is genuinely open, which is the `needs-triage` criterion:

- **Option A** — evaluate the presence guards once at CLI entry (call `argPort()`/`has("open")` in `main()` before dispatching on `process.argv[2]`). Catches both spellings everywhere, but makes `build` refuse flags that are meaningless to it rather than ignoring them, which is a behaviour change for any caller passing them today.
- **Option B** — refuse unknown-for-this-subcommand flags per subcommand, which is a larger design question about whether these scripts want per-subcommand arg schemas at all.

Nobody should pick between those inside a review deferral.

## Related, but not duplicates

- **#463** — `board.mjs`/`ci-state.mjs` silently ignore stray **positional** arguments. Different mechanism: that is a bare token with no leading `--`; this is a **known flag** that is simply never read on one path. A `parseArgs` strict-mode fix for #463 would not catch `build --port`, because `--port` is a legitimate flag of this script.
- **#365** — the unknown **flag name** sweep (`--basee` reading as absent). Also does not catch this: `--port` is a known name.
- **#366** — closed; added the non-numeric guard inside `argPort()`, which is exactly the guard the build path never invokes.

feigi:
## Summary

`argPort()`/`has("open")` used to run only inside `serve()`, so `board.mjs build` never evaluated `--port`/`--open` at all — both malformed spellings the shared guards exist to refuse were accepted and silently ignored at exit 0. Triage ruled Option A (#468 Agent Brief): evaluate the presence/shape guards once at CLI entry, ahead of the `build`/`serve` dispatch, not per-subcommand arg schemas (Option B, out of scope, overlaps #463's remedy).

Verified at `ee1c89d` (this PR's head), run directly against the built script:

```
$ node skills/fleet/scripts/board.mjs build --port
board: --port needs a value          (exit 2)
$ node skills/fleet/scripts/board.mjs build --port abc
board: --port wants an integer 0-65535, got abc   (exit 2)
$ node skills/fleet/scripts/board.mjs build --open=1
board: --open is a boolean flag, not --open=   (exit 2)
```

All three exit 2 with the exact wording `serve()` already refused them with — this changes *where* the guards run, not what they say.

## AC-4 decision (stated in the comment, `skills/fleet/scripts/board.mjs`)

A **well-formed** `board.mjs build --port 9999` is accepted and the value is simply unused: `build` has no server to bind or browser to open, so `argPort()`/`has("open")` are called for their validation side effect only and their return values are discarded on the build path. This is unchanged from before the fix — only the malformed spellings now refuse. Pinned by `build: a well-formed --port/--open is accepted and simply unused, per the #468 ruling` in `board-cli.test.mjs`.

## Enumerate-and-declare (the ticket's own class)

Covered — all three cases the ticket names: `build --port` (trailing, no value), `build --port abc`, `build --open=1`.
Deliberately left as **not this ticket's business** (per the triage ruling and existing `arg.mjs`/`board.mjs` comments): refusing `--port`/`--open` on `build` when they're *syntactically well-formed* — that's Option B, per-subcommand arg schemas, ruled out and overlapping #463.

Control group (ticket's own list) — `ledger`/`prev`/`spend-since`/`interval` already refused on the build path before this change and are unaffected: their existing guard call sites and ordering are untouched, and the full suite (below) still passes every test pinning them.

AC-2/AC-3 control — confirmed via the existing (unmodified) tests: `board-cli.test.mjs`'s `"no --spend-since at all is not an error"` (plain `build`, exit 0, unchanged shape) and `board.test.mjs`'s full `"CLI: serve ..."` block (port/interval/open all pinned) — both pass unchanged, confirming moving the guards to CLI entry didn't touch `serve()`'s own behavior.

## New edit's own risk (not just the bug class)

Placed the two guard calls in `main()` *after* `ledgerFile = arg("ledger")` and *before* both branches' own `stray()` call — same "value guards before `stray()`" ordering `arg.mjs` documents for every other guard in this file, so a malformed value still wins the specific wording over `stray()`'s generic one. Verified the existing ordering-sensitive test (`"CLI: trailing --prev names --prev, not the innocent value of the flag behind it"`) still passes unchanged.

## Review follow-up (`8dca28c`)

The review of this PR returned two `survived` findings and three `suggestion`-band findings that were checked here with one refuter each. All five are applied in `8dca28c`; nothing else from the review is in the tree.

**Applied — comment correction (`board.mjs`, survived 2-0).** The #468 call-site block justified `serve()`'s retained `argPort()`/`has("open")` calls with "keeps `serve()` callable directly (the tests do)". Nothing calls `serve()` in-process: `board.test.mjs` imports `{createBoardServer, mapCi, encodeProjectDir, findSubagentsDir, gatherSpend}`, `board-cli.test.mjs` only `{encodeProjectDir}`, and every `serve`-path test drives the real CLI, which enters `main()`. Two refuters measured it — one threw from `serve()` unless `main()` had set a marker (64/64 still green), one dropped the `export` (95/95 still green). The justification now rests on `serve()` being exported public surface.

**Applied — four new pins (`board.test.mjs`).** The three original malformed cases each pass the flag *alone*, so none of them can see the ordering the same comment claims. Four tests added:

| new test | mutant that reds it |
| --- | --- |
| `build --port abc extra` names `--port`, not the stray | guards moved below `build`'s `stray()` |
| a malformed `--port` with no subcommand names the flag, not the usage line | same mutant |
| `build --port=9000` refuses | hoist narrowed to `if (process.argv.includes("--port")) argPort();` |
| `build --port ""` refuses | hoist narrowed to skip an empty value |

The first two shapes were measured to change behaviour: with the guards below `build`'s `stray()`, `build --port abc extra` reports `unexpected argument 'extra'` instead of naming `--port`, and `board.mjs --port abc` printed the usage line before this PR. The last two are spellings the hoist closes beyond the three the ticket names; a refuter showed the narrowed hoist survives the whole 1492-test suite while reopening `--port=9000`.

**Applied — one blank line (`board-cli.test.mjs`).** This PR had attached its `#468 AC-4` block directly to the pre-existing `#365` block with no blank line, merging two comment runs so the `#365` comment headed the wrong test. A refuter measured the base (`#365` comment ended at 253, its test on 254) and scanned all 2127 contiguous comment runs in `skills/fleet/scripts/*.mjs`, finding no precedent for the merged shape.

**Not applied, deliberately.** The same finding also claimed the `#365` comment's *text* is stale. Refuted three times over: `board.mjs:502-509` still names the well-formed-accepted gap verbatim, so that sentence is still true and its wording is unchanged here. Two further `suggestion` findings (comment duplication in `board.mjs`; the header's word "once") were checked and refuted. Deferrals and refutations are filed — see the review comment on this PR.

## Testing

`node --test skills/fleet/scripts/*.test.mjs` (repo's full test command), from the worktree at `8dca28c` — **`tests 1498 / pass 1498 / fail 0`**, exit 0. (1494 at `ee1c89d`, plus the 4 review pins.)

All eight new tests are mutation-checked, in scratch copies rather than the worktree. Four mutants, each reddening exactly its intended tests and nothing else:

- guards moved below `build`'s `stray()` → reds exactly the ordering pin and the no-subcommand pin (68 tests, 2 fail)
- `if (process.argv.includes("--port")) argPort();` → reds exactly `--port=9000` (1 fail)
- hoist skipping an empty value → reds exactly `--port=9000` and `--port ""` (2 fail)
- **control:** the two guard calls' order swapped, an ordering nothing documents → **green at 68/68**, so the pins discriminate the defect rather than incidental ordering

The original three `build`-refusal pins were mutation-tested when they were written: commenting out the two guard calls in `main()` reddened exactly those three.

Sizing: light

Closes #468


feigi:
## Review adjudication — 13 findings

Applied in `8dca28c` (5), filed (6), no action (2). Full suite from the worktree at that SHA: `node --test skills/fleet/scripts/*.test.mjs` → **`tests 1498 / pass 1498 / fail 0`**, exit 0.

### Applied (5)

| finding | band | why |
| --- | --- | --- |
| `board.mjs:534` comment claims the tests call `serve()` directly | survived 2-0 | nothing calls `serve()` in-process; two refuters proved it by mutation |
| the guards' `stray()` ordering has zero coverage | survived 2-0 | mutant survives all 64 tests; `build --port abc extra` silently swaps which mistake is named |
| `build --port=9000` / `build --port ""` newly refuse, unpinned | suggestion, **checked, survived** | a narrowed hoist reopens `--port=9000` and survives the full 1492-test suite |
| the no-subcommand path changed and is untested | suggestion, **checked, survived** | pre/post measured; a live mutant restoring the old usage line stays 64/64 green |
| the `#468 AC-4` block merged into the pre-existing `#365` comment run | suggestion, **checked, survived** (structural half only) | no precedent in 2127 comment runs; the same finding's *staleness* half was refuted and is not applied |

### Filed (6)

- **#1092** `ready-for-agent` — `serve` never reads `--prev` at all, so a malformed one is swallowed with no shape guard. Real, pre-existing (a pre/post pair is byte-identical on all three `--prev` shapes), and out of #468's scope, which triage restricted to `--port`/`--open`. Cross-references #1076 and #468; explicitly **not** a duplicate of #1076 — that is a read sitting *below* `stray()`, this is a flag never read at all.
- **#1093** `needs-triage` — `main().catch((e) => die(e.message))` renders an internal fault as a `board: <text>` refusal at exit 2, stack dropped. Pre-existing, out of scope, and never adversarially checked.
- **#1094** `needs-triage` — `build --interval abc` does three `gh` reads before refusing; `--interval`'s guard is not hoisted the way `--port`/`--open` now are. Pre-existing, out of scope, never adversarially checked.
- **#1095** `wontfix`, **closed on filing** — the five findings that were checked and did not survive, each with its measurement. Closed because it is a record, not queued work; `ledger.mjs check` reads the tracker at `--state all`, so a future run re-deriving one of these lands on the evidence.

### No action (2)

- The lone red in the review's full-suite run is `unattended-git-sweep.test.mjs` failing at module load inside the `git archive` snapshot (`git rev-parse --show-toplevel` → `fatal: not a git repository`). A known, deterministic artifact of the review mechanism, not a regression; the suite is green in the real worktree, measured above. Not filed.
- The `--prev`-on-`serve` finding's *claim about this PR* (that it half-fixes the gap) is refuted and recorded in #1095; the underlying behaviour is #1092.

### Interaction recorded outside this PR

Commented on **#1076**: this PR closes the first of its three named cases. Re-measured at `8dca28c` — `serve --port --open x` now gives `--port needs a value` (was `unexpected argument 'x'`); `serve --interval --open x` and `build --spend-since --open x` are unchanged and still open. #1076's "still true at `6202af2`" line no longer covers all three.

PR #1091, opened 2026-08-30
fix(no-undo-audit): the content-corrupt tip object exits 1, not 0 (#494)

### LINKED ISSUE #494: no-undo-audit: two comments claim a content-corrupt TIP stash object leaves `stash list` at rc 0 — measured rc 1 on git 2.50.1
> *This was generated by AI during a PR review.*

**Dimension:** `types` (comment/prose accuracy) · **Deferred from PR #489 review** · pre-existing, **not** introduced or falsified by that PR's diff.

## The claim, and why it is wrong

Two comments assert that a **content-corrupted TIP stash object** leaves `git stash list` at **rc 0** with an empty list. Measured, it is **rc 1**, matching the missing-object row of the table above it.

- `skills/fleet/scripts/no-undo-audit.sh:369` — *"A stash object that is CORRUPT rather than missing (`echo junk > .git/objects/<xx>/<rest>`) reaches this same branch — **list empty at rc 0**, show-ref rc 0 — and git says it in 7 lines…"*
- `skills/fleet/scripts/no-undo-audit.test.mjs:454` — the same clause, word for word: *"list empty at rc 0, show-ref rc 0 — but git answers in SEVEN lines there…"*

Cite by phrase (`list empty at rc 0`) rather than by line number — this file's line citations are known to drift.

## Measurement

git 2.50.1 (Apple Git-155), macOS. Isolated fixture, one stash, tip object overwritten with `printf 'junk\n'`:

```
git version 2.50.1 (Apple Git-155)
$ git stash list
error: inflate: data stream error (incorrect header check)
error: unable to unpack fd41130e9ad39729a46fc0a16fc0a829ca4941e5 header
  (x3)
fatal: loose object fd41130e9ad39729a46fc0a16fc0a829ca4941e5 (stored in .git/objects/fd/41130…) is corrupt
rc=1
stdout=[]
$ git show-ref refs/stash >/dev/null 2>&1; echo $?
0
```

So: **rc 1**, empty stdout, `show-ref` rc 0.

## What is still true

Only the rc clause is wrong. The rest of both comments holds and was reproduced:

- the corrupt-object case **does** reach the `unknown` branch — the counting pipeline is `git stash list 2>/dev/null | wc -l`, whose status is `wc`'s, so `stash=0`; `show-ref` rc 0 ≠ 1, so the branch fires. The rc is simply discarded, which is why the wrong value never showed up as a test failure.
- git really does answer in **7** lines (10 with a bad `objects/info/alternates`) — counted in the same run.

## Why it matters

The paragraph is load-bearing for the neighbouring argument (*"`stash list` is silent in the other three (rc 0, no stderr — measured, git 2.50.1)"*). A reader reconciling the corrupt case against that sentence gets a wrong picture of which states are loud, which is the exact distinction #482 turns on.

## Why `needs-triage`

The defect is verified but the fix is not obviously a word swap. Correcting `rc 0` → `rc 1` in two comments is the minimal edit, but the same paragraph is what #482 is about, and a triager may prefer to fold this into #482's eventual change rather than land a one-word correction that the #482 work then rewrites. A triager should decide which.

## Why deferred from #489

PR #489 is a characterization test for the *truncated-reflog* ceiling (#306). This comment is diff-untouched by it and describes a different corruption shape, so correcting it there would have been an edit in a file the ticket has no business changing beyond its own paragraph.

feigi:
Closes #494

Sizing: light

Two comments said a content-corrupt stash TIP object leaves `git stash list`
`empty at rc 0`. It exits **1**. One token in each file; the paragraph is
otherwise untouched, and the mechanism sentences it carries are intact.

## Measured, not copied from the ticket

git 2.50.1 (Apple Git-155), macOS. Isolated fixture: one stash, tip object
overwritten with `printf 'junk\n'`.

```
$ git stash list 2>/dev/null; echo "rc=$?"      # rc=1, stdout empty
$ git show-ref refs/stash >/dev/null 2>&1       # rc 0
$ git stash list 2>/dev/null | wc -l            # 0  <- why the branch fires
$ git stash list 2>&1 >/dev/null | wc -l        # 7
  ... with `/no\clue/objects` in objects/info/alternates   # 10
```

The rc never surfaced as a failure because the counting pipeline
(`stash list … | wc -l | tr -d ' '`) reports the last stage's status.

The neighbouring `stash list` is silent in the other three sentence was
re-measured on the same git and left alone — it still reads correctly against
the corrected clause: unreadable reflog, unreadable `refs/stash`, and a
malformed `refs/stash` each give rc 0, empty stdout, empty stderr; the
missing-object case gives rc 1 with `fatal: bad object refs/stash`. So the
corrupt case belongs with the loud row, which is what the correction says.

## The pin

Inside the fixture that already builds the corrupt object:

- the rc is now **measured** — `stash list` must exit 1 there — so a wrong
  number in the prose cannot sit unnoticed again;
- the script's own comment is held to the same rc, in **its own test**, using
  `prose-pin.mjs`'s `stripHashGutter` + `between` + `phrase` — so a rewrap is
  not read as a regression, and the match is bounded to the paragraph under
  test.

Both points about the pin changed after review; see **Applied from review**
below for what moved and why.

Mutants, one file, `node --test`:

| mutant | result |
|---|---|
| script clause back to `rc 0` | fail 1, and only that test |
| clause rewrapped, words unchanged | tests 78, pass 78, fail 0 |
| clause deleted, a later occurrence added elsewhere in the script | fail 1 (the old unbounded match stayed green here) |
| `tr` fold at `:403` dropped **and** comment reverted | fail 2 — the behaviour test and the prose test red independently |
| fixture swapped for an unreadable reflog (rc 0 state) | fail 1, on the rc assertion's own message |
| fixture object deleted rather than corrupted (still rc 1) | rc assertion stays green; the `fatal: loose object … is corrupt` assertion fires |

## Out of scope, deliberately

- The rejected `simplify` finding on this block (replace the paragraph with a
  cross-reference) is not revived: it proposed deleting the lines being
  corrected, and asserted a ruling on #482 that does not exist.
- `run-merge-bot.md`'s three of its four causes leave it empty at rc 0 and
  silent enumerates the four causes of `unknown`; re-measured and correct as
  written. The corrupt tip object is a different shape and adding it there
  would be new prose, not this correction.
- `no-undo-audit.sh`'s absent-ref-with-reflog branch says git is silent here,
  empty list at rc 0 — re-measured on the same git: rc 0, empty stdout, empty
  stderr. Correct, untouched.

## Applied from review

The PR #1091 review returned 7 findings, all `suggestion` (0 refuters by
policy). Two were checked by refuters, survived, and landed in `cc85ec5` as one
restructure:

- The pin sat **ahead of `audit(c)`** inside the behaviour test, so `assert`'s
  abort meant a comment-only drift pre-empted this file's only coverage of the
  #304 fold and `printf` hazards — measured: with the fold at `:403` dropped and
  the comment reverted, only the comment was reported and the three behavioural
  assertions never ran. It now has its own test; both red independently.
- The hand-rolled `^[ \t]*#[ \t]?` strip plus `\s+` flatten matched the whole
  file **unbounded** — the false green `prose-pin.mjs`'s own comment warns
  about, and it reproduces. Now bounded to the paragraph via the shared slicer
  30 other prose tests already import.

The rest are recorded in #1096 (closed, `wontfix`): one refuted by measurement,
two positive results re-verified against the new pin, one snapshot artifact, and
one out-of-scope reproducer routed to #482.

Full suite at `cc85ec5`: `node --test skills/fleet/scripts/*.test.mjs` ->
tests 1491, pass 1491, fail 0.


feigi:
**Review findings adjudicated** — 7 findings, all `suggestion` (0 refuters by workflow policy, so all arrived unchecked). Scope-split, then one refuter dispatched per in-scope finding, biased to refuse.

| finding | dimension | disposition |
|---|---|---|
| `unverified[0]` `no-undo-audit.test.mjs:727` | correctness | **applied** — refuter `refuted=false`; false green reproduced |
| `unverified[1]` `no-undo-audit.test.mjs:726` | silent-failure | **applied** — refuter `refuted=false`; masking reproduced |
| `unverified[2]` `no-undo-audit.test.mjs:722` | silent-failure | **refuted** → #1096 |
| `unverified[3]` `no-undo-audit.sh:342` | silent-failure | **out of scope** → re-measured and commented on #482 |
| `unverified[4]` `no-undo-audit.test.mjs:722` | tests | positive result; re-verified against the new pin → #1096 |
| `unverified[5]` `no-undo-audit.test.mjs:728` | tests | positive result; re-verified against the new pin → #1096 |
| `unverified[6]` `unattended-git-sweep.test.mjs:1` | tests | `git archive` snapshot artifact, not a regression → #1096 |

- Checked-band record: #1096 (`wontfix`, closed — evidence, not queued work)
- Non-tip reproducer: https://github.com/feigi/claude-config/issues/482#issuecomment-5470680487

`[0]` and `[1]` landed together as one restructure — both rework the same assertion. Applied in `cc85ec5`.

Suite from the worktree, `node --test skills/fleet/scripts/*.test.mjs` → **tests 1491, pass 1491, fail 0**. No `--no-verify`; pre-commit hooks ran clean.

PR #1097, opened 2026-08-30
fix(fleet): the merge bot must re-derive the head the label certified (#493)

### LINKED ISSUE #493: run-team: every guard protecting ready-to-merge is upstream of the label — a push after the finisher audits lands on an unaudited head with nothing to stop it
Filed by the fleet controller, 2026-08-15. This is the **second halt-shape** recorded on #144 and deliberately scoped out of PR #488, which implemented #144's brief. #488's body names it and punts it here rather than building it; that ruling was made at phase 0, before dispatch, so the implementer never worked on an open design question.

## The shape

#144 is: *the finisher is dispatched, then the head moves, so the pin goes stale and the finisher halts.* The halt is the safety property and it works — it fired correctly several times in the run #144 measures, and again in this run.

This is the mirror image, measured on PR #426 in an earlier run:

1. Controller dispatched a finisher pinned to `141ad17`.
2. Finisher audited `141ad17` — clean worktree, `check` green, deferrals confirmed — and added `ready-to-merge`.
3. **After** that, the fix-applier surfaced a defect in its own new prose and asked the controller for a ruling.
4. Controller approved a one-line doc fix → new head `3aa5be8`.

Between 2 and 4, `ready-to-merge` was live on `141ad17` while a push to `3aa5be8` was authorized. **There is no pin left to go stale and nothing halts** — the label simply sits on a head no finisher ever audited, and a merge bot is entitled to merge it. Caught only because the controller pulled the label before approving.

The fix-applier's own observation is the reason this is worth a ticket:

> the window closed only because I had already stopped and reported; a member still working would have pushed straight through it.

**Every guard in the chain — the pin, the halt-on-moved-head, the worktree audit — is upstream of the label. Downstream, the label is unconditional and permanent until something removes it.**

## Why this is not #144, and not already fixed by PR #488

PR #488 implemented #144's two acceptance criteria: the controller gates finisher dispatch on outstanding relays, and every halt report carries two-cause evidence. Both are about the interval **before** labelling. Neither touches the interval after.

A related but distinct fact, from an existing measurement: **a GitHub label does not follow the branch.** `ready-to-merge` did not drop on push in a prior run (PR #180) and silently transferred to a commit nobody audited. So the label surviving a push is confirmed behaviour, not a hypothesis.

## Directions, none obviously right — hence `needs-triage`

Recorded from #144's thread plus this run, so triage does not re-derive them:

- **Finisher re-reads the head immediately after labelling and removes the label if it moved.** Closes the window; does not prevent it. Cheap.
- **Controller treats "a fix-applier has an outstanding question" as a dispatch blocker with the same weight as an outstanding ruling.** The current wording only names rulings the controller *owes*. This run hit exactly that case: `fix-pr-488` asked whether to add two pins, and the finisher was held until it acknowledged the answer. That worked, but it worked because the controller chose to hold, not because a rule required it.
- **Merge bot re-derives that the labelled SHA is still the head before merging.** Worth confirming rather than assuming — wave 1 and wave 2 bots in this run both re-derived every gate at the merge instant, including the head, so this may already hold in practice and only need writing down.

## Recovery that worked, for whoever takes this

Remove the label **before** approving the push. Removing the artifact a merge bot gates on is the reliable stop — messaging a bot races it, and a label has been observed holding until after an abort message arrived. Then re-run a **fresh** finisher against the new head; the first finisher's audit does not transfer, since it verified a different tree.

## Out of scope

- #144's own subject, the pre-label pin race — implemented in PR #488, merged/merging.
- **#389**, the finisher as a writer (its mutation gate edits files the fix-applier owns, with no lock). Adjacent, separately filed, do not fold in.
- Changing whether a moved head halts. It must always halt.

## Provenance

PR #426 sequence measured in an earlier run and recorded on #144. Scope ruling for #488 made at phase 0 of the 2026-08-15 run; #488's body records the punt. Related agent-brain record: a GitHub label does not follow the branch (PR #180).

feigi:
Closes #493

Sizing: heavy

## What was already true, and what was not

The brief said this is very likely writing down a guarantee that already exists in practice. **It is not.** What the wave bots already re-derive at the merge instant is *CI currency* — `ci-state.mjs` selects the workflow run whose `headSha` equals the current PR head (`r.headSha === prHead`) and refuses green otherwise. Nothing anywhere compared the head to the one the **label** certified.

Those are different properties, and the gap between them is the whole ticket. A member pushes after the finisher labelled; the push gets its own CI run; that run is green and its head *is* the PR head. Every gate in the per-PR sequence passes, and the bot merges a tree no finisher audited.

`#180` is that trace, read with the query this PR now prescribes:

```
$ gh api "repos/{owner}/{repo}/issues/180/timeline?per_page=100" --paginate \
    -q '.[] | select((.event=="labeled" and .label.name=="ready-to-merge")
          or .event=="committed" or .event=="head_ref_force_pushed"
          or .event=="merged" or .event=="unlabeled") | .event'
labeled
committed  (x5)
head_ref_force_pushed
merged
```

No `unlabeled` between the label and the merge. The label survived a head change and the PR merged on the moved head.

## The change

**`run-merge-bot.md`, new `## The labelled head` section** — before the per-PR sequence, because the bot's own rebase lands commits after the label by construction and a read taken afterwards cannot tell its commits from a member's. Confirmed on this repo's own merged PRs: #1090 (never rebased) lists `labeled ready-to-merge` after its commits; #1091 (rebased by its merge bot) lists three `committed` plus a `head_ref_force_pushed` after the label. A `committed` or `head_ref_force_pushed` after the last `labeled ready-to-merge` refuses the merge, reports `head-moved-after-label-#<pr>`, and leaves the label alone. It also records the pre-rebase head, which step 3 needs.

**`run-merge-bot.md` step 3** — the merge-instant leg. The head must be that recorded `pre` or the `post` step 1's own rebase produced. Any third SHA is a push that landed during the CI wait, which is exactly the case no gate above can see.

**`run-team/SKILL.md`** — the controller side. An unanswered question from a member is an outbox item and blocks finisher dispatch with the same weight as a ruling already made; the label-removal recovery, at the moment it applies; and a failure-table row so `head-moved-after-label-#<pr>` is placeable by the controller rather than existing only in the bot's runbook.

**`head-moved-after-label-prose.test.mjs`** — 15 pins across five `between()`-bounded slices, using the shared helpers in `prose-pin.mjs`. One of them is the gate's decision rule itself — that a `committed`/`head_ref_force_pushed` line counts only when it falls AFTER the last `labeled ready-to-merge` — pinned as one contiguous span through `**Refuse:` so the predicate stays bound to its verdict. Review of this PR measured that rule unpinned at 07589ee: inverting `after` to `BEFORE` left the file 14/14 green and deleting the sentence left 36/36 green across both prose suites, and inverted the document tells the bot to refuse the ordinary wave PR and to merge #180's shape.

## The class, and what is left out

Ways the label can outlive the head it certified:

| Case | Covered |
|---|---|
| Push authorized after labelling (the measured #426 sequence) | yes — timeline gate |
| Label does not follow a force-push (#180) | yes — `head_ref_force_pushed` is the same refusal |
| Push landing during the bot's own CI wait | yes — step 3's `pre`/`post` comparison |
| Label re-added after removal | yes — the gate reads the **last** labelling event |
| A head rebased by an earlier abandoned pass of this command | refuses, deliberately — that tree was audited by nobody either |
| `ready-to-merge` hand-added with no finisher behind it | **not covered, stated in the section** — the gate proves the head has not moved since labelling, never that anyone audited. `run-team/SKILL.md`'s reviewer-only rule owns that |
| #144's pre-label pin race | out of scope, merged in #488 |

## AC-4

The normal path is stated as unaffected in the runbook and pinned in its own test, together with the two cases the gate declines to answer. Mutation-tested in both directions on a copy outside the worktree: each deletion fails its own test and only its own; a rewrap of the whole section at 58 columns, with pinned phrases split across lines, stays green; and deleting a clause while re-adding the same sentence *after* the slice boundary reds the bounded pin where a whole-file match would have passed. The decision-rule pin added in review was measured the same way against the final tree: inverting the predicate, deleting the sentence, and splicing an exception clause between the predicate and the refusal each red exactly that one test, while a three-line rewrap of the same wording and a reword of the unpinned sentence above both stay green.


feigi:
Review deferrals filed: #1102 (`wontfix`, closed) — `refuted[0]` (correctness, `run-merge-bot.md:60`, refuted 2-0 with measurement) and `unverified[0]` (comments, `:50`, the same claim from another dimension, refuted by inheritance), plus the known two-read-gate ceiling recorded rather than filed. The review's one `survived` finding was applied in c874280, not deferred.

Issue #1064, opened 2026-08-30
release-ticket.sh: the stray suffix key names an unrelated same-basename worktree as "is this claim's" and tells the operator to hand-release it


feigi:
Deferred from PR #1054 review (issue #453). Dimension: **invariants expressed vs merely documented; casts that erase conformance** (surfaced as the residual of finding 1, and independently by the silent-failure refuter on finding 5).

Confirmed defect. **The remedy is open** — see below.

## The defect

`skills/fleet/scripts/release-ticket.sh` finds a claim's stray worktree with a directory-**suffix** key:

```sh
awk -v d="/$issue-$slug" '/^worktree /{n++; p=substr($0,10)
  if (n>1 && substr(p, length(p)-length(d)+1) == d) {print p; exit}}'
```

Any linked worktree whose directory basename is `<issue>-<slug>` matches, first match in porcelain (lexicographic) order wins, and the arms below then assert `worktree <path> is this claim's` about it. When an unrelated worktree carries that basename and sorts **before** the claim, the operator is told to hand-release somebody else's live tree.

Direction is safe — it blocks, rc 1, nothing is deleted, the branch survives. The defect is that the blocker names the wrong directory while asserting ownership the script never established.

## Measured

Fixture: decoy worktree added first at `.worktrees/0-decoy/9-release-ticket` on branch `decoy/9`, healthy, so it takes the registry entry `9-release-ticket` and the real claim at `.worktrees/9-release-ticket` registers as `9-release-ticket1`; the claim's own admin HEAD corrupted. The decoy's path sorts before the claim's.

```
BLOCKED: worktree /…/w/.worktrees/0-decoy/9-release-ticket is this claim's but is not on
fix/9-release-ticket — release it by hand
rc=1
```

Reproduced independently by two refuters during the PR #1054 review, each on its own fixture, and byte-identical against a pre-PR reconstruction of the script (`diff` confirming the only difference was that PR's hunk). So it is **pre-existing**, predating both PR #1054 and PR #450, and out of #453's scope — which is why it is filed rather than fixed there.

## Not the same as PR #1054's own regression

[...]

PR #1098, opened 2026-08-30
fix(release-ticket): name how to find the entry the unresolved-HEAD arm means (#455)

### LINKED ISSUE #455: release-ticket.sh:397 — the unresolved-HEAD blocker names the registry root but not which entry to open

feigi:
The unresolved-HEAD arm handed the operator the registry **root** and nothing else, so on a repo with several linked worktrees they had to scan `.git/worktrees/*/gitdir` themselves to find the entry being discussed. Every sibling arm hands over an exact command; this one handed over a directory to search.

It now appends that search **as a command**, on the suffix-key path only. Emitted text, captured from a real run at `4c5a97d` (paths elided):

```
… so its branch is unknown — inspect its entry's HEAD file under /…/w/.git/worktrees by hand;
find that entry with: grep -Fxl '/…/w/.worktrees/9-release-ticket/.git' '/…/w/.git/worktrees'/*/gitdir
```

Sizing: light

## Why a search and not a derived name

Per the ruling in the brief. `${stray##*/}` is measurably wrong, and the fixture in the new test shows it failing three ways at once — measured at `4c5a97d`, git 2.50.1 (Apple Git-155), on entries `9-x`, `9-x-renamed`, `9-x1`, `9-x2` where the claim's own is `9-x1`, a moved thief holds `9-x`, and `9-x2` is registered under a directory carrying the claim's whole path as a tail:

```
$ echo ${stray##*/}                         # -> 9-x         — the THIEF's entry
$ grep -Fl  "$stray"      $wtroot/*/gitdir  # -> 9-x1, 9-x-renamed, 9-x2
$ grep -Fl  "$stray/.git" $wtroot/*/gitdir  # -> 9-x1, 9-x2   — the anchor is not enough
$ grep -Fxl "$stray/.git" $wtroot/*/gitdir  # -> 9-x1, one line
```

`gitdir` holds exactly one line whose whole content is `<path>/.git`, so `-x` is the precise test and it subsumes the `/.git` tail anchor: a substring match is wrong in **both** directions — it hits a sibling whose registered path *extends* this one, and one registered under a path that *ends with* it. `-F` because the subject is a path and its metacharacters are not a pattern.

## The emitted string is shell text, so it is quoted as shell text

The operator pastes this. Interpolating the two paths raw broke it two ways, both measured end to end through the script:

- Inside **double** quotes a `$` in the worktree path expands *again* when the operator runs it. The search returns nothing at **rc 1** — reporting no entry while the entry is right there. That is this arm's own defect one step further out: a message asserting something the script never measured.
- An unquoted `$wtroot/*/gitdir` operand word-splits on a path containing a space (`grep: /…/my: No such file or directory`, rc 2).

Both are fixed by single-quoting each path with the glob left outside the quotes. A literal `'` in the path still defeats single quotes, and that residue is deliberate rather than overlooked: it breaks the pasted command **loudly** (`unexpected EOF while looking for matching`, rc 2, measured) instead of answering wrongly, which is the failure mode this ticket exists to avoid. Recorded on the line and filed as follow-up, not left implicit.

## Cross-product: four `block` calls × two key paths

| arm | suffix key | entry key |
|---|---|---|
| locked | unchanged | unchanged |
| directory gone | unchanged | unchanged |
| **unresolved HEAD** | **CHANGED — gains the search** | unchanged, byte for byte |
| branch-mismatch `else` | unchanged | unreachable by construction |

Exactly one cell moves. The entry-key path is left alone deliberately: it read `$wtroot/$issue-$slug/gitdir` itself and already names that entry in `$stray_own`, so searching for what it has in front of it would be a step backwards. `stray_find` is cleared there, and a test pins that it stays cleared.

**Deliberately not covered:** the `by hand` framing and the absence of a remedy command, both out of scope and both re-measured below.

## The `by hand` framing stands — re-measured, not copied

Against a corrupt-HEAD entry, git 2.50.1:

```
$ git worktree repair            # rc 0, entry unchanged
$ git worktree prune -v          # rc 0, prints nothing, entry unchanged
$ git worktree remove …          # rc 128, fatal: validation failed … '.git' is not a .git file, error code 7
$ git worktree remove --force …  # rc 128, same
```

One correction to the issue body: `git worktree repair` is rc 0 with the entry unchanged, but it is **not silent** — it prints `repair: .git file broken: <path>`. It still clears nothing, so the conclusion is unchanged.

## What this newly does

It changes what the operator is told to run — so every pin here **runs the printed command** rather than matching its text, and asserts it lands on exactly this claim's entry. A reworded instruction that still works stays green; one that stops working reds.

## AC-4 — the prose pins

`grep -rn 'could not read its HEAD' skills/fleet/scripts/*.test.mjs` found pins on the arm firing, but none on its tail wording. Now:

- **new** `the unresolved-HEAD blocker hands over a search for the entry, never a name derived from the directory (#455)` — the hostile fixture above, now carrying a **fourth** entry whose registered path ends with the claim's, so the whole-line anchoring is pinned in both directions rather than only the extending one.
- **new** `the entry search survives a $ in the worktree path (#455)` — a `$` slug; pins the quoting. git's refname rules accept `$`, `"`, a backtick and `;` alike (measured), so this path is reachable, not hypothetical.
- **new** `the entry search treats a regex metacharacter in the path as a character (#455)` — an `a.b` slug beside a `9-axb` sibling; pins `-F`, which **nothing in the suite discriminated before** (measured: dropping `-F` left all 116 tests green, before and after the `-x` change).
- `a MOVED worktree with an unresolvable HEAD blocks instead of releasing silently (#453)` — the entry key names the entry and never searches for it.
- `a worktree that is no longer on the branch blocks instead of releasing around it` — the `else` arm gains nothing.

## Verification at `4c5a97d`

```
$ node --test skills/fleet/scripts/*.test.mjs     # tests 1502, pass 1502, fail 0
$ shellcheck -x -S warning skills/fleet/scripts/release-ticket.sh   # rc 0, clean
```

Mutation testing, on a copy under the scratchpad — the worktree was never mutated. Every row is a full 118-test run of `release-ticket.test.mjs`:

| mutant | result |
|---|---|
| locator removed from the `block` call | the three #455 tests red, nothing else |
| `/.git` anchor dropped from the pattern | the three #455 tests red, nothing else |
| `-x` dropped (`-Fxl` → `-Fl`) | **only** the tail-sibling test reds |
| `-F` dropped (`-Fxl` → `-xl`) | **only** the regex-metacharacter test reds |
| single quotes reverted to double quotes | **only** the `$`-in-path test reds |
| entry-key `stray_find=` clear removed | **only** the #453 entry-key test reds |
| **control** — command rewritten equivalently (`grep -F -x -l`) | **stays green** |
| **control** — leading separator reworded, command unchanged | **stays green** |

One measured caveat worth stating rather than leaving for a reader to trip over: rewording the `find that entry with: ` anchor itself **does** red all three #455 tests, because that phrase is how they locate the command in the message. The anchor is the message's extraction contract, not free prose.

## Collision with #1064

#1064 is open and `ready-for-agent` on these same four `block` calls, about the surviving `is this claim's` literal naming an unrelated worktree that merely shares the `<issue>-<slug>` basename. **This PR does not fix it and does not touch `$stray_own`** — the literal is exactly as it was.

It does touch adjacent text: the message this PR extends is the one whose `is this claim's` #1064 will reword, so #1064 will land in the same string. The extension is a separate variable appended at the end, so the two edits do not overlap textually. Worth noting for whoever takes #1064: the search this PR prints is keyed on the measured path, not on the basename, so it stays correct under whatever #1064 does to the ownership wording — and it is arguably part of #1064's answer, since it gives the operator a way to check which worktree the message actually means.

Closes #455


feigi:
Review findings adjudicated and applied at `4c5a97d`.

**Applied** — the five findings on `release-ticket.sh:767` converge rather than conflict, so they landed as **one** edit (`grep -Fxl '$stray/.git' '$wtroot'/*/gitdir`), plus the two test-coverage findings and the two comment corrections. Suite green in the worktree: `node --test skills/fleet/scripts/*.test.mjs` → **tests 1502, pass 1502, fail 0** (1500 before, +2 new tests). Each new pin mutation-verified to red alone, with two equivalent-rewrite controls staying green. PR body reconciled against the final diff — the example command, the flag rationale, the test list, the mutation table and the verification counts all moved.

**Deferred and filed:**

| # | finding | state | label |
|---|---|---|---|
| #1100 | a literal `'` in the worktree path still defeats the single quoting (residual surfaced by the `unverified[3]` refuter, then measured) | confirmed defect, remedy open | `ready-for-agent` |
| #1101 | the `suggestion` band, checked — 4 checked, 3 applied, 1 refuted by mutation | refutation record | `wontfix`, closed |

The refuted one is `unverified[4]` ("the `doesNotMatch(/entry with: grep/)` pin at test:593 is vacuous"): a one-line mutation of the shipped script — appending `$stray_find` to the `else` arm — reds that assertion **and nothing else in the suite**, while line 589 stays green. The pin stays.

PR #1105, opened 2026-08-31
fix(no-undo-audit): a stash list that failed is not a count (#482)

### LINKED ISSUE #482: no-undo-audit: a corrupt loose object behind a non-tip stash entry prints a silently truncated count — git exits 1 and names the fault, the pipeline discards both

feigi:
Closes #482.

Sizing: heavy

## What was wrong

`stash=$(git -C "$wt" stash list 2>/dev/null | wc -l | tr -d ' ')` gave the
statement `wc`'s status, always 0. A corrupt loose object behind a **non-tip**
stash entry makes `git stash list` print the entries it could read, say
`fatal: loose object … is corrupt` and exit **1** — and the reflog is intact
there, so `show-ref` resolves the ref and the existing cross-check has nothing
to disagree with. The audit printed a short number with full confidence, on the
guard that decides whether a rebase can destroy work.

## What changed

Capture the list instead of piping it, count from the capture, and answer
`unknown` when the call itself failed. The empty-list states are asked first and
keep the sentence they already print, so a corrupt **tip** object — which
satisfies both tests — still gets the more specific `refs/stash` sentence it has
today. The `unknown` branch's existing second call already passes git's own
diagnostic through, so the operator gets `fatal: loose object …` on the audit's
own line without a second read.

Per the brief's ruling: the **exit code**, not stderr. No temp file and no
stdout/stderr split on the healthy path (AC-3) — the refusal in the comment
stands, reworded only where it described the pipeline that is gone.

## Enumerate and declare

The class is "the stash count reads a `stash list` that did not finish".

**Covered**
- Corrupt loose object behind a **non-tip** entry — nonempty list at rc 1. The ticket's case; new test.
- Any other nonzero `stash list` rc reaching this statement, whatever its cause — the guard is on the rc, not on the corruption shape.
- Corrupt **tip** object and the missing-object ref (empty list at rc 1) — already `unknown` via `show-ref`, and now via the rc as well; both keep their existing sentence and their existing tests.

**Deliberately left**
- **#306**, the truncated-but-parseable reflog: `stash list` succeeds at rc 0, so no rc exists to capture. Still prints the short count, still pinned as a known ceiling by its characterization test, still named in the comment as "a remaining ceiling". Out of scope per the brief, and the brief's reason — that detection there is a genuinely expensive cross-check — is unchanged by this.
- **#376**, the deleted `refs/stash`: its own branch, untouched.
- **#583**, the general pipeline-status class in the conflicts pipelines: same file, different statements, its own ticket.
- **#481**, `show-ref`'s own discarded stderr: named as a ceiling in the comment, untouched.

**What this can wrongly refuse** — the accept side, and it is tested:
a healthy repo's count must be unchanged and a never-stashed repo must still
answer `0` (AC-2). `git stash list` returns rc 0 in both, so neither reaches the
new condition; the existing tests for one entry, three entries, an honestly
empty stash in six spellings, and #306's undercount are the controls, and all
stay green. The count arithmetic itself changed shape, so both ends were
re-measured: `$()` strips the trailing newline (`printf '%s'`, two entries,
`wc -l` → 1), and `printf '%s\n' ""` on an empty stack emits a lone newline that
`wc -l` counts as one. Review follow-up replaced that pair, and the `[ -z ]`
guard between them, with `awk 'END{print NR}'` — see below.

**What my own edit newly does**: `msg` is now assigned before the branch rather
than inside it, and the emit is one block reached by two conditions. `msg` and
`sl`/`sl_rc` are used nowhere else in the script. The rc capture is a simple
command in an OR list, not a bare pipeline followed by `rc=$?` — measured under
both `/bin/sh` (bash sh-mode) and `/bin/dash` with `set -eu`: the assignment is
populated, `sl_rc` is git's, and the script reaches the end.

## Verified at 873c341

Fixture: bare origin + clone, branch `fix/1-thing`, three stashes, loose object
behind `refs/stash@{2}` overwritten with `junk`. git 2.50.1 (Apple Git-155), darwin.

```
$ grep -c '' .git/logs/refs/stash                    -> 3
$ git stash list 2>/dev/null | grep -c ''            -> 2
$ git stash list >/dev/null 2>&1; echo $?            -> 1
$ git show-ref refs/stash >/dev/null 2>&1; echo $?   -> 0

origin/main's script, same fixture:  rc 0, "stash":2, line `stash entries ...: 2`
this branch's script, same fixture:  rc 0, "stash":null, line says unknown and
                                     carries git's `fatal: loose object ... is corrupt`
```

```
$ ./agent-test                                       -> tests 1519, pass 1519, fail 0 (run twice, exit 0 both)
$ .github/scripts/check-tracked.sh '*.sh' shellcheck -x -S warning -> 17 files, exit 0
```

Shellcheck before/after compared on symmetric copies with `json.sh` beside both,
so `-x` resolves for each: identical finding set.

```
$ grep -rln 'stash entries (repo-global\|refs/stash is not absent\|do not stop at that list' \
    --exclude-dir=node_modules --exclude-dir=.git .
  -> run-merge-bot.md, no-undo-audit.sh, no-undo-audit.test.mjs, run-merge-bot-prose.test.mjs
```
No fifth file carries this prose, so the runbook update (AC-5) is complete.

## Mutation testing

Committed first, restored with `cp`, never `git checkout --`.

| mutant | expected | measured |
|---|---|---|
| drop `elif [ "$sl_rc" -ne 0 ]` from the script | the new test alone reds | pass 78, fail 1 — the #482 test |
| reword the deleted-ref branch's sentence | the #482 test stays green | pass 76, fail 3 — the three orphan-reflog tests, #482 green |
| drop the new cause sentence from `run-merge-bot.md` | the new pin alone reds | pass 22, fail 1 — the new pin |
| reword an unrelated instruction in the same section | the new pin stays green | pass 22, fail 1 — that instruction's own pin, new pin green |

## Prose

The runbook's `unknown` paragraph opened with "the list came back empty", which
is now one shape and not the only one, and told the operator that a tallied
share of the causes leave the list empty at rc 0 — a count this ticket made
wrong the moment it landed. It states the rule now instead: *the causes git
stays silent about leave it empty at rc 0*. Its prose pin was rewritten to pin
that rule rather than the tally, and a new pin covers the new cause.

The in-file paragraph that named #482 as an open, untriaged ceiling (AC-4) now
states what the code does instead. #306's ceiling paragraph and the phrase its
characterization test greps for ("a remaining ceiling") are untouched.

## Review follow-up — 5fbaccc

The review returned ten findings. One survived adversarial verification; six
`suggestion`s were in scope and each got its own refuter, three of which
survived. Four changes landed.

- **This PR's own new prose was wrong.** It called a corrupt loose object behind
  a non-tip entry "the other shape" of #306's ceiling and said git is loud about
  it. Only the corrupt half is. Measured (git 2.50.1): three entries, `rm` the
  loose object behind `refs/stash@{2}` → `stash list` exits **0**, prints two
  lines, writes nothing to stderr, and the audit reports the exact-looking `2`.
  That half stays under #306. Corrected in the script comment and in the test's
  own comment, which carried the same over-claim.
- **`awk 'END{print NR}'` replaces the two-statement count.** It counts the final
  incomplete record, so one statement answers both ends — measured under
  `/bin/sh` with `set -eu`: `""` → 0, and one/two/three entries with no trailing
  newline → 1/2/3. No `[ -z ]` guard, no padding to strip. `shellcheck -x -S
  warning` over all 17 tracked `*.sh`: exit 0.
- **`RC_FAILED_LINE`.** The new rc branch's operator line had no exact-string pin,
  unlike its siblings `UNKNOWN_LINE` and `ORPHAN_LINE`: the **full** suite stayed
  green through a reword of it. Pinned as a prefix, since git's own diagnostic is
  appended after it.
- **The design spec's `no-undo-audit.sh` row** still gave a closed either/or over
  empty-list states as the only causes of `stash: null`. It now names both
  shapes, this PR's nonempty-list-at-nonzero-rc one included.

Refuted by measurement, filed as one closed `wontfix` record: an else-clause for
an empty `diag` on the rc branch (no constructible repo state reaches it — ten
corruption fixtures tried, and `git stash drop` on the corrupt entry, the one
named trigger, itself exits 128 leaving the fault intact); a "stale" `diag`
comment (its own antecedents scope it to the empty-list causes, and both
entrances measure as it says); and a `corruptObject` test helper (net **+2**
lines across two call sites, and it de-parallelizes two of four sibling
fixtures).

| mutant | expected | measured |
|---|---|---|
| reword the rc branch's `msg` | the #482 test alone reds | pass 78, fail 1 — the #482 test |
| rename `sl_rc` → `list_rc` throughout | the #482 test stays green | pass 79, fail 0 |

```
$ node --test skills/fleet/scripts/*.test.mjs   -> tests 1519, pass 1519, fail 0 (exit 0)
```


feigi:
Review follow-up applied in 5fbaccc: the survived spec-row drift plus three in-scope `suggestion`s that survived their own refuters (the non-tip prose over-claim, the `awk 'END{print NR}'` count, and the `RC_FAILED_LINE` pin). Three refuted by measurement are recorded closed in #1109; two more were ruled no-action by the controller and are named there too.

`node --test skills/fleet/scripts/*.test.mjs` → tests 1519 / pass 1519 / fail 0, exit 0. `shellcheck -x -S warning` over 17 tracked `*.sh` → exit 0. PR body reconciled: it asserted the `[ -z "$sl" ]` guard the follow-up deleted.

Issue #971, opened 2026-08-28
Declare the implementer tier in frontmatter, and pair within each run


feigi:
Spec: docs/specs/2026-08-27-fleet-member-outcomes-instrumentation-design.md
Part 2 of 2. Depends on the Part 1 scraper being merged.

Declares the implementer's model+effort in an agent definition (the only
mechanism that decouples the two — the Agent tool has no effort parameter), then
moves one implementer per wave to the alternate tier so tier stops being
confounded with calendar date. Closes #864 and #472 along the way.

Implementers ONLY. Reviewer/finisher/merge-bot are not declared here — their
rows are recorded by the Part 1 scraper regardless, so the review side (84% of
run cost, the bigger prize) accumulates history without a decision being made
about it yet.

Issue #270, opened 2026-08-10
candidates.mjs reports a BROKEN query as an empty queue (exit 1, not 2) — non-deterministic, and only CI's node 22 has caught it


feigi:
Surfaced by CI on PR #251, which **cannot have caused it** — that PR's diff is `docs/specs/2026-07-23-fleet-plugin-design.md`, `skills/fleet/commands/run-merge-bot.md` and a comment in `skills/fleet/scripts/pr-overlap.mjs`. It touches neither `candidates.mjs` nor `candidates.test.mjs`.

## The failure

`skills/fleet/scripts/candidates.test.mjs:485` — *"the refusal survives a gh stderr larger than the pipe buffer — the case that motivated it"*. It fails at its **first** assertion, `:497`:

```
not ok 38 - the refusal survives a gh stderr larger than the pipe buffer — the case that motivated it
  location: 'skills/fleet/scripts/candidates.test.mjs:485:1'
  error: Expected values to be strictly equal:  1 !== 2
  expected: 2
  actual: 1
  operator: 'strictEqual'
  stack: TestContext.<anonymous> (…/candidates.test.mjs:497:10)
```

`:497` is `assert.equal(status, 2)`. The test drives a deliberately malformed jq program (`JQ_OVERRIDE: "z".repeat(100_000) + "((("`), so the query is **broken** and the contract requires exit **2**.

It exited **1**.

## Why that is the serious reading

`candidates.mjs`'s three-way exit contract is the whole point of the script's error surface, and `docs/specs/2026-07-23-fleet-plugin-design.md` documents it as:

```
0 = survivors returned      1 = the query succeeded and the queue is empty      2 = the query broke
```

Exit 1 means **"there is no work"**. So a broken query is being reported as *an empty candidate queue*. That is the silent queue-shrink this repository keeps closing from other directions (#3, #64, #111, #262): the caller is not told the question failed, it is told the answer is "nothing".

`/fleet:run-team` phase 0 gates its entire ticket supply on this script. A run that hits this reports "no candidates" and stops, with nothing in the log to distinguish it from a genuinely empty queue.

## Non-determinism — this is the part that needs a decision

The same code passes and fails depending on the run:

| where | node | result |
|---|---|---|
| PR #251 run `31406163776` attempt 1 | 22 (CI) | **fail**, status 1 |
| PR #251 run `31406163776` attempt 2 | 22 (CI) | **fail**, status 1 |
[...]

PR #1117, opened 2026-08-31
docs(issue-tracker): state the closing-keyword adjacency rule and the after-open check (#393)

### LINKED ISSUE #393: triage PR bodies write "closed #N" as prose and GitHub links it as a closing reference — twice in one day, on two different PRs
Two instances in a single day, both on `/triage`-authored PRs, both caught only because a reviewer or finisher happened to check `closingIssuesReferences`.

## The mechanism

GitHub's closing keywords are `close/closes/closed/fix/fixes/fixed/resolve/resolves/resolved`, and a keyword **immediately adjacent** to `#N` creates a closing link when the PR targets the default branch. Descriptive past-tense prose about what a triage pass did — "Triage closed #219…" — is indistinguishable from an intentional `Closes #219`.

## The two instances

| PR | body phrase | linked | actual subject of the linked issue |
|---|---|---|---|
| **#370** | `and closed #270 as already implemented` | #270 | `candidates.mjs` exit-code bug — nothing to do with the PR |
| **#382** | `Triage closed #219 and #220 as rejected enhancements.` | #219 | a duplication question the PR only *records*, does not resolve |

Both fixed the same way — insert a word to break the adjacency: `closed issue #270`, `closed issues #219 and #220`. Verified in both cases that `closingIssuesReferences` went to `[]` afterwards (GitHub recomputes lazily, so re-query after a pause).

**A detail worth keeping:** in #382 only `#219` linked, not `#220`. The keyword does **not** chain across `and`, unlike an explicit `Closes #219, #220`. So a body naming several issues links exactly the first one — which makes the damage look arbitrary and easy to miss.

## Why it matters, precisely

Bounded, but not nothing. Both linked issues were already closed, and GitHub does not re-close a closed issue, so no `ClosedEvent` is added on merge. The real effects:

- `closedByPullRequestsReferences` on the linked issue returns the unrelated PR **immediately**, before any merge — so the issue renders a linked-PR association that misdescribes its history.
- If the issue is ever **reopened before the PR merges**, merging re-closes it.
- A `/triage` pass that closes an issue *and* mentions it in a PR body produces a record where the PR appears to be the closer, which is exactly the attribution this repo cannot reconstruct afterwards — every agent writes through the same credentials.

## Why it is filed against triage rather than the fleet

The fleet's own PRs do not have this problem: `next-ticket` step 7 tells implementers to write a single explicit `Closes #N`, and everything else about the ticket goes in the commit body. The two affected PRs were both written by a triage pass narrating what it had just done. So the defect lives in how triage composes a PR body, and the fix belongs there.

- **Cheapest:** a line in the triage PR-body guidance — never write a bare closing keyword before `#N` in narrative prose; write `closed issue #N`, or name it without the `#`. One sentence, no tooling.
- **Detect rather than prevent:** have whatever opens the PR re-query `closingIssuesReferences` afterwards and refuse or warn on anything it did not intend. Catches hand-written bodies too, and this is what actually caught both instances — but only by luck of a reviewer looking.
- **Belt and braces:** add the check to the fleet's finisher gate for out-of-fleet PR classes, where an unintended link is most likely. #382's finisher already had it and that is how instance two was found.

The first is nearly free and prevents the common case; the second is the one that would have caught both without a human noticing. They compose.

## Scope searched

`ledger.mjs check` → `tracker-hit` naming **#3**, which its own scorer rated **0** and which is about implementer resupply — a false hit, of the kind filed as #388. Plus `gh issue list --state all --search "closing keyword PR body"` → no results. Read #370's and #382's review threads for prior art; neither filed this, both fixed their own instance in place.

feigi:
Closes #393

Sizing: light

A closing keyword immediately before an issue reference in a PR body creates a
closing link, so narrative prose about what a pass already did is
indistinguishable from a deliberate reference written to close. Two
triage-authored PRs shipped one, and both were caught only because someone
happened to check `closingIssuesReferences`.

Both remedies the ticket rules on are prose, and both land in
`docs/agents/issue-tracker.md` under a new
`## When a skill says "compose a PR body"` section: the guidance that prevents
the common case, and the after-open check that catches a body nobody wrote to
the guidance.

## Homed per the amending brief, not the issue body

`git ls-tree origin/main skills/` lists only `skills/caveman-compress` and
`skills/fleet`, and `ls -ld skills/triage` in this worktree returns
`No such file or directory` — an edit there escapes the worktree into an
unversioned directory and leaves a clean `git status`. `docs/agents/` is
tracked: `git ls-tree origin/main docs/agents/` lists `domain.md`,
`issue-tracker.md` and `triage-labels.md`. Nothing under `skills/triage/`,
`skills/grilling/` or any sibling symlink was touched.

## Why a doc in `docs/agents/` is the intended delivery surface

Raised in review, and worth stating rather than leaving implied.

The rule is not specific to one skill. It binds any agent that composes a PR
body, and this repo's convention for a rule with that shape is a single home
under `docs/agents/` rather than a copy inside each skill that needs it —
`CLAUDE.md` names the single-context layout, and the alternative has a measured
failure mode this repo already polices: `tracker-block-copy-prose.test.mjs` and
`caveat-unification-prose.test.mjs` exist because copies of a shared rule drift
apart, twice, in exactly this file.

The honest limit: nothing in the tree yet routes an agent here. This PR gives
the section a dispatch trigger in the form the file already uses for the
sections that carry one —

```
$ grep -o '^## When a skill says "[^"]*"' docs/agents/issue-tracker.md
## When a skill says "compose a PR body"
## When a skill says "publish to the issue tracker"
## When a skill says "fetch the relevant ticket"
```

— but no skill emits any of those three phrases, including the two that
predate this PR. Wiring the PR-authoring skills to cite the section is outside
#393's amended `Key interfaces`, which names `docs/agents/issue-tracker.md`,
and `run-team/SKILL.md` is the target of two other open tickets. Filed as
**#1118** so the reachability gap is tracked rather than implied by this PR's
silence.

## Every claim in the section re-derived at this commit

- **The in-place corrections held.** `gh pr view 370 --json closingIssuesReferences`,
  and the same call on 382 and 941, each return an empty list. The bodies now
  read `closed issue #270` and `closed issues #219 and #220`
  (`gh pr view 382 --json body`), which is the inserted-word fix end to end.
- **The regex over-reports, so the section forbids trusting it alone.**
  Scanning PR bodies for a keyword adjacent to a reference and comparing each
  hit against `closingIssuesReferences` leaves one hit GitHub did not link —
  1028's body carries such a string inside a code span, and
  `gh pr view 1028 --json closingIssuesReferences` returns 439 alone, the issue
  that PR set out to close.
- **Repeating the keyword is what links each reference in a multi-issue body.**
  973's body reads `Closes #971, closes #864, closes #472`, and
  `gh pr view 973 --json closingIssuesReferences --jq '[.closingIssuesReferences[].number]'`
  returns `[472,864,971]` — every one of them. This replaces an earlier draft
  of the section that prescribed `Closes #A, #B`, a single keyword before a
  list, as the form that links both; that contradicted the section's own
  non-chaining rule and was never measured. Caught in review.
- **The non-chaining behaviour is the one claim not re-derivable from the tree.**
  382's body was corrected in place, so the section states it as recorded on
  382 before that correction rather than as a live measurement.
- **The model the section points at is real, and now cited for what it does.**
  `skills/fleet/skills/next-ticket/SKILL.md` step 7 runs
  `gh pr create --base main --body "…\n\nCloses #N"` — narrative and the
  closing keyword in one `--body`. An earlier draft attributed a PR-body /
  commit-body split to that file;
  `grep -n -iE 'commit (body|message)' skills/fleet/skills/next-ticket/SKILL.md`
  returns nothing at exit 1, and the worked example above is the opposite of
  that split. Caught in review, and corrected to the rule that actually
  prevents the bug: one deliberate `Closes #N`, a word in front of every other
  mention.

## Acceptance criteria

- Adjacency rule, its fix, and the non-chaining detail — stated, and pinned as
  contiguous spans.
- Closing references re-queried after opening and compared against intent, with
  a mismatch surfaced rather than accepted.
- The lazy-recompute caveat sits in the same paragraph run as the check, and the
  check is described in exactly one place, so there is no second site for it to
  go missing from.
- Issue comments explicitly excluded — keywords bind in PR bodies and commit
  messages only.

## The pin, and what it discriminates

`skills/fleet/scripts/closing-keyword-prose.test.mjs` pins the rules that
decide behaviour, each as **one exact contiguous span** rather than two matches
on the same slice: N matches pin N facts and never the join, so a clause
spliced between them reverses a rule with every asserted token still present.

Mutation-tested both ways on a scratch copy, never in the worktree. Every
semantic mutant reds exactly its own pin and no other — dropping the fix
clause, splicing an exception into the rule, reversing the non-chaining
sentence, dropping the surface-the-mismatch clause, dropping the lazy caveat,
dropping the compare-against-the-linked-set rule, swapping the scope mapping so
comments bind and bodies are exempt (which keeps every asserted token in
place), truncating the nine-keyword enumeration, and amputating the preventive
rule's every-other-mention clause. Every benign control stays green —
unwrapping each paragraph to a single line, rewording an unpinned sentence,
appending a new unpinned sentence, and rewrapping the two spans pinned last.

The final two pins — the keyword enumeration and the preventive rule — were
added in review, which measured that deleting six of the nine keywords, or the
clause saying what to do with every mention other than the deliberate one, left
the whole suite green.

## What the restructuring could have broken

Which suites read the doc was measured, not grepped: replacing
`docs/agents/issue-tracker.md` with `MUTATED` in a scratch copy and running the
suite reddens `closing-keyword-prose.test.mjs`,
`tracker-block-copy-prose.test.mjs` and `issue-tracker-prose.test.mjs`, and no
others. Further files mention the path in comments only — which is why a
literal-path grep is not the settling command here, and why a stale comment
that relies on one is filed as **#1119**.

`tracker-block-copy-prose.test.mjs` compares the fetch-the-ticket block
byte-for-byte against `run-team/SKILL.md` and ends its slice at the **next
heading**, so this section is a sibling heading placed outside that block
rather than prose appended inside it. Renaming the section's heading in review
moves a heading, which is exactly what that slice is sensitive to, so it was
re-run rather than assumed. All three suites are green.

```
ℹ tests 1536
ℹ pass 1536
ℹ fail 0
```

## Review

Applied: the `next-ticket` citation, the multi-issue form, the dispatch
trigger, and the two pins above. Deferred and filed: **#1118** (routing the
section from the skills that author PR bodies), **#1119** (a stale grep claim
in an untouched test file). Two suggestion-band findings were checked by
refuters and did not survive; the measurements are recorded in **#1120**,
closed.


feigi:
Review applied at `b8303d40d979fc2a3df52952fef015a76c5007db`.

**Applied** — three survived findings, plus two suggestion-band findings that a refuter checked and could not refute:

- `docs/agents/issue-tracker.md` — the `next-ticket` citation named a rule that file does not state, and its worked example contradicts. Corrected (critical, dimension `comments`).
- `docs/agents/issue-tracker.md` — `Closes #A, #B` prescribed as "the form that links both" contradicted the section's own non-chaining rule. Replaced with the repeating form, measured on #973 (important, dimension `correctness`).
- `docs/agents/issue-tracker.md` — the section took a dispatch trigger in the file's existing `## When a skill says "<phrase>"` form (important, dimension `correctness`, applied in part).
- `skills/fleet/scripts/closing-keyword-prose.test.mjs` — pinned the nine-keyword enumeration (dimension `tests`) and the preventive rule's every-other-mention clause (dimension `correctness`). Both were unpinned; each new pin was mutation-checked to redden alone and to survive a pure rewrap.

**Deferred and filed:**

| Issue | Label | Finding | Why deferred |
|---|---|---|---|
| #1118 | `ready-for-agent` | Nothing routes an agent to the section; cite it from next-ticket step 7 and the run-team implementer prompt (`docs/agents/issue-tracker.md:28`, dimension `correctness`) | Outside #393's amended `Key interfaces`; `run-team/SKILL.md` is the target of two other open tickets |
| #1119 | `ready-for-agent` | `issue-tracker-prose.test.mjs:6` claims grepping `issue-tracker` matches only that file; eight files match (dimension `comments`) | File is not in this diff and the comment was already stale before this PR |
| #1120 | `wontfix`, closed | The suggestion band, checked — two findings refuted by measurement | Not deferred work; a durable record so a later run lands on the evidence |

**Refuted, with the measurement in #1120:** the missing default-branch qualifier (every one of this repo's PRs targets `main`, the fleet's only create path hard-codes `--base main`, and the remedy would have reddened two pins), and the claim that the regex caveat and the issue-comment exclusion are outside the ticket (both are in #393's `## Agent Brief` comments, which outrank the body).

**Tests** at the pushed SHA:

```
ℹ tests 1536
ℹ pass 1536
ℹ fail 0
```

Baseline before these edits was `tests 1534 / pass 1534 / fail 0`; the two added tests are the two new pins. No `--no-verify`; the pre-commit hook ran clean.

The PR body has been reconciled against the final diff — the heading name, the pin count, the suite count, and the list of suites that read the doc had all gone stale, and the body now records the two claims this review corrected rather than the ones it originally shipped.

Issue #1120, opened 2026-08-31
PR #1117 review: the suggestion band, checked


feigi:
Deferred from PR #1117 review (ticket #393). Durable record of the `suggestion`-band findings that were **checked and did not survive**. Closed on filing — `ledger.mjs check` reads the tracker at `--state all`, so a later run re-deriving one of these lands on the evidence instead of re-litigating it.

The review budgets the `suggestion` band 0 refuters, so every entry arrived unchecked. Each in-scope one got one refuter dispatched biased to refuse. Two survived and shipped in PR #1117 — the keyword-enumeration pin and the preventive-rule pin. The two below were refuted by measurement.

---

## Refuted — the missing default-branch qualifier

`docs/agents/issue-tracker.md:32` · dimension **correctness** · severity suggestion

**Claim:** the section drops #393's default-branch qualifier — closing links are created only when the PR targets the default branch — so the rule over-reaches on a PR based on a non-default branch.

**Refuted.** The qualifier names an empty category in this repo:

```
$ gh repo view feigi/claude-config --json defaultBranchRef --jq .defaultBranchRef.name
main
$ gh pr list --repo feigi/claude-config --state all --limit 500 --json baseRefName \
    --jq '[.[].baseRefName]|group_by(.)|map({b:.[0],n:length})'
[{"b":"main","n":333}]
```

Every PR this repo has had targets the default branch, and the fleet's only PR-create path hard-codes the base — `skills/fleet/skills/next-ticket/SKILL.md` runs `gh pr create --base main`.

The rule is also a prohibition ("do not write a bare keyword adjacent to `#N`"), so the omission errs conservative: applied on a non-default base it costs one extra word and creates no wrong link. The doc nowhere asserts a link *will* exist. The finding's second half — a clean re-query on a stacked PR being misread as the doc being wrong — is already covered in-section twice, by the lazy-recompute caveat and by "Compare a regex hit against the linked set before calling it a defect".

**And the remedy would have cost two pins.** The mechanism sentence is deliberately pinned as one contiguous span precisely so a spliced qualifier reddens. Applied as written:

```
baseline: ℹ tests 63  ℹ pass 63  ℹ fail 0
with fix: ℹ tests 63  ℹ pass 61  ℹ fail 2
  ✖ the adjacency rule names its fix in the same breath
  ✖ a reflowed section still matches
```

[...]

Issue #1115, opened 2026-08-31
PR #1114 review: the suggestion band, checked


feigi:
Record of the `suggestion` band from the PR #1114 review (`fix/529-guard-pin-tautology`, ticket #529, head `5593e277e7d17a0d7bc04ab75e01705cd94e921c`). Dimensions run: `correctness`, `tests`, `comments`; `dimensionsUnrun` empty. The workflow returned **0 survived, 0 refuted, 4 unverified** — all four in the `suggestion` band, which the workflow budgets 0 refuters, so all four arrived unchecked rather than cleared.

Closed as `wontfix`: one finding survived its refuter and was applied in the PR; the other three are no-action rulings. Nothing here is queued agent work. `ledger.mjs check` reads the tracker at `--state all`, so a future run re-deriving any of these lands on this record instead of re-litigating it.

## Checked and APPLIED — not deferred, recorded here for the band's completeness

**`skills/fleet/scripts/implementer-model-tier.test.mjs:303` — dimension `correctness`.** Claim: the comment's mutation recipe named the search string `one run.**`, which is not unique in `skills/fleet/skills/run-team/SKILL.md`, so a mechanical global replace applies a different mutation than the one documented.

**Refuter verdict: `refuted: false` — survived, and the defect is worse than the finding claimed.** Measured, not argued:

- `grep -c 'one run\.\*\*'` on that SKILL → **2**: the guard lead, and `**Why not decide inside one run.**`.
- That second hit is the `section()` **end anchor** of `skills/fleet/scripts/member-outcomes-prose.test.mjs:21`.
- Applying the old recipe with a global replace: `tests 1522 / pass 1519 / fail 3`, exit 1, three `member-outcomes-prose` failures under `AssertionError: anchor moved: **Why not decide inside one run.**`.

That red is the trap. A reader following the recipe mechanically reads an unrelated suite's red as *the deleted pin catching the hedge* — the exact conclusion the comment exists to prevent. This ticket is `class=correction`, and a settling command that does not do what it says is the defect class it exists to remove, so it was applied despite the `suggestion` band.

[...]

Issue #1112, opened 2026-08-31
PR #1110 review: the suggestion band, checked


feigi:
Record of the `/fleet:run-team` review of **PR #1110** (`fix/filing-claim-bar`, ADR 0002 — a second filing bar for `review-and-fix.md` step 5), reviewed at `2cbff0bcc2ecb9737acb7b37331a0a4fdbf68ab9`.

Dimensions run: `correctness`, `tests`, `comments`. `dimensionsUnrun` empty. Eight findings: 0 `survived`, 4 `refuted` by the workflow's own refuters, 4 `unverified` (the suggestion band, which the workflow budgets 0 refuters).

**Closed, not open.** Every entry below was CHECKED and did not survive, or was ruled no-action. None is a confirmed defect and none is a torn call, so none is `ready-for-agent` or `needs-triage`. Closed is still a durable home — `ledger.mjs check` reads the tracker at `--state all`, so a future run re-deriving one of these lands here rather than re-litigating it.

**Read the printed rows, not the exit code.** `check` scores tracker rows on TITLE only (`skills/fleet/scripts/ledger.mjs:607`, `overlap(scored, scoreTokens(h.title || ""))`), so an entry recorded in this body contributes nothing to this issue's score. A re-derivation will surface this record as a low- or zero-scored row at best, possibly under `soft-hit`'s exit 0 — measured this run at 0.00 for #1107 and #1109. The record is durable; the guard is advisory.

---

### 1. `important` / `correctness` — `docs/adr/0002-filing-second-bar-worth-a-claim.md:72`

**Claim:** ADR 0002's Trigger B cannot fire — its threshold is already false at the ADR's own measured baseline, so the "bar missed the load" half of the pre-chosen guard is dead on arrival, and this PR's pin test freezes it verbatim.

[...]

PR #1121, opened 2026-08-31
refactor(fleet): extract shared lift(code, name, signature) helper

### LINKED ISSUE #533: skills/fleet/scripts: the new Function source-lift is pasted into four test files — extract lift(code, name, signature)

feigi:
Closes #533

## What

`workflows/review-pr.js` runs a top-level `await pipeline(...)` and cannot be imported, so its pure functions get lifted out of the source text with `new Function` instead. That technique was pasted into a private copy in `review-pr-reads.test.mjs`, `review-pr-testcmd.test.mjs`, and `review-pr-snapshot-path.test.mjs` (the last deleted outright here, per the ticket's Agent Brief — the generic form already covered it unchanged), plus the `resolveDimensions` branch of `select-dimensions.test.mjs`'s dispatcher. Each of those private copies is replaced by one call into a new `skills/fleet/scripts/lift.mjs` exporting `lift(code, name, signature)`, so the first-match-wins ceiling is documented and pinned once, as a property of the shared helper, rather than once per copy.

`lift.mjs` gets its own direct suite, `lift.test.mjs`: the accept path (previously exercised only by side effect at each call site, never asserted to return a correct function rather than merely not throw), the refusal path (no original call site ever fed it a name/signature that fails to match, so that branch was dead under every green run), the first-match ceiling pinned as behavior rather than only documented in a comment, and — added during review — that a regex metachar in `name` or `signature` matches literally.

## What stays local, and why

- `select-dimensions.test.mjs`: `DEFAULT_DIMENSIONS` (array literal, no function signature), `selectDimensions` (closes over a preceding `const` declared immediately above it — lifting the function alone drops it out of `new Function`'s eval scope), `verifiersFor` (needs injected `new Function` parameters, which `lift()` has no channel for). Each already carried this reasoning in-file; a one-line pointer to why it can't route through `lift()` was added to each.
- `review-pr-unrun.test.mjs`'s three-function SEAM (`unrunReason` wrapped by `unrunEntries` wrapped by `unrunCrashed`) — **not named as an exception in the ticket, but measured to be one.** Every one of the three is exercised directly by a test in that file, and `unrunEntries`/`unrunCrashed` call their callee by name in the lifted source text, so each needs every function it calls present in the same `new Function` eval scope. Isolating `unrunEntries` alone through `lift(CODE, "unrunEntries", "review, dimension")` throws `unrunReason is not defined` the first time the returned function runs (measured). Left local with a comment explaining the incompatibility — same class as `selectDimensions`/`verifiersFor`, just interdependence instead of a closed-over const or an injected parameter.
- `review-pr-unrun.test.mjs`'s `findingsSchema()` lift of `FINDINGS_SCHEMA` — also not named in the ticket. It's a `const` object literal, not a `function name(signature)` declaration, same reason `DEFAULT_DIMENSIONS` stays local. Left local with a one-line comment.

## Sites found that are out of this ticket's class entirely

Re-measured the full population of `new Function(` lift call sites in `skills/fleet/scripts/` (the ticket's own body notes its file count had gone stale twice). Two more files use `new Function`, neither touched here:
- `review-pr-refuter-scratch.test.mjs` — renders a template literal with injected scope variables, not a function-declaration extraction.
- `spend-view.test.mjs` — lifts from `board.html`'s inline `<script>`, a different source file entirely, and (like the unrun SEAM) combines two interdependent declarations into one eval scope.

Both are a structurally different use of the `new Function` primitive than the review-pr.js source-lift duplication this ticket addresses, so out of scope here.

## Applied from review

**A regex-injection bug this diff would have introduced.** `lift()` interpolated `name` and `signature` into its pattern raw, so a metachar in either matched as pattern syntax. Three of the converted call sites previously used literal regexes with no interpolation at all, so the diff newly extended the hazard to them. Live case, measured on `86e1ba7`: `lift("function pick(a, b, rest) {…}", "pick", "a, ...rest")` returned the function — the `.` characters matched a comma and a parameter — where it must refuse. Both interpolations now go through `RegExp.escape`, and the same call at head prints `throws: review-pr.js no longer declares pick(a, ...rest) at top level — update this test`.

**A vacuous pin in this PR's own new test.** The refusal test's fixture was a single line, and `lift()`'s pattern ends at `^\}$`, so that fixture could never match whatever name or signature was asked: both `assert.throws` calls were satisfied by the terminator anchor rather than by the mismatch they name. Dropping `name` and `signature` out of the pattern entirely left the test green. The fixture now spans lines, and each half discriminates (matrix below).

**Three header claims that did not hold**, each corrected against a command run on the tree: the header pointed a reader to "this repo's domain notes" for a definition of TEXT-LIFT pinning that neither `CONTEXT.md` nor `docs/agents/domain.md` carries (`grep -rn "TEXT-LIFT" CONTEXT.md docs/agents/domain.md` → no output; the term itself is real and stays — `grep -rli "text.lift" skills/fleet/scripts/*.test.mjs` names arg, candidates and review-pr-reads); the duplicate-declaration guard was described as existing "per caller" when `grep -rn "each function is declared exactly once at top level" skills/fleet/scripts/*.test.mjs` returns a single hit, scanning review-pr.js's stripped source, so what it covers is any lift whose `code` is that text and nothing else; and the AC-4 comment stated counts of copies and call sites where the durable claim is the property.

**Both of the pattern's `^` anchors are now pinned.** They were not: dropping the leading `^` (what makes "top level" mean anything) and reducing the terminator `^\}$` to a bare `\}` each left the whole suite green. The second matters most — with a bare `\}` the non-greedy body stops at the first closing brace anywhere, so a function containing a nested block is lifted truncated, and that surfaces as a `new Function` SyntaxError thrown from module scope in whichever caller imports the helper: the opaque file-level error `lift.mjs`'s header says the once-declared guard is kept out of this module to avoid. Measured before the new cases existed, it reddened four sibling files as whole-file crashes. Two cases now catch both, named, in the helper's own suite.

**Two comments left citing `liftResolveTestCmd`,** a function this diff deletes, now cite the construct instead. `grep -rn "liftResolveTestCmd" . --exclude-dir=.git` leaves one hit, in a dated design spec that names it as precedent for what existed then — routed to #516 rather than rewritten as a live claim.

## Deferred from review

- **#1125** (open, `ready-for-agent`) — `select-dimensions.test.mjs` lifts `resolveDimensions` from raw `SOURCE` while every other caller passes stripped text, so a commented-out copy above the real declaration is what gets lifted. Measured against this head. Pre-existing (the same lift matched `SOURCE` on `main`) and a caller-side behavior decision, so out of this ticket.
- **#1124** (closed, `wontfix`) — the record of the review's suggestion band: five findings checked by refuters that each ran something, none of which survived.
- **#516** — comment carrying three citation residuals: the spec's dangling symbol, plus two line ranges into `select-dimensions.test.mjs` that were already wrong at this PR's base.

## Testing

`node --test skills/fleet/scripts/*.test.mjs` → `tests 1532` / `pass 1532` / `fail 0`.

Mutation matrix over `lift.mjs`, run on a scratch copy of the file (the worktree never mutated), each row `node --test lift.test.mjs` against the tests in that file:

| mutation | result |
|---|---|
| drop both `RegExp.escape` calls | `fail 1` — the metachar test alone |
| return the matched text instead of evaluating it | `fail 3` — accept, ceiling, and the metachar test's positive control |
| change the refusal message's wording | `fail 2` — both tests that quote it |
| take the LAST match instead of the first | `fail 1` — the ceiling test alone |
| ignore `name` in the pattern | `fail 1` — the refusal test alone |
| ignore `signature` in the pattern | `fail 2` — refusal and metachar |
| drop the leading `^` (top-level anchor) | `fail 1` — the nested-declaration case alone |
| reduce `^\}$` to `\}` (terminator anchor) | `fail 1` — the column-0-terminator case alone |
| rename the internal `m` binding (control) | `pass 6` |

Also verified dependency, re-measured at this head on a scratch copy of the tree: replacing `lift.mjs`'s body with a throw leaves `pass 1453 / fail 10`, and the failing files are exactly `lift.test.mjs`, `review-pr-reads`, `review-pr-testcmd`, `review-pr-snapshot-path` and `select-dimensions` — `review-pr-unrun.test.mjs` stays green, confirming it does not depend on the new module.

Sizing: heavy


feigi:
Review deferrals filed:

- **#1125** (open, `ready-for-agent`) — `select-dimensions.test.mjs` lifts `resolveDimensions` from raw `SOURCE`, so a commented-out copy above the real declaration is what gets lifted and evaluated. Measured against this head; pre-existing (the same lift matched `SOURCE` on `main`) and a caller-side behavior decision, so deferred rather than folded in here. `dimension: correctness`.
- **#1124** (closed, `wontfix`) — the record of this review's suggestion band: five findings, each checked by a refuter that ran something, none surviving. `dimension: types` (x2), `correctness`, `silent-failure`, `tests`, `comments`, `simplify`. Closed on filing per precedent (#1120, #1115, #1112); `ledger.mjs check` reads `--state all`, so a re-derivation lands on the measurements rather than re-litigating them.
- **#516** — [comment](https://github.com/feigi/claude-config/issues/516#issuecomment-5480989995) carrying three citation residuals: the design spec's now-dangling `liftResolveTestCmd` (left as a dated record, deliberately), plus two line ranges into `select-dimensions.test.mjs` that were already wrong at this PR's base. `dimension: comments`.

What survived was applied in `96a065a` and `2613d4c` and is described in the PR body.

PR #1122, opened 2026-08-31
fix(run-team): phase 3 reads phase 2's accumulated floor, not a set nothing records (#528)

### LINKED ISSUE #528: run-team phase 3: the tier guard gate fires "since the last check", a set nothing persists
Deferred from PR #523 review. Dimension: `correctness`. Band: `suggestion` — the workflow budgets that band 0 refuters, so nothing had looked at it; this fix-applier settled it on **scope**, not by refuting it.

**Location:** `skills/fleet/skills/run-team/SKILL.md:532-534` — the `**Pool empty**` bullet in phase 3.

## Finding (verbatim)

> Phase 3's gate for running the guard — "once three or more `class=routine` PRs have been ruled since the last check" — contradicts phase 2's floor, which the same commit reaffirms is "over the accumulated file, never one run", and names a set the controller cannot enumerate: nothing persists a last-check timestamp. Pre-existing text, not introduced by this PR, but this PR makes it bite — the accumulated floor is now permanently met (8 rows), so the incremental gate becomes the sole thing standing between the controller and the guard.

## Evidence recorded by the reviewer

- `SKILL.md:532-534` — "Run phase 2's tier guard here once three or more `class=routine` PRs have been ruled since the last check; nothing else in the loop owns it."
- Phase 2's floor — "the file holds at least three `class=routine` PRs spanning **two or more distinct `run_date`s**" … "That floor is over the accumulated file, never one run".
- Nothing persists the state: `grep -rniE "last check|last_check|lastCheck|guard.?(ran|run|checked)" skills docs` returns only `SKILL.md:534` itself plus two unrelated comments (`fleet-tick.test.mjs`, `board.mjs`). The TSV's columns are `run_date pr ticket class tier closed_own_ticket minted_false_claim note` — no check-state column, and no marker survives a compaction.
- Same defect shape the guard text itself already fixed one paragraph earlier: it rejected "compare this wave against the prior wave" because that "named two sets nobody can enumerate". The `/unit is the PR/` and `/no implementer waves/` pins exist for exactly that.

Outside PR #523's subject — the tier-guard revert and the phase-0 staleness probe. Confirmed against the pushed tree: the PR touches neither phase 3 nor this bullet.

```
$ git diff 512f7c7...9e8ee3d -- skills/fleet/skills/run-team/SKILL.md \
    | grep -E '^\+.*(last check|Pool empty|## Phase 3)'
(no output)
```

The remedy is undecided, and the two candidates are materially different builds:

- **(a)** drop the throttle and have phase 3 read the same accumulated floor phase 2 already defines, so the two cannot drift.
- **(b)** keep a throttle and give it a durable marker — a `checked` column in `tier-outcomes.tsv`, or a row count recorded in the ledger.

(b) is a schema change plus a new seam; (a) removes a throttle the maintainer may have wanted. Per phase 0's `decided?` test that is an undecided item, so this surfaces rather than guesses.

Deferred from PR #523 review.

feigi:
Closes #528

Sizing: light

## What changed

Phase 3's pool-empty event is the only place in the loop that dispatches phase 2's tier guard, so its wording IS the gate. It stated one of its own — *"once three or more `class=routine` PRs have been ruled since the last check"* — which failed twice over: a second threshold beside phase 2's, free to drift from it, and a key on a set nothing on disk records.

Remedy **(a)** as ruled in the Agent Brief: the gate now names phase 2's floor over the accumulated `docs/metrics/tier-outcomes.tsv` instead of restating it, so the two cannot disagree. **(b)** — a durable marker — is named in the prose as a schema change with its own ticket, rather than left implicit.

## Settling commands, run at this commit

The defect was live at this branch's parent (the string wraps a line break, so a single-line grep sees a false absence — flattened here):

```
$ git show HEAD~1:skills/fleet/skills/run-team/SKILL.md | tr '\n' ' ' \
    | grep -o 'guard here once three or more[^;]*;'
guard here once three or more `class=routine` PRs have been ruled since the   last check;
```

Nothing persisted that set, and after this change the only file carrying the wording is the pin that forbids it:

```
$ grep -rniE "last check|last_check|lastCheck" skills docs | sed 's/:.*//' | sort -u
skills/fleet/scripts/tier-guard-gate-prose.test.mjs
```

The metrics file has no check-state column to record it in:

```
$ grep -n 'run_date.pr.ticket' docs/metrics/tier-outcomes.tsv
74:# run_date  pr  ticket  class  tier  closed_own_ticket  minted_false_claim  note  sizing  profile  loc  files
```

AC-4 — phase 2's floor is unchanged. One hunk in `SKILL.md`, and it is the pool-empty bullet:

```
$ git diff origin/main...HEAD -U0 -- skills/fleet/skills/run-team/SKILL.md | grep '^@@'
@@ -927,2 +927,5 @@ multi-select**, and **a judgement the evidence cannot settle**.
```

That hunk is nowhere near the block `tracker-block-copy-prose.test.mjs` compares byte-for-byte against `docs/agents/issue-tracker.md` — that block opens at line 730 and ends at the `Re-derive the ticket's claims` rule, well above:

```
$ grep -n 'Read the issue with `gh issue view <N>' skills/fleet/skills/run-team/SKILL.md
730:> Read the issue with `gh issue view <N> ...
```

## AC-3 — the pin, mutation-tested both ways

`skills/fleet/scripts/tier-guard-gate-prose.test.mjs`, paragraph-tight slice via the shared `between()`.

The pin this PR first shipped was four word-presence checks over the slice plus two literal wording bans, and the review of this PR measured it green against the very defect it exists to catch. Replaced in 08b6942: the duty is now pinned as **one wrap-tolerant `phrase`** — the whole sentence — and the two defect classes are banned **by shape** rather than by wording: any count, once `phase N`, backticked spans and the pronoun *"the one that"* are removed; and a since/after clause reaching a recency word.

**Refuse.** Each shape below was applied to the real `SKILL.md` in a scratch copy, proven applied by diff, and reddens the pin it is aimed at. Each is also a fixture in the file asserting *which* pins fire, not merely that something went red:

| mutant | pin that reddens |
|---|---|
| the duty negated outright — *"Do NOT run phase 2's tier guard here…"* | `DUTY`, `NEGATED` |
| the floor rescoped — *"over this run's rows only rather than the accumulated `…tsv`"* | `DUTY` |
| phase 2's floor restated here in a paraphrase — *"at least three … two distinct `run_date`s"* | `DUTY`, `OWN_COUNT` |
| the incremental gate #528 removed, verbatim | `OWN_COUNT`, `UNPERSISTED_SET` |
| a count as *"at least four"*, *"more than three"*, *"six or more"*, *"3+"* | `OWN_COUNT` |
| an incremental key as *"after the most recent check"*, *"since it was last checked"*, *"not yet counted"* | `UNPERSISTED_SET` |

The first three and the last two of those were measured **green** against the pin this PR opened with; the review reproduced each with its own refuters.

**Accept.** Applied the same way, every pin stays green: the whole bullet reflowed onto one line, an unrelated clause in the slice reworded, and a `phase 1` sentence appended — the last being why the count ban strips `phase N` before it looks, and the reflow being why the fixtures search with `phrase` rather than with the line breaks the file has today.

**Ceilings, stated in the file rather than claimed away.** A count written inside backticks is exempt with them; a threshold carrying no number at all (*"once a handful have queued"*) is not covered, and no regex over prose will cover it; a recency set named without since/after (*"the rows this run appended"*) is caught only where it displaces the duty sentence.

**One review finding was refuted, not applied.** The `comments` finding on `SKILL.md:930` reads *"a throttle would need state nothing on disk records"* as a missing relative pronoun. It is a zero relative with `state` as the object of `records` — the construction this repo writes throughout, including the assertion message of the test that guards this very prose:

```
$ grep -rn 'nothing on disk records' skills docs
skills/fleet/scripts/tier-guard-gate-prose.test.mjs:125:  assert.doesNotMatch(gate(), UNPERSISTED_SET, "the pool-empty gate names a set nothing on disk records");
skills/fleet/skills/run-team/SKILL.md:930:  governs, and a throttle would need state nothing on disk records — a schema
```

`SKILL.md` is therefore unchanged by 08b6942 — the whole PR still touches two files:

```
$ git diff --stat origin/main...HEAD
 skills/fleet/scripts/tier-guard-gate-prose.test.mjs | 179 +++++++++++++++++++++
 skills/fleet/skills/run-team/SKILL.md               |   7 +-
```

## Enumerate and declare

The bug class is *a gate keyed on a set nothing persists*. Swept the whole document flattened, so a wrap could not hide a match:

```
$ perl -0777 -ne 's/\n/ /g; while (/(.{0,110}(since the last|since last|new since|not yet \w+ed|so far this run|since you last|since the previous|prior run|last run|previous run).{0,110})/gi) { print "$1\n---\n" }' SKILL.md
```

**Covered:** the phase-3 pool-empty gate — the only live member.

- *"the PREVIOUS run's own close-out PR"* (phase 0) — a diagnosis of a cause, not a gate, and the step it belongs to is already an enumerable command (`git merge --ff-only origin/main`).
- *"a worktree carried over from a prior run … predates the marker and has no `./agent-test`"* — enumerable by testing for the file, which is what the text tells you to do.
- *"pool — approved, not yet dispatched"* — enumerable from the ledger.
- Phase 2's floor and the tier guard's own thresholds — out of scope per the ticket, and AC-4 requires them untouched.
- The `checked` column / durable marker (option (b)) — explicitly out of scope; named in the prose as its own ticket.

## Tests

Run at 08b6942, on a clean tree:

```
ℹ tests 1536
ℹ pass 1536
ℹ fail 0
```


feigi:
Review findings adjudicated at 08b6942.

**Applied (3 of 5).** All three `survived` findings were `important` and 2-0; none was refuted. `correctness` at `tier-guard-gate-prose.test.mjs:38` and `tests` at `:41` are one defect — both regexes were literal-shaped — and are fixed once: the count ban now bans numbers rather than enumerating comparators, and the recency ban reaches the shapes measured green against the old one. `correctness` at `:51` (every positive pin a single-word presence check, so an outright negation of the duty passed all five tests) is fixed by pinning the duty as one wrap-tolerant `phrase`. Where a regex over prose cannot reach the whole class, the pin now names its ceiling instead of claiming coverage it lacks — the finding offered broadening and comment-narrowing as alternatives; both were done.

Correction to the `correctness` finding's own wording, measured: a **verbatim** restatement of phase 2's floor was already caught, because phase 2 reads *"at least three … spanning **two or more** distinct `run_date`s"* and "two or more" hit the old `OWN_THRESHOLD`. What slipped through was a one-word paraphrase of it, "at least N" / "more than N" / six-nine, and the incremental-keying half entirely. All of those are fixtures now.

**Filed (1 of 5).** #1123, closed `wontfix` — the `comments` finding on `SKILL.md:930` is refuted: the clause is a zero relative with `state` as the object of `records`, the same construction as the assertion message of the test that guards it.

**No action (1 of 5).** The `correctness` entry at `SKILL.md:926` carries no suggested fix and reports no scope creep — a clean bill of health, not a finding.

Every remedy was run before committing, and both pins were measured green together, since the two suggested fixes contradicted as written: the count ban had to exempt `phase N` for the duty pin's own `phase 2` to survive it.

```
ℹ tests 1536
ℹ pass 1536
ℹ fail 0
```

PR #1126, opened 2026-08-31
fix(review-pr): refuse a review whose tree is not the PR head, and stop re-creating a worktree that can be (#532)

### LINKED ISSUE #532: Phase 1: re-creating a worktree for a carried-over PR silently checks out a stale local branch, and the review snapshot is cut at the wrong commit
Measured this run by the `/fleet:run-team` controller. It cost a halted fix-applier and a review cut against a tree that was not the PR.

## What happened

PR #523 was carried over from a previous run; its worktree had been reaped, so the controller re-created one:

```
$ git worktree add .worktrees/pr523-tier-guard fleet/tier-guard-revert-and-staleness-probe
Preparing worktree (checking out 'fleet/tier-guard-revert-and-staleness-probe')
HEAD is now at cff7330 docs(fleet): revert class=routine to session tier, and probe ticket staleness in phase 0
```

A **local** branch of that name still existed in the checkout, left behind by the earlier run at `cff7330`. Meanwhile the branch had been rebased and `origin/` was at `9e8ee3d`. `git worktree add <path> <branch>` checks out the existing local branch and never consults the remote.

```
cff7330  parent 94acb8f   docs(fleet): revert class=routine ...
9e8ee3d  parent 512f7c7   docs(fleet): revert class=routine ...     <- gh pr view 523 --json headRefOid
```

Neither is an ancestor of the other. **Same subject line, different commits** — the same change on two different bases.

## Why it was not caught

Every signal available at that moment read normal:

- The `HEAD is now at` line quotes the commit **subject**, which was byte-identical to the PR head's. Reading it confirms the right *branch*, never the right *commit*.
- `git worktree list` showed the worktree present, on the expected branch.
- The worktree was clean.
- `claim-ticket.sh` was never involved — it only claims *fresh* tickets, so nothing in the fleet's tooling touched this path at all.

The review workflow was then handed that worktree and cut its immutable snapshot at `cff7330`, so **a full 22-agent review ran against a tree that was not PR #523**. A fix-applier was dispatched into it and began applying findings; it was halted before it could push, which would have been a non-fast-forward.

## What limited the damage — luck, not design

The two commits differed only in files carried by the base (`claim-ticket.test.mjs`, `drop-merged-label.test.mjs`, `inflight.test.mjs`, `no-undo-audit.sh`, `no-undo-audit.test.mjs`, `release-ticket.test.mjs`). All three files the findings targeted were **byte-identical** across both commits, verified by blob hash:

```
$ git rev-parse cff7330:skills/fleet/skills/run-team/SKILL.md
$ git rev-parse 9e8ee3d:skills/fleet/skills/run-team/SKILL.md      # identical
```

so the review's findings happened to remain valid. Had the rebase touched a finding-target file, every finding would have been about code that did not exist in the PR, and nothing in the pipeline would have said so.

## The gap

`run-team/SKILL.md` phase 1 already warns that a **reused** worktree may lack the `agent-test` runner, and names the carried-over case. It does not warn that the worktree may be on the wrong **commit**, which is the more damaging half — a missing runner fails loudly, a stale base fails silently and produces confident wrong output.

There is no check anywhere that a worktree's HEAD equals the PR head before a snapshot is cut.

## Candidate fixes — not choosing one

1. **Verify at creation.** After any `git worktree add` for an existing PR, assert `git -C <wt> rev-parse HEAD` == `gh pr view <N> --json headRefOid`, and refuse otherwise.
2. **Create from the remote ref.** `git worktree add <path> --detach origin/<branch>`, or delete the stale local branch first, so a leftover local ref cannot win.
3. **Verify at snapshot time**, inside `review-pr.js` — the snapshot agent already reports a head; have the caller compare it to the PR head and throw on mismatch. This is the same shape as #140 (`snap.path` handed out unchecked) and would cover hand-dispatched paths too.

(3) is the widest net; (1) is the cheapest. They are not exclusive.

## Scope searched

`git worktree add` invocations across `skills/fleet` (the fleet never scripts this for a carried-over PR — it is a controller action), `claim-ticket.sh` (fresh claims only, creates its branch from `origin/main`), and `run-team/SKILL.md` phase 1's reused-worktree paragraph. Not checked: whether `review-pr.js` already has any head-vs-PR assertion — believed not, but confirm before implementing (3).

feigi:
Closes #532

Sizing: light

The brief's ruling was **take (1) and finish (3)**, keeping the deliberate tolerance for a missing `prHead`. Both landed, plus (2) at creation time.

## Finish (3) — a mismatching head disqualifies the review, not just the diff

The comparison already existed, in `usableDiff`, where a mismatching `prHead` cost the review its **diff** and nothing else — so a review handed a tree that was not the PR ran to completion on the fallback read rules and returned findings about code the PR does not contain. `snapshotMissing` now refuses that snapshot, naming both commits. The diff drop stays where it is: it is the narrower guard and still correct for any caller that reaches it.

The compare is a second copy of one expression, because the Workflow sandbox forbids `import` and `lift()` evaluates a single declaration standalone. The copies are pinned **to each other** rather than to the comparison's own text: the pin anchors on the `if (snap.prHead && ` guard head and captures whatever comparison follows, so a change made to one side and not the other reds while a semantics-preserving rewrite of both stays green. Both directions are measured.

Both shas are normalized — trimmed and case-folded — once at the caller, ahead of the refusal. `git rev-parse` prints a trailing newline and `head` is relayed by an agent, and under the refusal an unnormalized `head` cancelled the whole review while naming two shas that look identical. Normalizing at the caller rather than inside either copy keeps the two copies byte-identical for the pin, and `usableDiff`'s three call sites read the same object.

## (1) + (2) — verify at creation, and create so it cannot happen

Phase 1 warned that a reused worktree may lack the runner and stopped there. It now also carries the commit half. Settled in a scratch repo against git 2.50.1 (Apple Git-155) rather than read off the ticket:

```
local feat  = bb8726d387e6f05aee1413d6aea7ab98f2484283
origin/feat = e8212407b83ff7503131d359c4d0722cf972dbf2
$ git worktree add "$D/wt" feat
HEAD is now at bb8726d same subject
ancestor: NO
$ git worktree add --detach "$D/wt2" origin/feat
HEAD is now at e821240 same subject
```

`worktree add <path> <branch>` takes the leftover **local** branch and never consults the remote. Both commits carry the same subject — which is what `HEAD is now at` quotes — and neither is an ancestor of the other: a fork, not a lag. Phase 1 now gives the `--detach origin/<branch>` form (removes the mechanism) and the verify-and-refuse pair (catches one anyway), reported naming both commits since the subjects will not separate them. It also states the asymmetry the brief asks for: **a missing runner fails loudly, a stale base fails silently and produces confident wrong output.**

## Acceptance criteria

- **Re-creating a worktree for a PR whose local branch is stale refuses, naming both commits** — phase 1, pinned by `worktree-head-prose.test.mjs`. This is a controller action with no script behind it (the ticket's own scope search found the fleet never scripts it), so the deliverable is the instruction plus its pin, not a new entrypoint.
- **A review handed a mismatching head refuses rather than proceeding without a diff** — `snapshotMissing`.
- **A missing `prHead` still proceeds, and a test pins the distinction** — `a snapshot with no prHead at all still proceeds` is the ACCEPT pin, and the mutant that makes an absent `prHead` refuse reds exactly it.
- **The normal path is unchanged** — `a matching head proceeds, abbreviated on either side`, which also holds the prefix tolerance in both directions. That tolerance got load-bearing: an abbreviated match used to cost a diff, and would now cost the whole review.
- **Full suite green.**

## Enumerate and declare

Covered: the `usableDiff` compare whose consequence was only to narrow; phase-1 re-creation for a carried-over PR; the ticket's passing case of an already-open PR's worktree a rebaser/resolver is sent into (same paragraph, same rule); hand-dispatched worktrees, via the review-pr refusal — and the prose says plainly that the backstop does **not** cover a worktree handed to an agent directly.

Deliberately left, each checked rather than assumed:

- **`claim-ticket.sh` fresh claims.** Out of scope per the brief; it builds from `origin/main`.
- **The phase-5 finisher's own `git worktree add --detach <scratch>/pr<N>/finish-<name> <dispatch pin>`.** Not a member of the class: already detached at an explicit SHA, so no local ref can win, and phase 5 already requires head-equality against the **dispatch pin** and argues explicitly against comparing to `headRefOid` there — a member that pushed after the pin would read as "settled". Adding a headRefOid check would break that.
- **`readRules`'s `skew` branch.** Now unreachable on this workflow's path, since the refusal runs first and a rejection with `diffPath` and `diffLines` both present is exactly the mismatch that already threw. Left in place with a comment recording it: `readRules` is a pure prompt builder pinned as one, and deleting the branch drops measured behaviour to buy nothing.

Residual risk, declared: if the snapshot agent ever relays `head` in a non-sha form that surviving trim and case-folding cannot rescue, the prefix compare fails both ways and the review now refuses where it previously only lost the diff. The failure is loud and names both values, and the same input already governed the diff drop.

Also declared: the head refusal is guarded on `snap.prHead &&`, so a failed `gh pr view` skips it. That is deliberate — absent and mismatching are different cases — but the run used to be indistinguishable from a checked-and-matched one. The snapshot run-log line now names the skip (`PR head (absent): head check SKIPPED`), and phase 1's prose says the backstop is conditional on `gh pr view` returning a head rather than claiming it unconditionally.

## Mutation verification

Every added test was broken and restored. On `snapshotMissing`: guard deleted, each half of the prefix tolerance dropped, absent-`prHead`-now-refuses, the two copies diverged, and the refusal message dropping either sha — all red their own pin, and the absent-`prHead` mutant reds the ACCEPT pin specifically. On the phase-1 prose: paragraph deleted, mechanism deleted, mechanism **negated**, `--detach origin/` downgraded to the bare branch, verify bullet deleted, refusal softened to a note, loud/silent **swapped**, subject-survives-rebase deleted, backstop widened to full coverage — all red their own pin. Benign controls stayed green: a full rewrap, and rewording each paragraph's lead.

On the two tests added under review (the normalization and the run-log line, both lifted as text and RUN rather than presence-pinned): normalization stops trimming, normalization moved after the refusal, its falsy-snap guard dropped, the log's absent-`prHead` marker removed — each reds its own pin with a message describing what actually happened. Benign controls stayed green: the two assignments reordered, and the log reworded around the marker.

https://claude.ai/code/session_01YC2ZkVMgSKjqLJDJ2bo39u


feigi:
## Review deferrals filed

| # | label | state | subject |
|---|---|---|---|
| #1130 | `ready-for-agent` | open | line citations in two docs rotted by this PR's +10 `readRules` insert (comments ×2) |
| #1131 | `ready-for-agent` | open | `readRules` says a rejected diff "is empty" when `wc -l` merely failed (silent-failure) |
| #1132 | `ready-for-agent` | open | the head-skew apparatus is unreachable after the refusal — three dead branches, one named (simplify) |
| #1133 | `ready-for-agent` | open | a `--detach` worktree cannot push with any push form the fleet instructs (carried over, checked here) |
| #1134 | `wontfix` | **closed** | `PR #1126 review: the suggestion band, checked` — five findings refuted by measurement, plus the 1/1561 snapshot artifact |

Applied in 8ac413a: the compare pin rewritten to capture rather than match a literal (Cluster A), sha normalization ahead of the refusal, the run-log naming a skipped head check, and phase 1's coverage claim corrected. `node --test skills/fleet/scripts/*.test.mjs` → **tests 1564, pass 1564, fail 0**.

Issue #1056, opened 2026-08-30
review-pr specialists measure the suite inside a `git archive` snapshot, which is not a git repo — every dimension reports one fewer test and one phantom failure


feigi:
Every dimension agent in a `review-pr` run reports a `test_run` block, and that block is the evidence the specialist offers that it validated the tree before reporting. On this repo it is **systematically wrong in a fixed direction**, and it was wrong in both reviews of one run.

## Measured

| PR | every dimension reported | controller measured, worktree at the same head |
|---|---|---|
| #1051 (3 dimensions) | `tests 1425, pass 1424, fail 1` | `tests 1426, pass 1426, fail 0` |
| #1052 (6 dimensions) | `tests 1447, pass 1446, fail 1` | `tests 1448, pass 1448, fail 0` |

Nine dimension agents across two independent runs, all reporting the identical wrong figure.

## Cause, reproduced directly

`workflows/review-pr.js` cuts its snapshot with `git archive | tar -x`. The result is not a git repository, and this repo has a test that depends on being inside one.

```
$ git archive 8787f4f | tar -x -C <scratch>/archtest
$ cd <scratch>/archtest && git rev-parse --show-toplevel
fatal: not a git repository (or any of the parent directories): .git
$ cd <scratch>/archtest && env -u FORCE_COLOR node --test skills/fleet/scripts/*.test.mjs
ℹ tests 1423
ℹ pass 1422
ℹ fail 1
ℹ skipped 0
✖ skills/fleet/scripts/unattended-git-sweep.test.mjs
```

The same commit measured in a **worktree**: `tests 1424  pass 1424  fail 0`.

**Delta: −1 test, −2 pass, +1 fail.** That delta is identical in the control above and in both reviews, which is what identifies it.

## What this rules out

The obvious hypothesis — the `tests` dimension mutating a shared snapshot while its siblings run their suites — is **false**. That predicts a race, and a race cannot produce the same count from nine agents across two runs. The determinism is the evidence against it.

## Why it matters

- A green PR reads as red. A controller or fix-applier chasing the phantom burns a cycle; this run avoided that only because the controller measured the head independently and told the fix-appliers to distrust the block.
- Worse in the other direction: the artifact **masks a real single-test regression**. From the payload alone, "one failure that is the artifact" and "one failure that is a genuine regression" are indistinguishable.
- It is silent. Nothing in the payload says the measurement environment differs from the tree under review.

## Prior art, same mechanism

[...]

PR #1127, opened 2026-09-01
fix(fleet): refuse a non-commit object at prove-merge's guard

### LINKED ISSUE #585: prove-merge.sh:71-72 — the guard peels ^{commit} but its consumers do not, so an annotated tag reports proved=false with every leg reading healthy
Deferred from the PR #577 review (`silent-failure`, severity `suggestion`, so **unrefuted** — the review budgets `suggestion` zero refuters by policy). **Pre-existing**; PR #577's diff touches only the guard block.

Filed by the fleet controller. The line citations below are **verified against `origin/main`**; the runtime measurements are the review's.

## The defect

`prove-merge.sh` peels `^{commit}` **inside the guard**, but not in the two lines that consume the same arguments.

Guard (`origin/main`, line 58):
```sh
for obj in "$pre" "$post" "$merge"; do
  git cat-file -e "${obj}^{commit}" 2>/dev/null || die "..."
done
```

Consumers (`origin/main`, lines 71-72):
```sh
pre_full=$(git rev-parse "$pre")
post_full=$(git rev-parse "$post")
```

So an **annotated tag** passes the guard — `cat-file -e "v1^{commit}"` peels and succeeds — and then `rev-parse "v1"` yields the **tag object's** sha, not the commit's. The later identity test (`[ "$second" = "$post_full" ]`) compares a tag sha against a commit sha and cannot match.

## Measured (review, real fixture, git 2.50.1)

Control, shas:
```
sh prove-merge.sh $H $H $M
rc=0  {"secondParent":"63b4611d…","headWasCurrent":true,"proved":true}
```

Same history, annotated tag `v1` (`git tag -a v1 -m t $H`) as pre and post:
```
sh prove-merge.sh v1 v1 $M
rc=1  {"preIsAncestor":true,"postIsAncestor":true,"secondParent":"63b4611d…",
       "firstParent":"3d0131dc…","parentCount":2,"proofPath":"no-rebase",
       "headWasCurrent":true,"proved":false}
stderr: headWasCurrent = true (want true, both paths)
        prove-merge: proved=false (path=no-rebase)
```

**Every emitted leg reads healthy and the verdict is false**, with nothing in the JSON or on stderr naming the real cause. Root cause isolated: `git cat-file -e "v1^{commit}"` → rc 0, while `git rev-parse v1` → `c8bd2a0f…` (the tag) vs `git rev-parse v1^{commit}` → `1690c0dd…` (the commit).

So an **argument-normalization mismatch is reported as a history disproof**.

## Why it is a suggestion, not a bug

It **fails closed** — exit 1, never a false proof — and the live call site passes shas: `skills/fleet/commands/run-merge-bot.md:138`. So no fleet path reaches it today.

It is filed anyway because the failure is *indistinguishable from a real disproof* by design: a disproof is exactly what an operator acts on, and here every diagnostic leg says the history is fine. Whoever hits it has no thread to pull.

## Why `needs-triage` — two defensible remedies

1. **Peel in the consumers**, matching what the guard already does:
   `pre_full=$(git rev-parse "${pre}^{commit}")`, same for `post`. Smaller diff, and it makes the guard's promise true for the rest of the script instead of only for the guard.
2. **Reject non-commit objects at the guard** — compare `git cat-file -t "$obj"` to `commit` — so a tag exits **2** ("could not evaluate") rather than 1 ("disproved").

These encode different contracts: (1) says tags are acceptable input and the script normalizes them; (2) says tags are out of contract and saying so is the script's job. That is a maintainer's call about the script's interface, not a mechanical fix — and it interacts with the exit vocabulary (`0` proved, `1` disproved, `2` could not answer) that #267 and #566 both worked to keep honest.

## Note on the diff that surfaced it

PR #577 reworded the guard's message to `cannot resolve $obj to a commit in this repository`. That wording is correct for the guard, but it **strengthens the impression that the script normalized the argument** — when only the guard did. Worth deciding alongside the remedy.

## Related

- #566 / PR #577 — the stderr-swallow and cause-assertion fix on the same guard block. Done.
- #267 / PR #556 — `prove-merge.sh`'s verdict-collapse defect (a probe failure reported as a disproof). **Adjacent and worth reading first**: this is the same *shape* — a non-history cause surfacing as a history verdict — reached through argument handling rather than probe status.

feigi:
## Summary

`prove-merge.sh`'s object guard peels `^{commit}` when checking that `pre`/`post`/`merge` exist, so an annotated tag clears it — a tag object dereferences to the commit underneath. The two `git rev-parse "$pre"` / `git rev-parse "$post"` lines right after do not peel, so they hand back the tag object's own sha rather than the commit's, and the identity test later in the script compares that against a commit sha and can never match. Reproduced directly against `origin/main` (343799c) in a scratch fixture: a control with plain shas returns `proved:true` at exit 0; the same history with an annotated tag as `pre`/`post` returns `proved:false` at exit 1, with `preIsAncestor`, `postIsAncestor`, `secondParent`, and `headWasCurrent` all reading healthy — nothing in the JSON or on stderr names the real cause.

Per the ruling on #585 (remedy 2): the guard now also checks the object's own, unpeeled type via `git cat-file -t "$obj"` and refuses anything that is not itself a `commit`, exiting 2 ("could not answer") rather than falling through to a disproof it was never equipped to make. Applied uniformly to all three positional arguments (`pre`, `post`, `merge`).

- A sha, a branch name, and a lightweight tag are all refs/objects that resolve straight to a commit with no tag object in between — `git cat-file -t` reports `commit` for each — so all three clear the new check and behave exactly as before. Verified: control (shas) and a lightweight-tag control both reproduce the pre-existing `proved:true` payload byte-for-byte.
- An annotated tag in any of the three positions now exits 2 with a message naming the object and its type (e.g. `v1 is a tag, not a commit — refusing to treat it as one`), never exit 1 and never `proved:false`.
- Remedy 1 (peel in the consumers, making tags supported input) was explicitly ruled out — the live call site (`run-merge-bot.md`) only ever passes shas, and widening the contract to fix a diagnostic was the wrong trade.

## Object-type class covered

`git cat-file -t` on the guarded objects can answer `commit`, `tree`, `blob`, or `tag` (or fail on an absent object). Coverage:
- **commit** (sha, branch, lightweight tag) — accepted, new test added for the lightweight-tag control specifically, per the ticket's explicit call-out.
- **tag** (annotated tag) — newly refused, new test added, asserting exit code `2` exactly (not a lazy non-zero check).
- **tree** / **blob** / absent — already covered by the pre-existing, unmodified `cat-file -e "${obj}^{commit}"` line (existing tests for a tree object and an absent sha both still pass unchanged); not duplicated for blob since it is the identical unmodified code path.

## Test plan

- [x] `./agent-test skills/fleet/scripts/prove-merge.test.mjs` — 25/25 pass (23 pre-existing + 2 new).
- [x] Full suite: `node --test skills/fleet/scripts/*.test.mjs` — 1554/1554 pass, exit 0.
- [x] Manual repro against `origin/main` before the fix, and against the fixed script after, in a scratch fixture (sha control, annotated-tag disproof, lightweight-tag control) — matches the ticket's measured symptom exactly, and matches the new exit-2 behavior after the fix.

Sizing: light

Closes #585

## Review follow-ups (commit `7c97f91`)

Three changes landed after the review, all inside the two files above:

- **The `post` position was not exercised.** `prove(w, "v1", "v1", merge)` dies on the guard loop's FIRST element, so it only ever tested `pre`; `prove(w, head, head, "vm")` tested `merge`. A guard skipping the type check for position 2 *alone* left the suite green at 25/25 — and that mutant is the live #585 bug (`claimed head` = the tag object's sha, every leg healthy, `proved=false`, exit 1). `post` is the position that matters most: `$post_full` is what `[ "$second" = "$post_full" ]` compares. Added `prove(w, head, "v1", merge)`; the mutant now reds exactly that assertion on `1 !== 2`, and a change it should not catch stays green.
- **`prove-merge.sh:115`'s `|| die` carried a verbatim copy of line 104's message**, so a reader could not tell which guard fired. Given its own words. Kept rather than deleted — four reviewers called it dead; under `set -eu` a bare failing assignment aborts with git's own 128, outside this script's 0/1/2 vocabulary (measured: rc=128, no diagnostic line).
- **Two comment fixes.** The `push -q origin main` in the ATTACK test is load-bearing, not scenery — without the merge on origin, the pre-fix script dies at `not reachable from origin/main` and 4 of 5 assertions pass on the bug they exist to catch; said so. And the stale `(22/22)` in the `jstr` comment is now the unqualified property, since a count rots by construction and that one had rotted twice.

Suite re-measured after the follow-ups: `node --test skills/fleet/scripts/*.test.mjs` → **1554/1554 pass, fail 0**.

The rest of the review's suggestion band was checked by refuters and recorded, closed, at #1137. One confirmed defect was out of this PR's scope and is filed open at #1136.


feigi:
Review deferrals filed: **#1136** (open, `ready-for-agent` — `run-merge-bot-prose.test.mjs:361` cites `prove-merge.sh:107` for a check now at `:154`; out of this PR's scope) and **#1137** (closed `wontfix` — the suggestion band, checked: four findings refuted by measurement, two residuals applied here, two entries below the claim bar). The `net.test.mjs` / `unattended-git-sweep.test.mjs` failures in the review snapshot are the known environment artifacts at #1056, not this branch — the suite is 1554/1554 here.

PR #1128, opened 2026-09-01
fix(fleet): state arg.mjs's refusal rules once and pin the copies against drift

### LINKED ISSUE #567: fleet scripts: ledger.mjs now hand-copies THREE of arg.mjs's refusal rules, and nothing fails when the copies drift
Deferred from PR #563 review (issue #362), which has since **merged** (`b2e7f90`); every line number below is against merged `main`.

Findings: `unverified6` (suggestion · **simplify** · `ledger.mjs:57` as reviewed), `unverified4` (suggestion · **types** · `arg.mjs:9`), `unverified5` (suggestion · **simplify** · `ledger.mjs:57`), and the second half of `refuted0`'s suggested fix (important · **silent-failure**). Dimensions are the verbatim ones; an earlier revision of this ticket named two of them from content because the finding text had not yet reached the fix-applier.

Two findings on the same subject, filed as one ticket because neither is actionable without the other.

## The duplication (`unverified6`, simplify)

`arg.mjs:76` holds the shared refusal rule:

```js
if (value === undefined || value.trim() === "" || value.startsWith("--")) die(`--${name} needs a value`);
```

`ledger.mjs` does not route through `arg()` — it splices `--file`/`--require-file` out of `argv` itself — so it carries its own copy at `ledger.mjs:73`:

```js
if (fileIdx !== -1 && file && (file.startsWith("--") || file.trim() === "")) die("--file needs a path");
```

PR #563's own `arg.mjs` header records the cost, in prose: *"the next change to the rule below is known to need a second edit there."*

**The finding got stronger during the review, not weaker.** PR #563 landed two more hand-copied refusals in the same file — `ledger.mjs:55-56` now restate `arg()`'s and `has()`'s `--flag=value` guards:

```js
if (argv.some((a) => a.startsWith("--file="))) die("--file needs a space-separated value, not --file=");
if (argv.some((a) => a.startsWith("--require-file="))) die("--require-file is a boolean flag, not --require-file=");
```

against `arg.mjs:72` (`--${name} needs a space-separated value, not --${name}=`) and `arg.mjs:87` (`--${name} is a boolean flag, not --${name}=`). So the count is now **three** copied rules, not one — and the two new ones were only written because the shared rule was out of reach.

Proposed shape from the review: export a predicate such as `isFlagLike(value)` from `arg.mjs`, so the *rule* lives once while each site keeps its own `die` wording — which is the constraint that made copying look necessary in the first place (`ledger.mjs`'s refusals are deliberately worded for `--file`, not generated from a flag name).

## The missing pin (`unverified4`, tests)

The cross-file sync requirement is asserted **only in prose** — `arg.mjs`'s header comment and `ledger.mjs`'s. No test fails when `arg.mjs`'s rule and `ledger.mjs`'s copy drift apart. Measured on the post-#563 tree: the copies can be edited independently and the suite stays at `tests 794 / pass 794 / fail 0`, because each file's tests pin only its own wording.

That is the same class as the defect #563 itself fixed — a requirement documented where nothing executes it.

1. **Out of #362's scope.** #362 is "`--file` swallows the next flag as its path". Exporting a new symbol from `arg.mjs` and rewiring `ledger.mjs` onto it is an interface change to the shared module, decided in a ticket that has no business editing it.
2. **It collides with #467.** #467 proposes collapsing `makeDie`/`makeArg`/`makeHas` into one `makeCli(NAME)`. Whether `arg.mjs` should also grow a fourth export, or whether `isFlagLike` should be a member of whatever #467 lands, is one design decision and not two — which is why this is `needs-triage` and not `ready-for-agent`. **Triage should rule these together.**
3. **Whoever takes it must not weaken `ledger.mjs`'s refusals.** They are what the #563 regression pins assert on: `ledger.test.mjs:716-851`, including the must-accept pin at `:762` (a path beginning with `-` or containing `--` is accepted **and actually read** — exit 1) and the two `=`-spelling pins at `:809`/`:836`, which are pinned in **both** argv positions. A guard widened from `--file=` to `--file` reds 35 of the 39 tests in that file (measured); a guard narrowed to first-position-only reds exactly the two `=` tests (measured).

---

## Also here: `arg.mjs`'s header now understates what `ledger.mjs` copies

Raised as the **second half of `refuted0`'s suggested fix**, which the fix-applier missed on the first pass and which PR #563 then made worse rather than better. `arg.mjs:12-13` still reads:

> #362 fixed that IN ledger.mjs, **with its own copy of the `--`-prefix rule**, because the splice has no equivalent here.

Singular — one copied rule. That was true of the tree the sentence was written against. It is not true of `main`: PR #563 added `ledger.mjs:55-56`, which copy `arg.mjs:72` and `arg.mjs:87` as well. `ledger.mjs` now carries **three** of `arg.mjs`'s refusals, and `refuted0` predicted exactly this reading — that the header *"reads as parity with a helper that refuses four spellings"*.

Fix it in the same edit as the deduplication above, so the header describes whatever is actually left after `isFlagLike` lands rather than being corrected twice.

## Also here: `unverified5` — fold the two `fileIdx !== -1` tests

> `ledger.mjs` tests `fileIdx !== -1` twice on consecutive lines; folding the new guard and the splice into one `if (fileIdx !== -1) { … }` block is byte-for-byte behaviour-identical and drops the repeated condition.

```js
if (fileIdx !== -1) {
  if (file && (file.startsWith("--") || !file.trim())) die("--file needs a path");
  argv.splice(fileIdx, 2);
}
if (!file) die("--file given with no path");   // unchanged, still its own wording
```

**Its stated claims are true and measured** — behaviour-identity confirmed over 22 hand cases plus 2500 randomised differential cases (fresh cwd and fresh ledger per variant, comparing exit code, stdout and stderr), zero divergence, suite green both ways. It is filed here rather than applied for two reasons, neither of which is a refutation:

1. **It rewrites the same two lines `isFlagLike` rewrites** (`ledger.mjs:73-74` in `main`). Doing it separately means doing it twice.
2. **Measured, it is not a simplification of size:** 487 → 489 lines, `+14/−0` → `+17/−1` against the PR base, plus a nesting level, and it splits the `--file` reader's shape away from the flat `if (requireFileIdx !== -1) argv.splice(requireFileIdx, 1);` three lines below. The finding never claimed brevity — so this is a cost to weigh here, not a defect in the finding.

Whoever takes this ticket should decide the shape of that whole region once, with `isFlagLike` and this fold on the table together.

feigi:
`ledger.mjs` hand-copied three of `arg.mjs`'s refusal rules and nothing failed when a copy drifted. This states each rule once, has every hand-rolled argv reader consume it, and adds the pin that reds on drift.

Sizing: heavy

## The ruling this implements

Per the Agent Brief: export `isFlagLike(value)` from `arg.mjs` now; it does not wait on #467. `makeDie`/`makeArg`/`makeHas` exist to bind a script NAME; these predicates take no name, so they are plain exports whichever way #467 lands.

## The class, enumerated

Found with `grep -n 'startsWith("--")\|trim() === ""' skills/fleet/scripts/*.mjs | grep -v test` at `origin/main` (`343799c`):

| Site | Copies | Covered |
|---|---|---|
| `ledger.mjs` `--file=` guard | `arg.mjs`'s `arg()` `=` guard | yes |
| `ledger.mjs` `--require-file=` guard | `arg.mjs`'s `has()` `=` guard | yes |
| `ledger.mjs` `--`-prefix/blank guard | `arg.mjs`'s value refusal | yes |
| `member-outcomes.mjs` `--file` guard | the same value refusal, **with the blank spelling missing** | yes — **not named by the ticket** |
| `fleet-tick.mjs` integer guard | the blank spelling only, inside an integer grammar | **no**, deliberately |
| `member-outcomes.mjs` unknown-flag sweep | nothing — `sweep()`'s class, narrower on purpose | **no** |

Two of the three named copies are a *different predicate shape* from the one the ruling names — over an argv and a flag name, not over a value — so `isFlagLike` alone leaves them in place and AC-1 unmet. Hence a second export, `hasEqualsForm(name, argv)`. This is the one place I went past the brief's literal wording, in service of its own acceptance criterion.

`member-outcomes.mjs` is included because AC-1 says the rule is stated **once**, and a surviving copy contradicts that. Its copy was the rule with the blank clause dropped, and the gap was live. Measured by running `origin/main`'s version of the file against this branch's dependencies:

```
$ node <origin/main member-outcomes.mjs> <session-dir> --file "   "
member-outcomes: scraped 0 of 0 members (0 dropped);     now holds 0 rows
exit=0                          # and a whitespace-named file appeared on disk
$ node skills/fleet/scripts/member-outcomes.mjs <session-dir> --file "   "
member-outcomes: --file needs a path
exit=2
```

That is the **only** intended behaviour change in this PR. Everything else is behaviour-identical. Nothing pinned the accepting behaviour (`grep -n '\-\-file' skills/fleet/scripts/member-outcomes*.test.mjs`).

## AC-2, demonstrated

Re-inline `ledger.mjs`'s guard **behaviour-identically** as `(file.startsWith("--") || file.trim() === "")`, then `./agent-test skills/fleet/scripts`:

```
tests 1555 / pass 1554 / fail 1
✖ every refusal that shares arg.mjs's rules calls a predicate instead of restating one
  AssertionError: ledger.mjs's --file value guard drifted from arg.mjs's rule or lost its own wording
```

One test red, 1554 green — the drift is invisible to every behavioural test in the repo, which is exactly why a source pin is the only thing that can see it. Reverted.

## AC-3, re-measured against this branch

`./agent-test skills/fleet/scripts/ledger.test.mjs` is `tests 81 / pass 81 / fail 0` unmutated. (#567 cites 35 of 39; that was measured before this file grew to the 81 the command reports. Same property, re-measured.)

- widen the `--file=` guard to match `--file` → **pass 6 / fail 75**
- narrow the `--file=` guard to the first argv position → **fail 3**: the pre-existing either-position test plus both new pins
- narrow the `--require-file=` guard the same way → **fail 2**: its either-position test plus the wiring pin

Both `=`-spelling pins still assert in **both** argv positions. All mutations reverted.

## AC-4

`CLI: a path beginning with '-' or containing '--' is accepted and actually read (#362)` — green, exit 1, i.e. read, not merely not-refused.

- `hasEqualsForm(name, argv = process.argv)` defaults to the full `process.argv`, the same array `arg.mjs`'s inline `some()` scanned. `ledger.mjs` passes its own `process.argv.slice(2)`, as its copy did. No scan widens or narrows.
- `ledger.mjs`'s `file &&` term is **kept** ahead of `isFlagLike()`, which answers true for an absent value. Folding it in would move a trailing `--file` off `--file given with no path` onto `--file needs a path` — reddening the pin that says #362 left that wording alone. `shared-refusal.test.mjs` pins that exclusion explicitly so it is on the record rather than a hole.
- `member-outcomes.mjs`'s `!file ||` term is **dropped** as redundant — that is the behaviour change above.
- New exports on a module every fleet script imports: `grep -rn 'isFlagLike\|hasEqualsForm'` was empty at `343799c`, so nothing collided.
- Import order and layout are **not** pinned. This bullet previously said the imports had to stay on one line because `arg.test.mjs`'s #367 wiring pin matches with `[^}]*`, "which cannot span a newline". That was wrong — a negated character class matches newlines in JS regardless of `/m`:

  ```
  $ node -e 'console.log(/import \{[^}]*\} from/.test("import {\n  a,\n  b\n} from \"./x.mjs\";"))'
  true
  ```

  The review caught it, and `07b09a5` now matches each imported name on its own and asserts only that it comes from `arg.mjs`. Measured: reordering `ledger.mjs`'s import redded the suite at `0a44404` and is green here, while dropping `hasEqualsForm` from it still reds.

## The pin can see the mode it guards

The new file's accept half asserts the value was **used as the path** (`arg()` echoes it; `ledger.mjs` quotes it back in its not-found warning), never merely that nothing refused it — a not-refused assertion is satisfied by any unrelated failure, and by a guard that swallowed the value and fell back to a default, which is the fail-open shape this whole family of guards exists to close.

- **#467** (`makeCli(NAME)`) — untouched; `isFlagLike` is unaffected whichever way it lands.
- **#584** — `ledger.mjs`'s argv handling wholesale. Not routed through `arg()`; only the rules are shared. Same file, so worth a rebase check before merging that one.

`./agent-test skills/fleet/scripts` → **tests 1555 / pass 1555 / fail 0** (baseline at `343799c` was 1552; the three added are this PR's). `node --test skills/fleet/scripts/*.test.mjs` → same.

## Applied from the review (`0a44404` → `07b09a5`)

- **`member-outcomes.mjs`'s refusal is now executed, not just matched** (critical). It had only a source-text pin, so a regression *above* the pinned `if (isFlagLike(file))` line — `const file = (raw ?? "").trim() || <default>` — left that line intact and slipped through: measured, that mutant passes 3/3 against the previous test file while `--file "   "` writes the production metrics path at exit 0. `probeFixture` now carries a `viaMemberOutcomes` spawn probe mirroring `viaLedger`, and it reds only the differential test on that mutant. Discriminators confirmed green: changing the default path string, and reflowing `member-outcomes.mjs`'s import.
- **Test 1 repinned to answer behaviour, not layout.** Beyond the import fix above, the two `$`-anchored guard pins dropped their end anchor — `stripComments` blanks whole-line comments only, so a trailing `// note` redded a line whose code had not changed (verified red at `0a44404`, green here). Re-inlining the blank rule still reds.
- **Both settling commands were self-matching.** `grep -ln 'function die('` reported `arg.mjs` — the file claiming it reports none — and `grep -l makeArg` listed `ledger.mjs`, the one script that paragraph says does *not* route through `arg()`. Anchored to `^` both now settle, and the "outside arg.mjs and the tests" hedges are gone because the output is the set.
- **Ceiling citation corrected** in `shared-refusal.test.mjs`: this is `member-prompt-prose.test.mjs`'s presence-pin ceiling, not `strip-comments.mjs`'s trailing-comment one. Measured — closing the latter closes a different escape and leaves this one open. The residual pin brittleness is now disclosed rather than implied.
- **One comment consolidation:** `arg.mjs` stated the #567 rationale in its header and again above the predicates; the second is now a three-line pointer. The per-file statements elsewhere stay.

Everything else in the suggestion band was checked by a refuter and did not survive — recorded in #1135 (`wontfix`, closed), with the measurements and a re-open trigger per entry.

`node --test skills/fleet/scripts/*.test.mjs` at `07b09a5` → **tests 1555 / pass 1555 / fail 0**, unchanged from `0a44404`.

Closes #567


feigi:
Review deferrals filed: **#1135** — `PR #1128 review: the suggestion band, checked` (`wontfix`, closed).

Six entries, all checked by a refuter that ran something, none surviving: `isFlagLike(null)` (silent-failure), the "BOTH readers" header claim (silent-failure), the redundant accept-loop spawns (simplify), the `probeFixture` `arg.mjs` copy (simplify), the `isFlagLike` shape objection (types, refuted 2-0 in review), and the comment-volume judgement (simplify, partially applied). Each carries its measurement and a re-open trigger.

`ledger.mjs check` returned exit 3 / `tracker-hit`, but every hit was another PR's record of the same title shape (#1134, #1124, #1123, #1120, #1115) — read and rejected as duplicates, so #1128's own record was filed.

What survived is in `07b09a5` and described in the PR body.

PR #1139, opened 2026-09-01
fix(review-pr): split snapshotMissing's no-tree branch by cause

### LINKED ISSUE #539: review-pr.js:622 — a snapshot agent that DIED reports as 'returned no tree', naming a missing tree when only the schema handshake failed

feigi:
## Summary

`snapshotMissing`'s first branch (`workflows/review-pr.js`, inside `function snapshotMissing(snap)`) collapsed three distinct causes into one message that named a missing tree for all of them:

```js
if (!snap || !snap.path || !snap.head) return "the snapshot agent returned no tree — nothing to review";
```

A snapshot agent that died, or whose harness exhausted its structured-output retries (see the `required:` comment inside `FINDINGS_SCHEMA` — a schema *rejection* is retried and every observed one recovered; only exhaustion emits `Failed to provide valid structured output after <n> attempts` and makes `agent()` return null), reported the same way as a report that came back with a tree path but no head. Only the first is actually "no tree"; the other two named a wrong cause.

Split into three branches, in the same left-to-right order the original `||` chain evaluated (`!snap`, then `!snap.path`, then `!snap.head`), so a report giving neither field still reads as giving no `path`, unchanged from before:

```js
if (!snap) return "the snapshot agent returned nothing (it died, or it exhausted its structured-output retries) — no tree to review";
if (!snap.path) return "the snapshot agent's report gives no `path` — no tree to review";
if (!snap.head) return "the snapshot agent's report gives no `head` — refusing to review a tree whose commit is unknown";
```

This is the full three-way split both PR #536's and PR #852's reviews converged on — not just peeling off the falsy-`snap` case. The `pathVerified` and `prHead` branches below it are untouched: same control flow, same refusal point, only this one message's wording splits.

### Each branch says only what it knows

The review of this PR's first commit found it had reintroduced, on one branch, the defect #539 exists to remove. The `!snap.head` message still ended `— no tree to review`, on the one branch where `path` is present and `pathVerified` may be true. Measured against a directory that really exists: the tree was there, `pathVerified` was true, and the message said there was no tree. For `!snap` and `!snap.path` that ending is defensible — the path is unknown, so nothing is reachable — but for `!snap.head` it is a false claim about the filesystem, in the exact register the ticket complained about. That branch now names what is actually gone: the sha to check the tree against.

Two further wording corrections from the same review:

- The falsy-`snap` message named *"its structured output was rejected"* as a cause. A rejection is precisely the event that does **not** produce a falsy `snap` — this same file records that rejections are retried and every observed one recovered, and only exhaustion returns null. The commit's own comments and message already said "exhausted"; only the operator-facing string carried the wrong cause.
- `is missing \`path\`` / `` `head` `` overstated a falsy check: the schema types both as `string`, so `""` validates and reaches the branch. Both now read `gives no`, accurate for absent and empty alike.

## Class enumerated (per the Agent Brief)

The branch covers exactly three inputs, matching the issue's own framing (a later comment corrects a claim of "four causes" — the agent-died case is a *cause* of `!snap`, not a fourth field):
- [x] `!snap` (agent died, or exhausted structured-output retries) — names the agent contract, not the tree.
- [x] `!snap.path` — names `path`, and still ends "no tree to review", which is true when the path is unknown.
- [x] `!snap.head` — names `head`, and does **not** claim there is no tree.

Deliberately left alone (out of scope per the Agent Brief): the `pathVerified` branch (#140) and the `prHead` mismatch branch (#532) — neither's wording nor control flow changes here.

Also left alone deliberately: the `!snap.path` and `!snap.head` guards are unreachable from the schema-validated production path only if you read `!x` as a presence test. It is a falsy test, and a relayed `head` of `"  \n"` validates against `{type:"string"}`, satisfies `required`, then normalizes to `""` at the in-place trim just above the call and hits the guard. Both guards are defensive, not dead, and both are killed by deleting them (measured); they stay.

## What the restructuring could newly violate, and what's tested for it

Splitting one `||` chain into three early returns changes nothing about *when* the function refuses or what it returns for the accept path — verified by the existing accept-path tests (a fully verified snapshot, an absent `prHead`, abbreviated-head matching) all still passing unchanged. The risk specific to this restructuring is the two-fields-missing case reading as the wrong field; `skills/fleet/scripts/review-pr-snapshot-path.test.mjs` now asserts `{}` and `{ head: "abc123" }` both name `path` (not `head`) for exactly that reason.

## Tests

`review-pr-snapshot-path.test.mjs`'s six-input loop asserted only `/no tree/` across all of them — the exact shape that let three different causes read alike. Replaced with three tests, one per cause, each asserting the message names its own field and explicitly *not* the other (`assert.doesNotMatch`) — proving the branch can actually distinguish what it now claims, not just that some string comes back. The head test additionally asserts `doesNotMatch(/no tree/)`, so the defect described above cannot come back silently.

The two `doesNotMatch` failure messages were narrower than the assertions they explained: the regex forbids *any* backticked mention of the other field, while the message said it forbids *claiming the other field is missing*. A strictly better production message would have red the assertion while satisfying its own failure text. The messages now state what the regex actually enforces, and why. The pattern was deliberately **not** narrowed to `/\`head\` — no tree/`: measured, that variant goes blind to the copy-paste defect it exists to catch the moment the head branch's trailing clause changes — which is what this PR just did.

`review-pr-testcmd.test.mjs`'s ordering check is re-anchored on the guard's **invocation**, `const missingReason = snapshotMissing(snap);`, not on its declaration and not on a return message. Four review dimensions independently found the declaration anchor near-tautological: it sits above every top-level statement in the file, so `callAt > guardAt` held even with `resolveTestCmd` moved ahead of the guard — the one regression the assertion's own message names. Measured both directions at this head: that mutation now reds *this* assertion (it stayed green before), and a pure reword of all three messages with call order unchanged leaves it green while only the three wording pins red. The comment justifying the earlier re-anchor claimed a reword could "silently point this at the wrong line"; it could not — a retired literal drives `indexOf` to `-1` and `guardAt !== -1` reds loudly. It now states what is true.

Every new and changed assertion was mutation-tested against this branch, each mutant killing exactly one test:

| mutation | result |
| --- | --- |
| `resolveTestCmd` moved above the guard | ordering test **red** (was green pre-fix) |
| all three messages reworded, call order unchanged | ordering test **green**; only the 3 wording pins red |
| `!snap` message reverted to "structured output was rejected" | falsy-snapshot test **red** |
| `!snap.head` ending reverted to "— no tree to review" | missing-head test **red** |
| `!snap.path` returning the head branch's message | missing-path test **red** |
| control: head clause reworded, still not claiming no tree | **green** |

Loop variable in the missing-path test renamed `dead` → `report`: `{}` and `{ head: "abc123" }` are live reports missing a field, which is what the test name, its comment and the asserted message all call them.

The citation chain's second hop pointed at "the `agent()` comment near this file's top", which does not carry the citation — the retry/exhaustion record lives in `FINDINGS_SCHEMA`'s `required:` comment. Both hops now name that block.

Full suite green at this PR's head (`node --test skills/fleet/scripts/*.test.mjs`, run at commit 9d12d0b): `tests 1571 / pass 1571 / fail 0`.

Sizing: light

Closes #539


feigi:
Review adjudication (fix-applier): **8 findings applied** in `9d12d0b`, **3 recorded** in #1142 (closed `wontfix` record — one refuted by measurement, one below the claim bar, one no-action note).

Applied: the `!snap.head` branch no longer claims there is no tree (the headline defect — #539's own register, measured against a real directory); the ordering assertion re-anchored on the guard's invocation, with its comment rewritten to a true claim (five findings, one edit); the `!snap` message's cause corrected from "rejected" to "exhausted", with its pin landed in the same commit; `is missing` → `gives no` on both field branches; the citation chain's second hop repointed at `FINDINGS_SCHEMA`'s `required:` comment; two `doesNotMatch` failure messages widened to state what their regex enforces; `dead` → `report` in the missing-path loop.

Recorded in #1142: the "two branches are unreachable" finding — **refuted**, they are defensive not dead (deleting either reds a test; a relayed `head` of `"  \n"` validates, normalizes to `""` at `review-pr.js:788-790` and reaches the guard on the production path); and the type-level verified/unverified distinction, out of diff scope and below the claim bar.

Suite: `tests 1571 / pass 1571 / fail 0`. Six mutants run, each killing exactly one test, plus a green control — table in the PR body.

PR #1140, opened 2026-09-01
fix(run-team): say a key absent from dimensionsUnrun ran a suite, not that it is covered

### LINKED ISSUE #535: run-team SKILL.md: "not in dimensionsUnrun is covered" overstates — a narrower test command still reads as covered
Deferred from PR #526 review (`unverified2`), and recorded here because **the controller ruling that dismissed it was partly wrong**. Filed rather than fixed in-PR: see the disposition note at the end.

## The claim

PR #526 introduces this sentence in `skills/fleet/skills/run-team/SKILL.md`:

> A key in `dimensionsRun` and NOT in `dimensionsUnrun` **is covered**

`unrunReason` reads only the **count** (`test_run.tests`), never the **command**. So a specialist that substituted a narrower runner — one file instead of the suite — reports `tests > 0`, is not classified unrun, and therefore "is covered" by that rule, having validated a fraction of what the reader assumes.

The finding's remedy is one word: `is covered` → `ran a suite`.

## Why this is filed rather than applied

**The controller ruled `unverified2` "do not apply — it restates `refuted1`, which was refuted 2-0". That conflated two different remedies.**

- `refuted1` asked for a **code** change: have `unrunReason` compare `test_run.command` against the dispatched `testCmd`. Two refuters killed it 2-0.
- `unverified2` asks for a **doc** change: soften a sentence this PR introduces.

Killing the code fix does not settle the doc wording. The doc-only remedy was never adversarially checked; it was dismissed on the strength of a refutation aimed at something else.

## The honest weighing, for whoever triages this

**Against fixing it:** "covered" in context means *a specialist ran and reported*, which remains true of a narrower run — so the sentence is **imprecise rather than false**. And this PR strictly improves the situation: `test_run.command` is a new required field, so the command a specialist actually ran is now recorded and readable, where before it was not captured at all.

**For fixing it:** this repo tracks minted false claims, and a correction-class PR introducing the strongest available reading of its own guarantee is exactly the pattern that discipline exists to catch. `ran a suite` is one word, strictly more accurate, and loses nothing.

## Related, and deliberately not merged into them

- **#143** — the `tests 0` floor being too narrow. Its PR #526 comment covers the `fail > 0` conjunction, **not** this. `unverified2` states explicitly that it wants no code change under #143's deferral, only the doc claim softened.
- **#140** — `snap.path` handed out unchecked. Different surface.

## Provenance

Surfaced by `finish-pr-526` while reconciling an off-by-one in the controller's adjudication comment (20 findings, 19 labels). It correctly declined to file it itself, on the grounds that filing against a standing "do not file" ruling on a coin-flip identification would contradict the controller — and escalated instead. The identification was then resolved from the review journal: the missing label was `unverified1` (the reject case, closed via #527), and `unverified2` is this.

Not a merge blocker for #526, which is labelled `ready-to-merge` on a `check`-green gate.

feigi:
`unrunReason` classifies from `test_run`'s counts and quotes `command` into its message; it never compares that command against the one the dispatch handed out. So a specialist that substituted a narrower runner — one file instead of the suite — reports a non-zero count, is not classified unrun, and read as "covered" under the old wording while having validated a fraction of the suite. The sentence now claims what the classifier establishes.

Sizing: light

## The change

One word in `skills/fleet/skills/run-team/SKILL.md`, plus the pin it had been missing.

Settled at the shipping commit `b886935`:

```
$ git show b886935:skills/fleet/skills/run-team/SKILL.md | grep -n 'NOT in `dimensionsUnrun`'
1085:in `dimensionsRun` and NOT in `dimensionsUnrun` ran a suite; every
```

## Doc half only — the code remedy stays refuted

Comparing the reported command against the dispatched one was refuted 2-0 and stays refuted. `review-pr.js` is untouched:

```
$ git show --stat b886935 -- workflows/review-pr.js
(no file row)
```

## The pin, and the mutation check both ways

The word had no pin — a revert would have shipped green. Neither existing pin over this paragraph covers it: `review-pr-unrun.test.mjs`'s Reviewers-section test asserts `dimensionsUnrun` is documented and that the pre-#137 workaround is gone, and `review-path-default.test.mjs` compares the `It returns …` field list against `review-pr.js`'s own `return`. None reads the verb.

`review-pr-unrun.test.mjs` now carries one, in two halves that divide the work:

- **The positive pin carries the claim.** Sliced to the paragraph with `between` so a failure prints the paragraph rather than the whole 2000-line doc, matched through `phrase()` so a reflow of the hard wrap cannot fire it, and run out to the sentence's `;` so the clause has to END there. That terminator is what catches the likelier regression — an editor softening rather than reverting, `ran a suite AND IS THEREFORE COVERED` — which a pin on the anchor alone reads as still present.
- **The exclusion then has nothing left to guess at**, so it stays whole-doc and case-insensitive (`/in\s+\`dimensionsUnrun\`\s+is\s+covered/i`): the retracted claim reds wherever in the file it comes back, including the lowercase paraphrase a `phrase` pin on `NOT` walks straight past.

The first half of that shape shipped in `b878c70`; review measured that it was green against the adjacent-clause softening, matched the whole doc on failure, and was case-sensitive on `NOT`. `b886935` is the fix, and it widens the exclusion's scope rather than narrowing it.

An intermediate formulation — a sentence-scoped `assert.doesNotMatch(para, /dimensionsUnrun\`?[^.]*covered/i)` — was measured and **rejected**: `[^.]*` spans newlines and does not care about polarity, so adding the semantically *correct* sentence ``No `dimensionsUnrun` entry, and no silence in its place, ever makes a dimension covered.`` turned the suite red. It also lost ground, missing an uppercase `NOT … is covered` planted outside the slice that the original pin caught.

Mutation matrix at `b886935`, run in the worktree, `node --test skills/fleet/scripts/review-pr-unrun.test.mjs` (20 tests), restoring from a `cp` copy between rows:

| Run | result |
|---|---|
| baseline, unmutated | 20 pass / 0 fail |
| kill — the word reverted to `is covered` | 19 / **1** |
| kill — softened in place, `ran a suite and is therefore covered` | 19 / **1** |
| kill — lowercase `not … is covered` restated *outside* the pinned paragraph | 19 / **1** |
| control — the hard wrap reflowed across the pinned phrase, the word intact | 20 / 0 |
| control — `; every` → `; each`, a benign reword right after the pinned terminator | 20 / 0 |

The controls are what prove the pin discriminates rather than reddening on any edit. Under the softening mutation the **full** suite reds exactly one test — the new pin, and nothing else — and its `actual:` value is the paragraph, not the document.

## The class, enumerated

Other places that state the same guarantee, checked at `b886935`:

- `skills/fleet/commands/review-and-fix.md` — names neither field (`grep -c 'dimensionsRun\|dimensionsUnrun'` → `0`), and its only unrun language is about a killed specialist whose report you do not hold. Not made to disagree; no change needed.
- `docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md` — names both fields and already says `dimensionsRun` "is not a coverage claim on its own", never asserting the inverse. Correct as written, and a dated design record. **Deliberately left.** (Review claimed the spec keeps the coverage framing; that finding is wrong — the spec pairs the "did not cover their ground" phrasing with the denial. Recorded in #1144.)
- `docs/metrics/tier-outcomes.tsv` — historical run records, not a statement of the guarantee. **Deliberately left.**
- `workflows/review-pr.js` — the code, explicitly out of scope. **Untouched.** Its own comment does still assert the refuted equivalence ("A key in the first and not the second is the only thing that means covered"), confirmed 2-0 by review. Deferred as **#1143**: pre-existing, outside #535's scope, and colliding with live sibling PR #1139, which is editing that file now.

## Review deferrals

- **#1143** (`ready-for-agent`) — the `review-pr.js` comment above.
- **#1144** (`wontfix`, closed) — PR #1140's suggestion band, checked: three findings applied here, three refuted or not-a-defect, with re-open triggers.

## Tests

`node --test skills/fleet/scripts/*.test.mjs` from the worktree at `b886935`: `tests 1570` / `pass 1570` / `fail 0`, exit 0.

Closes #535


feigi:
Review deferrals filed:

- **#1143** — `ready-for-agent`, dimension `correctness`. `workflows/review-pr.js` still carries the refuted equivalence in its own comment ("A key in the first and not the second is the only thing that means covered"), the exact claim this PR corrected in `run-team/SKILL.md:1085`. Confirmed 2-0 by refuters; nothing in the suite pins it. Deferred as pre-existing, outside #535's scope, and colliding with live sibling PR #1139 which is editing that file now. The issue says a fix must re-locate the line by construct, since #1139 may move it.
- **#1144** — `wontfix`, closed. PR #1140 review: the suggestion band, checked. Three of the six `suggestion` findings survived a refuter and landed here in `b886935` (adjacent-clause softening not caught, whole-doc match on failure, case-sensitive `NOT`); the other three are refuted or not-a-defect, with per-entry re-open triggers.

Not filed: the `unattended-git-sweep` failure the review saw is the `git archive` snapshot's missing `.git`, already tracked as #1056 — the suite is `1570 / 1570 / 0` in the worktree.

Issue #1129, opened 2026-09-01
review-pr: every snapshot lands on one fixed path, so a second review silently replaces the tree a live fix-applier is still citing


feigi:
Found by the `/fleet:run-team` controller mid-run, not by a review specialist.

## The defect

The `review-pr` workflow writes its immutable snapshot to a path derived from the caller's `scratch` argument and **nothing else** — no PR number, no head SHA, no run id. Both reviews this run reported the same destination:

```
run wf_a0946fd3-12f (PR #1126)  "snapshot": "<scratch>/snapshot"
run wf_4d655265-33c (PR #1128)  "snapshot": "<scratch>/snapshot"
```

The second review therefore overwrote the first review's snapshot. That is a collision between *sequential* runs sharing one scratch root, so it needs no concurrency to fire.

## Why it is not merely untidy

A fix-applier outlives the review that produced its findings — the controller dispatches it after the workflow returns, and it then runs for as long as it takes to refute and apply. Findings carry absolute `file` fields pointing into the snapshot; measured on the PR #1126 run, `survived[1]` was reported at

```
<scratch>/snapshot/workflows/review-pr.js:762
```

and the `unverified` cluster on `review-pr.js:753` likewise. While `fix-pr-1126` was live, that path came to hold PR #1128's tree — a different branch, a different diff, a different `review-pr.js`.

**The failure is silent in the worst direction.** The path still exists, still contains a plausible repository, and still has a `workflows/review-pr.js` with a line 753. A fix-applier that opens it to check a finding gets a confident wrong answer rather than an error, and nothing in the payload says which tree it is looking at.

The controller warned `fix-pr-1126` off the path by hand once it noticed. Nothing in the fleet would have caught it otherwise.

## Not the same as the known snapshot ticket

Distinct from the already-filed issue that the snapshot is a `git archive` extraction rather than a git repository, so suites relying on `git rev-parse --show-toplevel` report a phantom failure. That one is about what the snapshot *lacks*; this one is about **which PR's tree is in it**. A fix for either leaves the other standing.

## Directions, unmeasured

- Include the PR number, head SHA or run id in the path, so two runs cannot collide and a stale reference fails loudly instead of resolving to a stranger's tree.
- Return the snapshot path with the findings as something a consumer is told is ephemeral, and make findings carry repo-relative paths plus the head SHA rather than absolute snapshot paths — a consumer can then use `git show <sha>:<path>` and cannot be fooled.
[...]

PR #1147, opened 2026-09-01
fix(run-team): give implementers their own scratch dir, not the shared root

### LINKED ISSUE #581: run-team/SKILL.md: the per-member scratch rule is stated only for reviewers, so concurrent implementers share one mutation-harness directory
Observed during `/fleet:run-team` on 2026-08-18, reported by `impl-574`. Not a defect in any script — a gap in what `run-team/SKILL.md` tells the controller to put in an **implementer** dispatch prompt.

## What happened

Two implementers ran concurrently and both used the session scratchpad **root** as their mutation-harness directory. `impl-574` reported:

> the session scratchpad root is SHARED between fleet members. It already held `prove-merge.sh`, `prove-merge.sh.orig`, `m1–m4.sh` and `n1.sh` from `impl-566` when I arrived. My first copy wrote `verify-sha.sh`/`.test.mjs` (+`.orig`) into that root before I noticed; those are byte-identical to `origin/main`, and I moved all my work into a `574/` subdir and touched nothing of theirs afterwards.

No damage this run — the two members were mutating different filenames, and `impl-574` noticed and self-isolated. The near-miss is the point: a mutation harness works by writing a **broken** copy of a file, measuring, then restoring from `.orig`. Two members doing that in one directory is one filename collision away from a member restoring a sibling's `.orig` over its own file, or measuring a "clean baseline" that is actually a sibling's mutant.

That failure would be silent and would read as a real measurement.

## Why the existing rule did not prevent it

`run-team/SKILL.md` **does** carry the two-level scratch rule — `<scratch>/pr<N>/<finding>/`, with the measured rationale that finding ids restart at 1 every review so two fix-appliers both reach for `unv1`. But it lives in the **reviewer/fix-applier** section and is phrased for refuters and fix-appliers.

The **implementer** dispatch block in phase 2 says nothing about scratch directories at all. Implementers are explicitly instructed to mutate (*"a test you ADD must kill its own mutant"*, *"mutate a copy under `<scratch>/pr<N>/mutate/`"* appears only in the fix-applier prompt), so they reach for a scratch dir with no instruction on where to put it — and the natural default is the session root, which every member shares.

So the rule exists, is correct, and is scoped to the wrong half of the fleet.

## Suggested change

In phase 2's verbatim implementer block, name a per-member scratch path the same way phase 1 already derives per-ticket ports from the issue number — e.g. `<scratch>/impl-<N>/` — and state that the session scratchpad root is shared and must never be written to directly.

Two properties worth keeping from the existing rule: derive the path from something already unique per member (the issue number, exactly as `claim-ticket.sh` derives `postgres=16<N>`/`ollama=22<N>`), and say *why*, since "use a subdir" without the collision rationale reads as tidiness and gets skipped under time pressure.

## Not the same as the worktree isolation rule

Filesystem isolation of the **worktree** is already solved — `claim-ticket.sh` gives each ticket its own tree and its own `agent-test`. This is about the **scratch** directory, which is deliberately outside the worktree precisely so mutation harnesses do not dirty it, and which nothing partitions today.

feigi:
## Summary

Phase 2's verbatim implementer dispatch block in `skills/fleet/skills/run-team/SKILL.md` named no scratch directory at all — every existing scratch-path rule in that file lives in the reviewer/fix-applier half. Settled directly: grepping the implementer dispatch region (`and each of these verbatim:` through `Each rule in the enumerate-and-declare block`) at `origin/main` (`efd1989`) for "scratch" turns up no mentions in that slice.

Meanwhile every dispatched fleet member gets the same session-scoped scratchpad root injected into its own system prompt, unprompted, shared with every sibling. An implementer reaching for a mutation-harness directory (write a broken copy, measure, restore from `.orig`) defaults to that shared root before `SKILL.md` ever says otherwise. `impl-574` hit this on 2026-08-18 and only avoided a collision by noticing and self-isolating after already writing into the root.

## Change

Adds one block to the implementer dispatch prompt, inserted between the existing "commit incrementally" and "enumerate the class" blocks:

- Names the injected scratchpad root explicitly and says it is shared with every sibling member.
- Derives a per-member path the same way `claim-ticket.sh` already derives per-ticket ports from the issue number (`postgres=16<N>` / `ollama=22<N>`): `<scratch>/impl-<N>/` — no second scheme invented.
- Tells the member to `mkdir -p` it, since nothing else creates it.
- States the harm as a false measurement, not untidiness: two members sharing a directory are one filename collision away from restoring a sibling's `.orig` over their own file, or measuring a sibling's mutant as a clean baseline — silently.
- Explicitly distinguishes this from worktree isolation, which `claim-ticket.sh` already owns; the scratch root is deliberately outside every worktree.

The reviewer/fix-applier scratch rule (`<scratch>/pr<N>/<finding>/`) is untouched — it was already correct, only its scope (reviewer-only) was the gap.

`skills/fleet/skills/run-team/SKILL.md` is unchanged since `25fa8aa`; everything below landed in `dispatch-block-pins-prose.test.mjs`.

## Review follow-up (`cf64dde`)

Review of `25fa8aa` found the new test pinned every explanatory sentence in the block and **not the imperative itself**. Two refuters independently rewrote the mandate from *"goes under `<scratch>/impl-<N>/`, never into the scratch root by itself"* to *"may go under … or straight into the scratch root — either is fine"*, deleted the *"so writing to it directly … is the defect"* clause, and the file stayed **7/7 green** — one widened to every `*prose*.test.mjs` in the tree, **298 tests, still green**, so no sibling pinned it either. That is the failure mode this file's own header rules out at rule 2 (*a rule is pinned together with the command that carries it out*), and not the appended-exception ceiling it acknowledges.

Three changes, test file only:

- **Pin the imperative to its exclusion**, path and prohibition in one span, so an edit that keeps `<scratch>/impl-<N>/` and drops the "never" cannot pass.
- **Split the harm assertion's `/…{0,400}?…/` distance-capped regex into two contiguous `phrase()` pins.** Today's wording already spent 303 of the 400-character window, so one added clause would have redded the pin *with a message claiming the mechanism was removed* — a false red that misdirects whoever hits it. Widening the cap only moves the cliff; a contiguous span has no budget to approach. Both halves still cover every mutation the tests specialist measured against the old regex, including the *"indistinguishable from a real result"* tail.
- **Restate the rewrap test's coverage note as a property** — "every test in this file" — replacing the counts *"six"* and *"the five of them"*, which `25fa8aa` rotted by adding a seventh test one block above them. A count is false the moment the next test lands.

## Same defect class — what this covers and what it leaves

Per #581's brief, this is one of three measured instances of "shared harness directory → false measurement, not a crash":

- **Shared scratch root** (this PR) — covered.
- **Shared review-pr snapshot directory** — filed separately as #1129, out of scope here.
- **Shared process table** (`pkill -f` killing a sibling's run) — separate ticket, out of scope here.

## Deferred from review, filed not applied

- **#1150** (`ready-for-agent`) — `references/isolation.md:17` declares *"Per-member is the wrong axis"* without qualification, which now contradicts this PR's per-member rule, and `SKILL.md:1991`'s members-wide guard bullet names only `<scratch>/pr<N>/<finding>/`. A refuter falsified the reading that would have made this a docs-only tidy-up: `SKILL.md:1896-1897` states *"members fan out and their children consume slots you never dispatched"*, and the heavy `sizing-a-ticket` row routes implementers through `superpowers:subagent-driven-development`, which dispatches subagents. So implementers **do** have grandchildren sharing the injected root, and `<scratch>/impl-<N>/` is incomplete rather than wrong. Which axis governs which agents is a design question, not a typo — deferred rather than decided here, and neither rule reversed.
- **#1151** (`wontfix`, closed) — the checked suggestion band: one finding refuted by measurement (the harm example is #581's own incident prose compressed, and 7 of the 8 blocks in that region argue from mechanism alone), plus the `unattended-git-sweep.test.mjs` snapshot artifact (#1056, environmental) and one positive finding confirming the five original assertions all discriminate.

## Verification

At `cf64dde`, in the worktree (a real checkout, so `.git` exists):

- `env -u FORCE_COLOR node --test skills/fleet/scripts/*.test.mjs` → **tests 1573, pass 1573, fail 0**.
- **Kill control:** applied the refuters' exact mutant — mandate inverted, defect clause deleted — → **tests 7, pass 6, fail 1**, failing on the new assertion's own message (`the per-member path is no longer bound to the prohibition on writing into the scratch root itself`). Restored with `cp`; tree clean; back to 7/7.
- **Accept control:** reflowed the whole block at width 58, **18 lines → 25**, every wrap point moved, words byte-identical → **7/7 green**. The pins refuse drift, not reflow. Restored with `cp`; tree clean.

Sizing: light

Closes #581


feigi:
**Review deferrals filed** — #1150 (`ready-for-agent`, open) · #1151 (`wontfix`, closed record).

| Finding | Band | Dimension | Outcome |
|---|---|---|---|
| The new test pins every explanatory sentence but not the block's imperative | `survived`, important | correctness | **applied** — `cf64dde` |
| The file's coverage note still says "six tests" / "the five of them"; this PR made it seven | suggestion | correctness | **applied** — restated as a property, not a count |
| The harm pin's `.{0,400}?` window has ~97 chars of headroom (303 already spent) | suggestion | correctness | **applied** — split into two contiguous `phrase()` pins |
| Same claim, from the tests dimension | suggestion | tests | **applied** — same edit as above |
| `isolation.md` calls per-member the wrong axis; guard bullet not updated | suggestion | correctness | **deferred → #1150** — refuter showed implementers *do* fan out to grandchildren, so this is a live design question, not a docs tidy-up |
| The harm example is illustrative rather than a citation | suggestion | comments | **refuted → #1151** — it is #581's own incident prose compressed; 7 of 8 blocks in that region argue from mechanism alone; #1148 would be a mis-citation |
| The lone full-suite failure is `unattended-git-sweep.test.mjs` | suggestion | tests | **no action → #1056** — no `.git` in the archive snapshot; 1573/1573 green in a real checkout |
| Same failure, from the comments dimension | suggestion | comments | **no action → #1056** |
| All five original assertions genuinely discriminate | suggestion | tests | **no action** — positive finding, recorded in #1151 |

Every in-scope suggestion got one refuter before anything was touched; the band is budgeted 0 by the workflow. Verification at `cf64dde` — full suite 1573/1573, plus a kill control (the refuters' exact mutant reds the new assertion, 7/6/1) and an accept control (block reflowed 18→25 lines, same words, 7/7) — is in the PR body.

PR #1153, opened 2026-09-01
fix(fleet): refuse a stray flag folded into ledger.mjs's check/filed subject

### LINKED ISSUE #584: fleet scripts: ledger.mjs absorbs an unknown --flag into the check subject, flipping the duplicate-filing guard from exit 1 to exit 0

feigi:
## Summary

`ledger.mjs` splices its own `--file`/`--require-file` out of argv by exact
token match; every other `--` token lands in `check`/`filed`'s free-text
tail, where `rest.join(" ")` folds it straight into the duplicate-filing
subject. Measured against this tree before this change:
`check --requre-file "widget guard missing"` (a misspelled flag) searched
for subject `"--requre-file widget guard missing"` — a different subject
than the seeded one — and answered exit 0 (`unverified`) where the
correctly-spelled invocation answers `ALREADY FILED` at exit 1. A caller
gating on exit code files a duplicate.

Per #584's Agent Brief, the straightforward guard
(`rest.find(a => a.startsWith("--"))`) is refuted: it also refuses a
subject that legitimately opens with `--`, which is the routine shape of
issue titles in this repo (measured in the issue's own history).

## Fix

`refuseStrayInTail()` in `ledger.mjs` applies the Brief's ruling: the
documented calling convention hands the subject as ONE argument
(`check "<subject>"`, `filed <issue> "<subject>"`), so a stray flag reveals
itself only by splitting that argument into more than one. A `--`-prefixed
tail element is refused by name only when the tail has more than one
element; a one-element tail — including one that legitimately opens with
`--` — is accepted unchanged. Applied to both `check`'s tail and `filed`'s
subject tail (past its issue-number argument), per the ticket's own AC.

`arg.mjs`'s `makeSweep()` comment is updated: it no longer describes this
gap as open, and explains why the fix does not route through `sweep()`
(the tail is legitimately data, same reason `sweep()` already excludes
`ledger.mjs`).

## Enumerated cases

Covered (each has its own test):
- `check <stray-flag> "<subject>"` refused, naming the stray token
- `filed <issue> <stray-flag> "<subject>"` refused, naming the stray token
- `check "--<subject that legitimately opens with -->"` accepted unchanged
  (the pin the refuted remedy failed)
- `filed <issue> "--<subject...>"` accepted unchanged (same pin, filed side)
- an unquoted multi-word subject with no stray flag stays accepted (guards
  against the new check over-refusing by tail length alone)
- known residual, left deliberately per the Brief: a stray flag alone with
  no subject at all is a one-element tail, accepted as the (degenerate,
  harmless) subject — pinned so a future change doesn't "fix" it

Left alone, per the Brief's Out of scope: routing `ledger.mjs` through
`sweep()`, and the pre-subcommand position (already refused today).

Verified with two scratch mutants (not part of the diff): reverting to the
refuted naive guard (drop the tail-length gate) reds exactly the three
accept-pin tests and none of the refuse-pin tests; disabling the new guard
entirely reds exactly the two refuse-pin tests and none of the accept-pin
tests.

Sizing: light

Closes #584

## Test plan
- [x] `./agent-test skills/fleet/scripts/ledger.mjs` — 87/87 pass
- [x] `./agent-test skills/fleet/scripts` (full fleet scripts suite) — 1592/1592 pass
- [x] Manual repro of the measured defect against `origin/main`, confirmed fixed
- [x] Two scratch mutants confirm the new tests discriminate the fix from both the original defect and the refuted naive remedy


feigi:
Review findings applied in c81e792.

**Applied**
- `survived-0` `correctness` `ledger.mjs:276` — the id slot ahead of the guarded tail, which is where #584's own signature survived this branch. Guarded separately from the tail, per the finding's anti-remedy: widening `refuseStrayInTail` over `rest` would refuse the legitimate one-argument `--`-leading subject this branch pins. The finding's fix text extended the remedy to `row`'s and `ruled`'s id slots as a shared claim; each site was measured on its own instead, and each carries its own reason — `filed` loses the duplicate-filing answer, `row` loses the rewrite-in-place key, `ruled` loses only its decision record.
- `survived-1` `silent-failure` `ledger.mjs:258` — `refuseStrayInTail` on `row`'s and `ruled`'s tails.
- `survived-2` `tests` `ledger.test.mjs:1614` — a pin for a stray flag in a non-first tail position, derived from the post-fix tree and covering the newly guarded subcommands.
- `unverified-6` `simplify` `ledger.test.mjs:1672` — the one suggestion-band entry a refuter could not refute. The two inert seeds are gone, replaced by the stronger assertion they were masking.
- `unverified-0` and `unverified-4` needed no separate change: both named `row`/`ruled` being unguarded, which `survived-1` closes.

Each guard was deleted individually against a copy under a scratch dir: every deletion reddens exactly its own pin and nothing else, and a semantically equivalent rewrite of the id test is not caught.

**Deferred** — #1159, closed `wontfix` as the refutation record for this review. `unverified-1`, `unverified-2`, `unverified-3`, `unverified-5` and `unverified-7`, each checked by one refuter biased to refuse and each refuted by measurement. One residual a refuter surfaced *was* applied: the test block comment claimed only a misspelled flag can reach the tail, which is false for a repeated one, since the splice is `indexOf`-based.

`node --test skills/fleet/scripts/*.test.mjs` at c81e792: `fail 0`, exit 0.

feigi:
Comments-dimension re-run applied in 3e92e9f. That dimension produced no coverage on the original review, so none of this PR's prose had been checked.

Every sentence was re-derived from the tree as it stands after the code applies in c81e792, not copied from the report — the report measured the branch before the guards landed, so its suggested wordings describe a tree that no longer exists. `I8` is the clearest case: it asked for `row`/`ruled` to be documented as deliberately unguarded, and they are guarded as of c81e792, so applying it would have shipped a false sentence.

**Corrected**
- `C1` — "closed the stray-flag gap" was wrong; it is narrowed. A lone stray is a one-element tail and is still taken as the subject at exit 0. Both residuals now have an owner in the tree instead of only in an off-tree brief.
- `C3` — the paragraph priced the sweep's cost as "refuses working invocations" and presented this guard as carrying none. It carries the same cost, narrowed: `check the --basee flag is unread` answered at exit 0 on `origin/main` 9b313e7 and exits 2 here. Since c81e792 that also holds for `row` and `ruled`, so the cost is wider than the report measured. Stated rather than denied — #365's AC is what makes it matter.
- `C4` — the comment claimed a documented convention the tree contradicts. Quoted is documented for `check` alone; `filed`/`row`/`ruled` are spelled unquoted everywhere, including this script's own usage strings. Filed as #1161, not fixed here: the fix edits `run-team/SKILL.md`, which PR #1154 also edits, and that collision is not this PR's to take.
- `C2` — "reaches the tracker query unchanged" named the wrong value. The payload's `subject` field survives; the query is normalised and reordered (`the --basee flag is unread` → `unread basee flag`).
- `C5` — the "only a MISSPELLED one survives" claim. Already corrected in c81e792; the splice is `indexOf`-based, so a repeated correctly-spelled flag reaches the tail too.
- `I6`, `I9`, `I10`, `I11`, `I12` — mechanism wording, the "finds nothing" property that belongs to the stubbed gh rather than the invocation, the positional references (including two of my own from c81e792), the `startsWith("--")` distinction that is really the length gate, and the off-tree `Agent Brief` citations.

**Not applied**
- `I8` — stale on arrival, as above.
- `I7` — unsettleable in-tree, so the bare assertion was dropped rather than restated.
- The three clauses verified correct — the one-element/multi-element rule, the arg.mjs cross-reference, and the `subjectParts`-not-`rest` note — were left alone.

`node --test skills/fleet/scripts/*.test.mjs` at 3e92e9f: `fail 0`, exit 0.

Issue #1099, opened 2026-08-30
net.test.mjs's wall-clock budget tests fail nondeterministically under concurrent suite load — and the fleet's normal condition IS concurrent load


feigi:
Measured 2026-08-30 during a `/fleet:run-team 2 5` run, by `impl-493` while re-running its suite under a controller warning. Not introduced by any PR in that run — `impl-493`'s diff is markdown plus one new prose test.

## The defect

Two tests in `skills/fleet/scripts/net.test.mjs` are budgeted in wall-clock time and fail when the machine is loaded:

- `net_kill_tree signals the whole subtree…` — observed failing at **10008ms**
- `a fetch that is slow but WORKING…` — observed failing at **20254ms**

## Why it is a defect and not just a slow machine

**The same tree gave different answers on two consecutive full runs**, which is the property that makes a suite unusable as a measuring instrument:

| run | result |
|---|---|
| full suite at `07589ee` | `pass 1512 / fail 1` |
| full suite at `3052781` | `pass 1511 / fail 2` |
| `node --test skills/fleet/scripts/net.test.mjs` in isolation | **5/5 pass** |
| full suite again, machine quiet | **1513/1513, fail 0** |

Isolation passes. Load fails. Two runs of one tree disagree.

## Why this matters more here than in an ordinary repo

**Concurrent suite runs are the fleet's normal operating condition, not an edge case.** Implementers, fix-appliers, finishers and review specialists all run `node --test skills/fleet/scripts/*.test.mjs` from the same machine, often simultaneously — the isolation envelope partitions ports and worktrees, and deliberately does **not** partition the process table or CPU.

The consequences are all silent:

- **A mutation measurement taken under load is untrustworthy in the dangerous direction.** A baseline that fails for load reasons, compared against a mutant, reads exactly like a mutant that failed to be caught — or like one that was caught when it was not. Every `it and only it goes red` claim in this repo depends on a stable baseline.
- **A red here is unattributable.** A member seeing `fail 1` cannot tell a load artifact from a real regression without re-running, and the runbook tells members to treat red as a finding.
- It compounds with the existing rule that a member must never `pkill` a stalled run, which is correct — but leaves load-induced failure with no remedy except re-running and hoping.

## Approach is open — hence `needs-triage`

Several defensible directions, none obviously right:

1. **Raise the budgets.** Cheapest, but a wall-clock budget large enough for a loaded machine is large enough to stop detecting the regression it exists for.
[...]

PR #1154, opened 2026-09-01
fix(review-pr): tell a policy-skipped finding from one whose refuters crashed

### LINKED ISSUE #591: review-pr: the `unverified` band conflates policy-skipped suggestions with findings whose refuters CRASHED

Split out of #239's triage. Not part of that ticket's fix — filed because it is a hard blocker on one of the options #239 shelved.

## The defect

`review-pr`'s result payload puts two populations in `unverified`, and they read identically:

- **Policy-skipped** — the `suggestion` band, which the workflow budgets **0 refuters** by design. Nothing looked, on purpose.
- **Crashed** — findings whose refuters ran and died. Nothing looked, by accident.

A consumer cannot tell them apart. Measured on `feigi/claude-config`, one `/fleet:run-team` run: the workflow returned 33 agents, 18 done, **15 errored on `API Error: 529 Overloaded`**. Result bands: 4 survived, 2 refuted, 15 unverified. Every one of those 15 was a crash, and the payload presents them the same way it presents a policy skip.

`review-and-fix.md` step 2 already carries the rule that saves you *if you read it* — "**A finding in `unverified` whose refuters ran and crashed always defers, whatever its severity**" — but the step cannot mechanically tell which is which, so the rule rests entirely on the reader remembering the distinction exists.

## Why this blocks a severity floor

#239 shelved three throttling options behind a guard. Option 1 is a **severity floor** — file `important` and above, record the rest. Applied to this band, a floor silently drops `critical` findings that nobody ever looked at, because a crashed `critical` refuter is indistinguishable from a policy-skipped `suggestion`. The floor would read the band as "low severity, safe to drop" when it is in fact "unknown severity, nobody checked".

So: if #239's repriced label fails its guard and options 1-3 come off the shelf, **this ticket lands first**.

## Recovery path exists

The run is resumable — `Workflow({scriptPath, resumeFromRunId})` replays the unchanged prefix from cache and re-runs only the crashed calls. So the correct response to a crash-heavy `unverified` is **resume, not defer**. Nothing currently says so at the point of use.

## Why \`needs-triage\` and not \`ready-for-agent\`

The defect is confirmed by measurement. The remedy is genuinely open — distinguishing the two populations is a payload-shape question (a new band? a per-finding `refutersRun` count? an error carried on the finding?), which is API shape, and phase 0's inventory marks that undecided. This is an honest **remedy-open** ticket even under the filing bar #239 repriced.

## Acceptance criteria

- [ ] A consumer of the `review-pr` payload can mechanically distinguish a policy-skipped finding from one whose refuters crashed, without reading severity or guessing
- [ ] A crash-heavy result names the resume path at the point of use, not only in a memory
- [ ] `review-and-fix.md` step 2's existing rule is expressed against whatever discriminant lands, rather than resting on the reader

## Related

- #239 — the filing policy; this blocks its shelved option 1

feigi:
Closes #591

Sizing: heavy

## What was wrong

`review-pr`'s `unverified` band had two producers and one shape. A `suggestion`
skips the adversarial pass because the workflow budgets that band 0 refuters by
policy; a finding whose refuters were all dispatched and all died lands in the
same band with the same empty vote list. "Nothing looked, on purpose" and
"nothing looked, by accident" were byte-identical to a consumer.

Re-derived against `origin/main` before implementing — both branches assigned
the band with nothing to tell them apart, exactly as the Agent Brief's live
re-verification says.

## The change

`verdictFor(dispatched, votes)` is one pure top-level function both producers now
route through, and it returns `refutersDispatched` beside the verdict. Zero
dispatched is the policy skip; a non-zero dispatch with no surviving vote is a
crash. Additive, per the brief's ruling: no fourth band, no re-keying of the
three that exist.

The payload also carries `resume`, non-null only when some `unverified` finding
has refuters dispatched. A crash-heavy result is resumable, so the correct
response is resume rather than defer, and that was written nowhere the reader of
a result would meet it. `resumeFor(unverified)` is the second pure top-level
function, holding that crash population and the response it gates, for the same
reason `verdictFor` is one — see **Pinning**.

Settling command, re-run at `38bc502`:

```
$ grep -c 'verdict: "unverified"' workflows/review-pr.js
0
$ grep -n 'verdictFor(' workflows/review-pr.js
985:function verdictFor(dispatched, votes) {
1108:        if (n === 0) return Promise.resolve({ ...f, dimension: d.key, ...verdictFor(0, []) });
1159:          return { ...f, dimension: d.key, ...verdictFor(n, votes) };
```

No producer assigns the band outside `verdictFor`.

## Enumerate and declare

**The class — every site that produces `verdict: "unverified"`.** Both, and
there are no others: the 0-refuter early return and the all-votes-lost branch.
Settled by the `grep -c` above returning zero after the change.

**The sibling class — every file carrying the reading rule that rested on the
reader.** The ticket's AC names `review-and-fix.md` step 2; `run-team/SKILL.md`
carries the same rule verbatim in the fix-applier's dispatch prompt, and its
Reviewers section documents the return shape. All three covered — a rule
re-expressed in one copy and left resting on the reader in the other is the
defect moved rather than fixed.

Settling command, re-run at `38bc502`:

```
$ grep -rl "refutersDispatched" skills/ workflows/
skills/fleet/scripts/review-pr-unverified-discriminant.test.mjs
skills/fleet/skills/run-team/SKILL.md
skills/fleet/commands/review-and-fix.md
workflows/review-pr.js
```

**Deliberately left:** `docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md`
describes the band as it stood on its own date. A dated design record is not a
reading rule and is not rewritten to match later code. Also left, as the ticket
puts them out of scope: #239's shelved throttling options, the `suggestion`
band's 0-refuter budget, and any retry or auto-resume.

## What this could wrongly refuse, and what my edit newly does

The field refuses nothing — it is computed, not a guard. The risk this change
carries is the restructure: routing both producers through one function could
have changed the verdict for input it must still ACCEPT. The test feeds it live
vote lists and pins that the majority rule is unchanged, ties included
(`refuted * 2 >= live.length`), and that a crashed refuter among live ones is
filtered rather than counted in the denominator.

AC2 — existing consumers unchanged — is a reading obligation, and a green suite
does not discharge it, so I read them. No code consumes the payload at all:

```
$ grep -rnE '\.(survived|unverified|refuted|dimensionsRun)\b' skills workflows hooks agents | grep -v '\.test\.mjs'
workflows/review-pr.js:987:  const refuted = live.filter((v) => v.refuted).length;
```

That single hit is the workflow's own vote field, not the payload. The payload's
readers are agents reading prose: run-team/SKILL.md's Reviewers section (the
controller) and review-and-fix.md step 2 (the fix-applier). Both read the three
band keys by name, and all three keep their names and their contents.
`ledger.mjs`'s own `unverified` verdict is its tracker-check vocabulary and has
no coupling to this band.

## Pinning

Both shapes are driven through the lifted `verdictFor` and asserted to differ —
AC5's test, which the ticket asks for because both shapes previously produced
identical output, leaving the fix unpinned without it.

Every added assertion was mutation-checked against the review suites
(`review-pr-unverified-discriminant`, `review-pr-unrun`, `review-pr-reads`,
`review-path-default`, `select-dimensions`, `review-pr-refuter-scratch`), each
mutation applied to a scratch copy and never to the worktree:

| mutation | result |
| --- | --- |
| `refutersDispatched: dispatched` → `: 0` | the discriminant test and the accept test red, nothing else |
| tie flipped to `refuted * 2 > live.length` | the accept test red, nothing else |
| policy-skip branch builds its own verdict again | the call-site test red, nothing else |
| `resume` dropped from the return | the resume test red, plus the pre-existing documented-shape guard |
| `review-and-fix.md` rule stops naming the field | that file's prose pin red, nothing else |
| **control:** rename the local inside `verdictFor` | all green — the pins read behaviour, not incidental text |

### What the review found, and what this branch changed in response

Two of the call-site pins above were **vacuous**, and the review measured it.
Each read a SPAN — `[^}]*` in the policy-skip pin, an unbounded `[\s\S]*?` in
the post-refuter one — and a span finds `verdictFor(...)` wherever the token
lands, a comment included. `stripComments` documents that ceiling itself: an
inline `/* */` and a trailing `//` keep their text. So a revert that rebuilt the
object inline and left the call named in a surviving comment passed both pins.

Settling command, run at `38bc502` against a `git archive` of `6956e03`:

```
policy-skip branch, revert WITH inline comment      tests 7 pass 7 fail 0, exit 0
policy-skip branch, same revert WITHOUT it          tests 7 pass 6 fail 1, exit 1
post-refuter branch, revert WITH trailing comment   tests 7 pass 7 fail 0, exit 0
post-refuter branch, same revert WITHOUT it         tests 7 pass 6 fail 1, exit 1
```

The pins discriminated on comment text, not on the call. Both now match the
WHOLE returned expression, spread included, which no comment can supply.

The crash population also had no behavioural test: the predicate separating a
policy skip from a dead refuter gated both the log line and `resume`, and nothing
drove it. Same command, same archive: replacing that filter with the unfiltered
list left this file's tests green — `tests 7 pass 7 fail 0`, exit 0.
`resumeFor(unverified)` lifts the predicate and the response it gates into one
pure top-level function, so a test can DRIVE the classification over both
populations rather than pin it as text; the call is pinned too, so dropping it is
red rather than dead code.

Mutation-checked at `38bc502`, each row applied to a `git archive` copy in
scratch, each failing assertion named:

| mutation | result |
| --- | --- |
| policy-skip branch reverted, comment preserved | its own pin red, nothing else |
| post-refuter branch reverted, comment preserved | its own pin red, nothing else |
| `const { crashed, resume } = resumeFor(unverified)` bypassed | the call pin red, nothing else |
| `refutersDispatched > 0` → `>= 0` | the classification test red, nothing else |
| filter body → `return unverified;` | the classification test red, nothing else |
| **control:** log line reworded | all green |
| **control:** whole-line comment added inside the `.then` closure | all green |

Full suite green at `38bc502`:
`env -u FORCE_COLOR node --test skills/fleet/scripts/*.test.mjs` →
`tests 1594 / pass 1594 / fail 0`, exit 0. One more test than the run at
`6956e03`, which is the classification test this branch adds. The tree also
carries a pre-existing flake unrelated to this change — `net.test.mjs`'s
`net_kill_tree` failed once on the unmodified tree at the start of this work with
`ETIMEDOUT` (tracked as #1099), and passed in the green run above.


feigi:
Review deferrals filed (PR #1154 review of `6956e03`, all six dimensions run, `dimensionsUnrun` empty).

**Applied here** — the four `survived` findings, as one commit, `38bc502`:

- `survived-0` + `survived-1` (`critical`, tests): both vacuous call-site pins, repaired with one mechanism — each now matches the whole returned expression instead of a span a comment can occupy.
- `survived-2` + `survived-3` (`important`, correctness + tests): `resumeFor(unverified)` extracted so the crash population and the `resume` it gates are driven by a test rather than pinned as text, with the call pinned too.

**Deferred and filed:**

| finding | dimension | issue |
| --- | --- | --- |
| `unverified-0` + `refuted-0` residual | silent-failure, correctness | #1156 — `resume` stays null when a whole reviewer dimension dies |
| `unverified-2` + `refuted-4` residual | simplify, types | #1157 — `verdictFor`'s `dispatched` parameter, contested from two directions |
| `refuted-0` … `refuted-4`, plus `unverified-1`, `unverified-3`, `unverified-4` | all six | #1158 — refutation record, `wontfix`, closed |

`unverified-3` (delete the `if (n === 0)` early return) was held deliberately: it is mutually exclusive with `survived-0`, whose pin is that this branch routes through `verdictFor(0, [])`. Applying both would strengthen a pin and delete its subject in the same diff. Recorded in #1158 with that reason.

PR #1155, opened 2026-09-01
fix(no-undo-audit): read merge-tree's output with one command, not a pipeline

### LINKED ISSUE #583: no-undo-audit.sh:490 and :524 — pipeline stages ahead of the last one have their status discarded, so a fault under-reports at exit 0

feigi:
A POSIX pipeline's status is its **last** command's, so a fault in any earlier
stage is invisible to `set -e` and to a trailing `|| die` alike. Both of this
script's multi-stage answers had that shape:

- the conflicts split (`tr | tr | awk`) carried no guard at all;
- the at-risk list (`printf | sed | tr | xargs`) had a `|| die` that could only
  ever answer for the last of the four.

Either way a broken prefix stage yields short output, the pipeline exits 0, and
the audit reports a **smaller** answer with full confidence — on the script
whose stated standard is that an unanswerable question is exit 2 and never a
payload.

Both now go through one byte-oriented `python3` reader: stdout carries the paths
for the payload, `$ps_out` carries the same paths NUL-separated and
`:(literal)`-prefixed for `xargs`. One command, one status, nothing ahead of it
with a status to discard. The at-risk read is no longer a pipeline at all, so
its `|| die` — whose wording #522 settled and this PR does not touch — is now
accurate about every command that can produce the status it reads.

`set -o pipefail` is ruled out, and not only because POSIX sh lacks it. A
healthy conflicted run already ends with the prefix stages killed by SIGPIPE,
because the reader stops at the empty record by design and leaves merge-tree's
prose tail unread; `pipefail` would turn every large **correct** run into a
refusal. AC-3 asked for that to be a test rather than an argument, and it is.

Sizing: light

## Enumerate and declare

Covered — the answer-bearing pipelines, both of the ticket's sites:

1. the conflicts split, which had no guard at all;
2. the at-risk pathspec list, whose guard read only `xargs`.

Deliberately left, every one for the same reason: its only non-final stage is
the shell **builtin** `printf` on a string already in memory, which is the line
the ticket's own body draws ("`printf` and `sed` on in-memory data essentially
do not fail; `tr` demonstrably does"). Those are the stash count, the fold of
git's diagnostic, the `jarr`/`jarr_rewritten` pairs for both arrays, the
at-risk dedupe, and the three stderr listings — and the listings are not
answer-bearing at all: a fault there loses a diagnostic line, not a payload.

Settling command for that claim, re-run at this head:
`awk '!/^[[:space:]]*#/ && /\|/ && !/\|\|/' skills/fleet/scripts/no-undo-audit.sh`
— every line it reports has `printf` in the non-final slot.

The other half of the question. `python3` becomes required on the path this
audit takes through a conflicted branch; absent or broken, a run that used to
answer now refuses at exit 2 naming `python3`. The Agent Brief rules that
tradeoff in ("stage the output through a NUL-capable reader — `python3` or
`node`, both already dependencies of other fleet scripts"). A second temporary
is created, so a full TMPDIR is a new way to refuse; it refuses at exit 2 and
not at the exit 1 that means the worktree is dirty, and it removes the
temporary it had already created before refusing — which is what the
uncreatable-temporary test pins, on all three counts.

The accept-side controls are the large-run test, the broken-`sed` test, the
suite that was already here, and the byte-identity check below.

## Evidence

Mutation, each run at the head of this branch on a scratch copy, restored with
`cp`:

| mutant | reds |
|---|---|
| conflicts read loses its `\|\| die` | the unreadable-output test, alone |
| at-risk read reverted to `printf \| sed \| tr \| xargs` | the broken-`sed` test, alone |
| pathspec temporary loses its `\|\| die` | the uncreatable-temporary test, alone |
| cleanup trap moved back below the second `mktemp` | the uncreatable-temporary test, alone |
| whole file reverted to `origin/main` | those three; **the large-run test survives** |
| same, plus `set -o pipefail` | those three **and the large-run test** |

Every row above was re-run at this head after the review fixes landed, on
`skills`-only scratch copies restored by re-extracting the archive. Three
suites red in all of them, baseline included — the doc and design-spec prose
pins, which read files a `skills`-only copy does not carry — so the reds named
per row are the ones the mutant adds.

The last two rows are the AC-3 control doing its job: the large-run test is not
merely pinning this fix — it survives the old code and dies on `pipefail`, which
is the failure mode nobody could see by reading the pipeline. The
discrimination control is the other direction: rewriting the reader's loop as
`itertools.takewhile` leaves all 83 tests green, so the tests pin behaviour and
not the shape of the code.

Byte-identity for ordinary healthy runs (AC-4): `origin/main`'s script and this
one produce identical stdout **and** stderr on a clean no-conflict run and on
conflicting paths spelled with a quote and a space, with pathspec magic, and
with a control byte. Harness kept under this member's scratch directory; the
fixtures it builds are the same shapes the suite already carries.

Suites that read `no-undo-audit.sh`, settled by replacing the file with
`MUTATED` and mapping every red test back to its file rather than by a
literal-path grep — the sweeps assemble the path, so a grep would miss them:
`no-undo-audit.test.mjs`, `locale-pin-prose.test.mjs`,
`muted-git-guard-sweep.test.mjs`. All three green.

`./agent-test .` green at this head.

## Prose touched

The prologue's `#582` paragraph described the `tr | tr | awk` that this PR
removes, so it would otherwise have named a pipeline that no longer exists. It
now records the measurement as history, says the split has become a reader whose
status nothing discards, and points at what the locale pin still covers here —
the `sed` and `tr` that render a path or git's own diagnostic to stderr, with
json.sh noted as pinning per call rather than relying on the export. The same paragraph in `no-undo-audit.test.mjs` said it in the
present tense and had to be rewritten the same way; it now also records that
the mutant that test still kills dies by a route the paragraph does not
describe, so that a later tidy-up of the render it dies in cannot remove the
kill silently. The
`ponytail:` note that deferred this restructure is replaced by one that says
what is still deferred: `$ps_out` already carries a newline-holding path
unparked, so the at-risk half could answer for it, but `conflicts[]` reaches the
payload through a shell variable and NUL cannot travel there.

## Applied from the review

The review found the second `mktemp` sitting between the first one and the EXIT
trap, so a refusal in that window ran with no trap installed and left the first
temporary on disk. The trap is armed before the second call now, and the
uncreatable-temporary test — which drove exactly that window while asserting
only the exit code and the message, both correct either way — now reads back the
path the first call handed out and requires the file to be gone.

The `longTailConflictRepo` fixture's two parameters had no caller passing
either; inlined.

Deferred, each with the reason recorded: #1160 (a diagnostic stderr render can
abort a clean audit with the status that means dirty — sequenced, because the
locale regression test currently depends on that render dying) and #1162, the
closed record for the checked suggestion band.

Closes #583


feigi:
Review deferrals filed:

- **#1160** (open, `ready-for-agent`, dimension `silent-failure`) — `no-undo-audit.sh:565`: a diagnostic stderr render can abort a clean audit with exit 1, the status that means dirty worktree. Not fixed here because the invalid-UTF-8 regression test currently gets its mutant kill from that `sed` crashing; the issue carries that warning, so whoever takes it gives the test a new kill mechanism first.
- **#1162** (closed, `wontfix`) — `PR #1155 review: the suggestion band, checked`. Three suggestions refuted by measurement (`unverified-1` duplicate die message, `unverified-2` reader die label, `unverified-8` `/dev/stdout` shim variant), each with the run that refuted it. `unverified-7` (drop `ps_out`, one temp file) is recorded below the claim bar: mutually exclusive with the trap fix that landed here, with a re-open trigger.

Applied instead of deferred: the cleanup trap now covers both temporaries before the second `mktemp` can fail, the test that drives that window asserts the cleanup, the test-side #582 docstring is past tense and records what its mutant actually dies of, and the fixture's dead parameters are inlined. `node --test skills/fleet/scripts/*.test.mjs` from the worktree: tests 1590 / pass 1590 / fail 0, exit 0.

Issue #1162, opened 2026-09-01
PR #1155 review: the suggestion band, checked


feigi:
Durable record of the `suggestion` band from the PR #1155 review, which the review workflow budgets 0 refuters by policy. Each in-scope entry got one refuter, instructed to default to `refuted=true` and to verify by running rather than reasoning. Filed **closed**: nothing here is queued agent work. `ledger.mjs check` reads `--state all`, so a later run re-deriving one of these lands on the evidence instead of re-litigating it.

What survived is not here — it shipped in PR #1155 (03c563b): the cleanup trap armed before the second `mktemp`, the test extension that pins the cleanup, the rewritten #582 docstring, and the fixture's dead parameters inlined.

---

### `unverified-1` — dimension `correctness` — `no-undo-audit.sh`, the two `mktemp` guards
**Claim:** the new `ps_out` guard reuses `mt_out`'s exact die message, so the two temp-file failures are indistinguishable on stderr, against a convention the file states and tests elsewhere.

**Refuted.** The messages are byte-identical — that much holds. But the consequence does not. The refuter mutated the script to remove `|| die` from the `ps_out` guard alone and ran the test: it reds on its exit-code assertion (`expected 2, actual 1`), because under `set -eu` an unguarded `mktemp` aborts with mktemp's own status. So the test already pins the site it claims to; the stderr regex is a redundant second belt, not the load-bearing pin, and the finding's stated remedy delivers something already delivered.

The cited convention also does not cover this class. It reads "a single shared message could not tell the operator which **array** the run lost" — a rule about `conflicts[]` versus `atRisk[]`, two different answers a consumer pastes into a diff command. Both `mktemp` failures lose the whole run identically. Full operator-visible stderr was dumped for both sites and is identical apart from the temp dir name, with the same exit status, the same empty payload, and the same remedy.

*Re-open trigger:* a third temp file whose failure an operator could act on differently.

### `unverified-2` — dimension `silent-failure` — `no-undo-audit.sh`, the reader's `|| die`
**Claim:** that `|| die` covers two faults and names one — a failure *writing* `$ps_out` is reported as "could not read git merge-tree's output", pointing the operator at merge-tree instead of the temp dir.

[...]

Issue #1161, opened 2026-09-01
ledger.mjs filed/row/ruled are documented unquoted, but #584's guard requires the subject as one quoted argument


feigi:
Deferred from PR #1153 review (issue #584).

**dimension:** `comments`

## The finding

PR #1153 refuses a stray flag folded into `ledger.mjs`'s free-text tail. The rule rests on the subject arriving as ONE argument — a multi-element tail carrying a `--`-prefixed element is refused. That convention is documented for `check` and only for `check`:

- `skills/fleet/skills/run-team/SKILL.md:1422` — `ledger.mjs check "<subject>"` (quoted)
- `skills/fleet/commands/review-and-fix.md:10` — `ledger.mjs check "<subject>"` (quoted)

Every other spelling in the tree is UNQUOTED, including the one a caller is most likely to copy:

- `skills/fleet/skills/run-team/SKILL.md:2080-2081` — `ledger.mjs filed <issue> <subject>` and `ledger.mjs check <subject>`
- `ledger.mjs`'s own usage strings — `usage: ledger.mjs filed <issue> <subject>` (:292), `row <ticket> <text>` (:270), `ruled <pr> <decision>` (:306), `check <subject>` (:355)

Nothing in the tree documents `filed <issue> "<subject>"`, `row <ticket> "<text>"` or `ruled <pr> "<decision>"`.

## Why it matters

A caller following `SKILL.md:2080` literally, with a subject that contains a `--` word, now gets exit 2. Measured at PR #1153's head c81e792 against `origin/main` 9b313e7:

```
                                              main   c81e792
check the --basee flag is unread              exit 0  exit 2
filed 999 the --basee flag is unread          exit 0  exit 2
row 42 the --basee flag is unread             exit 0  exit 2
ruled 77 the --basee flag is unread           exit 0  exit 2
```

Quoting the subject accepts it unchanged in every case. The refusal is loud and recoverable and replaces a silent wrong-subject answer, which is the trade #584 ruled for — but the instruction a caller is reading tells them to do the thing that now fails.

## Remedy is OPEN

Two shapes, and the choice is a maintainer call:

1. Update the unquoted call sites to the quoted spelling, so the docs match the guard.
2. Leave the docs and narrow the guard for these subcommands.

[...]

Issue #1160, opened 2026-09-01
no-undo-audit.sh: a diagnostic stderr render can abort the audit with exit 1, the status that means dirty worktree


feigi:
**Deferred from PR #1155 review.** Dimension: `silent-failure`.

`skills/fleet/scripts/no-undo-audit.sh:565` — `printf '%s\n' "$conflicts" | sed 's/^/    conflict: /' >&2`

## The defect

That `sed` renders a diagnostic to stderr and decides nothing. It carries no `|| die`, so under `set -eu` a failure in it aborts the whole audit with the failing tool's own status. Out of this script exit 1 is the dirty-worktree refusal — the audit has by then already printed `clean` for the worktree it measured, so the caller is handed a refusal contradicting the script's own preceding output, with no `no-undo-audit:` line naming a cause. A render is deciding the verdict.

The reviewer reproduced it by shimming `sed` to fail only on the invocation carrying the `conflict: ` script, leaving the script's other `sed` calls untouched: stdout showed `clean`, then `sed: RIP`, then exit 1 with no payload.

The same shape stands on the other two diagnostic renders, `sh:246` (the porcelain dump) and `sh:631` (`at risk: `). Only `sh:565` was probed directly; the other two were enumerated by grep and share the construct.

## ⚠️ Read this before changing sh:565

**The invalid-UTF-8 regression test in `no-undo-audit.test.mjs` presently depends on that `sed` dying.** PR #1155's review measured the mutant that test exists to kill — `export LC_ALL=C` removed from the script — and found the mutant now dies via this very `sed` crashing on the bad byte (`sed: RE error: illegal byte sequence`, exit 1), not via the truncation mechanism the test was written for. That mechanism was deleted in #583 along with the pipeline it lived in.

So neutralising this `sed` — `|| :`, or replacing it with a shell loop — would silently defang that regression test: the mutant would stop dying and nothing would go red. **Whoever takes this must give that test a new kill mechanism first, and confirm the mutant still dies through it, before touching the render.** Neither PR #1155 review finding could see this from its own side; it is the reason this is filed rather than fixed there.

## Why deferred rather than fixed in #1155

Applying it inside #1155 would have removed a mutant kill that PR had no replacement for. #1155 instead records the dependency in the regression test's own comment, so the trap is visible from the test as well as from here.

## Remedy

[...]

Issue #1159, opened 2026-09-01
PR #1153 review: the suggestion band, checked


feigi:
Deferred from PR #1153 review (issue #584).

The review's `suggestion` band gets 0 refuters by policy, so every entry below arrived unchecked. Each was scoped first, then handed to one refuter biased to refuse, which verified by running something. All five were refuted by measurement. None was applied. Recorded here rather than dropped, so a later run re-deriving one lands on the evidence instead of re-litigating it (`ledger.mjs check` reads the tracker at `--state all`).

What *did* survive the band is in the PR: `unverified-6` was the one entry a refuter could not refute, and it shipped in c81e792. `unverified-0` and `unverified-4` were subsumed by the `survived-1` apply in the same commit, which guards `row` and `ruled` — the omission both of them named no longer exists.

---

## `unverified-1` — `correctness` — `skills/fleet/scripts/ledger.mjs:172`

**Claim:** a bare `--` end-of-options separator in a multi-element tail is now refused as `unknown flag --`, which is neither a flag nor a typo.

**Refuted.** The premise is false: `ledger.mjs` never implemented `--` as a separator. Pre-PR it folded into the subject as data — the payload carried `"subject":"-- widget guard missing"`, and `filed 999 -- "new widget thing"` persisted the row `#999 -- new widget thing`. The `check` case only appeared to work because `norm()` strips non-alphanumerics; the JSON subject callers consume was already wrong. So no supported invocation regressed, and the documented one-argument form with a legitimately `--`-leading subject behaves identically on both sides. This is the same premise `refuted-1` rested on; both were built with proper pre-PR controls and upheld as refuted.

Residual, cosmetic: for a tail whose only `--`-prefixed element is the bare `--`, the message misnames the token. **Re-open trigger:** a caller that actually needs to pass a `--`-leading subject positionally. The live capability question in that shape is already tracked in #818, which proposes an end-of-options separator for `staleness.mjs`; if that is granted, this file should follow it rather than be decided here.

## `unverified-2` — `silent-failure` — `skills/fleet/scripts/ledger.mjs:174`

**Claim:** the refusal calls the script's own known flags "unknown" — `--file`/`--require-file` are spliced at their first occurrence only (`indexOf`), so a repeated one reaches the tail and is refused under a message that misdiagnoses it.

[...]

PR #1163, opened 2026-09-01
chore(fleet): the 2026-09-01 run's three tier rows, and a settle window on the desync signature

feigi:
Close-out for `/fleet:run-team 2 5` on 2026-09-01. Four implementer PRs merged this run — #1153 (issue #584), #1154 (issue #591), #1155 (issue #583) and #1164 (issue #590), the last dispatched after the review backlog drained — and this carries the artifacts they produced plus one runbook defect the run measured.

## The tier rows

#1153 and #1154 are a **within-run tier pair**: same day, same dispatch prompts, both `class=routine`, one `sonnet` and one `opus`. That is the shape that makes tier orthogonal to prompt evolution, which the whole-file split cannot be.

```
$ awk -F'\t' '!/^#/ && $1=="2026-09-01" && $2>=1153 {print $2, $5, $6}' docs/metrics/tier-outcomes.tsv
1154 opus yes
1153 sonnet no
1155 opus yes
1164 opus yes
```

**#1153's `closed_own_ticket=no` is measured, not judged.** Its implementer guarded the *tail* of `check` and `filed` but never the id slot, so #584's own defect signature stayed reachable one token to the left of the new guard. The finisher established this two ways rather than inferring it:

```
$ git show 20c84bf | grep -c refuseStrayInId
0
$ node ledger.mjs --file L filed --requre-file 999 "widget guard missing"
{"issue":"--requre-file","subject":"999 widget guard missing"}   exit 0
```

**Do not read one pair as a tier result.** The guard's bar is ten pairs across five or more distinct `run_date`s. The alternate-tier member is also chosen as the most ordinary ticket in its wave and never the hardest, so the pairing trades the calendar confound for a difficulty one — condition on `sizing`/`profile`/`loc`/`files` before comparing anything.

`member-outcomes.tsv` is regenerated wholesale, never hand-edited. Every session directory for this cwd was scraped rather than only this run's, because the newest-transcript helper answers a different question when sessions overlap — and they do here. So the file carries rows from sessions other than this one, which is the property to check rather than any row tally: the tally moves every time the scraper re-runs, and it re-ran again for the fourth tier row.

```
$ awk -F'\t' '!/^#/ && $1 != "449b7161-729d-4c60-94ee-80a345e5d2f9"' docs/metrics/member-outcomes.tsv | head -1 | cut -f1,2,3
01f3c6b6-3b96-4bb3-b34c-a512ffb34650   2026-07-31   specialist
```

## The runbook fix

`run-merge-bot.md` calls `post != pre` with `pr_head` still on `pre` a desync that needs a controller. **Measured twice this run: that signature appears spuriously.** Immediately after `gh pr update-branch --rebase` returns `rc=0` and the branch ref moves, GitHub's PR object lags, so a `headRefOid` read taken in that instant reproduces the signature exactly while nothing is wrong.

| wave | cleared on re-poll |
|---|---|
| 2 | attempt 24 (~2 min) |
| 3 | attempt 14 (~84s) |

Both bots got it right — one on its own judgement, one because the controller had warned it. **Neither because the runbook said so.** A bot following the text literally reports a desync and stops, costing a wave for a state that fixes itself, and it is precisely the reading a bot gets when it does everything else correctly.

The signature now has to survive the same bounded cap the `ls-remote` poll already uses. A desync that outlives the cap keeps its existing handling and its close-and-reopen precondition unchanged.

**The new mandate is pinned, and only the mandate.** The review found the paragraph observed by nothing — the shape #1147 is scored for two sections up. Measured against the tree before the pin existed: inverting the imperative, deleting the paragraph, and reverting the prose hunk each left the whole suite green. The two imperative clauses are now anchored separately, because dropping "on the same bounded cap" and dropping "only if ... runs out" break the rule in different directions; each mutation turns exactly that one test red. The measurement sentence above is deliberately left unpinned, so re-measuring the lag does not turn a test red — rewriting it to drop both figures was mutated as the discriminator and everything stayed green.

## Verification

```
tests 1608 / pass 1608 / fail 0   (exit 0)
```

The prose-pin suites are included in that run, which is what the `run-merge-bot.md` edit needed to clear.

## Deferrals filed by this run

Open: **#1156** (`resume` null on a whole-dimension crash), **#1157** (the `verdictFor` `dispatched` seam, two findings pulling opposite ways), **#1160** (a cosmetic `sed` — its body carries the warning that the LC_ALL regression test currently depends on that `sed` dying), **#1161** (`ledger.mjs filed` documented unquoted while #584's guard requires quoting).
Closed records: **#1158**, **#1162**, **#1159**.
Evidence added to existing **#990** (zsh argv word-splitting silently exonerating a finding).

From this PR's own review, whose `suggestion` band arrived with 0 refuters by policy: **#1168** open (`run-team/SKILL.md`'s worktree head check reads the same lagging field once and refuses on mismatch — the sibling site, out of this close-out's scope, so filed rather than applied), and **#1169** closed as the band's checked record (the paragraph's elapsed times, refuted by timing the described loop body). The band's third entry, the missing pin, survived its refuter and is applied here.

**#910** asked for the PR-object lag distribution before any remedy at the merge-bot site. This PR carries that measurement and the settle window; it is left open rather than closed by this PR.


feigi:
**Deferred by the maintainer: do not merge this standalone — fold it into the next run's close-out PR.**

Leaving it open rather than merging. Recorded here so the next `/fleet:run-team` controller finds it without rediscovering it.

**What this branch carries**, and what a folding PR must preserve:
- Three tier rows for the 2026-09-01 run in `docs/metrics/tier-outcomes.tsv` — PRs #1153 (issue #584, `sonnet`), #1154 (issue #591, `opus`), #1155 (issue #583, `opus`). #1153 and #1154 are a within-run tier pair, which is the only kind of row that is unconfounded by calendar date.
- A regenerated `docs/metrics/member-outcomes.tsv`. **Regenerate rather than merge this file** — it is derived wholesale from subagent transcripts and a hand-merge would not survive the next regeneration.
- `skills/fleet/commands/run-merge-bot.md`: a settle window on the desync signature. Measured twice this run — after `gh pr update-branch --rebase` returns `rc=0`, `headRefOid` lags the ref move and reproduces the "desync, needs a controller" signature while nothing is wrong. Cleared on re-poll attempt 24 (~2 min) in one wave and attempt 14 (~84s) in the next.

**Known cost of leaving it open:** an unlabelled PR that closes no issue counts as review backlog permanently, because nothing in the fleet will ever label it. That is the third case in #590's Agent Brief, and PR #1164 is the fix for it — under #1164's predicate this PR would correctly drop out of the count.

<sub>Recorded by the `/fleet:run-team` controller on maintainer instruction.</sub>

feigi:
**Review deferrals filed.** The `suggestion` band arrived with 0 refuters by policy, so it was split by scope and each in-scope entry got one refuter dispatched against it.

- **#1168** — open, `ready-for-agent`. `run-team/SKILL.md`'s worktree head check compares the branch-derived HEAD against `headRefOid` once and refuses on mismatch, which the lag window this PR measured can turn into a false refusal. Out of this close-out's scope — it edits a file this PR does not touch — so filed rather than applied. Remedy open.
- **#1169** — closed, `wontfix`. The band's checked record: the elapsed times in the new paragraph, refuted by measurement. The prose reuses the `ls-remote` **cap**, not its tick, and the described loop body was timed at 5.44s/iter against this repo, putting attempt 24 at 125.3s.

The band's remaining entry — the new imperative pinned by no test — survived its refuter and is applied in this PR.

Related: **#910** raises the same single-sample-of-a-lagging-field concern at the merge-bot site and asked for the lag distribution to be measured before any remedy. This PR carries that measurement and the settle window. It is still open and is not closed by this PR.

PR #1164, opened 2026-09-01
fix(fleet-tick): take review and merge-hold state from the caller, not from labels

### LINKED ISSUE #590: fleet-tick.mjs: the reviewer row prints DISPATCH n from the review backlog, but on the default path reviews serialize on one workflow and those slots hold fix-appliers
Observed twice in one `/fleet:run-team` run, 2026-08-18. `fleet-tick.mjs`'s reviewer row prints a `DISPATCH n` action that cannot be acted on when the fleet is on its **default** review path.

## What it prints

```
$ node skills/fleet/scripts/fleet-tick.mjs --implementers 0 --reviewers 3 --merge-bots 0 --pool 0 --implementer-cap 2 --reviewer-cap 5
implementers 0/2 → HOLD   (pool=0 supply=58 review-backlog=4 — a review-bound pipeline gains nothing from more PRs)
reviewers    3/5 → DISPATCH 2   (review-backlog=4)
merge-bot    0/1 → IDLE OK   (merge-queue=0)
```

At that moment there was nothing to dispatch. Three fix-appliers were live, a fourth review was still running inside the `review-pr` workflow, and no further review output existed.

## Why the number is not actionable

`run-team/SKILL.md` makes two statements that together bound reviewer dispatch:

- **The controller runs each review itself**, via `Workflow({name: "review-pr"})`, and **one at a time**: *"One review workflow at a time … A queue is not a reason to start a second."*
- **The reviewer cap holds fix-appliers, not reviews**: *"reviews serialize on your own turn and the reviewer cap buys no review parallelism at all — those slots hold fix-appliers, which do the cheap half."*

A fix-applier can only be dispatched *after* a review workflow returns findings for its PR. So the number of dispatchable reviewer members is not `min(cap - live, backlog)` — it is the number of **reviews that have returned and not yet been handed to a fix-applier**, which is usually 0 and is never more than 1 on the default path, since only one review runs at a time.

`fleet-tick.mjs` computes the reviewer row from the review backlog (open PRs without `ready-to-merge`), which counts PRs waiting on a review that has not started. Those cannot be dispatched against.

## Why it matters

The skill's queue-depth section says explicitly that the reconcile exists so the deficit is *"computed, not remembered"* — because a table you must remember to consult is one you will not consult under a merge-side event storm. That argument applies to this row too: a controller trusting the printed ACTION under load could start a second concurrent `review-pr` workflow (which the skill forbids) or spawn reviewer members with no findings to apply.

It did not mislead this run, because the controller held the rule in context both times. That is exactly the reliance the script was written to remove.

## Note on the implementer row

The implementer row was **correct** both times, including its backlog-gate reasoning — the defect is specific to the reviewer row.

## Possible directions — not a ruling

1. Report the reviewer row as `HOLD` whenever the caller signals a review workflow is in flight, with a new flag (e.g. `--review-running`), since the controller is the only thing that knows.
2. Split the concept: report `review-backlog` as an observation and drop the `DISPATCH n` action for reviewers entirely, leaving fix-applier dispatch to follow a returned review — which is the only thing that can trigger it.
3. Leave the arithmetic and change the wording, so the row reads as "PRs awaiting review" rather than an instruction to dispatch.

Which of these is right depends on whether `fleet-tick.mjs` should model the workflow path at all, or stay a pure level-reconciler for the hand-dispatched fallback path. That is a maintainer's call, hence `needs-triage`.

feigi:
Three of `fleet-tick.mjs`'s rows printed an ACTION derived from a label proxy for
state labels cannot express, so each named work nobody could take. Applying the
script's own header contract — *the controller states what only it knows,
required, never defaulted* — settles all three.

Sizing: heavy

## The backlog-definition choice, which the ticket left open

The brief named one candidate ("excluding PRs whose body carries no `Closes #N`")
and left the decision to the implementer. **Chosen: open, without
`ready-to-merge`, and closing an issue — where "closes an issue" is GitHub's own
`closingIssuesReferences`, not a keyword regex over the body.**

Why that predicate. The gate's purpose is *a review-bound pipeline gains nothing
from more PRs*, so a PR under review or waiting on CI belongs in the count — it
is the pipeline being review-bound. What does not belong is a PR no member of the
fleet will ever review: a controller-authored chore PR is left unlabelled for the
maintainer deliberately, so it can never drain, and it floors the backlog at a
depth the run cannot reduce. That is the third case the ticket's last comment
measured, and it is a different failure from an over-count — permanent, not
transient.

Why GitHub's linked set and not the body regex the brief suggested.
`docs/agents/issue-tracker.md` records the measurement that rules it out: PR
#1028's body carries `fixes #77` inside a code span and GitHub linked #439, the
issue it actually closed — a keyword regex over-reports and "never settles this
on its own". The linked set is what GitHub itself computed. Settled against this
repo:

```
gh pr list --state merged --limit 25 --json number,title,closingIssuesReferences \
  --jq '.[] | [(.number|tostring), ([.closingIssuesReferences[].number]|tostring), .title[0:40]] | @tsv'
```

→ every row with an empty set is a controller-authored `chore(metrics)` or
`review-and-fix` PR; every implementer PR carries its issue.

Cost: one extra field on the single `gh pr list` the tick already makes — no
extra call, and the overlap probe the brief ruled out is not run. Known error
direction, stated in the code: GitHub recomputes closing references lazily, so a
PR queried seconds after its create can read as closing nothing and drop out of
the backlog for one tick. That can under-hold the refill by one, and the next
tick sees it. The permanent floor it replaces did not self-correct.

The code and `SKILL.md` now state the same definition, and `SKILL.md` also states
*why* the third clause is there, so the next narrowing pass does not read it as a
stray filter.

## The other two rows

**Reviewers** dispatched against the review backlog. A reviewer slot holds a
fix-applier and a fix-applier applies findings, so a slot is dispatchable only
once a review has RETURNED. `--reviews-ready <n>` is that count. A boolean
`--review-running` — the ticket's own direction 1 — would not have been enough:
in the last comment's six-tick table, row 5 has no workflow in flight and is
still non-actionable, because the returned review already had its fix-applier
live. The backlog stays in the detail as an observation.

**Merge bot** dispatched on `ready-to-merge` alone. run-merge-bot's hold rule
reads `pr-overlap.mjs` against every lower open PR and reports
`held-behind-#<lower>` without touching a label, so the queue read is blind to
it. `--merge-holds <pr,pr|none>` carries the held numbers. A held number the
queue no longer contains subtracts nothing rather than refusing — that is the
shape left behind when a hold clears normally — and it is named back on the row
as `ignored=`, so a mistyped number, or the issue number passed where the PR
number belongs, cannot read as `none`.

Both new inputs are required, and `none` is a word rather than a blank because an
unset shell variable arrives as `""`:

```
node skills/fleet/scripts/fleet-tick.mjs --implementers 0 --reviewers 0 --merge-bots 0 --pool 0
```

→ `fleet-tick: --reviews-ready is required. …` — exit 2, no reconcile line.

## The class, covered and left

The defect class is *a row whose ACTION derives from a label proxy for state
labels cannot express*. Its members:

- `reviewers()` — DISPATCH off the review backlog. **Covered.**
- `mergeBot()` — DISPATCH off `ready-to-merge` alone. **Covered.**
- `implementers()` — HOLD off a backlog inflated by PRs no member will review.
  **Covered.**
- `implementers()`'s pool, and every `AT CAP` row — already caller-stated and
  already required. **Left: they conform.**
- `implementers()`'s supply, read from `candidates.mjs`. A proxy too, and an
  upper bound by the script's own comment, but not a label read for
  unexpressible state, and its error lands on `RE-SHORTLIST` — which the script
  argues is the safe direction. **Left deliberately**, and out of the ticket's
  scope.
- The merge bot's cap of 1 — an invariant, not a proxy. **Left.**

## What the change can newly refuse, and what it can newly break

New inputs have a false-positive class of their own, so each has a test feeding
it input it must ACCEPT: `--reviews-ready 1` still prints `DISPATCH 1`;
`--merge-holds none` and a partly-held queue still print `DISPATCH merge-bot`; a
stale held number does not suppress a live dispatch; every accepted
`--merge-holds` spelling is exercised beside the refused ones. The restructuring
also touched `counts()`, where the existing required-flag contract lives — the
missing-flag test now covers all six flags, and a mutant that gives
`--reviews-ready` a default reds it.

A gh row without `closingIssuesReferences` refuses rather than reading as
"closes nothing", which would silently empty the backlog on a full pipeline.

## Applied from the review

Three defects the review surfaced, each in the same class as the one this PR
fixes — a caller-stated input whose rejection said nothing.

- **A `--merge-holds` number the queue does not contain was accepted and then
  dropped in silence**, so a one-digit typo, or the issue number passed where
  the PR number belongs, printed the same `DISPATCH merge-bot` the flag exists
  to suppress. It still accepts — the merged-away hold is the ordinary case —
  and now names the rejected input as `ignored=` on the merge-bot detail.
- **The refusal text for a required flag came out of a string-key lookup with a
  silent fallback**, so a flag added to `OPTIONS` alone borrowed the live-counts
  rationale and a renamed `WHY` key printed `is required. undefined` — both at
  exit 2, both reading like a working refusal. Each live count now carries its
  own key, the fallback is gone, and an unpaired flag is a load-time throw.
- **The review-backlog comment documented only the transient under-count.** The
  predicate stands, but the trade is now written down: a PR that goes through
  the fleet's review cycle closing no issue is invisible to this count for its
  whole life, and that is deliberate — an over-count starves implementers for a
  whole run, an under-count only admits an extra PR into a review-bound
  pipeline.

Two refuter-checked simplifications came with them: `holds()` parses the digits
its guard already proved present instead of re-splitting the string, and the
review backlog re-states the sign-off predicate by name rather than re-deriving
"not queued" by identity membership over the array beside it. Both were verified
behavior-identical over generated input, not by reading.

The rest of the review's suggestion band was checked and refuted, or refused on
scope, and is recorded closed in #1165.

## Verification

- `node --test skills/fleet/scripts/*.test.mjs` — `tests 1621 / pass 1621 /
  fail 0`, from the worktree at the head this body describes.
- Which suites read `fleet-tick.mjs` was settled by replacing it with `MUTATED`
  and running the whole suite: `fleet-tick.test.mjs` and `arg.test.mjs`, nothing
  else. Both carry the new argv. Re-run at this head, it still names those two
  and no others. Comments in them referred to this file by line number and this
  diff staled them; they now name the function. The review found the same rot
  twice more in `candidates.test.mjs`, where two citations of
  `run-team/SKILL.md` by line number had drifted onto lost-message prose; both
  now name the **Candidate scan** step instead, which greps to one line.
- Every new test was checked against its own mutant, each killed by the tests
  that pin it and nothing else: reviewers reverted to the backlog → 2 reviewer
  tests; merge-bot's hold subtraction removed → the 2 held tests; the
  closing-issue filter removed → the chore-PR test; `--reviews-ready` given a
  default → the missing-flag test; the SKILL.md clause removed → exactly one
  prose test, out of the whole 1618.
- And a change they should NOT catch: rewording the reviewer HOLD's reason
  string leaves both suites entirely green, so they pin the suppressed ACTION
  rather than the prose. The tests added for the review findings were held to
  the same pair — removing the `ignored=` field reds the two assertions that
  pin it and nothing else, blinding the pairing check reds only its own test,
  and an inert reorder of the `WHY` table with the rejected-hold filter's
  parameter renamed leaves the file's whole suite green.

Out of scope, per the brief: the one-review-at-a-time rule, `pr-overlap.mjs`'s
hold logic, and running the overlap probe inside the tick.

Closes #590


feigi:
Review applied at `30952d8`.

**Applied:** `survived-0` (a rejected `--merge-holds` number is now named as `ignored=`), `survived-1` (every required flag pairs with its own rationale; an unpaired one is a load-time throw), the `refuted-0` residual (the review-backlog comment now names the permanent under-count and says the trade is deliberate), `unverified-4` (two rotted `run-team/SKILL.md:<line>` citations in `candidates.test.mjs` re-anchored to the **Candidate scan** step — the refuter found the second one), `unverified-5` and `unverified-6` (both refuter-verified behavior-identical). `unverified-1` and `unverified-2` restated the two survived findings and are closed by those applies.

**Deferred:** all in #1165 — filed `wontfix` and closed, per the record-issue convention. `unverified-8` refuted by mutation (reordering `LIVE` reds the suite loudly, so the positional dependency is not silent); `refuted-1`, `refuted-2` and `refuted-4` upheld as refuted; `unverified-7` refused on scope and below the claim bar. `unverified-0`/`refuted-3` and `unverified-3` were ruled not to file and are recorded there so a re-derivation lands on the measurement.

`node --test skills/fleet/scripts/*.test.mjs` → `tests 1621 / pass 1621 / fail 0`.

PR #735, opened 2026-08-21
docs(fleet): give ci-and-staleness.md the no-ci verdict it never named

feigi:
`ci-state.mjs` has three verdicts. `skills/fleet/skills/run-team/references/ci-and-staleness.md` — the file `run-team/SKILL.md` sends a reader to for what `ci-state` means — described two. Nothing in it was false; it was an omission in the document that positions itself as this area's complete evidence trail, so a reader landing there first concluded `ci-state` only ever says green or not-green.

Adds the `no-ci` subsection beside the four-way-binding section, plus the one clause the header's own topic list was missing.

## Every claim settled against the tree, not the ticket body

The ticket is itself a correction of an earlier wrong account, so its accuracy note was re-derived rather than copied. All settled against `skills/fleet/scripts/ci-state.mjs` and `ci-state.test.mjs` at `origin/main`:

| Claim in the diff | Settled by |
|---|---|
| third verdict alongside `green`/`not-green` | the `verdict` ternary in `ci-state.mjs` |
| only genuine absence is `no-ci` — no `.github/workflows/`, or a dir holding no workflow files | `discoverWorkflowFile`'s two `return null` paths (ENOENT; no YAML present) |
| every failure to *read* a workflow is exit 2, never `no-ci` | the five `die()` sites: unreadable dir (non-ENOENT), unresolvable repo root, two files sharing the name, unreadable candidate when nothing matched, YAML present under other names |
| `no-ci` alone exits 1; exit 0 only under `--declare-no-ci` | `gateSatisfied`, plus the two tests pinning exit 1 and exit 0 |
| the flag is caller-side and reading it back verifies nothing | the `reasons.push` in the no-ci branch — wording taken from `run-team/SKILL.md`'s finisher gate rather than invented afresh |
| no run list / run view read under `no-ci` | the `if (noCi)` branch skipping the whole run-binding block, and the test asserting neither is reached |

## Deliberately out of scope

- **#286** (prose pin over the no-ci finisher gate) is not implemented here. Note that this PR moves that ticket's site count from three to four: `references/ci-and-staleness.md` now states the gate too, worded to agree with the three existing sites rather than to introduce a second phrasing.
- The ticket's own "points at it twice" is stale — `run-team/SKILL.md` carries **four** `See references/ci-and-staleness.md` pointers at `origin/main`. Background detail in the ticket, not an acceptance criterion, so it is reported rather than written into the diff.

## Tests

`./agent-test .` — 1027 pass, 0 fail.

[...]

PR #1166, opened 2026-09-01
fix(board): say so when a ci-state payload is garbage rather than absent

### LINKED ISSUE #605: board.mjs:113: mapCi swallows an UNPARSEABLE ci-state payload into the same "unknown" as no-run-yet — the red-ci flag is suppressed for that PR in silence

feigi:
Sizing: light

`mapCi`'s `catch { return "unknown"; }` collapsed an **unparseable** ci-state payload into the same `"unknown"` that a still-running run and a not-yet-created run produce, with nothing on stderr. Three states, one output.

The path is reachable exactly where the loud path was deliberately declined. `runCiState` treats non-empty stdout at any exit but 2 as a real verdict — that is by design, and its comment says why — so a truncated pipe write or a warning line printed ahead of the JSON is handed straight to `mapCi`. `gather()`'s carry-forward does not cover it either: that arm needs a `null` return, and this payload is not null, so a PR that read `red` last tick reverts to `unknown` and its red-ci flag, the top of the attention strip, stays down.

## The fix is additive, as the ticket requires

The return value stays `"unknown"` — a false red is worse than no verdict, and the regression gate pins it. Only the silence changes: the third state now leaves through stderr.

- **Warn-once keyed on the PR number.** `serve` rebuilds every ~15s and calls `mapCi` once per PR per tick, so a broken payload is broken on every tick. Per PR rather than globally, because a global flag would let the first broken PR mask every later one for the rest of the run — the same silence this ends.
- **`mapCi` takes the PR as a second argument, for the message alone.** The ticket left the placement open. Caller-side warning would mean `gather()` re-parsing the payload to learn it was garbage — two places that must agree on "unparseable". This keeps one parse.
- **`#603` has not landed** — `git show HEAD:skills/fleet/scripts/board.mjs | grep -nE 'function warnOnce'` returns no match at this commit, so no helper exists to call yet — and this is a hand-rolled gate of the same shape as the sidecar and transcript ones, and a caller for that helper when it arrives. (The identifier `warnOnce` *does* appear at this commit, in the comment that names #603 as its future home. The grep above is scoped to a definition for that reason.)

## Absent, empty, garbage — three different silences, and only two of them are the same

The guard ahead of the parse tests `ciJson == null`, not falsiness, and that distinction is the second half of this ticket.

- A **null** payload is a failed read `runCiState` has already reported on stderr. Warning again would report one failure twice, so it returns `"unknown"` silently.
- An **empty** payload is not that case. `runCiState` returns the child's stdout unconditionally at exit 0, emptiness untested — `sed -n '/^function runCiState/,/^}/p' skills/fleet/scripts/board.mjs` shows the `out.trim()` gate living in the catch, where it governs non-zero exits only — so a lost stdout write on a green verdict arrives as `""` with nothing yet said about it. A `!ciJson` guard swallows that as though it had been reported. It now falls through to the parse, where `JSON.parse("")` throws and the warn names the PR.
- A **garbage** payload takes the same arm, which is where this ticket started.

The return value is `"unknown"` for all three, so the empty case is additive in exactly the way the ticket requires: only its silence changed.

## Both halves are pinned

- An unparseable payload → `unknown` **plus** one stderr line naming the PR.
- An empty payload → `unknown` **plus** its own stderr line. That test is the only thing separating the two guards: its neighbours on either side pass under `!ciJson` and under `== null` alike.
- A legitimately unknown state — `status: null`, `in_progress`, `no-ci` — stays **silent**. Without that, every PR awaiting its first run prints a line every tick.
- An *absent* payload (`null`) stays silent too, for the reason above.
- Warn-once holds across ticks, and a second broken PR still gets its own line.
- One test drives the real `gather()` and matches the PR number on stderr. That wiring is invisible in-process: left as `mapCi(out)`, every other test stays green while the board warns about PR `undefined`.

Mutation-checked at this head — dropping `p.number` at the call site, removing the warn-once gate, warning on an absent payload, dropping the PR from the message, and reverting the null guard to `if (!ciJson)` each fail the suite, and the last of those fails the empty-payload test and no other. A negative control that reworded the warn text and swapped the green/not-green branch order stayed fully green, so the pins are not coupled to wording or branch order. Baseline restores clean.

## Also here

The `runCiState` comment glossed exit 1 as solely "not-green". `ci-state.mjs:604` emits three verdicts and `:622-623` puts `no-ci` on exit 1 unless `--declare-no-ci` is passed, which this call site omits. Prose only, deferred from PR #735's review onto this ticket, which quotes the block verbatim as its own rationale.

`mapCi`'s comment also credited "every caller and the regression gate" with pinning `"unknown"`. Mutating every `"unknown"` return inside `mapCi` puts every resulting failure in `board.test.mjs`, while the production caller propagates the mutant without complaint — it consumes the value, it does not pin it. The clause credits the regression gate alone.

## Deliberately left

`mapCi`'s trailing `return "unknown"` — a payload that parses, reports `status: "completed"`, and carries a verdict string `mapCi` has no branch for — stays silent. It has no producer today (a real `no-ci` carries `status: null`), and warning there would fire on the documented no-ci shape if `ci-state` ever set a status on it. Not part of this ticket's class, which is corruption of the payload itself.

A payload that parses to JSON `null` throws a `TypeError` out of `mapCi` rather than returning a verdict: the `d.status` dereference sits outside the `try`. That is pre-existing — the line is untouched by this diff — and changing it would not be additive, so it is filed as #1170 rather than fixed here.

The review's other findings are recorded in #1171, closed; the warn-once helper is #603.

Verified at this commit: `node --test skills/fleet/scripts/*.test.mjs` — every test passes, none fail, none skipped, exit 0.

Closes #605


feigi:
**Review deferrals filed.**

- **#1170** (`ready-for-agent`) — `mapCi` throws a `TypeError` on a payload that parses to JSON `null`, aborting the whole build tick rather than one CI cell. Confirmed by direct execution, pre-existing and untouched by this diff, and fixing it would not be additive, so it is deferred rather than applied. Two refuters split on reachability, not on the throw; both votes are in the body.
- **#1171** (`wontfix`, closed) — the durable record of what the review checked and did not apply: four findings refuted by measurement, one wrong on its scope premise, two below the claim bar, one confirmation. Each carries what was actually run, so a later run re-deriving one lands on the measurement instead of re-litigating it.
- **#603** — commented rather than filed: this PR's `warnedCiParse` is another caller for that ticket's `warnOnce(key, msg)` helper, and the comment at its declaration already says so.

Two findings survived refutation and are applied in the head commit: the `!ciJson` guard narrowed to `ciJson == null` so an empty payload earns a warn instead of borrowing the absent one's silence (with a test that no other test can stand in for), and the `"unknown" is what every caller and the regression gate pin` clause corrected to credit the regression gate alone.

PR #1167, opened 2026-09-01
fix(fleet): a worktree path git cannot round-trip is refused, not truncated — and a dangling link is not absent (#551, #725)

### LINKED ISSUE #551: fleet scripts: reap.sh, release-ticket.sh and worktree-audit.sh still split a worktree record on an embedded newline — the defect #185 fixed in inflight.sh
`inflight.sh` probe 3 was repaired under #185 by reading the listing with `git worktree list --porcelain -z` and swapping the separators with `tr '\n\000' '\001\n'` before awk sees it. Three sibling scripts parse the same listing with the same `substr($0,10)` shape and still carry the defect: a worktree path may legally contain a newline (APFS and ext4 both allow it), the plain porcelain terminates every attribute with a newline, so one record splits into two and the path truncates at the newline.

## Measured

Real repo, real linked worktree at `.../wt/fix-33<LF>slug` on branch `fix/33-slug`, git 2.50.1. Each parse run verbatim as the script runs it:

```
=== reap.sh:88-89 ===
wt=[.../wt/fix-33]

=== release-ticket.sh:117 (+ its awk at :127) ===
wt=[.../wt/fix-33]
listed=2   (expected 2 — the incompleteness guard at :122-124 does NOT fire)

=== worktree-audit.sh:58 ===
.../repo          refs/heads/main
.../wt/fix-33     refs/heads/fix/33-slug
```

All three hand a downstream consumer `.../wt/fix-33`, a path not on disk, and none of them refuses. `release-ticket.sh`'s own `listed - 1 == registered` cross-check is blind to it here: the orphaned second line is `slug`, which does not start with `worktree `, so the count still agrees.

## Sites

- `skills/fleet/scripts/reap.sh:88`
- `skills/fleet/scripts/release-ticket.sh:117` (path consumed by the awk at `:127`)
- `skills/fleet/scripts/worktree-audit.sh:58`

The shape landed in `inflight.sh` is `--porcelain -z` into a temp file, then `tr '\n\000' '\001\n'`, then `gsub(/\001/, "/", b)` in the matcher so the substituted byte counts as a path separator for matching while the evidence keeps the raw bytes. Whether that whole shape ports to all three, or whether each needs less, is not settled here:

- `worktree-audit.sh:58` pipes into a `while IFS=<TAB> read -r` loop, so `-z` changes the loop's separator problem as well as awk's.
- `reap.sh:88` matches on the `branch` line rather than the path, so only the reported path is wrong there, not the match.
- `release-ticket.sh:117` also needs the count question answered, since its cross-check does not catch this.

Two portability facts already measured under #185 and worth carrying: `awk -v RS='\0'` is NOT an option — macOS BWK awk 20200816 reports ONE record for a listing of any length, all three spellings; and `git worktree list --porcelain -z` requires **git 2.36.0** (2022-04-18), a floor this repo already exceeds via `no-undo-audit.sh`'s `git merge-tree --write-tree --name-only -z` (2.38/2.39).

## Not a duplicate of #395

#395 tracks `release-ticket.sh`'s divergence from the `inflight.sh` cross-check for **three other** defects PR #384 fixed (stray registry directory, `grep -c … || true`, and the recount). The newline record split is a fourth, and it is the one the other two scripts share. Fold or keep separate as triage decides.

Deferred from PR #548 review — `dimension: correctness`, severity `suggestion`, `skills/fleet/scripts/release-ticket.sh:117`. Deferred because it is outside #185's ticket scope (`inflight.sh` probe 3) and the approach above is unsettled.

### LINKED ISSUE #725: gone() reads a dangling symlink as established-absent — triplicated in worktree-audit.sh, reap.sh, release-ticket.sh, and only the last compensates

feigi:
Sizing: heavy

Closes #551
Closes #725

Both tickets in one branch, as both briefs' **COLLISION** notes require: they
rewrite the same reads in the same three scripts, and splitting them means
resolving a three-file conflict after the fact.

**This branch is a recovery.** A previous run's member wrote it and its session
ended before it was finished; the controller preserved 244 lines of uncommitted
regression tests as a `wip` commit so they could not be lost. That work is
carried forward, not redone — see *Route taken* below.

## The two defects

**#551** — a worktree path may legally hold a newline, and `git worktree list
--porcelain` terminates every attribute with one, so a record splits in two and
`substr($0,10)` truncates the path. Every consumer downstream is handed a path
that is not on disk, and nothing refuses. Re-measured on today's tree, git 2.50.1
(Apple Git-155):

    git worktree add -b fix/33-slug "../wt/$(printf 'fix-33\nslug')"
    git worktree list --porcelain |
      awk '/^worktree /{w=substr($0,10)} /^branch /{print "["w"]"}'
    -> [.../wt/fix-33]        and [ -d .../wt/fix-33 ] is false

**#725** — `gone()` decides *established-absent* as against *could not look*.
Every `test` primary except `-L` stats, so on a dangling symlink `-e` is false
while `-L` is true and the predicate answered established-absent for a path
`git worktree add` treats as occupied. It was triplicated, and the one copy that
compensated did so at a call site rather than in the predicate.

## What changed

New `skills/fleet/scripts/worktree.sh`, a sourced helper beside the existing
`json.sh` seam — #725's brief rules this remedy (option 3) and rules it out of
`json.sh`, which is scoped to escaping:

- `wt_listing` reads `--porcelain -z` into a temp file and swaps separators with
  `tr '\n\000' '\001\n'`, the shape #185 landed in `inflight.sh`.
- `nl_path` is the single test for the substituted byte, so a caller refuses a
  path it cannot hand back to the filesystem rather than truncating it.
- `gone` gains `[ ! -L "$look" ]` in its **walk** condition, which also covers a
  dangling symlink standing in for an *ancestor* — a fourth clause on the result
  would not have.
- `gone` also refuses a substituted path outright. Its walk asks about that
  path rather than the file git named, and the substituted path is reliably not
  there, so the predicate returned *established absent* — the confident answer —
  for the one input it had no business answering about. Measured:
  `release-ticket.sh` emitted its newline refusal and then a second blocker on
  the same path saying the directory "is gone" and naming `git worktree prune`
  as the remedy, for a worktree on disk holding an uncommitted file; `git
  worktree prune -v` against it is rc 0, no output, entry still registered.
  Refused in the predicate rather than at the call sites, for the reason the
  rest of this section gives.

The three scripts source it and drop their own copies. `gone()` now has one
definition in the tree.

`worktree-audit.sh` carried the same truncation one delimiter over, and the
branch's own rewrite is where it survived. The awk emitted `path<TAB>branch`
into `IFS=<TAB> read -r wt br`, so a path holding a TAB split at it — and
`wt_listing` substitutes NUL and newline only, so `nl_path` is false and nothing
refused. A present, clean worktree at `wt/fix-44<TAB>slug` was reported `MISSING
on disk` at a truncated path with its tail swallowed into the branch field,
which is the same wrong answer the newline case gave the fleet controller. Fixed
by emitting the path **last**, so `read` assigns the whole remainder of the line
to it whatever bytes it holds. The branch cannot take that slot and that
asymmetry is why this works rather than an accident of ordering: a ref name
rejects a tab outright. Widening `nl_path` to a second byte class was the
alternative and is strictly more to keep true.

## Enumeration — the named sites were not all the sites

Both tickets name one parse site per script. Re-enumerated against today's
`origin/main` rather than against the branch's 51-commit-old assumptions, the
scripts read the listing at more sites than that — including `reap.sh`'s two
post-removal registry re-reads and the fresh listing `release_outcome` judges a
halted release from, which no ticket mentions. **All are covered**; no raw read
is deliberately left:

    for f in $(git ls-files '*.sh'); do grep -n 'git worktree list' "$f" \
      | grep -vE ':[[:space:]]*#' | grep -v '"[^"]*git worktree list'; done
    -> only worktree.sh and inflight.sh, both `--porcelain -z`

`inflight.sh` is **already fixed** on `origin/main` by #185 and is named
out of scope by #551 — reported, not re-fixed. `no-undo-audit.sh`'s
`[ -d "$wt" ] || die` is explicitly out of scope in #725's brief: it already
fails closed with only a slightly wrong diagnosis.

So the sweep pins the **property**, not the sites: no tracked shell script
outside those two invokes `git worktree list` at all. Pinning only the named
sites is how a fix ships without closing its own ticket.

The sweep in the tree is stricter than the shell approximation above, because
the first version of it pinned one way of *typing* the read rather than the read
itself. Three spellings of a raw listing read walked straight through it, each
verified by reintroducing it and watching the suite stay green: `git -C "$dir"
worktree list` (the interposed option breaks the literal phrase, and `git -C` is
the dominant git spelling in these scripts); `raw=$(cd "$HOME" && git worktree
list …)` (the exclusion asked whether a quote appeared *before* the phrase,
which is a different question from whether the phrase is *inside* a string and
agreed with it only by luck); and the three words split across a
backslash-newline, valid POSIX sh and invisible to any scan over physical lines.
The gate now folds continuations, allows options between `git` and its
subcommand, and scans quote parity. The companion awk scan matched a literal
`\0` and missed `sep=$(printf '\0'); awk -v RS="$sep"`, which reintroduces the
identical one-record read — measured on BWK awk 20200816, NR 1 for both
spellings against 3 through `tr` — so it now flags any `RS=`.

The same shape held for #725's redefinition guard, which required the opening
line to be byte-for-byte `gone() {`: the pre-PR body reintroduced under `gone()
{ # local override` left every test green, and a redefinition arrives with a
comment excusing it far more plausibly than without one.

Every widened guard was verified in both directions — each missed spelling now
red, each already-caught spelling still red, and a banner, a diagnostic string,
a single-quoted banner, a full-line comment, an unrelated `git -C` invocation, a
comment naming `gone() {` and an extra `gone` call all still green. A guard that
only refuses has not been shown to discriminate.

A reader that refuses a path it cannot name has a false-positive class, so it is
pinned in both directions. `worktree-audit.test.mjs` drives an **ordinary**
worktree and requires the reported output byte-identical across the change
(#551's third acceptance criterion), and the pre-existing `reap.test.mjs` and
`release-ticket.test.mjs` suites drive real worktrees at ordinary paths
throughout. The sweep also asserts the two allowed readers *do* still read the
listing — without that it would pass just as happily over a tree where nothing
reads it at all.

Two **new** false refusals were found this way rather than reasoned about, and
fixed. `release-ticket.sh`'s top-level guard paired `[ ! -e "$wt" ]` with a
`gone` that had just stopped following links, so a dangling link died at exit 2
with no receipt. It now composes `! occupied && ! gone`, the tri-valued pairing
`gone`'s own contract prescribes.

That same guard then took the same trap by a second route, which is why the
`gone` change was traced through every caller rather than only the site that
reported it: with `gone` refusing a substituted path, the pairing reads that
refusal as "cannot tell" and dies — downgrading a receipt-carrying `NOT
released` verdict to an exit-2 die that says nothing the receipt did not, for a
path the newline blocker above has already answered in the operator's own terms.
`! nl_path` restores the verdict. `reap.sh` and `worktree-audit.sh` need no such
exemption: both refuse a substituted path earlier and `continue` before any
`gone` call, which was established by reading each caller rather than assumed
from the one that surfaced the defect.

## What routing through a shared helper newly does

Every `wt_listing` call sits in a condition context (`if !`, `||`), so `set -e`
never fires on it; `wt_listing` ends in an explicit `return 0` so a temp file
that could not be unlinked is never reported as a failed listing. `wt_listing`
writes a shared `$wt_list`, and `release_outcome` needs a *fresh* listing while
the guards above it read the pre-mutation capture — so it saves and restores
`$wt_list`. All four of its verdicts are exercised by the suite.

That save/restore is load-bearing rather than defensive, and the reason is a
second defect this routing introduced. `wt_listing` captures git's stderr into
`$wt_err`, and `release_outcome` dropped it — so the one halt whose headline
degrades to "what landed could not be measured" was the one halt reaching the
operator with no cause attached, on stderr or in the receipt. The halt does name
why the *release* stopped (`git worktree remove`'s own refusal text, both
places); what was lost is why the follow-up **measurement** failed, which is
exactly the case where nothing else is available to act on. The `$( )` around
the call was the mechanism — the capture died with the subshell — so
`release_outcome` assigns `$wt_outcome` and `$wt_why` instead of echoing a
state, and `halt` appends the cause to the Indeterminate report. With the
subshell gone there is nothing else keeping the two listings apart.

Pinned on **both** channels in the existing Indeterminate case, because the
receipt is the only one a caller that never sees stderr can read.
Mutation-verified in both directions: dropping the cause fails the stderr
assertion, and keeping it while removing the newline-flattening stays green — so
the pin is that the cause arrives, not how it is plumbed.

`reap.sh`'s branch sweep keeps `wt_listing || :` deliberately, preserving the
swallow #622 documents: an unreadable listing leaves `$wt` empty and a
worktreeless `[gone]` branch is still reaped.

## Acceptance criteria

None is contradicted by the current tree. Two are answered differently from the
literal wording, both in the direction the criterion intended:

- *"`release-ticket.sh`'s cross-check either catches the newline case or is
  documented as not covering it"* — documented. It counts records against
  registry entries; the orphan continuation line does not begin `worktree `, so
  it moves no count over this listing. `listed - 1 == registered` **agrees**
  while the path is truncated (measured: `listed` 2, `registered` 1). The check
  is structurally blind, not merely wrong.
- *"`gone-walk.test.mjs`'s caller inventory states the correct count"* — the
  inventory is now **derived from the tree** instead of stated. The stated count
  had already drifted once, which is what #725's own review caught; an assertion
  cannot.

## Route taken

`git rebase origin/main` over 51 commits, and it applied **cleanly** — so the
original branch is carried forward, not superseded. A clean rebase is not
evidence of a correct one, so everything above was re-derived against today's
tree rather than trusted: the site enumeration, the reproduction, and every
factual claim in the commit bodies.

That re-derivation found real drift, fixed here:

- Prose inherited from `origin/main` in `release-ticket.sh` argued from "the
  three copies of `gone()` are pinned byte-identical". This branch is what made
  that false. The conclusion survives and gets *stronger* — one shared
  definition means a third value added for one caller lands in all of them — so
  the reason is restated rather than the rule relaxed.
- Every per-file test count asserted in the preserved commit bodies had drifted
  (118→119, 68→70, 18→20), and one cited a fixture figure whose shape was not
  recorded and could not be re-run. All re-measured at the commits that ship
  them, or replaced with the property they were cited for.
- The controller's `wip` preservation commit is reworded; its content is intact.

One existing test was removed, because its subject stopped existing:
`test("all three copies of gone() are still byte-identical")`. Three stronger
tests replace it — one definition, every caller reaching it, inventory derived.
That is still the only removed test.

Two test files carry deleted lines, and both are comment or guard rewrites
rather than lost coverage — settled from the diff rather than asserted:

    git diff $(git merge-base origin/main HEAD)..HEAD --numstat -- '*.test.mjs'

`gone-walk.test.mjs` (the removed test above, plus the widened redefinition
guard) and `printf-die-sweep.test.mjs`, which this branch does not otherwise
touch. It is in scope only because this branch **falsified its comment**: the
comment gave two counts as exact re-runnable output and a new `keep` call site
made both wrong, silently, since nothing goes red when a number in a comment
stops being true. The numbers are not corrected — a corrected count rots on the
next commit the same way, and this comment's own history is the argument, having
already shipped one wrong count and a list of sites cited by line numbers that
had drifted off every one of them. It states the property they were cited for
instead: `keep()`'s reason is the one line reap.sh's refusals converge on, and a
worktree path reaches it from more than one of them — which is why the escaping
belongs in `keep()` and not at any call site. The re-derivation commands stay,
without answers attached.

## Verification

    node --test skills/fleet/scripts/*.test.mjs
    -> tests 1631, pass 1631, fail 0

    .github/scripts/check-tracked.sh '*.sh' shellcheck -x -S warning
    -> clean, 18 files

Both re-run at the commit this ships, not carried over from an earlier one.

The suite total is unchanged from the previous head, and that is the expected
reading rather than a sign nothing landed: the new coverage is assertions added
to an existing case and guards widened in place, neither of which creates a
`test()`. What the guards now catch is established by mutation in both
directions above, which is the only evidence a count could not carry anyway.
Deleted lines in test files are accounted for under *Route taken*.

Per-defect coverage, each driving a real linked worktree rather than a synthetic
listing: `worktree-audit.sh` refuses a newline path and reports a dangling link
as occupied rather than `MISSING on disk`; `reap.sh` **keeps** a `[gone]` branch
at either kind of path rather than reaping it; `release-ticket.sh` blocks rather
than releasing. The `reap.sh` direction matters most — it is the script that
deletes, and both defects made it confident about a path it could not name.


feigi:
Review pass applied at c833e3d — 18 findings adjudicated.

**Applied (6).** `S0` worktree-audit.sh's tab truncation (survived 2-0); `S1` `gone()` refusing a substituted path, plus the caller-side exemption that trace turned up in release-ticket.sh (survived 2-0); `S2` printf-die-sweep.test.mjs's falsified comment, restated as the property rather than re-counted (survived 2-0); `R0` narrowed to *an Indeterminate `release_outcome` result reaches the operator with no cause attached* and re-checked under that wording (refuted=false); `U0`/`U2`/`U3`/`U4` the four test-guard blind spots, each reproduced by a refuter in both directions (refuted=false).

**Deferred (9), all in #1173** — closed `wontfix`, the durable record. `U1`, `U5`, `U7`, `U11` refuted by measurement; `U8`, `U9`, `U10`, `U12`, `U13` below the claim bar per `docs/adr/0002-filing-second-bar-worth-a-claim.md`, recorded with re-open triggers. No entry was dropped and none warranted an open issue: nothing left over is an unconfirmed defect or a torn call.

`U8` is worth one note — it called `release_outcome`'s `$wt_list` save/restore unobservable, correctly, because a command substitution wrapped the only call site. Fixing `R0` removed that subshell, so the save/restore is now load-bearing rather than dead. The finding was right and the resolution went the other way.

`node --test skills/fleet/scripts/*.test.mjs` → tests 1631, pass 1631, fail 0. `check-tracked.sh '*.sh' shellcheck -x -S warning` → clean, 18 files. Both re-run at c833e3d.

feigi:
**Correction to the comment above, and one further commit.**

That comment's applied list omitted **U6**, which was applied, not deferred. The full split is:

**Applied (9 ids):** S0, S1, S2, R0, U0, U2, U3, U4, **U6**.
**Deferred (9 ids) → #1173:** U1, U5, U7, U8, U9, U10, U11, U12, U13. That list and the issue are unchanged.

U6 flagged worktree.sh's justification for the `\001` substitution byte — it claimed *"nothing else in the listing can hold one"* and backed it with a roster of git's attribute keywords. A refuter settled it against a real fixture rather than a grep: `git worktree lock --reason "$(printf 'why\001here')"` is accepted unsanitized and the listing hands the byte straight back (`od -c` → `l o c k e d   w h y 001 h e r e`), so the universal was false. `prunable`'s reason is git's own fixed prose, so the hole is `locked` only.

The first correction of it, in 8383400, **shipped its own wrong claim** — it said no field any caller parses can hold a `\001`, when a path is such a field and may hold one natively, which `nl_path`'s own comment three lines below says outright. Dropping the word "else" from the original quantifier is what broke it. Fixed in a01af92, which drops the roster rather than extending it — a corrected roster is the same completeness claim with more items — and states the property instead: no part of the listing's *structure* carries a control byte, so a `\001` reaching `$wt_list` can never be read as a delimiter or an attribute. Both non-claims are now explicit.

Head is now **a01af92**. `node --test skills/fleet/scripts/*.test.mjs` → tests 1631, pass 1631, fail 0. `check-tracked.sh '*.sh' shellcheck -x -S warning` → clean, 18 files. Comment-only change; no test pinned the old wording.

PR #1172, opened 2026-09-01
fix(claim-ticket): refuse a contaminated ndeps capture by shape

### LINKED ISSUE #752: claim-ticket.sh:71 — node stderr merges into $ndeps, so an exotic env reports chatter as a dependency count (and dropping the 2>&1 breaks a tested path)
**Dimension:** silent-failure (from PR #750's review, `refuted[0]`)
**Deferred from PR #750 review** — 1-1 split, upheld as refuted on *severity and scope*, not on mechanism. The defect is confirmed; it is filed because the remedy is open, not because anyone doubts it reproduces.

## The defect

`skills/fleet/scripts/claim-ticket.sh:71`

```sh
elif ! ndeps=$(printf '%s' "$pkg" | node -e '…console.log(String(…reduce(…)))' 2>&1); then
  die "could not read origin/main:package.json — $ndeps"
elif [ "$ndeps" = 0 ]; then install="true"
else die "origin/main declares $ndeps dependencies but has no lockfile — refusing to guess an install command"
fi
```

The `2>&1` merges node's stderr into `$ndeps` on the **success** path as well as the failure path. Any node stderr chatter therefore becomes the "dependency count", and a manifest declaring **zero** dependencies is refused with a message naming a count that does not exist.

Reproduced by both reviewers independently, against a fixture (`git init`, `package.json` = `{}`, `t.test.mjs`, `origin/main` ref):

```
$ NODE_DEBUG=module sh .../claim-ticket.sh 42 slug fix
claim-ticket: origin/main declares MODULE 53643: Module._load REQUEST module parent: <eval>
MODULE 53643: Module._load checking cache for module false
… 0 dependencies but has no lockfile — refusing to guess an install command
exit=2

$ NODE_OPTIONS=--inspect=0 sh .../claim-ticket.sh 42 slug fix
claim-ticket: origin/main declares Debugger listening on ws://127.0.0.1:51696/… 0 dependencies but has no lockfile…
exit=2
```

Controls on the same fixture: no env → `lockfile → install: true`, exit 0. `FORCE_COLOR=1` → `install: true`, exit 0 (PR #750's own fix, working as advertised).

## Why it was deferred rather than applied

- **Exotic trigger.** It needs `NODE_DEBUG` or `NODE_OPTIONS=--inspect`. Unlike `FORCE_COLOR` — which any developer terminal or CI sets routinely, and which is why PR #750 exists — no normal caller sets these.
- **Fails loud.** Exit 2 with the chatter printed in the refusal. Nobody proceeds on a wrong install command and no worktree is built, so it is not the silent-plausible-refusal shape PR #750 fixed.
- **Environment-wide, not this line.** Under `NODE_DEBUG` every node stderr in the fleet is polluted; that is an environment condition rather than something this PR introduced or mis-attributed.

## ⚠️ The remedy trap — read this before "fixing" it

**Do not simply drop the `2>&1`.** The merge is deliberate and load-bearing. `$(...)` captures stdout only, so the child's failure reason travels on **stderr**; without the merge, `die "could not read origin/main:package.json — $ndeps"` prints an **empty reason**. That path has a passing test:

> `install: an unparseable manifest refuses, and says so` — `skills/fleet/scripts/claim-ticket.test.mjs`

The same rationale is written out at `claim-ticket.sh:91-99` for the sibling `derive-testcmd.sh` capture at `:101`, which merges for exactly this reason. Removing the merge trades a loud, exotic-env cosmetic defect for a **silent loss of the real error text** — a strictly worse bug.

**Root of the difficulty:** this capture is *dual-purpose*. On the failure path its value is a human-readable reason; on the success path it is a number that gets compared. That is what makes it harder than the sibling site fixed in PR #750's follow-up commit (`claim-ticket.sh:148`), where the captured value is **only** data and the merge could simply go.

## Suggested direction (unverified — the remedy is open)

Validate the captured value instead of removing the merge, so a contaminated capture is refused by shape rather than interpreted:

```sh
elif ! ndeps=$(printf '%s' "$pkg" | node -e '…' 2>&1); then
  die "could not read origin/main:package.json — $ndeps"
elif case $ndeps in ''|*[!0-9]*) true ;; *) false ;; esac; then
  die "could not read origin/main:package.json — unexpected output: $ndeps"
elif [ "$ndeps" = 0 ]; then install="true"
```

Reported as verified legal under `/bin/sh` (`case` as an `elif` condition; fires on the contaminated string, falls through on a bare `0`). This also subsumes the `String()` fix's class — an SGR-wrapped `0` is non-numeric — so it is a root-cause guard rather than a second patch beside it. Worth a sibling test asserting the refusal names the unexpected output rather than a dependency count.

## Related

- #625 — `reap.sh:219,245`, the same family from the other direction (`2>/dev/null` discards git's diagnosis), and already carrying the mirror-image warning that "`2>&1` is not a safe fix there".
- #27 (closed) — the dependency-free exception's semantics; the current code is its fix, and it does not touch the stderr merge.
- PR #750 commit `3ce8ca5` fixed the *data-only* sibling at `claim-ticket.sh:148`, where dropping the merge was safe; see that commit message for why the two sites differ.

feigi:
## Summary

`claim-ticket.sh`'s `ndeps` capture in the lockfile/install derivation does `node -e '…' 2>&1`, merging node's stderr into `$ndeps` on the success path as well as the failure path. Stray chatter then got read as part of the "dependency count" — a manifest declaring zero dependencies could be refused with a message naming a count that was never really there.

The hazard is **not** an enumerable set of env vars. #752 arrived through `NODE_DEBUG` and `NODE_OPTIONS=--inspect`, but the class is anything that writes to node's stderr and still exits 0 — a `--require` preload or a version-manager shim on `PATH` as much as a debug flag.

The `2>&1` is load-bearing and stays: `$(...)` only captures stdout, so without the merge the failure path's `die "could not read origin/main:package.json — $ndeps"` prints an empty reason (pinned by the existing test `install: an unparseable manifest refuses, and says so`).

Fix: guard by shape instead of touching the merge. A clean count is always bare digits, so anything else — empty, or holding a non-digit — refuses up front with the contaminated text quoted, before it reaches the `= 0` comparison or the `declares $ndeps dependencies` refusal. This also subsumes the existing `String()`-wrap regression fix's class, since an SGR-wrapped digit is non-numeric too.

## What the guard buys, and what it does not

`[!0-9]` is a **necessary-condition** filter: it rejects only a capture holding a non-digit byte. So the guard delivers **no false refusals** — measured across the manifest matrix, the success path emits nothing but digits — and that is the whole of what it establishes. It does **not** make a contaminated capture trustworthy: digit-only chatter ending without a newline runs together with the count and still reads as one.

```
node stubbed to `printf 1 >&2; printf "0\n"`, dependency-free manifest, no lockfile:
  claim-ticket: origin/main declares 10 dependencies but has no lockfile   (rc 2)
same stub with a trailing newline on the stderr write:  correctly refused (newline is a non-digit)
control, `printf x >&2`:                                 unexpected output: x0
```

The comment at the capture site now says this rather than claiming the stronger thing. The guard itself is unchanged and that residual is filed as **#1174** — a loud refusal in every case, never a wrong install command, so it is a ceiling of shape validation rather than a live tree-corrupting bug. The obvious remedy (redirecting node's stderr to a temp file) was measured and refuted: it cannot retire the shape guard, since a `node` shim writing to **stdout** walks straight past stream separation, and it inverts this PR's own regression test. Recorded in #1176.

## Class enumeration (per the ticket)

- **Fixed** — the `ndeps` capture, where node's stderr reaches the captured variable directly. (It is not the only dual-purpose `2>&1` capture in this file — the `derive-testcmd.sh` capture is one too, and its own entry says what separates them. The earlier claim that it was the only one is corrected there.)
- **Corrected, not fixed — the `derive-testcmd.sh` capture.** An earlier revision of this PR claimed it was *not a member* of the class, on the grounds that `derive-testcmd.sh`'s own success path never reads its captured `$pkgerr`. **That reasoning is about the wrong variable and the conclusion is wrong.** `claim-ticket.sh`'s `$testcmd` *is* consumed on the success path — echoed as `test entrypoint → …`, compared against `"node --test"`, and `exec`'d in the emitted runner — so the capture is the same construct and carries no guard. What defends it is its callee, which folds its own node's stderr into `$pkgerr` rather than emitting it:

  ```
  $ NODE_DEBUG=module sh skills/fleet/scripts/derive-testcmd.sh . origin/main 2>&1 1>/dev/null
  (empty)
  ```

  Adequate, but contingent on a sibling file's behavior, which nothing pins. The comment now states that instead of pointing at the sibling and calling it exempt. Guarding it is out of #752's scope and is filed as **#1175**.
- **Left alone — the `tmpl_stamp=$(cksum "$0")` capture** (the ticket's "already-fixed sibling" from PR #750's follow-up): confirmed no `2>&1` present — data-only capture, out of this class entirely.

## Test coverage

- `install: non-numeric node stderr merged via 2>&1 refuses by shape, not misread as a dependency count` — stubs `node` on `PATH` to emit chatter plus a bare `0` (a deterministic stand-in for `NODE_DEBUG`'s actual output, which is unpinned across node versions), and asserts the refusal names "unexpected output" rather than a dependency count. Verified to fail without the fix, reproducing the exact bug text `declares MODULE 12345: chatter\n0 dependencies but has no lockfile`. The name says *non-numeric* because that is what the fixture pins; the digit-only case the guard admits stays visibly uncovered rather than papered over by a name that claims the class is closed.
- `install: a node that prints nothing refuses by shape, not as a dependency count` — covers the guard's `''` arm, which mutation showed had no coverage anywhere in the file. Both refusals exit 2, so it asserts the message, not the status.

  ```
  $ node --test skills/fleet/scripts/claim-ticket.test.mjs
  shipped                            -> tests 81 / pass 81 / fail 0
  dropping the '' arm                -> tests 81 / pass 80 / fail 1  (the new test alone)
  reordering the arms (a no-op)      -> tests 81 / pass 81 / fail 0
  ```

- Accept path — a clean environment with a zero-dependency manifest still reaches `install: true`, exit 0 — was already pinned by the existing install matrix rows (no manifest at all, empty dependencies object, null dependencies, all four fields empty, empty workspaces array) plus the `FORCE_COLOR'd caller` regression test. Unaffected, still green.
- Full suite:

  ```
  tests 1623 / pass 1623 / fail 0
  ```

Sizing: light

Closes #752


feigi:
## Review deferrals filed

Nine findings adjudicated: four from the review workflow's `unverified` band (`U0`–`U3`, `refutersDispatched: 0` — the suggestion band gets no refuters by policy) and five from a controller-run `comments` re-analysis (`CC0`–`CC4`). The workflow's size tier never assigned the `comments` dimension to this PR even though the `.sh` half of the diff is majority prose — that trim is #218.

Each in-scope finding was handed to one refuter, biased to refuse and required to verify by running something.

| # | State | Dimension |
| --- | --- | --- |
| **#1174** | `ready-for-agent` — confirmed defect, remedy open | `correctness` / `comments` (`U0`, `CC0`) |
| **#1175** | `ready-for-agent` — confirmed defect, out of #752's scope | `comments` (`CC4`) |
| **#1176** | `wontfix`, closed — durable record | `silent-failure` / `comments` (`U1`, `U2`, `CC3`-partial) |

- **#1174** — the shape guard is a necessary-condition filter, so digit-only node stderr ending without a newline still merges into the count. Measured: a dependency-free manifest reported as `declares 10 dependencies`. This PR names the ceiling in the comment; closing it needs a policy decision first, which is why the remedy is open.
- **#1175** — the `derive-testcmd.sh` capture is the same unguarded construct. An earlier revision of this PR asserted it was exempt; that was wrong, and both the comment and the PR body are corrected here. What defends it is its callee swallowing node's stderr — adequate, but a cross-file invariant nothing pins.
- **#1176** — the proposed stream-separation remedy, refuted by measurement (it cannot retire the shape guard, and it inverts this PR's own regression test). Also records, below the claim bar, the two sweep test files that crash at import outside a git repo, and the two **pre-existing** positional references in `claim-ticket.sh` that this diff did not introduce and did not sweep.

Applied here: the comment's overclaim (`CC0`/`U0`), the trigger roster restated as a property (`CC2`), the `above`/`below` positional references this diff added (`CC3`), the cross-reference correction (`CC4`), the test name and messages that generalized past their fixture (`CC1`), and a companion test for the guard's `''` arm (`U3`) — mutation-verified: dropping the arm reds that test alone, reordering the arms stays green.

Issue #1176, opened 2026-09-01
PR #1172 review: the suggestion band, checked


feigi:
Durable record of the PR #1172 review's suggestion-band entries that were **checked and not applied**. Closed on purpose: `ledger.mjs check` reads the tracker at `--state all`, so a later run re-deriving one of these lands on the evidence instead of re-litigating it.

The review workflow ran `correctness`, `silent-failure` and `tests`; its size tier never assigned `comments` (that trim is #218), so the controller re-ran that dimension by hand. Nine findings total. What was applied, and the two confirmed defects filed open, are listed on PR #1172.

---

### The `ndeps` capture should separate streams instead of shape-guarding the merge
**Dimension:** `silent-failure` (`U1`), plus the remedy half of `correctness` `U0`.
**File:** `skills/fleet/scripts/claim-ticket.sh`, the `ndeps` capture.

**Claim:** `2>&1` is only needed on the failure path, so a two-file capture (stdout to the variable, stderr to a temp file read only when node exits nonzero) would keep both the reason and a clean count; the PR instead converts every stderr-emitting environment into a hard refusal whose answer was fully recoverable.

**Refuted.** The premises are true and were reproduced — under `NODE_DEBUG=module` and `NODE_OPTIONS=--inspect=0` the script refuses, under `NODE_DEBUG=timer`, `--trace-warnings` and `FORCE_COLOR=1` it succeeds, so the difference really is only whether node chose to write to stderr. The conclusion does not follow:

- Stream separation **demotes the guard, it does not remove it.** With the patch applied and a `node` shim printing `Now using node v22.0.0` to **stdout** before exec'ing the real node, the shape guard is the only thing that catches it. Separation closes the stderr class alone.
- It **inverts PR #1172's own regression test**: `node --test skills/fleet/scripts/claim-ticket.test.mjs` → 79/80, the sole failure being the #752 test, which asserts rc 2 for a stderr-emitting stub. That makes it a policy change, not a defect fix.
[...]

Issue #877, opened 2026-08-24
PR #868 review: the suggestion band, checked — 5 rulings, all refuted with measurement, none applied


feigi:
Record ticket. **No action expected** — filed so the measurements are not redone, per the pattern of #589.

The review workflow budgets the `suggestion` band 0 refuters by policy, so nothing had looked at these. All are in scope (each concerns code or fixtures PR #868 itself adds), so the fix-applier dispatched a refuter against each, biased to refuse and required to verify by running something. **Every one came back refuted, each with zero blast radius.** Nothing was applied.

Full-suite baseline, established independently by four of the five refuters against the pinned head `682bee1`: `tests 1126 / pass 1126 / fail 0`.

---

## 1. The comment/constant pair — `comments` + `types` — REFUTED

Two findings, ruled together because they are alternative remedies for one issue and applying both would be redundant: *"nothing pins the correspondence"* between `board.mjs`'s and `staleness.mjs`'s `maxBuffer`, and *"the bound is a copy-pasted literal, not a shared constant"* across `board.mjs`, `staleness.mjs` and `board-cli.test.mjs`.

The substrate is real — three literals, no shared constant, no test asserting equality. Neither remedy earns its diff:

- **The extraction was built and run.** An exported constant in `arg.mjs`, all three sites rewired: `tests 1126 / pass 1126 / fail 0`. Mechanically fine, and still rejected — the three sites cap three *different* payloads (a ledger tool's stdout, git's read of tracked content, and the test harness's own read of the subject's output, whose comment says so in as many words: `// This harness's own ceiling, not the subject's`). One constant would assert they are one quantity. They coincide; they are not the same thing. It also adds a module edge `board-cli.test.mjs -> arg.mjs` for a number, and `arg.mjs` is chartered narrowly (`// Shared CLI-boundary helpers ... die(), arg(), has(), sweep().`) by #367, which extracted drifting copies of *behavior*.
- **The comment does not over-claim.** Verbatim: `// exhaust the box. Same headroom staleness.mjs takes, for the same reason.` A present-tense fact plus provenance. It does not say kept in sync, must match, is shared, or see-shared-constant — so there is no asserted invariant for a pin to protect. If `staleness.mjs` changed, `board.mjs`'s behavior is untouched and one sentence goes stale.
- **The second finding's own remedy is "do nothing"**: *"If a fourth call site needs the same headroom, hoist one exported constant."* There is no fourth call site.

[...]

PR #1178, opened 2026-09-01
fix(run-team): give the merge gate jq's third outcome, not a boolean

### LINKED ISSUE #1177: run-merge-bot.md: the merge gate's two named traps don't cover jq's THIRD outcome — a compile error is neither pass nor fail, and both obvious readings are wrong
Surfaced by `merge-bot-3-3` during `/fleet:run-team 2 5` on 2026-09-01, wave 3 (PR #1172). Controller-verified independently before filing.

## The defect

`run-merge-bot.md` tells a bot to gate on the `ci-state` payload's own fields with `jq`, and names two ways a gate can be silently wrong:

- **`ci-state --quiet` is unsatisfiable-FALSE** — it drops `jobs` and `missing`, so a per-job gate can never be satisfied.
- **zsh has no `PIPESTATUS`** — `${PIPESTATUS[0]}` is always empty and `[ "" -eq 0 ]` passes, so the gate opens without reading an exit code.

Both are about a gate that returns a *wrong verdict*. Neither covers the case where the gate **does not run at all**, and `jq` has a distinct exit code for exactly that.

`jq` returns **0** when the last output was truthy, **1** when it was false or null, and **3** on a compile error — a malformed program that never evaluated. Measured:

```
$ echo '{"a":1}' | jq -e '.a｜length' >/dev/null 2>&1; echo $?
3
$ echo '{"a":1}' | jq -e '.missing' >/dev/null 2>&1; echo $?
1
```

(The first uses a full-width `｜` where `|` was intended — which is how it happened.)

## Why both obvious readings are unsafe

A gate written as a `jq -e` conjunction is consumed by testing its exit code, and every natural spelling is wrong in one direction:

- **`[ $rc -eq 0 ]`** → a compile error reads as *not green*, and the bot blocks a mergeable PR forever. Fails safe on merging, but stalls the queue on a typo, and the runbook's own guidance elsewhere is that a poll which cannot succeed "does not look like a bug; it looks like patience".
- **`[ $rc -ne 1 ]`** → a compile error reads as *green*, and the bot merges on a gate that never parsed. That is the dangerous direction: no field was ever compared.

Neither is a reading a careful author would notice is wrong, because the two-outcome model the runbook establishes is what invites treating the exit code as a boolean.

## What actually happened

The bot's first conjunction returned `rc=3`. It read the code as *neither pass nor fail*, re-ran, and got a real `0` on the corrected program — so #1172 merged correctly. That was the bot's judgement in the moment, not something the runbook told it to do; a bot that had written either spelling above would have been wrong in one of the two directions and would have had no signal that anything was amiss.

## Desired behavior

The gate section should state the third outcome and the rule that follows from it: **any `jq` exit other than 0 or 1 means the gate did not evaluate — re-run it, and report if it repeats. Never read it as a verdict in either direction.** That is the same principle the fleet already applies to `inflight.sh` exit 2 and `verify-sha.sh` exit 2 ("a probe that could not look is not an answer"), which is why it belongs stated rather than left to a bot's judgement.

Worth considering alongside it, though the remedy is open: the gate is a single long conjunction, so a compile error anywhere kills the whole thing, and the bot cannot tell which leg it lost. Validating the program once against a known-good fixture before pointing it at the real payload would separate "my gate is malformed" from "this PR is not green" — but that is a design call, not part of the confirmed defect above.

## Out of scope

- The two existing traps, which are correct as written.
- `ci-state.mjs` itself — the payload was well-formed; the fault is entirely in how a gate's exit code is read.

## Not filed, recorded here to save a rediscovery

The same bot reported a second finding: that `runStatus` is `null` in the payload and a `runStatus == "completed"` leg is therefore unsatisfiable, "which the skill's step-3 prose invites". The first half is true and the second is not. `grep -rn 'runStatus' skills/fleet workflows` returns nothing — the runbook names `status`, which is a real field the payload carries. The bot introduced that field name itself. Its leg was unsatisfiable because it named a field nobody defines, not because the runbook led it there, so there is no tooling defect in that half.

Provenance: `/fleet:run-team 2 5`, 2026-09-01, wave 3 merging PR #1172 as `c5b9aab9c9e9b35d2cd9e08f0ab0dc5c5a34ce28`. Deduped through `ledger.mjs check` (`verdict: soft-hit`; nearest rows #1057, #847, #910 read and each a different defect).

feigi:
Sizing: light

Closes #1177

## What changed

`run-team/SKILL.md`'s **Merge bot** trap list gains one entry: under the boolean comparison this spec prescribes, a `jq` exit outside 0 and 1 means the gate never compared a field, so it is not a verdict in either direction. The response is scoped to what the exit can change — an exit that can clear on a re-read gets one re-run, and an exit that is a property of the program or the invocation is reported without one. Plus `merge-gate-jq-exit-prose.test.mjs`, pinning the imperative clauses.

The list preamble said "both gate traps" and "Both are measured" over a list that already carried more than two entries. It states the property now instead of counting.

## Review corrections folded in

This PR is a correction, and review found the correction itself minting claims the tree does not support. All were fixed here; each is settled by its command below.

- **The re-run was undifferentiated across every non-0/1 exit.** A compile error and a usage error are properties of the program and the invocation, so a second run returns them by construction, which made "if the same exit repeats" dead prose on the exits the bullet names as motivating. The rule now scopes its response by exit.
- **"Every other exit means the gate never compared a field at all" was false for exit 4** under a filtering gate — an ordinary way to write a gate over the fields this spec names. A correct refusal was being reclassified as a gate that never looked, re-run, then escalated to a human. The table is now bound to the boolean form, and the filtering shape is named as outside it.
- **The probe-citation gloss inverted the files it cited in the same sentence.** The citation was right and stays; the gloss now matches what those files say.
- **The test file recorded the #1172 incident as rc=3 then a re-run,** which teaches the opposite of the rule. It now states the correction step that actually cleared it.
- **Counts in the added prose come out,** replaced by the property — the same rule this PR enforces on the preamble it fixed.
- **The bounding rationale claimed `phrase()` pins terms.** It builds one regex from the whole quoted sentence, so an unbounded pin would not go green off a stray `jq`; what the bound buys is placement.

Refuted findings from the same review are recorded, with their measurements, in #1179 (closed).

## The claims, and what settles each

Re-run on this tree at `228ca2e`. Every exit code here is established on **jq-1.7.1-apple** (macOS) and is not asserted as universal — exit 5 was 2 in jq 1.6.

```
$ jq --version                                                    # jq-1.7.1-apple
$ echo '{"a":1}' | jq -e '.a｜length'  >/dev/null 2>&1; echo $?   # 3  (full-width ｜)
$ echo '{"a":1}' | jq -e '.missing'    >/dev/null 2>&1; echo $?   # 1
$ echo '{"a":1}' | jq -e '.a'          >/dev/null 2>&1; echo $?   # 0
$ echo '{"a":1}' | jq -e --badopt '.a' >/dev/null 2>&1; echo $?   # 2
$ echo '{"a":1}' | jq -e 'empty'       >/dev/null 2>&1; echo $?   # 4
$ echo 'not json' | jq -e '.a'         >/dev/null 2>&1; echo $?   # 5
```

**The exits that cannot change on a re-run** — which is what the scoping rests on:

```
$ for i in 1 2 3; do echo '{"a":1}' | jq -e '.a｜length' >/dev/null 2>&1; echo $?; done
3
3
3
$ for i in 1 2 3; do echo '{"a":1}' | jq -e --badflag '.a' >/dev/null 2>&1; echo $?; done
2
2
2
```

**Exit 4 is a real verdict under a filtering gate,** which is why it could not stay in the never-evaluated class:

```
$ printf '{"verdict":"fail"}' | jq -e 'select(.verdict=="pass")' >/dev/null 2>&1; echo $?   # 4
$ printf '{"verdict":"pass"}' | jq -e 'select(.verdict=="pass")' >/dev/null 2>&1; echo $?   # 0
$ printf '{"verdict":"fail"}' | jq -e '.verdict=="pass"'         >/dev/null 2>&1; echo $?   # 1
```

The gate compared the field, the field said `fail`, and the refusal came back as 4. Under the boolean form the same payload is a plain 1, and an empty payload is 4 there too (`printf '' | jq -e '.verdict=="pass"'` → 4), so the table holds for the shape the spec prescribes.

**The probe citation and its gloss.** The rule follows the house form for an unanswerable probe rather than a new one — `sed -n '22,23p' skills/fleet/scripts/instruments.sh` names `inflight.sh`, `verify-sha.sh` and `staleness.mjs` as the probes that hand exit 2 back for the caller to decide. What those files say about that exit is that it is an answer, not an absence of one:

- `staleness.mjs:139` — `"Could not check" is a first-class ANSWER, not a failure`, and `:136` maps `{ live: 0, fixed: 1, unknown: 2 }` as three peers, with `:28` warning against "collapsing `unknown` into either neighbour".
- `verify-sha.sh:8` — `Exit 0 reachable, 1 not reachable, 2 the question could not be answered`; `:58` — "exit 2 is a verdict the controller acts on".
- `inflight.sh:14` — its exit 2 "carries a payload — the same shape as" the other two.

The bullet's gloss now says could-not-look is a third answer carried beside pass and fail rather than "not an answer", which is the framing those files were written to reject.

`gh pr view 1172 --json state,mergeCommit` → `MERGED`, `c5b9aab9c9e9b35d2cd9e08f0ab0dc5c5a34ce28`, matching the ticket's provenance.

## The ticket names the wrong file, and this fixes the right one

The ticket puts the gate section in `run-merge-bot.md`. It is not there:

- `grep -c PIPESTATUS skills/fleet/commands/run-merge-bot.md` → `0`
- `grep -n PIPESTATUS skills/fleet/skills/run-team/SKILL.md` → inside the **Merge bot** section, which is where the named traps and the "gate on the payload's own fields … read with `jq`" spec actually live.

`run-merge-bot.md` reads `ci-state.mjs`'s own exit code, never a `jq` one — its single `jq` is a `gh --jq` on `gh pr list`. Patching it would have left the paragraph that invites the misreading untouched.

## Enumeration: every other site that touches jq, and why it is left

No site in the repo **reads a standalone `jq` exit status as a verdict**. `grep -rn 'jq -e\|jq --exit-status' --include='*.md' --include='*.sh' --include='*.mjs' .` matches only prose — this PR's own bullet and its test header — and no invocation whose status is branched on. The gate is written by the bot at run time, which is why the rule is prose and there is no script behavior to test. The other `jq` mentions, all deliberately left:

- `gh --jq` in `next-ticket/SKILL.md`, `sizing-a-ticket/SKILL.md`, `docs/agents/issue-tracker.md`, `run-merge-bot.md`, and in `drop-merged-label.sh`, `inflight.sh`, `release-ticket.sh`, `candidates.mjs` — that is gojq inside `gh`, so the status read is `gh`'s, and each script already reads it with `if ! x=$(…)`. `candidates.mjs` goes further and validates the output shape rather than trusting a status. Review checked whether the new bullet misroutes onto this path and it does not: `gh` has no `-e` and returns 0 on a genuine `false`, so the bullet's mapping describes an outcome `gh` cannot produce, and the gate never invokes `ci-state.mjs` through `gh`. Recorded in #1179.
- `tail -1 <output-file> | jq -r '…'` transcript extraction in `review-and-fix.md` and `references/member-lifecycle.md` — those documents already say to judge by whether a report came out, explicitly not by a status.
- `json.sh` runs no `jq` at all.

## What this change could wrongly cause

A "re-run on a non-0/1 exit" rule loops forever against a genuinely broken gate — a bot that mistypes its program once mistypes it every time. That is now closed structurally rather than by a cap: the exits that cannot change on a re-read are told to report without re-running at all, and the exits that can get one re-run, then report and stop. The other misreading — taking "not a verdict" as licence to merge anyway — is foreclosed by "is not a verdict" sitting beside "report it and stop".

## Deliberately not implemented

The ticket's open design idea — validating the gate program against a known-good fixture before pointing it at the real payload — is out of scope and stays out. It is a real improvement and it is also a second mechanism to maintain; the confirmed defect is that a three-outcome code was read as two, and this fixes that alone. Worth its own ticket if a bot loses a leg to a compile error again.

## Verification

```
tests 1641 / pass 1641 / fail 0, exit 0
```

The pinned response sentence changed in this revision, so its pin changed with it and both pins were re-mutated against the new text. Mutants were applied to a copy under a scratch dir, never to the worktree, and each was diffed against a pristine snapshot to confirm it applied before the run:

- replacing the scoped response with the undifferentiated re-run it superseded → red on the scoping pin alone, headline pin green;
- gutting the mandate headline → red on the headline pin alone, scoping pin green;
- control, rewording an unpinned sentence in the exit-code list → green;
- relocating the whole bullet out of the trap list into an appendix → both pins red, while the same pins run **unbounded** against that same relocated text go green. That is the property the `between()` bound actually buys, and it is what the bounding comment now claims.

`merge-gate-jq-exit-prose.test.mjs` is the only test file that pins these sentences or the trap-list bounds (`grep -rln` over `skills/fleet/scripts/*.test.mjs`), so the discrimination above is the whole of it.


feigi:
Review findings adjudicated at `228ca2e`.

**Applied (5 of 9):** the undifferentiated re-run (**W0**+**CC1**, one defect — the re-run rule is now scoped by exit, and the #1172 record states the correction step that actually cleared it); exit 4 wrongly in the never-evaluated class (**CC2** — the table is now bound to the boolean form, with the filtering shape named as outside it); the probe-citation gloss inverting the files it cites (**CC4** — citation kept, gloss fixed); counts in added prose replaced by the property (**CC7**); the bounding rationale's wrong account of what `phrase()` matches (**CC6** — it buys placement, not term-collision protection).

**Deferred (3 of 9), all refuted by a refuter that ran commands:** **W1** (jq exit 2 also covers a missing input file — refuted on the scenario: an unpiped `ci-state` with stdout redirected yields an *empty* file, which exits 4, not 2); **CC3** (`gh --jq` boundary — refuted: `gh` has no `-e` and returns 0 on a genuine `false`, so the bullet's mapping describes an outcome `gh` cannot produce, and the gate never routes through `gh`); **CC5** (the test header's characterisation of the prior trap list — refuted: every pre-existing trap hands a boolean gate a code inside {0,1}, so all of them are wrong verdicts and none is the never-evaluated class).

Recorded with their measurements in #1179, filed `wontfix` and closed per the precedent in #877 and #1176. A one-word residual (`a usage error` → `a usage or system error`) sits there under **Below the claim bar** with a re-open trigger.

Every exit code asserted in the diff and the body is established on **jq-1.7.1-apple** (macOS) only, and is not claimed universal — exit 5 was 2 in jq 1.6.

`node --test skills/fleet/scripts/*.test.mjs` → tests 1641 / pass 1641 / fail 0, exit 0.

Issue #786, opened 2026-08-22
board.mjs: the gh issue/pr list rows get no per-row shape guard, so a partial row renders as undefined on the operator's page


feigi:
Surfaced by `impl-232` while implementing #232, as a sibling of the same class in a different script. Flagged to the controller rather than filed by the member; **verified independently by the controller** against `origin/main` before filing.

**Not a review deferral** — no specialist generated it, so it carries no `dimension`.

## The defect

`board.mjs`'s `build` maps `gh issue list` and `gh pr list` rows straight out of `tryParse` with no per-row shape guard:

```js
const issues = tryParse(issuesJson, [], "gh issue list").map((i) => ({
  number: i.number, title: i.title, labels: (i.labels || []).map((l) => l.name),
}));

const prs = tryParse(prsJson, [], "gh pr list").map((p) => ({
  number: p.number, state: p.state, title: p.title, labels: (p.labels || []).map((l) => l.name),
}));
```

`tryParse` establishes that the payload is JSON and that it is an array. Nothing establishes that a row is an object, or that it carries the fields the map interpolates. `(x.labels || [])` is the only defaulting present, and it covers exactly one of the fields consumed.

So a row missing `number` or `title` produces `undefined` in the board's own model, which reaches the operator's page as a ticket or PR the payload cannot describe.

## Relationship to the ledger siblings

This is the same class as #232 and #643, in a different script, and it is **not** covered by fixing either of them — there is no shared guard the three route through. `impl-232` checked the other gh-JSON consumers rather than assuming, and reported the rest clean:

- `candidates.mjs` validates every field it consumes (`n`, `t`, `l`, `d`, `spec`) — not defective
- `fleet-tick.mjs`'s issue read validates `number` + `labels`, exactly what it requests and consumes — not defective
- `fleet-tick.mjs`'s row read checks `Array.isArray` only, but consumes only `rows.length` — not defective

`board.mjs` is the one site where the fields consumed and the fields validated genuinely diverge.

## Why the severity is lower than #232's, and why it is still filed

The failure direction is opposite. #232 fails **closed** into a blocking exit code, where a confident-but-undescribable hit can stop a filing. This one fails **soft** into a dashboard: the board still renders, and the wrong value is visible rather than acted on. Nothing gates on it.

Filed under the repriced bar from #239 — the defect is confirmed, the remedy is open.

## The remedy is open

At least three shapes, none obviously right, and the choice is the reason this is not a one-liner:

[...]

PR #1180, opened 2026-09-01
fix(board): guard mapCi against a JSON payload that parses to null

### LINKED ISSUE #1170: board.mjs mapCi: a ci-state payload that parses to JSON null throws a TypeError out of the whole build tick
**Dimension:** correctness (code-reviewer) and silent-failure — raised independently by three dimensions of the PR #1166 review. · **Deferred from PR #1166 review**

`skills/fleet/scripts/board.mjs:177` — the `d.status` dereference immediately after `mapCi`'s `try { d = JSON.parse(ciJson); }` block.

## The defect

`JSON.parse("null")` succeeds and yields `d === null`. The `d.status` read sits **outside** the `try`, so the parse guard does not cover it and `mapCi` throws a `TypeError` instead of returning a verdict string.

Reproduced by direct execution:

```
node -e 'import("./skills/fleet/scripts/board.mjs").then(m=>{try{console.log(m.mapCi("null",1))}catch(e){console.log("THREW:",e.constructor.name,e.message)}})'
→ THREW: TypeError Cannot read properties of null (reading 'status')
```

The blast radius is not one CI cell. `mapCi` runs inside `gather()`'s per-PR loop, and `serve`'s tick is `try { computeBoard(gather(...)); … } catch (e) { console.error(\`${NAME}: build tick failed: …\`) }`. One throw aborts `gather` for the whole tick, so `board.json` is not rewritten and **every** PR's row goes stale — under a message that names neither the PR nor `ci-state`. The `build` path has no catch at all.

## What is settled, and what is not

Two refuters looked at this during the PR #1166 review and split 1-1.

**Both agree** the throw reproduces by direct execution, and that the code is **pre-existing** — the `d.status` line is untouched by PR #1166's diff.

**They disagree on reachability.** `ci-state.mjs` always emits `JSON.stringify({pr, …})`, so its own output starts with `{`; a truncated write from it fails to parse (and now warns) rather than yielding the four bytes `null`. No producer that emits a bare `null` was identified. One refuter read that as "unreachable, therefore not a defect"; the other read the unguarded dereference as a defect regardless of which producer reaches it today, since `mapCi` is exported and `runCiState`'s contract is "whatever the child printed".

That dispute is why this is filed rather than closed: **the defect is confirmed, its priority is not.**

## Why it was deferred rather than fixed in PR #1166

#605 constrained its fix to be **additive** — the return value pinned, only the silence changed. Making `mapCi` answer `"unknown"` for a payload that parses to a non-object changes behaviour on a path #605 does not own, so PR #1166 left it alone deliberately.

## Remedy — open

Two shapes were proposed and neither was measured:

- `if (!d || typeof d !== "object") return "unknown";` after the parse, or
- move the `d.status` check inside the `try`, which is the smaller diff and closes the same state.

Either way it needs a test that kills its own mutant: `mapCi("null", N)` must not throw, and a near-miss change must stay green. Decide there whether the non-object case should also warn — it is a corrupt payload by the same argument #605 makes, and PR #1166's `warnedCiParse` gate is right there — or stay silent like the other legitimately-unknown states.

## Related, not duplicates

- #875 — `runCiState`'s exit-code gate accepting a truncated or signal-killed read as a verdict. Same function, different failure: that one is about which reads are promoted to verdicts, this one is about a verdict that parses and then throws.
- #603 — the `warnOnce(key, msg)` helper, if the remedy chooses to warn.

feigi:
## Summary

`mapCi` in `skills/fleet/scripts/board.mjs` threw a `TypeError` on a CI payload that parses to JSON `null`, instead of returning a verdict string. `JSON.parse("null")` succeeds and yields `d === null`, and the status gate then dereferenced it. Confirmed live against `origin/main` before the fix, by direct execution:

```
node -e 'import("./skills/fleet/scripts/board.mjs").then(m=>{try{console.log(m.mapCi("null",1))}catch(e){console.log("THREW:",e.constructor.name,e.message)}})'
→ THREW: TypeError Cannot read properties of null (reading 'status')
```

**Nullishness, not object-ness, is the discriminator.** `null` alone among parsed payloads has no properties to read; every other one boxes and reads its status as `undefined`, answering `"unknown"` at the status gate without throwing. `[]` and `{}` are not counterexamples to that and never were — `node -e 'console.log(typeof [], typeof {}, typeof null)'` prints `object object object`, so an object-ness axis does not separate the throwing payload from the safe ones. The fix is a one-line guard, `if (d === null) return "unknown";`, with its own return rather than a path into the status gate.

`mapCi` is exported and has one caller, `gather()`'s per-PR loop. The consequence differs by entry point, and both were read off `board.mjs` directly: under `serve` the throw is uncaught until the whole-tick catch, which logs `build tick failed: ...` — naming neither the PR nor `ci-state` — and leaves the previous `board.json` standing; on the one-shot `build` path it exits through `main().catch` and dies loudly.

**Stated as a conditional, because it is one.** No producer in this repo can emit a bare `null`: `grep -n 'emit(1\|process.stdout' skills/fleet/scripts/ci-state.mjs` returns two writes, both `JSON.stringify` of an object literal. The guard is defence in depth against a producer that can, and `mapCi` is exported, so any future caller inherits it. Two refuters split 1-1 during the #1166 review on this defect's *priority*, not on whether it is real; both agreed the throw reproduces and the code is pre-existing. Not re-litigated here.

Silence is the correct treatment and is now pinned rather than asserted. Measured across the payload space, every parseable payload already answers `"unknown"` with zero stderr; the warn arm covers unparseable payloads only (`""` warns). A bare `null` joins the class it structurally belongs to.

Sizing: light

## Review follow-up

A comment-dimension re-run found eight defects in this PR's own prose and tests, all fixed in `b6955e8`. The review workflow never assigned that dimension — its size tier drops it when no docs-classified *file* is in the diff, and `board.mjs` classifies as `src`, though this diff's `board.mjs` half was almost entirely comment lines (#218).

- **Three returns were described as one.** The old comment said the sibling payloads "fall through to the same trailing `unknown`". They answer at the status gate; the guard has its own return; the trailing return is reached only by an object with status `"completed"` and an unrecognised verdict. A maintainer deleting that trailing return on the comment's authority would have broken the `no-ci` case this suite pins.
- **Reachability was asserted as history** ("aborted that tick's rewrite"), now conditional and completed with both entry points.
- **Positional references replaced by a name, not by another position.** `grep -n 'd\.status' skills/fleet/scripts/board.mjs` reads one line, so "the status gate" needs no positional word.
- **The over-guard rationale is dropped, not reworded** — see the test plan.
- **Counts and tallies out** ("the one JSON scalar", "the lone outlier"), replaced by the property.

Deferred: **#1181** — `tryParse` (`board.mjs:104-109`) carries the same unguarded shape and is out of #1170's scope. Refuted and recorded, not applied: **#1182**.

## Test plan

- [x] Reproduced the throw against `origin/main` by direct execution before changing code (see above)
- [x] `board.test.mjs` "a JSON payload that parses to null → unknown, not a thrown TypeError" — now wraps in `withStderr` and asserts **zero** stderr lines. Every other `mapCi` test in the file captures stderr; these two did not, so the "Silent, not warned" contract rested on nothing. Asserting a warn here would have been backwards.
- [x] The companion case pins that a payload which already answered `"unknown"` still does — i.e. the guard narrowed nothing. **It does not discriminate an over-guard, and no longer claims to.** Enumerated over every JSON value shape plus a 200,000-case random sweep, a guard "rejecting every non-plain-object, not just `null`" returns the same verdict as this one for every input, zero divergence — reaching a verdict at all takes a plain object, which such a guard passes through. The earlier claim that this case caught that mistake was vacuous, so it is gone rather than reworded.
- [x] Mutation-verified, each mutant applied to a copy outside the checkout and confirmed by diff against a pre-mutation snapshot: guard deleted → the null test alone reds; guard starts warning → the null test alone reds; no-status payloads start warning → the sibling test alone reds; an inert local rename → 62/62 green.
- [x] `node --test skills/fleet/scripts/*.test.mjs` — tests 1641 / pass 1641 / fail 0, exit 0, at `b6955e8`

Closes #1170


feigi:
Review deferrals filed:

- **#1181** `ready-for-agent` — dimension `correctness`. `tryParse` (`board.mjs:104-109`) carries the same unguarded-null shape this PR fixed in `mapCi`; its `json == null` guard covers the input string, never the parsed result. Deferred because #1170's acceptance is about `mapCi` and `tryParse` is a different function with different callers. The remedy is left open in the body: the file now carries three policies for one payload class (`mapCi` guards null only, `readAgent` rejects every non-object, `tryParse` guards neither), and whether they should converge is the question worth answering.
- **#1182** `wontfix`, closed — the checked-and-not-applied record. Holds the CC5 refutation (both charges, with the measured payload/stderr partition and the widened `legitimately unknown` search) and the refuted half of the test-name finding.

Also commented on **#786** to correct a premise of its body: it states that `tryParse` establishes the payload is an array, and it establishes neither that nor non-nullness. Distinct defect from #1181 — a wrong value rendered on the page versus a throw that aborts the tick — so neither fix covers the other, but they land in the same two `.map` call sites.

Issue #1171, opened 2026-09-01
PR #1166 review: the suggestion band, checked


feigi:
**Deferred from PR #1166 review.** Durable record so a later run re-deriving one of these lands on the measurement instead of re-litigating it. Closed on purpose — nothing here is queued agent work.

The PR #1166 review produced fifteen findings. Two survived refutation and landed in the PR (see the end). The rest are recorded here with what was actually run.

---

## Refuted by measurement

Each was handed to a refuter biased to refuse, instructed to verify by running something.

### `runCiState() routes this payload straight here by design — at any exit but 2, non-empty stdout is a real verdict`
**Dimension:** comments · `skills/fleet/scripts/board.mjs`, in `mapCi`'s comment, and its twin in `board.test.mjs`

Claimed the clause is mis-scoped: the `e.status !== 2 && out.trim()` test lives in `runCiState`'s **catch**, so it governs non-zero exits only, and exit 0 applies no emptiness test — so a rule stated over "any exit" is wrong. Secondarily, a signal kill leaves `status` null, which `!== 2` admits.

**Refuted.** A refuter drove the real `gather()` over stubs at every exit: exit 0 empty → `"unknown"`; exit 0 non-empty valid → `"green"`; exit 0 non-empty unparseable → routed to `mapCi` and warned there; exit 1 non-empty → `"red"`; exit 2 non-empty → carry-forward, not routed. The proposition the sentence states — for any exit but 2, non-empty stdout reaches `mapCi` as a verdict — is true at every exit tested and false only at 2, exactly as written. It never claims the converse. The grammatical subject is "this payload", the non-empty unparseable one named in the preceding sentence, so the clause is scoped to non-empty stdout by its own subject.

On the signal case the refuter measured `e.status === null`, `e.signal === "SIGKILL"`, guard admits — factually correct, and the comment's very next clause predicts that outcome ("a truncated write … reads exactly like a PR whose first run has not started"). The observed behaviour matches the prose.

### The false-positive test's comment credits discrimination for flood control
**Dimension:** comments · `skills/fleet/scripts/board.test.mjs`, the comment above the legitimately-unknown-stays-silent test

Claimed that "a warn that cannot tell them from a garbage payload prints a line for every PR awaiting its first run, on every ~15s tick" credits the wrong mechanism, because the shipped `warnedCiParse` gate would hold an indiscriminate warn to one line per PR.

[...]

PR #1184, opened 2026-09-02
refactor(board): collapse every hand-rolled warn-once gate into one warnOnce(channel, key, msg)

### LINKED ISSUE #603: board.mjs: three hand-rolled warn-once idioms collapse into one warnOnce(key, msg) helper

feigi:
Collapses every hand-rolled warn-once gate in `board.mjs` into one
`warnOnce(channel, key, msg)` helper over a single Set.

Sizing: light

## What the sweep covered

The ticket's title says three idioms and its body names three; both later
comments correct that upward. Enumerated against the merge base rather than
taken from the ticket:

```
$ git show "c3d065dc:skills/fleet/scripts/board.mjs" | grep -nE '^(const|let) warned'
146:const warnedCiParse = new Set();
305:const warnedMeta = new Set();
307:const warnedLines = new Set();
383:let warnedNoSpendDir = false;
390:const warnedSkips = new Set();
```

All of them are collapsed — `ci-parse`, `meta`, `lines`, `no-spend-dir`,
`skips`. Nothing is deliberately left. The class does not extend past this
file: no other fleet script carries a warn-once gate.

```
$ grep -rlnE 'console\.error' skills/fleet/scripts/*.mjs | grep -v test \
    | xargs grep -lnE '(let|const) [a-zA-Z]*(warn|Warned|reported|Reported|announced)'
skills/fleet/scripts/board.mjs
```

After the sweep the only such declaration is the helper's own, and no gate
hand-rolls the membership test any more:

```
$ grep -nE '^(const|let) warned' skills/fleet/scripts/board.mjs
147:const warnedOnce = new Set();
$ grep -n '\.has(\|\.add(' skills/fleet/scripts/board.mjs
150:  if (warnedOnce.has(k)) return;
151:  warnedOnce.add(k);
```

## The channel prefix, and why it is not optional

The stored key is `${channel}\0${key}`, never the caller's key alone. The
torn-line and skipped-transcript gates are keyed on the **same** transcript
path:

```
$ grep -n 'warnOnce("lines", file\|warnOnce("skips", file' skills/fleet/scripts/board.mjs
353:        warnOnce("lines", file, `${file} has an unparseable line that is not its last; ...
461:        warnOnce("skips", file, `skipping ${f}: ${e.message}`);
```

A bare-key Set would let a mid-file-tear warning consume that same file's
later skip warning — a behaviour change, not a refactor. NUL joins the two
halves because neither a channel name nor a path can hold one.

This hazard was invisible to the suite as it stood: every existing test hands
each gate a path no other gate has seen, so a channel-less key stays green.
The new separation test is what closes that, and it kills exactly its own
mutant — see the test commit body for the runs.

## `warnedNoSpendDir`

It was a bare `let` boolean with no key. **Picked: fold it in with the empty
key** rather than leave it hand-rolled. The key is constant either way, so the
gate is behaviour-identical and module-private. Nothing in the suite drove that
mode, so this PR adds the test that does.

The empty key here is a **choice, not an absence** — corrected in review, where
the first version of these comments called it the "nothing to key on" case.
`dir.error` is not constant: `findSubagentsDir` words an unresolvable project
dir differently from a lookup that threw.

```
$ sed -n '241p;248p' skills/fleet/scripts/board.mjs
    if (!existsSync(projects)) return { error: `no transcript dir for cwd ${cwd} (looked in ${projects})` };
  } catch (e) { return { error: `transcript lookup failed: ${e.message}` }; }
```

So there is something to key on, and the gate declines to use it — which is
what preserves the base behaviour #603 asks for. What that costs is only the
repeated stderr LINE: a fault that differs still reaches the browser on the
tick it happens, through the `{ error }` `gatherSpend` returns for it.

```
$ grep -n 'sp.error' skills/fleet/scripts/board.html
107:  if (sp.error) return { kind: "error", text: "spend unavailable: " + sp.error };
```

Both comments this PR wrote about the empty key say that now — the gate's own,
and `warnOnce`'s header, which described the empty key by a motive that is
false for its only caller and now describes the mechanism instead.

## What the restructuring could newly have broken

- **The torn-line position check.** It stays OUTSIDE the `warnOnce` call. A
  legitimate torn tail must never reach the gate, or it consumes that file's
  one warning and permanently silences the real fault when the tear later
  moves mid-file. `a tail tear that later moves mid-file is reported on the
  tick it moves` pins this and stays green.
- **`mapCi`'s per-key gating.** Unchanged: warn once per PR across ticks, a
  second PR still gets its own line. Pinned by `an unparseable payload warns
  ONCE per PR across ticks, and a second PR is not masked`.
- **`pr` staying optional at `mapCi`'s warn site.** The refuted finding from
  the #1166 review (recorded in #1171) is **not** reintroduced.
- **Over-silencing.** The separation test's second tick is input the gate must
  still WARN on, alongside the existing per-PR and two-transcript tests that
  feed it distinct keys.

## Prose

Comments that named the removed identifiers are renamed to the channel they
now refer to, in both `board.mjs` and `board.test.mjs`. The reason the
torn-line gate is safe is deliberately **not** merged into the reason the skip
gate is safe: the skip message carries a count and needs `skipped` reaching the
browser every tick to keep that number live, and the torn-line message carries
none. Sharing `warnOnce` collapses how the gates are written, not why each is
allowed to stay quiet.

## Tests

```
# tests 1645, pass 1645, fail 0
```

## Review deferrals

Found by the PR review and deliberately **not** fixed here, both because they
change stderr bytes and #603's contract is byte-identical stderr:

- **#1190** — key the `no-spend-dir` gate on `dir.error` instead of `""`, so a
  second, genuinely different spend-dir fault gets its own line. Ships with the
  test extension that axis needs: the test this PR adds binds one `dir` object
  and calls `gatherSpend` twice, so it discriminates once-per-process vs
  once-per-tick and is silent on empty-key vs message-key.
- **#1191** — the `skips` gate keys on the full path but prints only the
  basename, so two broken transcripts under different session dirs produce
  byte-identical lines.

The rest of the review's `suggestion` band was checked and did not survive; it
is recorded closed in #1197.

Closes #603


feigi:
Review deferrals filed:

- **#1190** `ready-for-agent` — `board.mjs:425`, dimension correctness. Key the `no-spend-dir` gate on `dir.error` instead of `""`, plus the test extension for the keying axis. Deferred: changes stderr bytes, and #603's contract is byte-identical stderr.
- **#1191** `ready-for-agent` — `board.mjs:461`, dimension silent-failure. The `skips` gate keys on the full path but prints the basename, so two broken transcripts under different session dirs give byte-identical lines. Deferred for the same reason.
- **#1197** `wontfix`, closed — the checked `suggestion` band: three findings refuted with the runs, two below the claim bar, three that alleged no defect, and two snapshot artifacts dismissed against a green worktree run.

Applied here instead: `159db18` and `8ad3674`, both comment-only. The gate's comment claimed it had "nothing to key on", which is false — `findSubagentsDir` words an unresolvable project dir differently from a lookup that threw — and correcting it falsified `warnOnce`'s own header, which introduced the empty key by that same motive.

Issue #927, opened 2026-08-25
ci-state.mjs: the no-ci payload emits "jobs":[] and "missing":[] for a repo whose workflow set was never read


feigi:
**Deferred from PR #913 review** (ticket #288). `dimension: silent-failure`, severity `suggestion`.

**PRE-EXISTING and unchanged by PR #913's diff** — recorded here so it is not lost, not because that PR introduced it.

`skills/fleet/scripts/ci-state.mjs` — the `no-ci` payload emits `"jobs":[]` and `"missing":[]` for a repo whose workflow set was never read. `missing` is initialised to `[]` (at `:427`, re-derived at `50d63f8`) and assigned only inside the non-no-ci arm.

**Measured.** Probe with the pristine script against a `git init`-ed repo holding no `.github/workflows`:

```json
{"pr":42,…,"verdict":"no-ci","reasons":["no workflows configured under .github/workflows/ — pass --declare-no-ci …"],"jobs":[],"missing":[]}
```

exit 1. Identical before the PR: the removed line bound `expected = []` under `noCi`, so the diff changes nothing here.

This is the same empty-array-reads-as-clean shape the file's own rate-limit comment calls out as the reason the rate-limited payload OMITS `missing` — a consumer that gates on `missing.length` sees an assertion that nothing was missing, from a run that never looked.

**Suggested remedy — UNVERIFIED, the review never ran it.** Only if a payload reader ever gates on `missing`: drop `jobs`/`missing` from the no-ci payload the way `emitRateLimited()` does, so a caller sees absence rather than an empty array. Any such change needs the payload's consumers checked first — `board.mjs`'s `mapCi()` at minimum.

**Why deferred, and why `needs-triage` rather than `ready-for-agent`:** out of #288's scope (pre-existing, and a payload-shape change rather than a simplify item), and the finding itself does not establish a defect — `verdict: "no-ci"` already disambiguates for any reader that checks it, which is why it was raised as a suggestion. Torn on whether anything is broken, so a maintainer should decide whether any consumer gates on `missing` at all before this is worth doing.

feigi:
> *This was generated by AI during triage.*

**Moving to `ready-for-agent`.** The ticket asks a maintainer to decide *"whether any consumer gates on `missing` at all before this is worth doing"*. **They do, and two of them have already been burned.**

- A merge bot gating on `ci-state --quiet` polled forever, because `--quiet` drops the very fields the gate reads — the loop was unsatisfiable and looked like patience.
[...]

PR #1185, opened 2026-09-02
fix(ci-state): state the expected-job derivation as a property, not a tally

### LINKED ISSUE #847: ci-state.mjs:314 — the job-count comment names four and five; the workflow defines two
Surfaced during PR #842's merge wave by `merge-bot-11`, which refused to accept `missing: []` from `ci-state.mjs` without checking whether that emptiness was vacuous. Filed by the controller.

## The defect

`skills/fleet/scripts/ci-state.mjs:314` reads:

> `// Never hardcoded. The fleet's prose names four jobs; the workflow defines five.`

Measured against `origin/main`, `.github/workflows/ci.yml` defines **two** jobs, `rebase-check` and `check`, with no job-level `name:` overrides. Both figures in that sentence are wrong.

```
$ git show origin/main:.github/workflows/ci.yml | awk '/^jobs:/{f=1;next} f && /^  [a-zA-Z_-]+:/{print $1}'
rebase-check:
check:
```

## Why it is harmless today, and why it still matters

The sentence's own first clause is the reason nothing breaks: **`Never hardcoded`** — the expected-job list is derived live from the workflow file, so `missing` is computed correctly regardless of what the comment claims. No behaviour depends on the numbers.

What it costs is a reader's ability to sanity-check the derivation. A merge bot gating on `missing: []` reasonably asks "empty because nothing is missing, or empty because the derivation broke?" — and this comment is the only in-file statement of how many jobs to expect. It answers that question wrongly in the direction that makes a broken derivation look plausible. `merge-bot-11` had to go read `ci.yml` itself to establish that its empty `missing` was real.

## The class, which is the actual point

This is a **count written into prose**, and it is stale. The repo already carries the rule that a count is false the moment the next commit lands and that prose should state the property instead — this comment predates that rule and demonstrates exactly why it exists. The fix is not to update the numbers to `two`; that would rot the same way the next time a job is added.

## Suggested direction (approach open — hence `needs-triage`)

State the property the sentence is actually reaching for: that the expected-job list is derived from the workflow rather than hardcoded, so it tracks the file automatically and a reader should read `ci.yml` for the current set. No number.

Whoever takes this should also check whether the "fleet's prose names four jobs" half is a live claim about `run-team/SKILL.md` and `review-and-fix.md`, or itself already stale — that half was not measured here, only the workflow half.

Deferred from PR #842's merge wave — surfaced by a gate vacuity check, not by a review finding.

feigi:
Sizing: light

The comment introducing `expectedJobs()` in `skills/fleet/scripts/ci-state.mjs`
read:

> `// Never hardcoded. The fleet's prose names four jobs; the workflow defines five.`

Two figures about files the comment does not own. This PR states the **property**
instead, and pins it with a prose guard in the shape #348 established.

## What was actually wrong with that pair

Measured against `origin/main`:

- The **workflow** figure never described this repo. Every commit in this repo's
  `.github/workflows/ci.yml` history carries the same job set, and it is not the
  figure the comment named:
  ```
  for c in $(git log --format=%h origin/main -- .github/workflows/ci.yml); do
    git show "$c:.github/workflows/ci.yml" \
      | awk '/^jobs:/{f=1;next} f&&/^  [a-zA-Z]/{n++} END{print n+0}'
  done | sort | uniq -c
  ```
  -> one distinct count across the whole history. That figure described the
  agent-brain repo's workflow, which is on an internal GHE host and so cannot be
  settled from this repository in either direction.

- The **prose** figure was a true claim about this repo's own documents at the
  time it was written.

So the defect was **the pairing**, not either half alone: a figure measured here
sitting beside one measured somewhere else, both written as though measured here.
And no job was ever added to this repo's workflow, so an earlier revision of this
branch was wrong to say the figures rotted *because* one was — that clause is
gone from the comment.

## The fix

The comment states the property: `expectedJobs()` parses the `jobs:` block of the
workflow file this run resolved, so the expected set tracks that file across
repos, and a reader asking whether an empty `missing` is real should read that
workflow. **No count**, in either direction — a count would rot exactly as its
predecessor did.

The comment is now **shorter than the version this branch first pushed**: the
correction is made by deleting minted sentences, not by adding qualifying ones.

### Deviation from the acceptance criterion — unchanged, but its reason corrected

The ticket asks the comment to point a reader at `ci.yml`. It points at *the
workflow file this run resolved* instead, because writing a fixed path into prose
re-mints in a comment what `discoverWorkflowFile` exists to remove.

An earlier revision of this body and of the comment justified that with "the path
is not hardcoded in the code". **That was false and both are now corrected.**

```
$ grep -n 'arg("workflow") || \|WORKFLOWS_DIR = ' skills/fleet/scripts/ci-state.mjs
235:const workflow = arg("workflow") || "CI";
302:const WORKFLOWS_DIR = ".github/workflows";
```

Both are literals. #111 removed the hardcoded *filename default*, not the
directory and not the workflow-name default. The deviation stands on the
portability argument alone; the sentence asserting otherwise is deleted rather
than repaired, since its only other half restated the discovery comment above
`WORKFLOWS_DIR`.

### The cross-repo half, narrowed to what its own command settles

The comment keeps `integration-docker` as the reason the list is derived rather
than read off the fleet's documents, but the claim now matches the command
offered as proof of it:

```
$ git grep -l integration-docker -- skills/fleet
skills/fleet/scripts/ci-state.mjs
```

An earlier revision claimed the token "appears in no fleet document" and cited
`-- 'skills/fleet/**/*.md'`. That pathspec matches nothing, has a `**/` depth hole
that skips any `.md` sitting directly in `skills/fleet/`, and could not see the
file that falsifies the wider claim — `docs/specs/2026-07-23-fleet-plugin-design.md`
names the token. The directory pathspec above has no depth hole and settles
exactly the sentence it now sits beside.

## Test

`skills/fleet/scripts/expected-jobs-prose.test.mjs` pins that the comment states
the property and carries no job tally.

It bans a **tally, not a number** — a cardinal offends only in a sentence that
also says "job" — because a pin coupled to a moving figure reds on every
legitimate edit and earns its own deletion (#348, and PR #554 paid for it).

Mutation-verified in both directions. The first row is a full-suite run at this
head; the rest are scoped to this test file:

| mutation applied | result |
| --- | --- |
| a stale tally restored to the comment | **1644 pass / 1 fail**, and the one red is this guard |
| the derivation claim inverted to "IS hardcoded" | red |
| the whole argued comment deleted, section banner left | red |
| a digit tally added | red |
| benign reword of the comment, no tally | **green** |
| issue citations sitting beside the word "job" | **green** |

The guard's own new logic is load-bearing too: dropping `\d+` from `CARDINAL`,
dropping the `#\d+` issue-citation strip, or reverting the derivation check to its
polarity-blind form each turns the refusal test red.

Two of those rows are the fixes this PR's own review produced. The check accepted
the bare word `hardcoded` in **any** polarity, so inverting the comment's central
claim left both tests green; and the slice included the `// --- ` section banner,
which says "derived" by itself, so the entire argued comment could be deleted with
that half still satisfied. Both are closed, and the file now states its ceiling as
what the guard actually holds rather than as proof the comment is sound.

Full suite green at this head: `node --test skills/fleet/scripts/*.test.mjs`
-> **1645 tests, 1645 pass, 0 fail**.

No behaviour change — `expectedJobs()` is untouched:

```
$ git diff origin/main...HEAD -- skills/fleet/scripts/ci-state.mjs \
    | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -vE '^[+-]\s*//' | wc -l
0
```

## Review deferrals

Filed from this PR's review, none dropped: **#1193** (the cardinal vocabulary
stops at "ten" — a convention-wide question across four sibling prose guards),
**#1194** (the tally rule refuses a cardinal that counts something other than
jobs), and **#1195** (closed record: the suggestion band, checked — what was
refuted and what sits below the filing bar).

## Scope

`#927` (the no-ci payload emitting empty `jobs`/`missing`) is **not** touched. It
is the same file and the same reader's question — "is this emptiness real?" — from
the payload side; this PR only changes the comment side. Left for #927.

Closes #847


feigi:
Review deferrals filed from this PR's review — every finding accounted for, none dropped:

- **#1193** `ready-for-agent` — dimension **tests**. `expected-jobs-prose.test.mjs` CARDINAL spelled-number list stops at `ten`, so a tally spelled eleven or above evades the guard. Deferred rather than fixed here because the identical `one..ten + \d+` vocabulary is house convention across four sibling prose guards (`line-distance-prose`, `verify-sha-exit2-prose`, `tier-guard-gate-prose:85`, this file); changing it in one file diverges from the repo and leaves the hole open in the other three.
- **#1194** `ready-for-agent` — dimension **types**. The tally rule is scoped to a period-delimited sentence, so any cardinal sharing that span with the word "job" reads as a job count (`3 workers`, `job 2 of a matrix build`). Deferred because the finding's own suggested remedy measurably makes it worse: requiring the cardinal to adjoin `job(s)` stops detecting the exact rot #847 was filed for in its non-adjacent phrasing, and still false-positives on the matrix-index case.
- **#1195** `wontfix`, closed — the record for the suggestion band: one finding checked and refuted (the `git grep` scope and the GHE-host citation in the derivation comment are load-bearing, not padding — `cross-repo-citation-prose.test.mjs:95` pins that marker for foreign evidence), one upheld as refuted, and one below the filing bar (`docs/adr/0002`).

Applied in `e5b896a`; the full suite reads 1645 pass / 0 fail at that head.

Issue #375, opened 2026-08-11
a timed-out `gh pr create` still creates the PR but drops `--label`, so the release label is silently absent


feigi:
Hit live during a `/fleet:run-team` run by `impl-79` while opening PR #373 (ticket #79).

## What happened

`gh pr create ... --label patch` exceeded the caller's 120s tool timeout and was backgrounded. **The PR was created. The label was not applied.** `gh pr list` showed empty labels afterwards. The member noticed, ran `gh pr edit 373 --add-label patch`, and re-verified — but only because it happened to re-read the labels.

## Why it is not merely cosmetic

The release label (`patch`/`minor`/`major`) is the fleet's only per-PR semver signal, and nothing downstream re-derives it:

- `next-ticket` step 7 / `run-team` phase 2 instruct the implementer to create the PR *with* exactly one release label, and treat that as done once `gh pr create` returns.
- The finisher checks the worktree, the CI gate and the deferral issues. It does not check for a release label.
- The merge bot merges a `ready-to-merge` PR. It does not check for a release label either.

So a timed-out create produces a PR that reads as correctly opened, passes every later gate, and merges with no release label — and the timeout is more likely on exactly the PRs with the longest bodies.

It is also the failure shape this repo keeps closing: a partial success reported as a success. The exit path is a timeout rather than a non-zero status, so nothing in the chain has a code to act on.

## Possible directions (why `needs-triage`)

- Have the implementer's step-7 instruction require a label re-read after create (`gh pr view <n> --json labels`), and add it if absent. Cheapest, but it is a fifth thing to remember in a step that already has four.
- Split the create: `gh pr create` without `--label`, then `gh pr edit --add-label` as its own command, so the label write has its own exit status. Costs one extra call per PR and makes the failure loud.
- Have the finisher assert exactly one release label as part of its gate, alongside the deferral check. Catches it wherever it was lost, including a hand-created PR, but only at the end.

The three are not exclusive, and which one is right depends on whether the label should be guaranteed at creation or merely before merge.

## Not reproduced deliberately

Reported by the member that hit it; the controller did not re-run a 120s-timeout create to reproduce. The observable state — PR #373 open with empty labels, then labelled by a follow-up edit — is consistent with the report, and the ledger row for #79 records it.

## Scope searched

`ledger.mjs check` (`verdict: clean`, tracker reached) plus `gh issue list --state all --search "gh pr create label timeout"` — no existing issue covers this.

feigi:
[...]

PR #1187, opened 2026-09-02
test(candidates): pin next-ticket's candidate-scan flag to the flag OPTIONS declares

### LINKED ISSUE #851: next-ticket/SKILL.md:15 — the live --require-label rule is pinned by nothing, while its run-team twin is pinned by candidates.test.mjs
Surfaced by `impl-240` while implementing #240 (PR #850), scope-checked and left out of that PR as creep into a live skill unrelated to the spec correction. Filed rather than dropped; the implementer did not invent the work.

## The gap

`skills/fleet/skills/next-ticket/SKILL.md:15` carries the mandatory candidate-scan invocation:

```
~/.claude/skills/fleet/scripts/candidates.mjs --require-label ready-for-agent --allow-fallback
```

The spelling is **correct**. Nothing asserts that it stays correct.

Its sibling copy in `skills/fleet/skills/run-team/SKILL.md` *is* pinned — `candidates.test.mjs`'s `phase0Step1()` slices `RUN_TEAM` between `1. **Candidate scan**` and `\n2. ` and asserts the flag there. That pin was added by PR #228 when #173 fixed the same wrong spelling in the same place. The `next-ticket` copy was never brought under it.

## What was checked, and how

Verified independently by the controller against `origin/main`, not taken from the implementer's report:

- `git show origin/main:skills/fleet/skills/next-ticket/SKILL.md | grep -n -F -- '--require-label'` → one hit, `:15`.
- `phase0Step1()` reads `RUN_TEAM` (`candidates.test.mjs:1148`, `join(…, "skills", "run-team", "SKILL.md")`). It never opens `next-ticket/SKILL.md`.
- `inflight-exit2-prose.test.mjs` *does* read `next-ticket/SKILL.md` (`:46`), but its assertion slices `between(NEXT_TICKET, "## 3. In-flight check", "## 4. Suggest")` — the in-flight section, which does not contain `:15`.
- Every other `next-ticket/SKILL.md:15` occurrence in `candidates.test.mjs` (`:176`, `:741`, `:1062`) is inside a **comment** — a citation of the documented caller, not an assertion on it.

Scope searched: `skills/fleet/scripts/*.test.mjs`, tracked files, at `origin/main`.

## Why it matters

`--label` is the spelling `candidates.mjs` **refuses at exit 2**, and it is also the correct `gh issue list` flag — so the wrong spelling is the plausible one to write, and a reader copying from a `gh` example lands on it. That is the exact defect #173 fixed in `run-team/SKILL.md` and #240 fixed in the design spec. `next-ticket/SKILL.md` is the third live carrier and the only one with no assertion behind it.

## Open decisions for triage

- Whether a pin belongs here at all, or whether `phase0Step1()`'s slice should be generalised to cover both skills rather than duplicated.
- Whether to derive the flag from `candidates.mjs`'s own `OPTIONS` declaration, as PR #850 does for the spec copy, so a rename reddens instead of minting the next stale copy — or to assert the literal string, as `phase0Step1()` does.

Labelled `needs-triage` rather than `ready-for-agent`: the defect is confirmed, but the approach above is a real fork and `ready-for-agent` is triage's verdict to give, not the fleet's.

## Explicitly not this ticket

- The dated design specs' `--require-label` mentions — PR #850 deliberately left those unpinned, on the ruling that a historical record must not be made to track the code.
- `docs/specs/2026-07-24-fleet-cockpit-design.md` and `board.mjs`, whose `--label` is `gh`'s own flag and correct. PR #850 pins the cockpit spec's usage as the accept-side control.

feigi:
`next-ticket/SKILL.md` names `candidates.mjs`' label flag in its candidate-scan
step, and nothing asserted that spelling. `--label` is `gh issue list`'s own
correct flag, so a reader writing from a `gh` example writes the one spelling
`candidates.mjs` refuses at exit 2 — the defect already fixed in run-team's rule
(#173) and in the run-team design spec (#240). This is another live carrier of
it, and not the last one nothing asserts on: see **Carriers** below.

Sizing: light

## Both forks, as the Agent Brief rules them

**Fork 1 — generalise the slice helper, do not duplicate it.** The generalised
helper already existed: `prose-pin.mjs` exports `between(text, from, to, what)`,
which `candidates-exit3-prose.test.mjs` already uses for these two exact slices.
`candidates.test.mjs` had its own inline `indexOf`/`indexOf`/`slice` pair,
written out twice. This PR imports the shared helper and deletes both, so the
new pin adds no copy and the file is left with fewer than it had.

Verified the run-team pin still passes for the reason it passed before, not by
luck: `between` searches the end anchor from `at + from.length`, and neither
start anchor contains its own end anchor, so it cannot skip a hit the inline
`indexOf(to, at)` would have found — the slice is byte-identical.

**Fork 2 — derive the flag from `OPTIONS`.** The pin calls the existing
`declaredLabelFlag()`, which reads `candidates.mjs`' own declaration, so a
rename of the flag reds here instead of minting the next hand-copied spelling.
Guarded the same way it already was: called from a test body, every step
asserted with a message naming `OPTIONS` as the thing that moved, because an
unguarded match at module scope takes the whole file down at import on a
zero-behaviour reformat.

## What the pin asserts, and what it does not

**It is scoped to the TOKEN, not to the invocation that carries it.** Measured
both legs at this commit: deleting the candidate step's only fenced block and
naming the flag in prose instead leaves the suite green; deleting that same
fence with no mention of the flag reds, with this pin's own message. So the pin
holds "the flag named here is the one `OPTIONS` declares" and says nothing about
whether a runnable command is present. It was deliberately **not** narrowed to
the fenced block: that would buy a different invariant from the flag spelling
#851 is about, one nothing else in this suite claims, and widening the ticket to
it is not this PR's business.

## What the pin would wrongly refuse, and what was done about it

A derived pin can red on a change that is perfectly correct, so it is bounded
and its accept side is tested rather than assumed.

- **Bounded to the candidate step, and that bound is load-bearing.** Step 7 of
  this same skill says "Never fold `--label` into the create", where `--label`
  is gh's own flag and correct (#375). A file-wide negative would red on a
  sentence it has no business reading.
- **A reflow of the skill must stay green.** The first version of the accept-side
  control reddened on one, and that was found by running it rather than by
  reasoning about it: its non-vacuity guard asserted its own input had changed,
  and unwrapping an already-unwrapped file changes nothing, so it refused the
  exact maintainer edit it exists to bless. It now reflows in both directions
  and requires only that one of them differ, which cannot go vacuous whatever
  the file's current wrapping and cannot red on a reflow either.
- **A second `*label` flag in `OPTIONS`** reds `declaredLabelFlag()` with a
  message naming the declaration. That refusal is pre-existing and deliberate:
  indexing the first hit instead would pin silently to the wrong flag.

The pin is applied through a function taking text, so its own test puts it in
front of the refused spelling (must red), the live file carrying that legitimate
`--label` (must stay green), and both reflows (must stay green) — each control
asserting it is not itself vacuous.

## Carriers: covered, and deliberately left

Enumerated by text, not by the ticket's line numbers. `git grep -l --
--require-label docs skills` lists them all.

Covered: the candidate-scan invocation in `next-ticket/SKILL.md`.

Already pinned, unchanged: run-team's phase-0 rule; the run-team design spec's
candidate-scan step; the cockpit spec's `gh issue list --label`, which stays the
accept-side control.

Deliberately left:

- The dated design specs' mentions — the Brief's exclusion, on the ruling that a
  historical record must not be made to track the code. **This PR's review found
  that exclusion leaves real unpinned carriers, and one of them is value-bearing:**
  `docs/specs/2026-07-23-fleet-plugin-design.md`'s `candidates.mjs` synopsis row
  spells `[--require-label L]`, and rewriting it to `[--label L]` leaves the
  whole suite green (measured at this commit).
  `docs/specs/2026-07-30-fleet-trust-the-label-design.md` carries a value too.
  Filed as **#1196** rather than pinned here — #851's subject is
  `next-ticket/SKILL.md`, and pinning a docs carrier it does not name widens the
  PR past its ticket.
- The cockpit spec and `board.mjs`, whose `--label` is gh's own flag and correct
  — the Brief's other exclusion.
- `fleet-tick.mjs`'s `supply()`, which passes the flag as an argv element. A
  live carrier, and not prose: a wrong spelling there fails when the tick runs
  rather than when a reader copies it, so it is a different class from the one
  this ticket is an instance of. Named here rather than swept in.
- `review-and-fix.md` and `docs/adr/0001-…`, which name `candidates.mjs
  --require-label` with **no value**, as the thing an unlabelled issue is
  invisible to. Nothing there is copied into a shell. That is the category line
  #1196 records, and it is why the plugin-design synopsis row does not sit on
  this side of it.

## Verification, re-run at this commit

- Suite: `node --test skills/fleet/scripts/*.test.mjs` → every test passes, none
  failed, none skipped.
- Mutant, run against the real file and restored with `cp`: rewriting the
  candidate scan to the spelling `candidates.mjs` refuses reds the tests this PR
  adds and no other test in the suite.
- Accept-side control, same method: reflowing the skill leaves the suite green.
- Token-scope, both legs: fence deleted with the flag named in prose → suite
  green; same fence deleted with no mention → reds with this pin's own message.
- Unpinned carrier: `[--require-label L]` → `[--label L]` in the plugin-design
  synopsis row → suite green, the measurement behind #1196.
- Guard, both legs: one-lining `OPTIONS` reds only the tests that derive the
  flag, each with `OPTIONS block no longer matches — update this test`, and the
  rest of `candidates.test.mjs` still runs and passes; adding an unguarded
  module-scope `.match(...)[0]` on top of that reformat instead collapses the
  file to a bare `TypeError` with zero passes.

Closes #851


feigi:
Review deferrals filed: **#1196** — `candidates.mjs`' `--require-label` spelling has unpinned carriers in `docs/`, and the `docs/specs/2026-07-23-fleet-plugin-design.md` synopsis row among them is value-bearing (`[--require-label L]`), so it is not in the no-value category this PR's body exempts. Measured at `1210fbf`: rewriting that row to `[--label L]` leaves `node --test skills/fleet/scripts/*.test.mjs` green. Deferred rather than pinned here because #851's subject is `next-ticket/SKILL.md`.

Nothing was refuted in this pass, so there is no `wontfix` record issue. The other two findings were confirmed by measurement and applied in `1210fbf` as comment corrections — the pin itself is unchanged, and the PR body has been reconciled against the final diff.

PR #1198, opened 2026-09-02
fix(staleness): end-of-options separator for a --gone/--present needle starting with --

### LINKED ISSUE #818: staleness.mjs cannot check a ticket whose defect string starts with `--` — the refusal is the safe path, the capability is what is missing
Deferred from PR #813 review (issue #238).

**dimension:** `correctness`

## The finding

`staleness.mjs` takes the ticket's quoted string as the value of `--gone`/`--present`. A value that starts with `--` is refused by `skills/fleet/scripts/arg.mjs:87` (`value.startsWith("--")`), so a ticket whose defect IS a flag spelling cannot be probed at all. Measured on the pushed head:

```
$ node skills/fleet/scripts/staleness.mjs --path skills/fleet/scripts/arg.mjs --gone '--require-file'

staleness: --gone needs a value
exit=2
```

The class is in this backlog rather than hypothetical: #240's defect is the spelling `--label ready-for-agent`.

## Why deferred, and what PR #813 did apply

The DEFECT half was ruled refuted. Exit 2 is the designed safe path: the ticket is offered and annotated as unchecked, which is exactly what the exit-code table instructs for exit 2. No ticket is dropped and no false verdict is emitted.

The DOCUMENTATION half was applied on the PR: `skills/fleet/skills/run-team/SKILL.md:228` now states the restriction in the usage block, and `skills/fleet/scripts/arg.mjs`'s header no longer claims no caller passes such a value.

What remains is a capability, not a defect: letting a `--`-prefixed needle be checked at all — an end-of-options separator (`--gone -- '--require-file'`), reading the needle from stdin, or something else. Which shape is wanted is a maintainer call, so this is `needs-triage`.

feigi:
## Summary

- `staleness.mjs`'s `--gone`/`--present` read through `arg()`, whose value guard refuses any value starting with `--` (#61/#169) — so a ticket whose defect IS a flag spelling, #240's `--label ready-for-agent` being one this backlog carries, could never be checked.
- #818's Agent Brief ruled the shape: an end-of-options separator, `--gone -- '<value>'` / `--present -- '<value>'`, scoped locally to `staleness.mjs` rather than touching `arg.mjs`'s shared value guard used by every other fleet script.
- Given bare, with no `--` immediately before it, `arg()`'s refusal is unchanged and unweakened — the exact reproduction from the Agent Brief still refuses: `node staleness.mjs --path src.mjs --gone --require-file` → `exit 2`, `staleness: --gone needs a value`.
- `run-team/SKILL.md`'s usage block and `arg.mjs`'s own comment on the refusal are reworded to describe the separator instead of stating the restriction as absolute.

## The class this feature exists for, and how the first cut missed it

Review found the separator did not work for a member of its own motivating class: a needle that spells one of this script's own flag names. Resolution ran by fixed code order — `needleArg("gone")`, then `needleArg("present")`, with `arg("path")` ahead of both — each scanning the whole argv for one name, so a separator-quoted needle was matched as a control token.

Measured at `8a14614` (fixture: bare origin + clone, `src.mjs` tracked and carrying the literal strings `--gone` and `--path`; each row is `node staleness.mjs <args>` from the clone):

| argv | at `8a14614` | at `84b3d2a` |
|---|---|---|
| `--path src.mjs --present -- --gone` | `exit 2` `--gone needs a value` | `exit 1`, needle `--gone`, verdict `fixed` |
| `--path src.mjs --present -- --gone extra` | `exit 2` `--gone and --present are opposite questions; give one` | `exit 2` `unexpected argument 'extra'` |
| `--gone -- --path --path src.mjs` | `exit 2` `--path needs a value` | `exit 0`, needle `--path`, path `src.mjs`, verdict `live` |

Every one of those names a flag the caller never typed, and the middle row reports the wrong error entirely — `--gone` having silently taken `--present`'s trailing token.

`separatedNeedles()` replaces `needleArg()`: **one** left-to-right pass over argv carrying both names, consuming each `<flag> -- <value>` triple before `arg()` reads anything. Position in argv decides which flag owns a token, never the order the reads are written in, so a token already handed to a preceding flag can no longer be recognised as a control token — and `--path` is read after the pass for the same reason.

## Collision check

#567 and #584 (the two tickets that made the shared `arg.mjs` module "busy" per the Agent Brief) are both CLOSED and already merged — PR #1128 and PR #1153. No collision; this PR does not touch `arg.mjs`'s exported functions or their behavior, only a comment.

## Class enumeration

- `staleness.mjs` has exactly two needle-bearing flags — `grep -n 'separated\.\|arg("' skills/fleet/scripts/staleness.mjs` shows `--gone`, `--present` and `--path`, of which `--gone`/`--present` are the needle pair. Both are wired through the same `separatedNeedles()` pass and each has its own accept-path test.
- `arg.mjs`'s own comment names `staleness.mjs` as the one caller this restriction was written against (`grep -n 'staleness.mjs' skills/fleet/scripts/arg.mjs` → lines 89 and 93). `ledger.mjs`'s free-text tail is a different, already-narrowed mechanism (#584) and is out of scope.

## What this could wrongly accept, and the controls against it

Loosening a guard that exists for a measured reason (#61/#169: a flag swallowing the next flag as its own value) risks re-opening it if the separator is too permissive. Four mutants were applied to the tree, run, and reverted with `cp` — never `git checkout`. `node --test skills/fleet/scripts/staleness.test.mjs` under each:

| mutant | failing tests |
|---|---|
| separator accepted with **or** without the `--` (guard disabled) | the no-separator control, and the malformed-order case |
| the pre-review fixed-order resolution restored (`staleness.mjs` at `8a14614`) | the two flag-shaped-needle cases, and nothing else |
| the both-flags guard reverted to `process.argv.includes("--gone") && process.argv.includes("--present")` | the both-flags-through-the-separator case, and nothing else |
| `--` taken from anywhere later in argv rather than immediately after the flag | the malformed-order case, and nothing else |

The third and fourth are gaps this review opened and this commit closes. The `includes()` revert left the whole file green beforehand, while `--path src.mjs --gone -- a --present -- b` exited 0 with a full-confidence `live` under `--gone` and `--present` silently dropped — the separator pass splices both flag tokens out of argv, so a raw scan finds neither. The scan-anywhere mutant was undetectable beforehand because the malformed-order case's argv (`--gone --require-file --`, nothing after the trailing separator) produced byte-identical exit, stdout and stderr under both readings, `exit 2 / staleness: --gone needs a value`; a real value token after that `--`, carried by the fixture, separates them.

## Input shapes tested

- Separator immediately after the flag, `--`-prefixed value → accepted, real verdict.
- Separator immediately after the flag, ordinary value → accepted, same verdict as without the separator.
- No separator, `--`-prefixed value → still refused (the control).
- Needle whose literal text is the sibling flag's own name (`--present -- '--gone'`) → answered, needle carried verbatim into the payload.
- Needle spelling `--path`, with the real `--path` behind it → answered, and `--path` still resolves to its own value.
- Both flags given, at least one in separator form → refused, `opposite questions`.
- `--` after the value instead of before it, with a real token behind it → still refused, `needs a value`.
- `--` as the very last token, nothing after it → refused, `needs a value`.
- The flag given twice, both times with the separator → refused. This is now deliberate rather than incidental: `separatedNeedles()` steps over a repeat instead of overwriting the first, which strands that repeat's `--` for `sweep()` — `--path src.mjs --gone -- a --gone -- b` exits 2 `unknown flag --`, the same refusal that shape earned before the pass existed, and a case pins it.

## Also in this commit

The both-flags guard is spelled `gone && present`, matching the truthiness test on the line below it. Neither `arg()` nor `separatedNeedles()` can return a falsy non-null value — both refuse an empty-or-blank one — and a differential run over 2028 argv shapes (both flags in bare and separator form, values including `"0"`, `""`, `"  "`, `"--x"`, `"--"`, three orderings) found the two spellings byte-identical on exit code, stdout and stderr, with 364 of those shapes reaching this guard.

## Test plan

- [x] `node --test skills/fleet/scripts/staleness.test.mjs` → `tests 25 / pass 25 / fail 0` at `84b3d2a`
- [x] `node --test skills/fleet/scripts/*.test.mjs` → `tests 1657 / pass 1657 / fail 0` at `84b3d2a`; the same command at `8a14614` reported `tests 1653 / pass 1653 / fail 0`
- [x] Each of the four mutants applied, run, and confirmed to red only its intended tests, then restored with `cp`

Sizing: light

Closes #818


feigi:
Review applied at `84b3d2a`.

**Applied — three defects, six `survived` findings.** The `types`/`correctness`/`silent-failure` findings on the separator resolution were one defect (needle resolution ran by fixed code order, so a separator-quoted needle spelling a sibling flag was read as that flag); the two on the both-flags guard were one (the guard's resolved-value reading was pinned only in its bare form); the `tests` finding on the malformed-order case was one (it did not discriminate the bug class its own comment names). Each remedy was run before it shipped, and each pin was mutation-verified to red only its intended case — the table in the PR body carries the four mutants and what each killed.

**Deferred — the `suggestion` band, all of it checked.** The workflow budgets that band 0 refuters, so each entry was scoped and given one refuter, biased to refuse, before the tree was touched. Four came back refuted on measurement; one was not refuted and shipped in `84b3d2a` (the both-flags guard's `!== null` spelling). Record: #1203, closed `wontfix`, with two below-the-bar residuals and a re-open trigger per entry.

`ledger.mjs check` returned `verdict: "tracker-hit"` at exit 3 before filing; every row it printed belongs to a different PR and scores on the shared title template, and `gh issue list --state all --search "1198 in:title"` returned nothing.

`node --test skills/fleet/scripts/*.test.mjs` → `tests 1657 / pass 1657 / fail 0` at `84b3d2a`; the same command at `8a14614` reported `tests 1653 / pass 1653 / fail 0`.

Issue #1199, opened 2026-09-02
ledger.test.mjs: two independent sightings of subprocess/pipe tests failing under concurrent load — precedent says this class was a REAL race, not a flake


feigi:
Two independent `/fleet:run-team` members hit failures in `skills/fleet/scripts/ledger.test.mjs` during the 2026-09-02 run, in different tests, both only under concurrent suite load, both passing on isolated re-run. Filed by the controller because **the obvious reading — "flaky test, ignore" — is the reading this repo has already had wrong once.**

## The two sightings

**1. `impl-851`, working #851.** `ledger.test.mjs`'s *"the search query carries at most three sanitised terms and no qualifiers"* failed with `actual: undefined` — a subprocess that produced **no output at 20s**. Passed in both the clean baseline and the final run. The member reported it rather than acting on it; it does not read `next-ticket/SKILL.md` and was untouched by that PR.

**2. `impl-818`, working #818.** `ledger.test.mjs`'s **oversized-payload pipe test** failed on the first full-suite run. Re-run 3x in isolation, clean each time (90/90). Never touched `ledger.mjs` or `ledger.test.mjs`.

Different tests, same file, same trigger shape: concurrency. The fleet runs several suites at once **by design**, so this is the normal operating condition, not an exotic one.

## Why this is not obviously a flake — the precedent

`ledger.mjs check` returned `verdict: tracker-hit` on this subject, naming **#298** and **#302**. Both are closed, both concern `candidates.test.mjs` rather than this file — so they are the same *class* on a different file. Reading how #298 closed is what makes this worth filing:

> *"this is not a flaky test. It is a real, load-sensitive production race that the test correctly reports."*

#298 closed as **already implemented**, with the fix landing in the shared `makeDie` rather than in one script — a `writeSync` throwing **EAGAIN when queued on a non-blocking pipe** (#299 / #328). So the last time this repo saw "two tests in one file go red only under parallel load", the tests were right and the production code was wrong.

**And sighting 2 is a pipe test in `ledger.mjs`** — a file whose own history already includes a payload being abandoned at the pipe buffer and still exiting 0 (#246, which is why `read`, `row`, `filed`, `ruled` and both `check` arms are now pinned to reach a pipe whole). That is the same organ and the same failure family.

## What is NOT established

- **No reproduction.** Both members re-ran clean. Nobody has driven it deliberately.
[...]

PR #1200, opened 2026-09-02
docs(triage-labels): drop the filing criterion from the Meaning column

### LINKED ISSUE #836: docs/agents/triage-labels.md:9 glosses ready-for-agent as "Fully specified", which the repriced filing bar contradicts
`Deferred from PR #834 review`. Dimension: `correctness`. Severity as raised: `important`.

## The finding

`docs/agents/triage-labels.md:9` glosses `ready-for-agent` as:

```
| `ready-for-agent`          | `ready-for-agent`    | Fully specified, ready for an AFK agent  |
```

The repriced filing bar routes a **confirmed defect whose remedy is still open** to that same label. A remedy-open ticket is by construction not "fully specified", so a triager reading the table and a filer reading the bar resolve the same label to opposite standards.

## Why it was not folded into #834

Two independent reasons, both settled by running:

**1. The contradiction predates #834 — it is not introduced by this diff.** The filing bar is stated inline in `skills/fleet/commands/review-and-fix.md`, and that half landed before #834 branched:

```
$ git log -1 --format='%H %ad %s' 6d0c1f6
6d0c1f6a9000ad6415f1c229c71e8cc73b122a13 Tue Aug 18 14:46:54 2026 +0200 docs(fleet): state step 5's filing bar inline instead of pointing at phase 0

$ git show --name-only --format= 6d0c1f6
skills/fleet/commands/review-and-fix.md

$ git merge-base --is-ancestor 6d0c1f6 origin/main && echo "ancestor of origin/main"
ancestor of origin/main

$ git diff --name-only origin/main...d6873b4 -- docs/agents/triage-labels.md
(empty — #834 does not touch the file)
```

So both sides of the contradiction were already on `origin/main` before #834 existed. #834 adds the ADR that *records* the reprice; it does not create the mismatch.

**2. It is outside #239's acceptance criteria.** #239's ACs name `review-and-fix.md`, `run-team/SKILL.md`, `CONTEXT.md` and a new `docs/adr/` entry. `docs/agents/triage-labels.md` is not among them.

## Scope, and why this is a wording fix rather than a behavioural one

The file scopes itself to vocabulary, in its own words:

- `:3` — "This file maps those roles to the actual label strings used in this repo's issue tracker."
- `:13` — "All five roles resolve — label string equals role name."
- `:19` — "Edit the right-hand column to match whatever vocabulary you actually use."

`CLAUDE.md` cites it only for "Label string equals role name for all five roles", i.e. it delegates the naming, not the filing criterion. Nothing executable reads the Meaning column — the mapping that scripts depend on is the label *string*, which is unaffected.

That is why this was raised as `important` and is being filed rather than applied: it is a one-cell doc-staleness fix in a file #834 has no business editing, not a live routing bug.

## Suggested wording, as raised

> Defect confirmed and agent-actionable; the remedy may still be open

Not applied and not verified here. Whoever picks this up should re-derive the current gloss and the current bar from the tree rather than trusting the quotes above, and should decide whether the Meaning column is the right home for a criterion the file otherwise disclaims.

feigi:
The `ready-for-agent` gloss promised a fully specified ticket; the repriced filing bar routes a confirmed defect with an open remedy to that same label.

```
$ git show 1fcabf5:docs/agents/triage-labels.md | grep -n 'ready-for-agent'
9:| `ready-for-agent`          | `ready-for-agent`    | Fully specified, ready for an AFK agent  |

$ grep -c 'The bar here is whether the DEFECT is confirmed — never whether the remedy is settled' skills/fleet/commands/review-and-fix.md
1
```

## Root cause, which is not the one the ticket assumed

The ticket reads this as one badly worded cell. It is not. The Meaning column is a **transcript of the tracker's own label descriptions**, and a transcript has no owner, so it drifts against whichever side moves. ADR 0001 moved the filing bar; the transcript stayed.

Run at this branch's head, comparing every cell to the live label it was copied from — this is also the row-by-row check the ticket asks for:

```
$ for L in needs-triage needs-info ready-for-agent ready-for-human wontfix; do
    cell=$(awk -F'|' -v l="\`$L\`" '$2 ~ l {gsub(/^ +| +$/,"",$4); print $4}' docs/agents/triage-labels.md)
    live=$(gh api repos/feigi/claude-config/labels/$L --jq '.description')
    [ "$cell" = "$live" ] && m=SAME || m=DIFFERS
    printf '%-16s %-7s cell=%-38s live=%s\n' "$L" "$m" "$cell" "$live"
  done
needs-triage     SAME    cell=Maintainer needs to evaluate this issue live=Maintainer needs to evaluate this issue
needs-info       SAME    cell=Waiting on reporter for more information live=Waiting on reporter for more information
ready-for-agent  DIFFERS cell=Ready for an AFK agent                 live=Fully specified, ready for an AFK agent
ready-for-human  DIFFERS cell=Needs a maintainer, not an AFK agent   live=Needs a maintainer session, not an AFK agent
wontfix          DIFFERS cell=Will not be actioned                   live=This will not be worked on
```

The tracker is mutable and nothing re-derives that output; it is an as-of reading, not a live invariant. The `DIFFERS` on the two corrected rows is this PR. Rows still matching byte-for-byte are the ones where neither side has moved — which is what identifies the column as a copy rather than the file's own prose.

## The design question the ticket asks: restate, point at, or drop

**Drop.** The Meaning column is not the right home for a filing criterion, and the durable form is not a pointer either.

- **Restating** — the wording the ticket suggests, *"Defect confirmed and agent-actionable; the remedy may still be open"* — reproduces the exact mechanism that produced this ticket. A second copy of a criterion drifts the same way the first one did.
- **Pointing at the bar** adds a cross-document dependency nothing pins. ADRs are superseded by name, so a pointer at `docs/adr/0001-…` silently addresses a superseded record once the bar moves again, and no reader arrives at this file needing the bar in the first place: `:3` scopes the file to the role/label-string mapping, and `CLAUDE.md` cites it only for that. The consumers who do need the bar already state it inline (`review-and-fix.md` step 5, ADR 0001).
- **Dropping** leaves the cell carrying only what the file owns — the role's name in words. It cannot drift, because it no longer asserts a filing standard in either direction.

That last clause is deliberate. Over-correcting into *"the remedy may still be open"* would let a reader conclude a remedy-open ticket needs no specification at all — a new false claim pointing the other way. `Ready for an AFK agent` says nothing about specification, so it makes neither claim.

## The other four rows

- **`needs-triage`**, **`needs-info`** — no change. Neither states a filing criterion, both still match the live description exactly (above), and both agree with how the repo routes them.
- **`ready-for-human`** — **corrected**, same class of defect, and an addition rather than a cell #836 named. The ticket raises `ready-for-agent` alone; this row is corrected under its acceptance criterion that *"the other four rows are checked for the same drift while the file is open — a table with one corrected cell and four unexamined ones is a worse artifact than before"*. Its drift vector is also not the re-sync named above: the tracker's own description reads `Needs a maintainer session, not an AFK agent`, so this row had narrowed toward implementation on the repo side alone. The gloss said `Requires human implementation`, which excludes work needing a human *before* implementation, and the label routes exactly that:
  ```
  $ grep -c 'ready-for-human` needs a human to brainstorm first' skills/fleet/skills/run-team/SKILL.md
  1
  $ grep -c 'needs human hands — external access, manual testing, judgment during the work' skills/fleet/skills/run-team/SKILL.md
  1
  ```
- **`wontfix`** — no change. `Will not be actioned` differs in wording from the live `This will not be worked on` but makes the same claim, and it states no criterion. Rewording it would be churn on a cell that is not wrong.

## The class, and what this covers

The class is *a cell in this table restating a standard that is owned elsewhere*. Enumerated from the mechanism (a transcript of an external source), not from the ticket's prose, which frames it as one cell:

- **Covered:** `ready-for-agent` (filing bar), `ready-for-human` (routing criterion).
- **Left, deliberately:** the three rows above, none of which restates a standard.
- **Left, cannot be fixed from this repo:** the tracker-side label descriptions. `ready-for-agent`'s live description still reads `Fully specified, ready for an AFK agent`. Editing it is an outward-facing change to a shared system, outside this ticket's scope, and this branch has no approval channel for one. It is reported rather than applied — and it is precisely why this PR carries a pin.

Dropping the criterion adds no cross-document dependency, which is the main thing the pointer alternative would have added. What it does add is a **divergence between this file and the tracker description it was copied from**, on a column whose whole history is copying. Nothing outside this repo will fix that source, so the re-addition vector stays live indefinitely: the next person syncing this column from the tracker restores the contradiction verbatim.

`skills/fleet/scripts/triage-labels-gloss-prose.test.mjs` is what pins that. One row slice per corrected row, and on each slice: the role-to-label-string mapping, that the gloss still turns on whether an AFK agent can take the work, and the absence of the restated criterion.

That middle assertion was added after review measured the pin vacuous in the direction it most needed to hold — **both pins stayed green with the Meaning cell they guard emptied entirely.** The label-string companion pins columns 1 and 2, so an emptied column 3 left the negative passing on nothing at all, which is the exact vacuity that companion's own comment claimed to prevent. On a PR whose whole subject is a gloss that drifted, the gloss could silently vanish. `assert.match(r, /AFK agent/)` pins the load-bearing content — agent versus not-an-agent — without re-pinning the wording this PR set out to strip.

Both negatives pin a **stem, not a phrase**, because the adjacency form measurably failed: `/[Ff]ully\s+\w*\s*specified/` stayed **green** against `Ready for a fully and precisely specified AFK agent` — a one-word gap does not span two. Replaced with `/specifi/i`, and `/implementation/` with `/implement/i` for the same reason.

Mutation matrix re-derived on a copy of the tree, `cp` restore after each: every mutation reds its own pin and no other, and every control stays green — including both emptied-cell mutants, which did not red before `3172d41`. Each mutant and control is named in the test file's header comment and tabulated in `3172d41`'s body.

```
$ env -u FORCE_COLOR node --test skills/fleet/scripts/*.test.mjs 2>&1 | grep -E '^ℹ (tests|pass|fail)'
ℹ tests 1649
ℹ pass 1649
ℹ fail 0
```

That run is at the head this body describes. An earlier full run, on the pre-fix tree, lost `net_kill_tree signals the whole subtree, however the snapshot is ordered` under concurrent load from sibling agents — `pass 1648 / fail 1`. Re-run in isolation the same tree gave `tests 5 / pass 5 / fail 0` at exit 0, and the test passed again in the green run above. That is the already-filed #1099, where this sighting is recorded; it is not a flake and not this diff's.

The label strings are untouched — that mapping is what scripts read.

Sizing: heavy

Closes #836


feigi:
## Review adjudication — the `suggestion` band, checked

The review returned **0 survived, 1 refuted, 4 unverified**, every one of the four at `refutersDispatched: 0` — the workflow's 0-refuter policy budget for the `suggestion` band, not a crash. So nothing adversarial had looked at any of them. Each was scoped, and each in-scope one got a refuter dispatched under *"default to refuted if uncertain; verify by RUNNING something"*.

| entry | dimension | outcome | where it went |
|---|---|---|---|
| `refuted[0]` — spec gloss-mismatch note falsified | `comments` | upheld as refuted (2-0) | #1202 |
| `unverified[0]` — pins green on an emptied gloss cell | `correctness` | **survived its refuter → applied** | `3172d41` |
| `unverified[1]` — `ready-for-human` row out of scope | `correctness` | refuted on three grounds; its final clause applied | #1202 |
| `unverified[2]` — baseline `fail 3 / cancelled 3` is a snapshot artifact | `tests` | dismissed, settled by measurement | #1202 |
| `unverified[3]` — the `:3` citation overstates line 3 | `comments` | **survived its refuter → applied** | `3172d41` |

**#1202** — *PR #1200 review: the suggestion band, checked*, `wontfix`, **closed**. Holds the three checked-and-not-surviving entries with their refutations, plus one residual below ADR 0002's claim bar (`/AFK agent/` matches both rows, so a cell-swap stays green; both narrower forms that kill the swap redden a benign-reword control, which is the brittleness this file's design avoids).

**#1099** — commented, not filed. The full run on the pre-fix head lost `net_kill_tree signals the whole subtree, however the snapshot is ordered` at 43.7s under sibling load; isolation gave `tests 5 / pass 5 / fail 0` at exit 0, and it passed in the green run below. That test is named eight times in #1099's body, so that is its home. (The controller brief attributed it to #1199, which covers `ledger.test.mjs` — same class, different file.)

Nothing was dropped, and nothing new was opened: no entry alleged wrong behaviour that is still standing.

### What changed in `3172d41`

Both pins stayed green with the Meaning cell they guard emptied entirely — the label-string companion pins columns 1 and 2, so an emptied column 3 left the `doesNotMatch` negative passing on nothing at all, which is exactly what that companion's comment claimed to prevent. Adding `assert.match(r, /AFK agent/)` per row closes it. The `:3` citation now quotes what line 3 actually says, and the header states that the `ready-for-human` cell is an addition made under #836's check-all-rows criterion rather than a cell the ticket named.

Matrix re-derived on a copy of the tree, `cp` restore after each — every mutation reds its own pin and no other, every control green, tabulated in `3172d41`'s body.

```
$ env -u FORCE_COLOR node --test skills/fleet/scripts/*.test.mjs   # at 3172d41
ℹ tests 1649
ℹ pass 1649
ℹ fail 0
EXIT=0
```

PR #1201, opened 2026-09-02
test(workflows): guard `export const meta` first, and record the trap it guards

### LINKED ISSUE #853: workflows/: a statement before `export const meta` silently drops a workflow from the registry — no error, it just is not there
Measured by `impl-538` while implementing #538 (PR #852), as a **control** rather than as the thing it set out to test. Surfaced to the controller and filed here rather than by the member, to avoid unilateral scope expansion.

## The trap

A workflow script whose body has any statement **before** `export const meta = {...}` is silently absent from the workflow registry. No error, no warning, no diagnostic — it simply does not appear in the available list, which is indistinguishable from a file that was never written, a bad path, or a harness that rejected it for the reason you were actually testing.

The `Workflow` tool's own contract states the requirement (*"Every script must begin with `export const meta = {...}`"*). What is undocumented anywhere is the **failure mode**: silent omission rather than a load error.

## How it was isolated, and why that matters

This is the interesting part, and it is the reason the ticket exists rather than a one-line note.

impl-538 was measuring whether the Workflow sandbox permits `import`/`require` of `node:fs` (#538). Its first probe file carried a static `import` and did not appear in the registry. The obvious reading — *the import was rejected* — is **wrong**, and it produces the right verdict for the wrong reason.

The control that caught it: a twin file with the import **removed** was also absent. That falsified "the import caused the absence" and exposed the real rule. Re-running with `export const meta` first isolated the import cleanly, and the actual refusal turned out to be a different mechanism entirely (all import forms fail at syntax level — see PR #852's in-tree record).

So the cost of this trap is not a broken workflow. It is a **misattributed cause** in any measurement that uses registry presence as a signal, and it reads exactly like a successful negative result.

## Current exposure

Latent, not live. `workflows/` holds exactly one file, `review-pr.js`, and it begins with `export const meta` — verified at `origin/main`:

```
$ git show origin/main:workflows/review-pr.js | head -1
export const meta = {
```

So nothing is broken today. The exposure is that `workflows/` is expected to grow, `review-pr.js` is the fleet's **default review path**, and a file that silently fails to register produces no red — the fleet would fall back to the hand-dispatched reviewer, where `selectDimensions` does not run and nothing budgets the adversarial pass, and the only symptom is a review that is quietly thinner.

## Open decisions for triage

- Whether a guard is worth it, and where: the cheap form is a test asserting every `workflows/*.js` file's first non-comment statement is `export const meta`. That is a shape assertion, so it wants the usual both-ways mutation check.
- Whether to record the failure mode in prose instead — the fleet's own docs are the place a future measurement would look before using registry presence as a signal.
- Whether this belongs to this repo at all, given the behaviour is the harness's and not ours. Filing it here because the consequence lands on `workflows/review-pr.js`, and because the misattribution risk is what bit a fleet member.

Labelled `needs-triage`: the behaviour is confirmed and controlled for, but whether it warrants a guard, prose, or nothing is a judgement, and `ready-for-agent` is triage's verdict to give.

feigi:
A workflow script with any statement before `export const meta` was observed absent from the workflow registry — no error, no warning, nothing that tells it apart from a file that was never written, a wrong path, or a harness that refused the construct you were probing. Latent, not live: `workflows/` holds only files that begin correctly.

Sizing: light

## What ships

**The guard.** `skills/fleet/scripts/workflow-meta-first.test.mjs` asserts every `workflows/*.js` file's first non-comment statement is `export const meta`. Files are DISCOVERED with `readdirSync`, never listed — a hardcoded list reproduces this ticket's own failure mode one level up, where the workflow nobody remembered to add reads exactly like a workflow that passes.

**The recorded trap.** The prose half is what actually cost time, and a passing test does not teach it: #538 came within one control of the right verdict for the wrong reason, reading its probe's registry absence as "the import was rejected". It is recorded in `workflows/review-pr.js` at the re-measure instruction — the sentence a future measurer reads immediately before writing the throwaway probe that would trip it — and pinned, slice-bounded, so deleting it reds.

The behaviour is the harness's and cannot be settled from this repo, so it is cited as observed rather than as a specification. The guard is ours: asserting the shape of our own files needs no cooperation from the loader.

## The guard shipped with the defect it exists to catch

Found in review, fixed in `fb90a07`. `stripComments` is line-based and blanks a line WHOLE when it opens with `/*` or when it closes a block — code on that line included. So a real statement sharing a physical line with a block comment's close was **erased before `firstStatement` looked**, and the guard went GREEN on a file carrying a statement before `export const meta`. `/** @type {{a:number}} */ const cfg = { a: 1 };` is an ordinary JSDoc one-liner that reaches it.

`firstStatement` now breaks every `*/` onto its own line before stripping, so the tail lands where the stripper cannot blank it. Fixing the shared stripper instead was refused for the reason `strip-comments.mjs` already states: a span regex opens a comment at a glob inside a string literal and swallows the code after it.

Three further corrections in the same commit, each measured rather than argued:

- The negative table had **no same-physical-line case**, so unanchoring `META_FIRST` shipped undetected. That row is now the only one that discriminates the `^` — every other row refuses under the mutant anyway, because none of them puts `export const meta` on the line the matcher is handed. The `*/` split alone does **not** close it; the table below measures that.
- The four prose pins asserted presence but never that nothing **between** them reverses the guidance. Collapsed into one contiguous span running from the control through the positive signal to the sentence's closing period.
- An empty or comment-only workflow reported "has a statement before `export const meta`" when it has no statement at all. #853's whole provenance is a misattributed cause, so the guard now names its own two causes apart.

## Both directions, run at the shipping commit

Whole suite, not one file, since a sibling can red on the same mutant. Mutations applied in the worktree and restored with `cp` — a copy of this tree cannot run the suite validly, because `derive_workspace_id` shells out to `git rev-parse` and a non-git copy falls back to the directory basename.

| mutation | result |
|---|---|
| statement prepended to `workflows/review-pr.js` | tests 1651 / pass 1649 / fail 2 — the meta-first guard, **and** `review-pr.js parses` in `review-pr-reads.test.mjs` |
| CONTROL: licence block + line comment + blank lines prepended, export not moved | tests 1651 / pass 1651 / fail 0 |
| `*/` split reverted in `firstStatement` | tests 1651 / pass 1650 / fail 1 — the only red is the refusal table, on "code on the line that closes a block comment" |
| `META_FIRST`'s `^` dropped | tests 1651 / pass 1650 / fail 1 — the only red is the refusal table, on "a statement before the export on one line" |
| exception clause spliced into the throwaway-probe instruction | tests 1651 / pass 1650 / fail 1 — the only red is the prose pin |

An earlier revision of this description claimed the first row's only red was the meta-first guard. Re-measured at `fb90a07`: it is two, the second being the pre-existing parse test. Corrected rather than carried.

On the new test file alone, where a whole-suite run adds nothing:

| mutation | result |
|---|---|
| `META_FIRST`'s `^` dropped **and** the new same-line row deleted | pass 4 / fail 0 — so the row kills that mutant, not the `*/` split |
| an empty workflow file, and a comment-only one | both red naming the no-statement cause; a file with a real statement first still reds naming the other one |
| CONTROL: the recorded trap's words reflowed at another width | pass 4 / fail 0 |
| CONTROL: a legal licence-block + line-comment workflow added | pass 4 / fail 0 |

Full suite at this commit: `node --test skills/fleet/scripts/*.test.mjs` → tests 1651 / pass 1651 / fail 0.

## What the guard accepts, and what it refuses

Accepts, each pinned as a fixture: a leading licence `/* */` block, a JSDoc block, a run of `//` lines, blank lines, a comment block followed by blank lines, a trailing comment on the export line, irregular spacing in the declaration, and a commented-out statement ahead of the real export.

Refuses, and correctly: a `const`, a static `import` or a call before the export; the export present but preceded; the export appearing only inside a comment; an empty file; a file of only comments. Plus the four rows the review added — a statement on the line that closes a block comment, code after a one-line block comment, a JSDoc one-liner declaration, and a statement before the export on one physical line.

`"use strict"` is refused too, and the refusal is correct — settled with `new AsyncFunction('"use strict";\nconst meta = {};')`, which compiles, so a directive prologue is legal as a function body and nothing local rejects it; it is still a statement before the export, so the shape rule drops the file anyway. Likewise a shebang, settled with `new AsyncFunction("#!/usr/bin/env node\nconst meta = {};")` → `SyntaxError: Invalid or unexpected token`; the hashbang grammar is legal only at the start of a Script or Module source, and a workflow body is neither, so refusing it costs nothing that could have worked.

**Two ceilings remain, and both are LOUD** — a red with an obvious fix, never a silent pass. That is the trade, and after the `*/` split it runs in that direction only:

- `/* c */ export const meta = {` on one physical line reds: the split leaves the export indented and `META_FIRST` anchors at `^`. Refusing an indented export costs nothing here, because `.github/workflows/ci.yml` and `review-pr-reads.test.mjs` both strip with `/^export /m` before compiling, and that pattern does not match an indented line either — such a file already reds with `SyntaxError: Unexpected token 'export'`.
- A `*/` inside a string literal or inside a `//` comment **above** the export splits that line as well, so its tail reads as a statement. Every file under `workflows/` reaches `export const meta` before its first `*/`.

## The empty glob

A glob-driven test newly asserts over files that do not exist yet, so the question is what it does over none. Measured both ways against an empty `workflows/`: with the floor, `fail 1` naming the vacuity; with the floor removed, `tests 4 / pass 4 / fail 0` — green having asserted over nothing, which is this ticket's own silent-absence defect arriving through the guard written to prevent it. The floor is a floor, not a count: a pinned total would have to be edited every time a workflow is added, and that is the step nobody remembers.

## Deferred from review

- #1204 (`ready-for-agent`) — discovery is flat and `.js`-only, so a `.mjs` or nested workflow is silently uncovered. Declared deliberately by the implementer and left as declared; the remedy needs a fact this repo cannot settle.
- #1205 (`wontfix`, closed) — the `suggestion` band, checked: two findings refuted with evidence, two recorded below the claim bar.

Closes #853


feigi:
Review applied at `fb90a07`. 14 findings, all accounted for.

**Applied (6 findings, one commit).** The headline three — `silent-failure` survived[0], `correctness` unverified[0], `types` unverified[7] — were one defect: `stripComments` blanks a line whole when it opens with `/*` or closes a block, so a statement sharing a physical line with a block-comment close was erased before `firstStatement` looked and the guard went **silently green** on exactly the file it exists to red. Fixed once, in `firstStatement`. Plus `tests` survived[1] (the missing same-physical-line row, the only row that discriminates `META_FIRST`'s `^`), `tests` survived[2] (four presence pins collapsed into one contiguous span), and the residual on the refuted `correctness` finding (an empty or comment-only workflow was reported as "has a statement before `export const meta`" when it has none). The `comments` finding on `"use strict"`'s unbacked "settled here" was checked, survived, and applied. The `simplify` finding on the four repeated `assert.match` blocks was applied incidentally — the contiguous-span rewrite uses the loop form the file already uses twice.

**Deferred (2 issues).**

- #1204 (`ready-for-agent`) — `correctness` unverified[1] + `silent-failure` unverified[4], filed as one: discovery is flat and `.js`-only, so a `.mjs` or nested workflow is silently uncovered. Two dimensions reached the same construct independently. The implementer declared both limitations deliberately, with reasons, and the remedy needs a fact this repo cannot settle — so it is deferred as declared rather than widened into this PR.
- #1205 (`wontfix`, closed) — the `suggestion` band, checked. `correctness` unverified[2] (comment tolerance assumed from wording) and `correctness` unverified[3] (indented export is a false red) both **refuted with evidence**: the first by a direct two-way registry measurement showing comments neither register nor drop a file, the second by measuring that this repo's own `/^export /m` syntax checks already red an indented export, so the guard's red is true and the remedy would only silence it. `types` unverified[8] and `simplify` unverified[9] recorded below the claim bar with re-open triggers.

**Not filed.** `types` unverified[6] is the reviewer reporting it could not complete a full-suite run under sibling-agent load — a coverage statement about the review, not a defect in the diff. The count it could not get, at `fb90a07`: **tests 1651 / pass 1651 / fail 0**.

The `suggestion` band arrives with `refutersDispatched: 0` by policy, so each in-scope entry got its own refuter before anything was applied. Every changed pin was mutation-tested with a control that must not red; the table is in the description, which has been reconciled against the final diff — including one claim it carried that re-measurement contradicted.

PR #1209, opened 2026-09-02
fix(reap): stop discarding the worktree status probes' stderr (#625)

### LINKED ISSUE #625: reap.sh:219,245 — the two worktree status probes still send git's diagnosis to /dev/null, and `2>&1` is not a safe fix there
**Dimension:** silent-failure (from PR #620's review, `unverified[1]`)
**Deferred from PR #620 review** (issue #391) — out of that ticket's scope, which was scoped to the `git worktree remove` refusal.

## The two sites

`skills/fleet/scripts/reap.sh:219`
```sh
if ! status_out=$(git -C "$wt" status --porcelain 2>/dev/null); then
  keep "$b" "worktree $wt could not be read"
  continue
fi
```

`skills/fleet/scripts/reap.sh:245`
```sh
if ! ignored_raw=$(git -C "$wt" status --porcelain --ignored 2>/dev/null); then
  keep "$b" "worktree $wt unreadable (git status --ignored failed)"
  continue
fi
```

Both are the class PR #620 fixed at the `worktree remove` and `git branch -D` sites: the guard suppresses git's stderr and keeps under a label that names the step, not the fault. `grep -n '2>/dev/null' skills/fleet/scripts/reap.sh` returns exactly these two after #620.

## What is discarded

Reviewer's measurement, `chmod 000 $wt` on a merged `[gone]` branch's worktree, `--apply`:

```
KEEP feature/merged — worktree <wt> could not be read
```

Run by hand immediately after:
```
$ git -C <wt> status --porcelain
fatal: not a git repository: <w>/.git/worktrees/feature-merged
rc=128
```

The message names the **admin** path, i.e. which of the two directories is actually broken. That is the whole remedy signal, and it reaches only the terminal.

## Why this is NOT a copy of #620's one-liner — the constraint that makes it a triage question

At #620's sites the capture is pure diagnosis: `err` is used only to build a reason. At **these** two sites git's **stdout is the payload**:

- `:219` — the very next line is `if [ -n "$status_out" ]; then keep "$b" "dirty worktree $wt"`. Fold stderr in with `2>&1` and any git warning on a healthy run makes `$status_out` non-empty, so a **clean worktree reads as dirty and is never reaped**. That is a worse failure than the one being fixed: it is silent, it looks like correct conservative behaviour, and it strands every affected worktree.
- `:245` — `ignored_raw` is filtered by `awk '/^!! /'`, which stderr cannot match, so the dirty-flip does not apply there. But the two sites are the same idiom four lines apart and should not diverge without a stated reason.

So the remedy has to separate the two streams rather than merge them — `git … 2>"$tmp"` with a temp file, a second capture, or a `sh` fd juggle — and each of those has a cost in a `set -eu` POSIX script that currently creates no temp files at all. **That choice is the open question**, which is why this is `needs-triage` and not `ready-for-agent`.

## Related

- #578 — same *family* (guards that swallow git's reason) but a different shape: `rev-parse --verify --quiet … || die`, including `reap.sh:42`. Not a duplicate; that one is about `die` messages asserting an unknowable cause, this one is about `keep` reasons on a probe whose stdout is load-bearing.
- #620 fixed the two sites where the capture was diagnosis-only.

feigi:
## Summary

Fixes #625. Three `git status --porcelain` capture sites in `skills/fleet/scripts/reap.sh` sent git's stderr to `/dev/null`: the branch sweep's dirty check (`reap.sh:328` on `origin/main` at `fa3e497`), its `--ignored` sibling (`:351`), and the branchless sweep's copy of the dirty check (`:614`). A `fatal:` at rc 128 named only the step ("could not be read") and a `warning:` at rc 0 (measured on PR #726's review: a `chmod 000` ignored directory) reached nobody — no keep, no message — and the run proceeded to reap a worktree it had not finished reading.

`2>&1` is not the fix (per the Agent Brief's ruling): the dirty check's very next line tests whether the captured text is non-empty, so folding a warning in would make any git warning on a healthy run misread a clean worktree as dirty forever.

Added `git_probe()` — an fd-juggle (no temp file, confirmed by `grep -c mktemp skills/fleet/scripts/reap.sh` = 0) that captures stdout and stderr into separate variables plus git's own exit status from one invocation — and used it at all three sites, with two helpers built on it:

- **`gp_why`** — the `: <git's message>` suffix, at every site. rc != 0 now carries git's own message, with no dangling `": "` when git dies writing nothing to stderr (measured: a signal-killed git exits 137 silently) and no trailing space in the JSON reason.
- **`gp_cut_short`** — the rc-0 gate, at **both** probe kinds and all three sites. It matches `could not open directory`, the message git emits when its directory walk was cut short and only then.

### Why the rc-0 gate is on both probes, and why it is not "any stderr"

The first revision of this PR gated rc-0-with-stderr on the `--ignored` probe alone, on the reasoning that a plain scan skips ignored paths without opening them. Its review measured that false in both directions (git 2.50.1, Apple Git-155):

- **too narrow** — `git status --porcelain` opens **untracked** directories. An unreadable one makes it warn at rc 0 and return an **empty** listing, so the dirty check reads a worktree holding uncommitted work as clean and `--apply` removes it. `.worktrees/` is exempt from the `--ignored` probe, so the plain scan is the only probe a fleet worktree ever reaches.
- **too broad** — a gate on stderr merely being non-empty fires on a global gitconfig with a key outside any section (`error: key does not contain a section:`, printed by every git command at rc 0 — the same fault this file's `rev-parse` probe is already redirected for), permanently keeping every worktree outside `.worktrees/` with a reason that blames the worktree for a fault in `~/.gitconfig`.

`could not open directory` was measured at rc 0 with entries missing on an unreadable untracked directory, an unreadable ignored directory, and an unreadable directory holding a tracked file — and measured **absent** for both noise cases above (the config fault, and an unreadable `core.attributesFile`), which leave the listing complete. The gate runs **before** the dirty check, since a cut-short walk answers empty.

### Also from the review

- The `git worktree remove` comment claiming it "refuses on modified and untracked files" is corrected. Measured with a control: `status.showUntrackedFiles=no` disables the untracked half of that refusal (rc 0, the untracked file deleted with the worktree); the identical fixture without the config exits 128 refusing. Pinning the probes against that config remains **#730**'s deliverable and is deliberately not done here.
- The `git_probe` header no longer says "the two worktree status probes" — three existed before this PR, all discarding stderr.
- Dropped `&& [ "$#" = 4 ]` from the two `.worktrees/` shims: measured over every git call those shims see, the only 5-argument status call belongs to the outside-`.worktrees/` fixture, whose shim discriminates on `$5`.

The `rev-parse --absolute-git-dir` probe is untouched, per its own existing comment and the dispatcher's explicit site list: its stdout is used as a path, not message text, so this ticket's fix doesn't apply there.

**Sizing: light**

## Test plan

- [x] `node --test skills/fleet/scripts/*.test.mjs` — **1673/1673 pass** (1668 before this PR; 5 net new tests)
- [x] `node --test skills/fleet/scripts/reap.test.mjs` — 78/78 pass
- [x] `shellcheck -x -S warning skills/fleet/scripts/reap.sh` — exit 0; `dash -n` clean
- [x] Behaviour re-run by hand under real `/bin/dash` (not just macOS's bash-based `/bin/sh`, which is what the Node harness's `spawnSync("sh", …)` resolves to locally — CI's bare `sh` is dash) for all three shapes: cut-short warning at rc 0 → kept, git dies at rc 128 → kept with its message, ambient config noise at rc 0 → still reaped. No trailing space in any emitted reason.
- [x] Mutation-tested every new test on a scratch copy — each kills its own mutant (gate deleted at site 1, gate reverted to any-stderr, `gp_why` dropped at the `--ignored` and branchless sites, gate deleted at site 3, `gp_why` reverted to the naive inline flatten) **and** stays green under a change it should not catch. Working tree restored by `cp` from a pristine copy, never `git checkout`.

The eight tests covering #625 now pin: the rc-128 message at the branch sweep's plain probe, at the `--ignored` probe and at the branchless sweep's probe; the rc-0 cut-short keep at the plain probe, the `--ignored` probe and the branchless probe; and two controls — ambient stderr on the plain probe and on the `--ignored` probe must still reap.

Closes #625


feigi:
Review deferrals filed:

- **#1212** (`needs-triage`, open) — `reap.sh:193` `git_probe`: separator-byte (`0x02`) uniqueness asserted only in a comment, `gp_rc` returned unvalidated. Refuters split 1-1 — the truncation reproduces literally, but it fails **closed** (a garbled `gp_rc` cannot yield rc 0, so it can never turn a KEEP into a REAP) and git C-quotes control bytes for both current callers. Torn, not settled; out of #625's scope, so not applied here.
- **#1213** (`wontfix`, closed record) — `PR #1209 review: the suggestion band, checked`. Three suggestion-band entries refuted by measurement: the `git_probe` doc-comment asymmetry claim (the comment makes a routing claim, not a symmetry one, and documents the asymmetry explicitly — and its own suggested fix flips a keep verdict), the `2>&1` comment at `:698` (still literally true, and propped up by **four** surviving `2>&1` reasons above it, not one), and flattening `gp_err` inside `git_probe` (−76 bytes but +1 line, fights the 20-site house idiom, and makes `gp_err` lossy for the very gate this PR adds).

**Held, deliberately not filed and not applied:** the `status.showUntrackedFiles=no` rc-0-empty data-loss hole at the dirty gate. It is critical and real — demonstrated with a control — but no stderr gate can reach it (rc 0, empty stdout **and** empty stderr), and `-uall` is **#730**'s deliverable, which is open. Absorbing it here would close another ticket by stealth. The `git worktree remove` comment that wrongly claimed an unconditional refusal on untracked files **was** corrected in this PR, with a pointer to #730.

PR #1214, opened 2026-09-02
docs(fleet): state the reflow ceiling as break_on_hyphens, not width 40

### LINKED ISSUE #676: staleness-qualifier-prose.test.mjs: failure message claims the twin document still names the qualifier, which is false when both drift together
Deferred from PR #673 review (closes #196). **Refuted by an adversarial refuter — filed as `needs-triage` because the defect is not confirmed, not because the remedy is open.**

**dimension:** `tests` — the controller's relay did not carry the per-finding dimension label for this one; this attribution is the fix-applier's, from the finding's location on the new test file.

## Finding

`skills/fleet/scripts/staleness-qualifier-prose.test.mjs:54` (now `:93` after PR #673's follow-up commit `39b87a8`). The assertion message reads, unconditionally:

> `${name} no longer says "${CLAUSE}". Bare "staleness" names no cause, and its twin document still names one — that one-sided drop is exactly #196. …`

The clause *"its twin document still names one"* is **false when both documents drift together** — the exact case the file's own header comment anticipates ("a deliberate reword reddens this"). The failure then names a one-sided drop that did not happen.

Proposed remedy: compute the twin's match state and word the message off it, or drop the twin claim and keep the rest.

## Why deferred

Refuted on measurement. The refuter reproduced all three states in a scratch copy of the tree:

| state | tests | pass | fail |
|---|---|---|---|
| baseline | 975 | 975 | 0 |
| one-sided (`review-and-fix.md` only) | 975 | 974 | 1 |
| two-sided (both documents) | 975 | 973 | 2 |

In the two-sided state the message *is* literally false in both printed failures (verified verbatim in the spec reporter's output, which node selects even with stdout redirected, and which CI uses at `.github/workflows/ci.yml:199` — so the message does reach a reader). But the reader is not misdirected in practice:

1. `ℹ fail 2` plus two adjacent `✖` lines naming **both** documents put the twin's state on screen ahead of either message.
2. The two messages visibly cancel each other — each claims "its twin still names one" while the twin's own failure prints 16 lines below.
3. The message's next sentence is already correct for this state: "re-anchor CLAUSE in this file to the new wording in **BOTH** documents at once."

Ruled prose imprecision in a diagnostic, not a defect in what the test pins or in how it reddens.

## If someone picks this up

The proposed remedy was sketched and measured and does work — 4 added lines, accurate in **four** states (clean, one-sided each way, two-sided), suite green:

```js
const [twinName, twinText] = DOCS.find(([n]) => n !== name);
const twinState = phrase(CLAUSE).test(twinText)
  ? `its twin ${twinName} still names one — that one-sided drop is exactly #196`
  : `its twin ${twinName} has dropped the qualifier too, so this is a both-sided reword, not the one-sided drop of #196`;
```

Note it needs adapting to `39b87a8`, which bounded each document to the paragraph carrying the rule — the twin's state must be computed over the same bounded slice (`rule(name, anchor)`), not over the whole file, or the recomputed message reintroduces the false-green surface the bound removed.

It also upgrades the vague "its twin document" to the twin's actual path, which is a small real gain. The cost is precision bought in a case the existing output already resolves for free.

feigi:
Rescoped #676: the titled finding (the failure message's "its twin document still names one", false when both documents drift together) **stays refused and is not touched here**. The work is the header's reflow-ceiling note.

Sizing: light

## What changed

`skills/fleet/scripts/staleness-qualifier-prose.test.mjs`'s header stated the reflow ceiling as a width map — `behind-count` as the split-vulnerable token, red at 40 and 72, green at 60/80/100, with the width-80 green credited to the slice bound "doing its job, not luck".

Review measured all three parts wrong, and the header now states the ceiling as a **property** instead:

- **Any hyphenated token inside a pinned clause is vulnerable**, not one named token. `CLAUSE` carries `behind-count`; `MECHANISM` carries `behind-count` **and** `rebase-check`.
- **The slice bound does not protect against this.** Splits of either token land *inside* review-and-fix.md's step-6 slice, so the "bound doing its job" framing is gone.
- **A width quoted without its wrap discipline and its document is not reproducible.** Per-source-line and per-paragraph rewrapping red at disjoint widths on the same document — which is the root of the whole finding cluster, and why this note has been read as false. Every width the header keeps now names the discipline and the document it was measured under.

The irony was the point: this ticket existed to replace a WIDTH with a MECHANISM, and the first attempt shipped a width map without the discipline that makes it reproducible.

Comment-only: no assertion, no slice bound, no fixture touched.

## Settling commands, re-run at `7fe38a1`

**1. Both tokens red the MECHANISM test from inside the slice.** `review-and-fix.md` rewrapped **per source line** with Python `textwrap.fill(break_on_hyphens=True, break_long_words=False)`, then the suite run on the rewrapped tree:

```bash
# in a scratch copy of the tree, never the worktree
python3 -c 'import sys,textwrap; p=sys.argv[1]; w=int(sys.argv[2]); s=open(p).read(); \
open(p,"w").write("\n".join(textwrap.fill(l,width=w,break_on_hyphens=True,break_long_words=False) if l.strip() else l for l in s.split("\n")))' \
  skills/fleet/commands/review-and-fix.md 60
env -u FORCE_COLOR node --test skills/fleet/scripts/staleness-qualifier-prose.test.mjs
# then, for the split location: awk the hyphen-ending lines between the
# "6. Diff-check green" anchor and the next blank line
```

| discipline · document | width | only split inside step 6's MECHANISM clause | suite |
|---|---|---|---|
| per-source-line · `review-and-fix.md` | 60 | its own `behind-count` — `` `skipped` off the behind- `` / `count …` | `pass 2 / fail 1` — ✖ *step 6 keeps the mechanism behind "a behind PR never reaches" exit-0 green* |
| per-source-line · `review-and-fix.md` | 87 | its `rebase-check` — ``never reaches: a `rebase-`` / ``check` …`` | same test reds |
| per-source-line · `review-and-fix.md` | none (unwrapped control) | — | `pass 3 / fail 0` |

Width 87 is the clean single-token case: `behind-count` does not split there, `rebase-check` does, and the test still reds — so `rebase-check` is a second split-vulnerable pinned token on its own.

**2. The retained `break_on_hyphens`-off claim still holds.** Both documents paragraph-rewrapped, `break_on_hyphens=False`, `seq 60 5 400` (69 widths), suite run at each: **no reds**.

## Tests

`node --test skills/fleet/scripts/*.test.mjs` at `7fe38a1`, from the worktree — `tests 1668 / pass 1668 / fail 0`. Baseline measured the same way at `9a0f89d` (this PR's previous head): `tests 1668 / pass 1668 / fail 0`. Comment-only, so the count is unchanged by design.

That total includes the sweeps that read every comment in this directory, which a header reword can redden: `line-distance-prose.test.mjs` (no comment locates a construct by counting lines — the new text names occurrences semantically, no line numbers) and `snapshot-runner-audience-prose.test.mjs`.

## Deferred / filed

- **#1218** (open, `ready-for-agent`, dimension `tests`) — pre-existing coverage gap: the suite pins the `solely` qualifier in review-and-fix.md step 6 and in `run-team/SKILL.md`, but step 4's own instance of that same load-bearing qualifier is unpinned. Mutation-verified (`solely` → `mostly` at step 4 leaves the suite `3/3` green). Out of this comment-only PR's scope.
- **#1219** (closed, `wontfix`) — review record. The `critical` finding against the note's width-80 claim was **refuted by both refuters**: its red was an artifact of re-wrapping the numbered step list as a single `textwrap` paragraph, a reflow no Markdown wrapper performs. The overlapping `important` finding reproduces in line mode, which that objection does not touch, and is what this diff applies.

## Out of scope

The failure-message rewrite and the slice bound itself. The recorded 4-line patch for the message must compute the twin's state over the bounded slice `rule(name, anchor)`; adapted carelessly it reintroduces the false-green surface that bound removed, so it is not a ride-along inside a comment edit.

Closes #676


feigi:
Review deferrals filed:

- **#1218** — `ready-for-agent`, dimension `tests`. `skills/fleet/commands/review-and-fix.md:9` — step 4's own `solely` qualifier on the behind-count is unpinned by any prose test, while step 6's and SKILL.md's are pinned. Mutation-verified (`solely` → `mostly` at step 4 leaves `staleness-qualifier-prose.test.mjs` at 3/3 green). Pre-existing gap, out of this comment-only PR's scope; remedy left open.
- **#1219** — `wontfix`, closed. Review record for the refuted `critical` finding against the note's width-80 claim (probe artifact: the numbered step list re-wrapped as one `textwrap` paragraph), plus the disposition of the checked suggestion band.

Applied here: the survived `important` finding and the two suggestion-band entries that share its root cause, as one edit. Head `7fe38a1`, suite `tests 1668 / pass 1668 / fail 0`.

PR #1223, opened 2026-09-02
docs(arg): state the second-edit-site purpose apart from the reach test, and name candidates.mjs

### LINKED ISSUE #712: fleet scripts: arg.mjs header paragraph names second-edit sites by reachability, but candidates.mjs has its own parseArgs AND is reached by arg()
`arg.mjs`'s header paragraph exists to name every caller that carries its OWN flag parser, so that the next change to the refusal contract below is known to need a second edit there.

Its stated membership test is reachability: a caller is named when **`arg()`'s refusals never reach it** (`ledger.mjs` imports `makeDie` alone; `fleet-tick.mjs` likewise). `candidates.mjs` is the case that test cannot express:

| Claim | Settling command | Result |
|---|---|---|
| binds `makeArg`/`makeHas`, so `arg()`'s refusals DO reach it | `grep -n 'from "\./arg\.mjs"' skills/fleet/scripts/candidates.mjs` | `:30 import { makeDie, makeArg, makeHas } from "./arg.mjs";` |
| carries its own `parseArgs` anyway | `grep -n 'parseArgs' skills/fleet/scripts/candidates.mjs` | `:29` import from `node:util`, called at `:92` |
| that call refuses unknown flags | `sed -n '87,92p' skills/fleet/scripts/candidates.mjs` | `strict` is `parseArgs`' default and is left as the default |

So `candidates.mjs` has **both**: `arg()`'s refusals reach it *and* it hand-writes an unknown-flag refusal of its own. It fails the paragraph's stated membership test while satisfying the paragraph's stated purpose — a change to the unknown-flag rule in `arg.mjs` needs a second edit in `candidates.mjs`, and the paragraph does not say so.

PR #710 deliberately left it unnamed, correctly on the paragraph's own terms: naming it under the current wording would assert "`arg()`'s refusals never reach it", which is false, and that false-claim class is exactly what #696 corrected. The exclusion is reasoned, not an oversight — but it leaves the second-edit site unrouted.

## Two shapes, neither settled here

1. **Split the test from the purpose.** The paragraph currently conflates "refusals don't reach it" with "has a second edit site". Separate them, and name `candidates.mjs` under the second.
2. **Leave it.** Argue the reachability framing is the useful one and `candidates.mjs`'s own parser is adequately documented at its own site — then say so explicitly, so the next reader does not re-derive this.

Whichever, the deciding fact is that the paragraph's membership test and its purpose are not the same predicate, and one caller separates them.

## Provenance

Declared by `impl-696` during PR #710 and ruled a separate ticket by the run controller. Filed by `finish-pr-710` at labelling time: the ruling was not present in the run ledger (`ledger.mjs check` → `found: false`; `ledger.mjs read` → no `candidates` entry) and #710 carries `Closes #696`, which would have buried it at merge.

Not a defect in PR #710. #710 is correct as it stands.

feigi:
Sizing: light

`arg.mjs`'s header names the callers that need a second edit when the refusal contract changes. Its membership test was reachability — `arg()`'s refusals never reach the caller — and `candidates.mjs` is the caller that test cannot express: it binds `makeArg`/`makeHas`, so the refusals DO reach it, and it carries its own `parseArgs` anyway. Naming it under the old wording would have asserted something false, which is why #710 correctly left it unnamed; the site stayed unrouted.

This states the purpose first — a script holding a flag refusal of its own is a second edit site — and demotes unreachability to one of two ways to qualify. `candidates.mjs` is named under the other way, with its reachability stated as the true thing it is. `makeSweep()`'s existing `candidates.mjs` passage is pointed at, not restated.

Review then found that stating a criterion and naming exemplars still reads as a roster, and the roster was short one script. The second commit answers that by making the roster executable instead of extending a hand-list. Comment-only, one file, both commits.

## The membership test, settled by mutation

The review's two refuters both mutated `isFlagLike` — the **value guard** — to answer a question about the **unknown-flag rule**, and returned a confident wrong verdict. Two distinct rules live in this file and a finding about one says nothing about the other:

- the value guard — `isFlagLike()`/`hasEqualsForm()`, which `ledger.mjs` and `member-outcomes.mjs` both import and call, so a change there reaches them with no second edit;
- the unknown-flag rule — `makeSweep()`/`makeStray()`, which neither imports.

Mutating the rule actually at issue settles membership. In a scratch copy, changing only `makeSweep()`'s refusal wording to `MUTATED-SWEEP` and re-running every script with a flag nothing accepts:

| message rewritten by the mutation — routed | message unchanged — own refusal |
|---|---|
| `board.mjs`, `ci-state.mjs`, `diff-stats.mjs`, `pr-overlap.mjs`, `staleness.mjs` | `candidates.mjs`, `fleet-tick.mjs`, `ledger.mjs`, `member-outcomes.mjs` |

Reaching the sweep in `diff-stats.mjs` and `pr-overlap.mjs` takes their required flags first (`--pr 42`, `--a 1 --b 2`). A bare bogus flag exits on the required-flag error and misreads as an own refusal, which would have put both in the wrong column.

`compute-board.mjs`, `compute-spend.mjs`, `lift.mjs`, `prose-pin.mjs` and `strip-comments.mjs` import nothing from `arg.mjs` and refuse nothing — they exit 0 on a bogus flag. No `skills/fleet/scripts/*.sh` carries a refusal either.

## What the header now says instead of a list

The right-hand column is exactly what the header's own recommended grep yields, once the reader is told to read the imported **symbol** list rather than the filenames:

```
$ grep -n '^import .* from "./arg.mjs"' skills/fleet/scripts/*.mjs \
    | grep -v test | grep -v makeSweep
candidates.mjs:30:import { makeDie, makeArg, makeHas } from "./arg.mjs";
fleet-tick.mjs:146:import { makeDie } from "./arg.mjs";
ledger.mjs:14:import { makeDie, isFlagLike, hasEqualsForm } from "./arg.mjs";
member-outcomes.mjs:251:import { makeDie, isFlagLike } from "./arg.mjs";
```

So the header states that reading rule and stops naming a roster. The scripts it does name are exemplars of a way of qualifying, and it now says so, rather than leaving the set to be inferred from who got mentioned. That is deliberate: this header elsewhere tells the reader not to trust a list here to have aged well, and #712 exists because the previous list had not.

`member-outcomes.mjs` was the omission review caught. It hand-rolls its refusal and no helper here reaches it:

```
$ node skills/fleet/scripts/member-outcomes.mjs /tmp --bogus-zzz
member-outcomes: unknown option --bogus-zzz (did you mean --file <path>?)   rc 2
```

`ledger.mjs` belongs to the set for the same rule, which a `grep -l makeSweep` hides: its one mention of `makeSweep` is prose, a comment pointing at that function's own passage, while its import row never takes it. Why its refusal is not the sweep's is argued at `makeSweep()`, which owns that trade-off, so the header points there rather than restating it.

```
$ node skills/fleet/scripts/ledger.mjs --file <scratch> row 123 the --basee flag is unread
ledger: unknown flag --basee in subject — quote the subject as one argument   rc 2
```

The `^import` anchor is not decoration. Unanchored, the command matches this header's own text — the trap the header already records for `^function die(`, now with more lines to match:

```
$ grep -c 'from "./arg.mjs"' skills/fleet/scripts/arg.mjs
3
```

## Verified at 8749182

```
$ node skills/fleet/scripts/candidates.mjs --limit
candidates: --limit needs a value                                (exit 2)
  -> arg()'s GENERATED wording: this file's refusals reach candidates.mjs

$ node skills/fleet/scripts/candidates.mjs --bogus
candidates: Unknown option '--bogus' — accepted: --require-label,
--allow-fallback, --limit                                        (exit 2)
  -> candidates.mjs' own parseArgs refusal: its own edit site

tests 1676  pass 1676  fail 0  skipped 0
```

## What this edit newly does

The purpose sentence generalises a claim that was previously made only about the scripts out of `arg()`'s reach. Checked against the AC that guards it: the `fleet-tick.mjs` paragraph, including "the one script still outside the rules" and its reachability rationale, is carried through byte-identical, and the `ledger.mjs`/`member-outcomes.mjs` reachability passage is untouched. No claim that `arg()`'s refusals fail to reach `candidates.mjs` is made anywhere.

## Can the suite see the mode that changed?

Measured, not assumed. Positive control: injecting a line-count idiom into the new paragraph reddens `line-distance-prose.test.mjs`, which reports the hit as `arg.mjs:1` because it stitches comment blocks. So this file's header prose is inside a gate, and both commits were written to stay outside it — every pointer names its target semantically rather than locating it.

Ceiling, also measured, and unchanged by the second commit: deleting the new paragraphs leaves the suite green. Nothing pins the header's roster, and nothing executes the runnable probes it now carries, in a file whose own argument is that a requirement recorded where nothing executes it is the defect. **Deliberately left** — #712's AC is comment-only and both remedies add or extend a test file. Filed as #1227.

The review's checked-and-refuted band is recorded closed at #1228.

Closes #712


feigi:
Review deferrals filed for this PR:

- **#1227** (`ready-for-agent`, dimension `correctness`) — nothing pins the header's second-edit-site roster, and nothing executes the runnable probes it carries. Confirmed defect, remedy open; deferred because #712's AC is comment-only and both remedies add or extend a test file. The unpinned-roster ceiling `impl-712` declared and the review's unpinned-probes finding are the same defect at two sizes and are filed as one issue, not two.
- **#1228** (`wontfix`, closed) — the checked band. Two `correctness` findings and one `silent-failure` finding, each measured and refuted, with the mutation table that settles the membership question and the two measurement traps that produced the wrong verdicts.

Applied rather than deferred: the omission of `member-outcomes.mjs` from the roster, fixed in `8749182`.

Separately, the `comments` dimension did not run on this PR — a comment-only single-file `.mjs` diff scores `docs: 0`, the size tier trims it, and `dimensionsUnrun` came back empty so nothing reported the hole. Recorded as a fourth measured instance on #218 rather than as a new issue.

Issue #1149, opened 2026-09-01
muted-git-guard-sweep.test.mjs and unattended-git-sweep.test.mjs throw at module load without an ambient .git — 2 of 93 files, and every review snapshot pays for it


feigi:
Deferred from the review of #1145 (issue #578), refuted and confirmed rather than accepted on the finding's word.

## The defect

Both files resolve their file list at MODULE SCOPE:

```js
const REPO = execFileSync("git", ["rev-parse", "--show-toplevel"], { cwd: DIR, encoding: "utf8" }).trim();
const SHELL_SCRIPTS = execFileSync("git", ["ls-files", "*.sh"], { cwd: REPO, encoding: "utf8" }) …
```

With no ambient `.git`, `rev-parse --show-toplevel` exits 128 and `execFileSync` throws before any test body runs. node reports it as `test at <file>:1:1` — a module-level throw, not an assertion failure, which reads confusingly out of context.

## Measured

`git archive 6735324ec4ad21316ce5b26b66ef670a77fe151a | tar -x -C <scratch>` (no `.git`; confirmed `git rev-parse --show-toplevel` → `fatal: not a git repository`, exit 128, and the scratch dir is not nested under any repo), then `env -u FORCE_COLOR node --test skills/fleet/scripts/*.test.mjs`:

```
EXIT=1   tests 1572   pass 1570   fail 2
✖ skills/fleet/scripts/muted-git-guard-sweep.test.mjs
✖ skills/fleet/scripts/unattended-git-sweep.test.mjs
Error: Command failed: git rev-parse --show-toplevel   status: 128
```

Control, same tree with `git init` + one commit in scratch: `EXIT=0  tests 1582  pass 1582  fail 0`. The arithmetic closes exactly — 1582 − 1570 = 12 = the two files' own 12 tests, and the 2 "failures" are the two synthetic module-throw entries.

**2 of 93** `*.test.mjs` in that directory fail there; 91 pass. Seven siblings mention `show-toplevel`/`ls-files` but only in comments, prose-assertion regexes, or fixture repos they create themselves.

Adjacent case also probed: an extraction nested under an *unrelated* repo. There `git ls-files` succeeds and returns nothing, so 10/12 tests pass vacuously — but each file's own non-vacuity guard fires (`the sweep sees the scripts it is supposed to police`, `every script that made one of those calls still sources net.sh`), exit 1. So the wrong-repo case is caught, not silently green.

## Why it matters beyond the two files

[...]

PR #1224, opened 2026-09-02
test(release-ticket): pin linkage die's fatality, not just its phrase

### LINKED ISSUE #742: release-ticket.test.mjs:687 — the RESOLVES test pins the linkage guard's PHRASE, not its fatality (refuted 1-1, residual real)
Deferred from PR #737 review. **dimension: `tests`** (severity `important`, adversarially verified 1-1, refutation upheld on the votes).

## The finding, as raised

`skills/fleet/scripts/release-ticket.test.mjs:687` — the test *"a symlink over HEAD that RESOLVES is not swept into the unresolved-HEAD arm"* claims in its own docstring that "the refusal below is pinned by the one phrase that separates it from the HEAD arm rather than by that arm's silence". The finding argued that phrase does not in fact pin the linkage guard at `skills/fleet/scripts/release-ticket.sh:753-754`, because a downstream guard reproduces the same observable shape.

## Why this is filed rather than applied — both votes

The two refuters split 1-1 and tested **different mutants**. That difference is the whole ruling:

- **Dissenting vote (refuted=false)** reproduced only the finding's own mutant: replace the linkage `die` with an echo of the *identical* message plus `toplevel="$wt_canon"`, so execution continues. The test stays green — because the downstream `git -C "$wt" status --porcelain` die at `release-ticket.sh:760` fails on the same fixture and reproduces the identical shape (`json === null`, `code === 2`, no `/could not read its HEAD/`, artefacts untouched), while the retained echo still satisfies the `/cannot read the git linkage of/` match. Conclusion: the assertions pin only that the phrase is emitted *somewhere*.
- **Refuting vote (refuted=true)** tested the refactor the finding actually **names** — "a future change that treats the linkage rev-parse check as redundant with the status check below it". Replacing both linkage lines with `:` makes the test **RED**, measured:
  ```
  AssertionError: The input did not match the regular expression /cannot read the git linkage of/.
  Input: 'fatal: not a git repository: .../worktrees/9-release-ticket
          release-ticket: cannot read the status of ..., so whether it holds uncommitted work is unknown'
  at release-ticket.test.mjs:687
  ```
  The downstream status die emits `cannot read the status of`, **not** `cannot read the git linkage of`, so the phrase does separate the two.

**Ruling:** the finding's *stated* risk — that dropping the linkage check as redundant would ship green — is measured **false**. The only surviving mutant is one that deliberately keeps the guard's exact stderr string verbatim while downgrading it to an advisory, which no plausible refactor produces. The docstring is accurate as written.

## The residual, stated honestly

What genuinely remains unpinned is guard **fatality** when the message is identical: if someone downgrades the linkage `die` to an advisory *while keeping its exact wording*, the test cannot tell. That is a real gap, but it is **not the risk the docstring claims to cover**, and it is not a shape the codebase drifts into on its own.

This is also **not** the `verify-sha.sh` pattern recorded in agent-brain (`yRzXRWsiHp95AH81yaJd-`): there the downstream guard reproduced the *same stderr substring*; here it does not.

## Cheap remedy if anyone later wants fatality pinned

Not applied on PR #737 — deliberately, since it guards against a mutation nothing produces:

```js
assert.doesNotMatch(stderr, /cannot read the status of/,
  "the linkage guard itself must be what refused this, not a downstream guard reached because it went advisory");
```

**Interacts with the sibling ticket on the die's wording:** the uniqueness of `cannot read the git linkage of` is exactly what makes this test discriminate. Rewording that die without updating the pin at `:687` reds the suite.

Tracker rows reviewed before filing (`ledger.mjs check`): #135, #189, #74, #126 — all CLOSED and all about a `.git` file naming a foreign gitdir, a different defect from test-pinning guard fatality.

feigi:
Closes #742

## What

`release-ticket.test.mjs`'s "a symlink over HEAD that RESOLVES never reaches
the unresolved-HEAD arm — the linkage probe refuses first" test pinned the
linkage `die`'s stderr *phrase* but not its *fatality*. #742's residual names
the gap exactly: an advisory linkage guard whose wording is left intact.

Test-only change; no change to `release-ticket.sh`.

**This PR was reviewed and its first commit corrected.** That commit added the
assertion #742's "cheap remedy" section names verbatim,
`assert.doesNotMatch(stderr, /cannot read the status of/, …)` — and that
assertion is inert for the refactor #742's body actually names ("treat the
linkage rev-parse check as redundant with the status check below it").
Downgrading the `rev-parse --show-toplevel` die to a non-fatal `printf` with
identical wording leaves `toplevel` empty, so the `does not point at` mismatch
die fires next and the status probe is never reached: `cannot read the status
of` is never emitted and the assertion passes vacuously. Worse, its message
named a failure mode it could not detect.

The shipped assertion is a widened pattern over **both** exits below the
linkage die, with a message naming exactly the two guards the pattern matches:

```js
assert.doesNotMatch(stderr, /does not point at|cannot read the status of/,
  "the linkage die must END the run: neither the `does not point at` mismatch guard nor the `cannot read the status of` probe may be reached");
```

Sizing: light

## Verification (re-run at `f9669bf`, this PR's head)

- New assertion present:
  `git show "f9669bf:skills/fleet/scripts/release-ticket.test.mjs" | grep -c 'assert.doesNotMatch(stderr, /does not point at|cannot read the status of/'` → `1`
- Absent on `main`:
  `git show "main:skills/fleet/scripts/release-ticket.test.mjs" | grep -c 'does not point at|cannot read the status of'` → `0`
- The two `die` sites the pattern names, both in the block below the linkage die:
  `git show "f9669bf:skills/fleet/scripts/release-ticket.sh" | grep -n 'does not point at\|cannot read the status of'` → the `.git does not point at` mismatch die and the `cannot read the status of` status die.

### Mutation, all four directions

Mutants applied to `release-ticket.sh` inside `git archive` extractions of
this tree (never the worktree), each run with
`env -u FORCE_COLOR node --test --test-name-pattern "linkage probe refuses first" skills/fleet/scripts/release-ticket.test.mjs`:

| mutant | result | vs. the first commit's assertion |
|---|---|---|
| unmutated | `pass 1  fail 0` | same |
| linkage `die` → non-fatal `printf`, wording kept, `toplevel` left empty (**single point** — the refactor #742 names) | `pass 0  fail 1`, `operator: 'doesNotMatch'` | was **GREEN** — this is the gap closed |
| same downgrade **plus** `toplevel="$wt_canon"` (two point) | `pass 0  fail 1`, `operator: 'doesNotMatch'` | also red — nothing traded away |
| linkage `die`'s trailing clause reworded, left a real fatal `die` (benign control) | `pass 1  fail 0` | same — not wording-brittle |

The two-point mutant is the one the first commit measured. It is what made
that commit's message and this body's original text describe the assertion as
catching a downgrade that "let execution continue" — true only because the
mutation also restored `toplevel`, which no refactorer writes.

- Full suite at this commit:
  `env -u FORCE_COLOR node --test skills/fleet/scripts/*.test.mjs` →
  `tests 1676, pass 1676, fail 0`, exit 0.

## Enumerate-and-declare (bug class: a `die` pinned by phrase but not fatality)

Covered: the linkage `rev-parse --show-toplevel` die, in the one test that
pins it. #742's Out of Scope section restricts this ticket to that assertion —
no change to `release-ticket.sh`, and the sibling wording-ownership ticket
stays out too. Left deliberately: a broader audit of whether the script's
other `die` sites carry the same phrase-only-pin gap — a real question, out of
scope by the ticket's own text.

**False-positive class for the widened pattern.** It reds any run of *this*
test whose stderr carries either phrase. Both phrases have legitimate
accept-cases elsewhere in the file, each in its own test scoped to that guard,
and neither is touched by this change:

- `cannot read the status of` — asserted positively in *"an unsearchable
  worktree is not reported as having no .git"*
- `does not point at` — asserted positively in *"a .git linkage naming another
  repository's worktree refuses instead of leaking that repo's clean status"*

Both are green in the full-suite run above.

## Review deferrals

The review's other three findings all sat in the `suggestion` band (0 refuters
by policy) and were adjudicated here rather than filed:

- The three sweep files failing from a `git archive` snapshot are **#1149**'s
  defect, not a regression from this PR — routed there as a comment, which
  also corrects that issue's population from two files to three. `ledger.mjs
  check` → `verdict: tracker-hit`, `#1149 (OPEN, score 0.62)`.
- A `tests`-dimension record of the red/revert/green protocol run against the
  two-point mutant — a positive record, no action.
- A second, independent sighting of the single-point gap — the same defect as
  the finding applied above, closed by the same commit (measured: that mutant
  now reds).


feigi:
## Review adjudication — applied / deferred

Head at adjudication: `f9669bf716f3ea5165ea8c56d42aafbe19da2236`.

### Applied

**`survived[0]`** — dimension `correctness`, severity `important`, verified
2-0. *The one assertion this PR added is inert for the refactor #742 names.*

Re-derived from this tree rather than taken on the review's word. Confirmed in
all four directions with mutants applied to `release-ticket.sh` inside
`git archive` extractions (never the worktree):

| mutant | shipped assertion (`a68052e`) | corrected assertion (`f9669bf`) |
|---|---|---|
| unmutated | pass | pass |
| linkage `die` → non-fatal `printf`, wording kept, `toplevel` left empty (**single point**) | **pass — vacuous** | **fail**, `operator: 'doesNotMatch'` |
| same, plus `toplevel="$wt_canon"` (two point) | fail | fail, `operator: 'doesNotMatch'` |
| linkage `die`'s trailing clause reworded, left fatal (benign control) | pass | pass |

The review's `suggested_fix` was a single `doesNotMatch(/does not point at/)`
added alongside the shipped line. Shipped instead as one widened pattern
replacing it, `/does not point at|cannot read the status of/` — same kill set,
one assertion rather than two, and the message can then state the whole
property.

**Beyond the finding.** The finding called the shipped `cannot read the status
of` line "inert either way — keep or drop as taste dictates". That
under-reads it: its message ("the linkage guard itself must be what refused
this, not a downstream guard reached because it went advisory") *named the
single-point failure mode it cannot detect*. A message asserting a guarantee
the assertion does not provide is worse than no assertion, because the next
reader trusts it instead of re-deriving. So the message was replaced, not just
the pattern, and the new one names the two guards its own pattern matches —
verifiable on the same line, with no positional reference to rot.

The first commit's message and this PR's original body carried the same
mis-description (they described the two-point mutant as a downgrade that "let
execution continue"). Both are corrected: the body is rewritten, and the
correction is stated in `f9669bf`'s message.

### Deferred / routed

**`unverified[0]`** — dimension `correctness`, severity `suggestion`. Three
sweep files fail from a `git archive` snapshot. **Not a regression from this
PR** — it is #1149's defect. Routed as a comment there rather than a new
issue, and it **corrects that issue's population from two files to three**:
`worktree-listing-sweep.test.mjs` landed by `a4c46cc` about ten hours after
#1149 was filed.
`ledger.mjs check` → `verdict: "tracker-hit"`, `#1149 (OPEN, score 0.62)`.
→ https://github.com/feigi/claude-config/issues/1149#issuecomment-5517456155

**`unverified[1]`** — dimension `tests`, severity `suggestion`. A **positive**
record of the red/revert/green protocol, run against the two-point mutant. No
defect alleged, nothing to apply, nothing to file.

**`unverified[2]`** — dimension `tests`, severity `suggestion`. The same
defect as `survived[0]`, found independently. Closed by the same commit, not
filed as a duplicate — measured: that finding's mutation is exactly the
single-point row above, and it now reds.

No `wontfix` record issue was filed: the `refuted` band is empty, and every
`suggestion`-band entry resolved into a durable home (an existing tracker row,
or this PR's own diff and body) rather than into a residual needing one.

### Dimension coverage, stated rather than assumed

`dimensionsUnrun` names **`comments`** — the snapshot run reported three
failing tests and that reviewer filed nothing. So **no dimension checked the
prose on this PR**, and the prose is where the defect was. Those three
failures are the #1149 environmental case above, not a real red.

### Suite

`env -u FORCE_COLOR node --test skills/fleet/scripts/*.test.mjs` at `f9669bf`,
run in this worktree (a real repository) → `tests 1676, pass 1676, fail 0`,
exit 0, zero `✖` lines.

One earlier run of the same command reported a single failure in
`net.test.mjs` — `net_kill_tree` `spawnSync` `ETIMEDOUT` after 19.6 s — while
four mutation suites were running concurrently on the same machine. It did not
reproduce on a quiet re-run. Recorded rather than discarded: that is the class
of #1199 (subprocess/pipe tests failing under concurrent load), not something
this diff touches.

PR #1225, opened 2026-09-02
fix(fleet): refuse in the interpreter's name when node cannot be resolved

### LINKED ISSUE #1141: claim-ticket.sh:71 invokes bare `node`, and an intermittent resolution miss reds three tests on an untouched tree
Reported by `impl-535` during a `/fleet:run-team` wave, corroborated by the controller. **The cause below is a lead, not a confirmed diagnosis** — the failure did not recur, so nothing here reproduces it on demand.

## What was observed

`impl-535`'s **first** baseline run, on a clean worktree at `origin/main` with zero edits of its own:

```
tests 1569 / pass 1566 / fail 3
```

All three failures were `claim-ticket.test.mjs` install cases, each reporting:

```
claim-ticket: could not read origin/main:package.json — .../claim-ticket.sh: line 71: node: command not found
```

The member's final run and its `git archive` export runs were all green, so the failure is **intermittent** and was not caused by its change.

## Why it is worth a ticket rather than a shrug

**It reds an untouched tree.** A member that hits this while measuring a mutation baseline has a red it did not cause, in a file it did not touch, and the natural reading is "I broke something". This run already carries two measured instances of the same shape — a killed baseline from a sibling's `pkill -f`, and a phantom suite failure inside a snapshot with no `.git` — and in both the harm was a *measurement that looks real and is not*.

## The lead

`claim-ticket.sh` invokes **bare `node`**:

```
$ grep -n 'node -e' skills/fleet/scripts/claim-ticket.sh
71:elif ! ndeps=$(printf '%s' "$pkg" | node -e '...')
```

and on this machine an asdf shim directory precedes the real interpreter on `PATH`:

```
$ which node
/opt/homebrew/bin/node
$ echo "$PATH" | tr ':' '\n' | grep -n 'asdf\|homebrew/bin'
4:/Users/chris/.asdf/shims
5:/opt/homebrew/bin
```

An asdf shim resolves per-directory. `claim-ticket.test.mjs` runs the script inside throwaway fixture repositories, which carry no `.tool-versions`, so shim resolution there does not depend on the same state as an interactive shell in the repo root. That is a *plausible* mechanism for an intermittent `command not found` and is offered as the first thing to test — **not** as an established cause. It does not explain why the same command succeeded on later runs in the same session, and that gap is the part any fix has to close.

## What a fix would have to establish

- A reproduction, with a positive and a negative control. Without both, an intermittent failure will be attributed to whatever was changed last. The reproduction has to distinguish "the shim could not resolve node" from "node was resolvable and something else failed", since `command not found` is emitted by the shell in both a missing-binary and a failed-shim case.
- Whether the remedy is to resolve the interpreter once at script start rather than per-invocation, to invoke it by an absolute path derived from the caller's environment, or to make the guard's `die` distinguish *the interpreter was unavailable* from *the manifest was unreadable* — the current message conflates them, which is why the report reads as a manifest problem when the interpreter is what went missing.
- Whether sibling scripts share the exposure: `grep -ln 'node ' skills/fleet/scripts/*.sh` is the starting scope, and any fix keyed to `claim-ticket.sh` alone should say what it deliberately leaves.

## Scope of the observation

One machine, one session, one occurrence, three test cases. Not reproduced since. `node --version` from an interactive shell in the repo root resolves normally, which is exactly why this went unnoticed until a member ran the suite from a fixture directory.

<sub>Observed by `impl-535`; `PATH` ordering and the bare-`node` call site verified by the `/fleet:run-team` controller. Cause unconfirmed.</sub>

feigi:
Sizing: heavy

`claim-ticket.sh` and `derive-testcmd.sh` both read a manifest through a bare
`node -e` whose capture merges stderr. When the interpreter cannot be used, its
own message lands inside that capture and each script refuses in the manifest's
name. A guard that **invokes** the interpreter now sits on the arm that reaches
each of those captures, so each refuses in its own voice.

## The defect, measured before the fix

Driven under a `PATH` holding real binaries with the interpreter deliberately
absent, at `84ffde8` (the branch point):

```
claim-ticket:    could not read origin/main:package.json — .../claim-ticket.sh: line 71: node: command not found
derive-testcmd:  could not read origin/main:package.json — .../derive-testcmd.sh: line 70: node: command not found
```

## The probe is an invocation, not a name lookup

Review measured that a `command -v node` guard closed only three of the four
ways an interpreter goes unusable. `command -v` answers whether a PATH entry
named `node` exists and is executable — which a version-manager shim that
resolves and *then* fails satisfies, so its own stderr still arrived under the
manifest's name. That is the ticket's defect surviving in its own fix, in the
case `claim-ticket.sh`'s comment already names as a live hazard. The probe is
`node -e 0` instead.

Four modes x both scripts, measured on the merge tree (fixture: shimmed PATH of
real binaries, repo with `{,,broken` and no lockfile):

| mode | `command -v` | refusal now |
|---|---|---|
| name absent | caught | `claim-ticket: node is unusable, … — sh: node: command not found` |
| present, mode 644 | caught | `… — <shim>/node: Permission denied` |
| dangling symlink | caught | `… — sh: node: command not found` |
| resolves, exits 127 | **missed** | `… — N/A: version "v20" is not yet installed` |

The three it already covered do not regress; the fourth is the one it could not
see. The control — a working interpreter on that same PATH — still earns the
manifest refusal word for word.

## The enumerate-and-declare list

The ticket's own starting scope, and then the mechanism:

```
$ grep -ln 'node ' skills/fleet/scripts/*.sh
skills/fleet/scripts/claim-ticket.sh
skills/fleet/scripts/derive-testcmd.sh
skills/fleet/scripts/inflight.sh
skills/fleet/scripts/instruments.sh

$ git grep -nE '(^|[|;&(]|\$\(|\bexec |\bthen |\bdo |\|\| |&& )[[:space:]]*node[[:space:]]' -- '*.sh'
skills/fleet/scripts/claim-ticket.sh:71:...
skills/fleet/scripts/derive-testcmd.sh:70:...
```

**Covered — both interpreter invocations in the fleet's shell scripts.**
`claim-ticket.sh`'s dependency count and `derive-testcmd.sh`'s manifest
outcome. Both are fixed; neither is deliberately left.

- `inflight.sh` and `instruments.sh` match the ticket's grep on prose alone —
  the graph sense of the word, and "node instrument". Neither invokes an
  interpreter, so neither can misattribute a refusal.
- The `exec node --test` inside the `agent-test` runner `claim-ticket.sh`
  writes. That is the payload the runner exists to run, not a derivation
  reporting on a manifest; an interpreter missing there fails at `exec` under
  node's own name, which is already attributable. Guarding it would be a
  different contract on a different file.
- The cockpit's spawning of siblings by interpreter name — the brief puts it
  out of scope, and its read helper already has a tested degrade-and-warn
  contract.

The guard is a new arm in an existing `if`/`elif` chain (claim) and a new first
statement inside an existing `if [ -n "$pkg" ]` block (derive). Nothing is
moved, reordered or rewrapped, so no script's last command changes, no `set -e`
coverage changes, and no guard is hoisted ahead of another. Placement is the
load-bearing part: every lockfile arm and the no-manifest arm settle the
install *without* an interpreter, so the guard must not fire on them.

## Discrimination, mutation-tested

Mutants applied to copies under a scratch dir, never to the tree.

| Mutation | Result |
|---|---|
| drop the claim guard | exactly `install: an unresolvable interpreter refuses…` reds; derive suite untouched |
| drop the derive guard | exactly the derive suite's own case and claim's delegated case red |
| reword each guard's message tail, keeping the interpreter named | **all green** — the tests pin the property, not a spelling |
| downgrade the claim guard's `die` to a warning, with a compensating `install="true"` | **red** — the fatality assertion |
| hoist either guard above its manifest gate | **red**, and only the silence case reds |
| rewrite the probe to an equivalent form, placement and wording unchanged | **all green** |

The pre-existing assertion this had to sharpen, driven with the interpreter
absent: **green without the guard, red with it.** It could be satisfied by an
interpreter fault; it no longer can.

Two properties the guards have that a message-shape assertion cannot reach are
pinned in their own cases:

- **The refusal is fatal.** `derive-testcmd.sh` carries the same guard and is
  delegated to moments later, so a downgraded `die` in `claim-ticket.sh` still
  lands exit 2 carrying a refusal that names the interpreter. Only stopping *at*
  the refusal tells them apart, so the case pins the refusal as the last thing
  on stderr.
- **The guards stay silent where no interpreter is needed.** No manifest, node
  absent, exit 0. Nothing but the guards' placement under their manifest gate
  holds that, and every refusal assertion in the file stays green while a hoist
  breaks it.

Each refusal case also carries its must-ACCEPT half — a resolvable interpreter
over the same unparseable manifest still earns the manifest refusal word for
word — plus a false-positive control proving the stripped PATH is not itself
what refuses: under exactly that PATH with the interpreter linked in, both
derivations run to completion.

## The open design choice, and how it went

Whether the interpreter's identity gets one home. **Decided: two homes.** A
shared home means `derive-testcmd.sh` sourcing a library it otherwise does not
need, adding a refusal path to a script whose exit 1 two consumers read as a
verdict — for one word, when the two refusals are deliberately *not*
byte-identical because each speaks in its own voice. This repo already answers
the question for this exact script pair: `testfile_re` is documented as "same
pattern, two necessarily separate homes".

## Prose this diff owns, settled against the tree

- Both test-name citations now name their target by verbatim title with no
  direction — a positional pointer rots the moment a case is inserted between
  them, which is how the first one broke.
  `grep -c 'test("install: an unparseable manifest refuses, and says so"' skills/fleet/scripts/claim-ticket.test.mjs` → 1;
  `grep -c 'test("a manifest that does not parse refuses instead of degrading to node --test"' skills/fleet/scripts/derive-testcmd.test.mjs` → 1.
- Both `SHIMMED` arrays claimed to hold "the REAL binaries the script reaches
  for" while listing names no path under test invokes. A wrapper logging every
  exec under each fixture named `sh git sed tr dirname cksum grep` for
  `claim-ticket.sh` (`sed`/`tr` transitively through `json.sh`'s `jstr`,
  `cksum` for the runner hash) and `sh git grep` for `derive-testcmd.sh`, which
  sources nothing. The arrays are trimmed to those, which makes the sentence
  true.
- The design-doc row said `node` unresolvable on `PATH`; the widened probe
  outgrew it, and it is retargeted with the fix.

## Verification

```
$ shellcheck -x -S warning skills/fleet/scripts/claim-ticket.sh skills/fleet/scripts/derive-testcmd.sh
(clean)

tests 1682 / pass 1682 / fail 0
```

Closes #1141


feigi:
Review findings applied and deferred. Filed:

- **#1230** (`ready-for-agent`) — `derive-testcmd.sh:96` takes the pipeline's status, so a dying `grep` refuses as "has no scripts.test and no test files". Same misattribution family as #1141, pre-existing and outside this diff. Measured, remedy open.
- **#819** — commented rather than filed: that ticket already names `derive-testcmd.sh` as missing from the design doc's script-surface table (`ledger.mjs check` scored it 0.60). This PR supplied the concrete instance it was waiting on, and the review's scoping clause — a claim completes at exit 0 with no interpreter when neither derivation reads a manifest — went with it.
- **#1231** (`wontfix`, closed) — the checked record: the "PATH-shadow convention" provenance findings, re-settled independently and refuted; the `shimPath()` simplification, below the claim bar with a re-open trigger; and the correctness dimension's clean result.

Issue #765, opened 2026-08-21
fleet test comments: four remaining "N lines earlier/later/above/below" distances — one sweep, and neither enumeration found them all


feigi:
Deferred from PR #762 review. **Dimension: tests.** Out of scope for #762 — different files, none in that diff.

PR #762 fixed one instance of an unmeasurable relative-distance idiom in a test comment. Four more are live. They are a class, not four independent nits, so this is one sweep ticket.

## The sites

All verified present at `b763fdd`:

- `skills/fleet/scripts/board.test.mjs:35` — `// gate swallows two lines earlier; written that way this test took the same`
- `skills/fleet/scripts/no-undo-audit.test.mjs:432` — `` //   `git stash drop` two lines later. Removing the gate must remove the ``
- `skills/fleet/scripts/review-pr-snapshot-path.test.mjs:18` — `// byte-identity 'Verify it' step a few lines above it in the prompt.`
- `skills/fleet/scripts/review-pr-snapshot-path.test.mjs:163` — `// Delete the two lines below in review-pr.js and #140's refusal is dead code`

## Neither enumeration was complete on its own — a sweep needs both forms

This is the part worth carrying forward. Two independent passes each found sites the other missed:

- The review's regex `~[0-9]+ lines|[0-9]+ lines (earlier|later|above|below)` catches **numeral** forms. It found `board.test.mjs:35` and `no-undo-audit.test.mjs:432`, and missed both `review-pr-snapshot-path.mjs` sites because they use **word** forms (`a few lines above`, `the two lines below`).
- `impl-721`'s pass caught the word forms, and missed `no-undo-audit.test.mjs:432`.

Only the union is the real list. A future sweep that runs one regex and calls the tree clean will under-report — that is the same "a probe that could not look returns what a probe that looked and found nothing returns" shape. Run both:

```
grep -rnE '~?[0-9]+ lines (earlier|later|above|below|apart)' skills/fleet/scripts/*.test.mjs
grep -rniE '(a few|several|two|three|four|five|some) lines (earlier|later|above|below|apart)|lines (above|below) it' skills/fleet/scripts/*.test.mjs
```

**Scope is confirmed to test files.** The same sweep widened across `skills/` and `.github/` (`*.mjs`, `*.sh`, `*.yml`, `*.md`, excluding `*.test.mjs`) returns nothing, so there is no non-test tail to chase.

**Two hits are out of class, do not "fix" them:**
[...]

PR #1229, opened 2026-09-03
fix(verify-sha): add --verify to the origin/<branch> tip capture

### LINKED ISSUE #1146: verify-sha.sh: the tip capture omits `--verify`, so a file named `origin/<branch>` resolves to a path at exit 0
Found by `impl-578` while sweeping for the muted-guard class in #578, and **deliberately not fixed there** — different mechanism, different site, and #578 named neither. Reproduced independently by the `/fleet:run-team` controller before filing.

## The defect

`verify-sha.sh` captures the branch tip with a bare `git rev-parse`, no `--verify`:

```sh
tip=$(git rev-parse "origin/$branch")
```

`git rev-parse` without `--verify` falls back to treating an unresolvable argument as a **path**, prints it, and exits **0**. So if a file or directory named `origin/<branch>` exists relative to the working directory, the capture carries a non-sha onward at exit 0.

Controller-reproduced from a fresh fixture:

```
$ git init -q . && git commit -qm init
$ mkdir -p origin && echo hi > origin/weird
$ git rev-parse "origin/weird"
origin/weird
rc=0
$ git rev-parse --verify "origin/weird"
fatal: Needed a single revision
```

The `--verify` form refuses, which is the behaviour every guard in this family already relies on.

## Why this is narrower than it first looks, and still worth fixing

**It is not a silent-corruption bug of the worst kind.** The value is `jstr`-wrapped downstream, so the JSON payload stays parseable — the field is simply *wrong* rather than malformed. And unlike the guards #578 swept, this line suppresses nothing, so git's own diagnosis is not being discarded; there is just no diagnosis to emit, because git considers the call a success.

**It is worth fixing because the failure is a wrong answer at exit 0.** A caller comparing a reported SHA against `tip` gets a comparison against a path string, and every surrounding signal reads healthy. That is the same shape this repo has now fixed twice in the guard family — a probe that cannot answer reporting as though it did.

## What a fix has to establish

- **`--verify` is the remedy, but confirm the exit-code contract per call site.** `git rev-parse --verify <bad>` exits **128**, where the bare form exits 0. Where the capture's status is read, or is the last command in a function, that change is observable and must be enumerated rather than assumed. `impl-578` measured exactly this for the five guards it did fix: all five kept a script-boundary exit of 2 despite `rev-parse` moving 1 → 128, but that held **because** each was followed by `|| die`, which is not the shape here.
- **Whether sibling captures share the exposure.** The starting scope is every `$(git rev-parse …)` capture without `--verify` under `skills/fleet/scripts/`; a fix keyed to `verify-sha.sh` alone should say what it deliberately leaves.
- **A positive control.** A well-formed sha and a real branch name must behave exactly as today — `--verify` refusing something it should accept is a worse failure than the one being fixed.

## Related, distinct

- **#578** — the muted-guard sweep that found this. Different mechanism: those guards discarded git's stderr with `--quiet`; this one gets a successful exit and no stderr at all. `impl-578` narrowed its new sweep test to guards that discard stdout, and recorded this exclusion with its measurement in the test comment rather than silently dropping it.
- **#884** — same file, different defect (the JSON-escape guard failing open). Worth reading together by whoever takes either.

<sub>Measured by `impl-578` during the #578 implementation; reproduced from a fresh fixture by the `/fleet:run-team` controller. Filed rather than folded into #578, on that member's own recommendation.</sub>

feigi:
## Summary

`verify-sha.sh` captured the branch tip with `tip=$(git rev-parse "origin/$branch")` — no `--verify`. Without it, an unresolvable `origin/<branch>` falls back to treating the argument as a path: a file or dir named `origin/<branch>` in the working directory clears the capture at exit 0, and the `|| die "origin/$branch does not resolve after fetch"` beside it never fires.

Fix: add `--verify` to that one capture (`skills/fleet/scripts/verify-sha.sh:80`, commit `c7c7077`).

A second commit (`ef0376b`) applies the review of the first: it restores a pin the first commit destroyed, corrects a claim the first commit got wrong, and clears the `line-distance-prose.test.mjs` red the first commit introduced. Details in that commit's body; the claims below are re-measured at `ef0376b`.

## Measured, not assumed

The fallback's blast radius in this script is narrower than the general bug class suggests — but **which guard catches it depends on the sha**, and the first commit stated only half of that, in both its prose and one of its assertions. Re-measured at `ef0376b` against a single-point mutant (that capture with `--verify` removed, nothing else changed), on the ref-absent + `origin/<branch>`-path-collision fixture:

- **sha is a real commit in the object store** — `git cat-file -e` passes, `git merge-base --is-ancestor` re-resolves the identical `origin/$branch` string, and merge-base has no path fallback, so it refuses (rc 128). Observed: `verify-sha: git merge-base --is-ancestor failed — cannot tell reachable from unanswerable`.
- **sha is absent from the object store** (the `0`×40 the colliding-path test uses) — the run never reaches merge-base at all. Observed: `verify-sha: cannot resolve 0000000000000000000000000000000000000000 to a commit in this repository`, from the `git cat-file -e` guard.

So the script was fail-closed before the fix either way — never the malformed-but-parseable payload at exit 0 the issue's general description of the class raised. What the bug cost is the **diagnosis**: whichever guard fired, it named something other than the real cause (`origin/<branch> does not resolve after fetch`). `verify-sha.test.mjs`'s mutant test pins the merge-base half of that directly; the colliding-path test's negative assertion pins the `cat-file` half.

## Sibling scope

Ticket's own starting point: every `$(git rev-parse …)` under `skills/fleet/scripts/` without `--verify` (checked via `grep -rn '\$(git rev-parse' skills/fleet/scripts/*.sh`, non-test files):

- `inflight.sh`, `claim-ticket.sh`, `release-ticket.sh` each capture `git rev-parse --git-common-dir` — takes no ref argument, categorically not exposed to this mechanism.
- `prove-merge.sh`'s two captures (`pre_full`, `post_full`) resolve `$pre`/`$post`, both already refused upstream (`prove-merge.sh:104,120-121`) by `git cat-file -e "${obj}^{commit}"` plus a commit-type check before either capture runs.
- `verify-sha.sh`'s one capture (this PR) was the only live site.

## Sweep exclusion comment

`muted-git-guard-sweep.test.mjs` named this ticket in its capture-exclusion comments and one `REFS` fixture while #1146 was open.

The **prose** is updated: captures stay out of that sweep on mechanism (they suppress nothing; a capture's failure already reaches the operator), not on this ticket's status — widening the sweep into a general capture audit is left an open, separate decision, per the ticket's own "out of scope."

The **`REFS` fixture is deliberately NOT updated** and keeps the bare spelling `tip=$(git rev-parse "origin/$branch")`. It is the only row in that table excluded *solely* by the `>/dev/null` clause of `isUnverifiedGuard`; written with `--verify` it is excluded by the `--verify` clause as well, and that clause can then be deleted with the whole file still green. The first commit did rewrite it, destroying that pin; `ef0376b` restores it and records the measurement both ways.

While editing that file's header, its `measured, 2 of 93 test files in this directory` count is replaced by the property it was standing in for — the files that die in a `git archive` extraction are the ones resolving a repo root at module scope. Both of its numbers had rotted: measured at `ef0376b`, three such files exist (`muted-git-guard-sweep`, `unattended-git-sweep`, `worktree-listing-sweep`), not two. Stated as the property rather than as a new number, which would rot the same way. The related defect itself stays tracked in #1149 and is untouched here.

## docs/specs row

`docs/specs/2026-07-23-fleet-plugin-design.md`'s `verify-sha.sh` exit-2 cause list already names "`origin/<branch>` does not resolve after it" — confirmed still true, no change needed there.

## What I deliberately left out

`docs/specs`'s row and the exit-2 prose test (`verify-sha-exit2-prose.test.mjs`) forbid tallying causes by design (Rule 4 in that file's own header) — nothing to update there since the fix adds no new exit-2 cause, only corrects which existing guard produces it.

## Test plan

- [x] Full suite at `ef0376b`: `node --test skills/fleet/scripts/*.test.mjs` → `tests 1679 pass 1679 fail 0`. At `c7c7077` the same command was `pass 1677 fail 2` — both tests of `line-distance-prose.test.mjs` (#765), reddened by two comments the first commit added. Construct named, count dropped, at both sites.
- [x] `node --test skills/fleet/scripts/verify-sha.test.mjs skills/fleet/scripts/verify-sha-exit2-prose.test.mjs skills/fleet/scripts/muted-git-guard-sweep.test.mjs` at `ef0376b` → `tests 39 pass 39 fail 0`.
- [x] Test `a rev-parse that would fall back to a colliding path is still fatal, and names the tip guard as the cause`: ref absent + `origin/<branch>` path collision → still refused, now with the correct cause and git's `--verify` diagnosis (`fatal: Needed a single revision`) on stderr. Its negative assertion is `doesNotMatch(/cannot resolve .* to a commit/)` — the guard the mutant actually reaches. The first commit asserted `doesNotMatch(/cannot tell reachable from unanswerable/)`, which that fixture's sha can never produce; it was vacuous and is replaced.
- [x] Positive control `a colliding path does not shadow a real remote-tracking ref`: ref **and** same-named path both present → resolves the ref, `tip` is the real sha, run succeeds — the acceptance criterion's explicit "worse failure than the one being fixed" case.
- [x] Mutant test `mutant: without --verify, the same fixture is still caught, but by the wrong guard`: the fixed capture line stripped of `--verify` on a scratch copy of the script (siblings `json.sh`/`net.sh` copied alongside so it still sources them), run against a real-commit sha with the same ref-absent + collision fixture — still exit 2 (fail-closed), but caught by the merge-base guard with the wrong stated cause. Its `git push` was dead setup and is dropped: only local existence of the commit is required, and the deleted tracking ref is what makes merge-base refuse.
- [x] Two pre-existing tests updated (`a fetch that succeeds but leaves origin/<branch> unresolvable is exit 2 at the rev-parse guard`, `a rev-parse that cannot resolve origin/<branch> is fatal — the script stops rather than answering off an empty tip`): `--verify` changes git's own diagnostic wording for an already-covered "ref absent, no collision" case from `ambiguous argument 'origin/main'` to `fatal: Needed a single revision` — exit code and guard attribution unchanged, only git's own message differs.
- [x] Mutation proof of the repaired assertions, run at `ef0376b` on a copy under a scratch dir: control → `tests 21 pass 21 fail 0`; `--verify` removed (single point) → `fail 4`, including the colliding-path test; a change these tests should *not* catch (reflowing the explanatory comment above the capture) → `tests 21 pass 21 fail 0`.
- [x] Mutation proof of the restored sweep fixture, both directions, at `ef0376b`: bare fixture + `>/dev/null` clause present → `tests 13 pass 13 fail 0`; bare fixture + clause deleted → `fail 1` (`the --verify half keys on shape, not on a list of variable names`); `--verify` fixture + clause deleted → `tests 13 pass 13 fail 0`, which is the hole the first commit opened.

Sizing: light

Closes #1146


feigi:
Review deferrals filed: **#1235** (closed `wontfix` record) — the two `suggestion`-band items that were checked by refuters and did not survive (the ref-name pairing on the `fatal: Needed a single revision` brackets; the mutant test's `assert.ok` guard wording), one calibration note recorded as no-action, and one below-the-claim-bar entry (deleting the mutant test) decided against with the measurement. Every other in-scope finding is applied in `ef0376b`. No open follow-up issue: nothing survived that was not applied.

Issue #578, opened 2026-08-18
fleet scripts: five more guards swallow git's reason — #566's claim that prove-merge.sh was the last instance is wrong


feigi:
Surfaced by `impl-566` while implementing #566, as an explicit refutation of that ticket's own claim. #566's body states that `impl-559` swept `skills/fleet/` and that `prove-merge.sh:58` was **the one remaining instance** of the stderr-swallowing guard class. That is not accurate — five sibling guards still suppress git's own diagnosis.

Filed by the fleet controller. The five call sites below were **verified verbatim against `origin/main`** before filing; the git-behaviour measurements are `impl-566`'s, attributed as such.

## The class

A guard that suppresses git's stderr (`--quiet`, `2>/dev/null`) and then `die`s with a message the script cannot actually know to be the cause. This is what PR #565 fixed in `verify-sha.sh` (#559) and PR #577 fixed in `prove-merge.sh` (#566).

## The five sites, verified on `origin/main`

**The sharp one — swallows *and* asserts an unknowable cause list:**

`skills/fleet/scripts/no-undo-audit.sh:237-238`
```sh
git -C "$wt" rev-parse --verify --quiet "origin/$branch" >/dev/null \
  || die "origin/$branch does not resolve — fetch it, or it was never pushed"
```

`fetch it, or it was never pushed` is a two-item cause list the guard cannot distinguish between — structurally the same defect as `verify-sha.sh`'s `branch missing, or no network`, which PR #565 deleted for exactly this reason. Its own preceding comment names a **third** cause the message omits: a caller who already passed a name prefixed `origin/`.

**Four byte-identical `$base` guards**, same shape as the one PR #577 just fixed:

| file:line | line |
|---|---|
| `no-undo-audit.sh:232` | `git -C "$wt" rev-parse --verify --quiet "$base" >/dev/null \|\| die "$base does not resolve"` |
| `worktree-audit.sh:47` | `git rev-parse --verify --quiet "$base" >/dev/null \|\| die "$base does not resolve"` |
| `release-ticket.sh:61` | `git rev-parse --verify --quiet "$base" >/dev/null \|\| die "$base does not resolve"` |
| `reap.sh:42` | `git rev-parse --verify --quiet "$base" >/dev/null \|\| die "$base does not resolve"` |

## Measured (by `impl-566`, git 2.50.1, throwaway repo)

[...]

PR #1232, opened 2026-09-03
fix(fleet): sweep suites decline without an ambient git working tree (#1149)

### LINKED ISSUE #1149: muted-git-guard-sweep.test.mjs and unattended-git-sweep.test.mjs throw at module load without an ambient .git — 2 of 93 files, and every review snapshot pays for it

feigi:
The sweep suites resolved the repository root at module scope with a bare `execFileSync`, so where there is no `.git` at or above them the call threw during module evaluation. Node cannot attribute a throw at module load to any test, so it synthesised one entry at line 1 of the file and the run reported a single failure in place of the file's real tests. That is the environment failing the file's precondition, but it reads like a regression in the tree under test — and it is exactly how review specialists measure this suite (#1056).

They now **decline, with a reason**, instead of throwing — but only for the one condition that licenses it.

## The class, derived rather than inherited

The defining property is a module-scope `const` bound from `git rev-parse --show-toplevel`. Against `origin/main`, before this change:

```
$ git grep -n '^const .*show-toplevel' origin/main -- skills/fleet/scripts/'*.test.mjs'
skills/fleet/scripts/muted-git-guard-sweep.test.mjs:109:const REPO = execFileSync("git", ["rev-parse", "--show-toplevel"], …
skills/fleet/scripts/unattended-git-sweep.test.mjs:53:const REPO = execFileSync("git", ["rev-parse", "--show-toplevel"],
skills/fleet/scripts/worktree-listing-sweep.test.mjs:38:const ROOT = execFileSync("git", ["rev-parse", "--show-toplevel"], …
```

All of them are covered here. The issue body named a smaller set — `worktree-listing-sweep.test.mjs` joined via `a4c46cc` after it was filed. Re-derived here rather than taken from either the body or its correcting comment, because the population had already moved once.

**Siblings checked and deliberately left**, each named:

```
$ git grep -l 'show-toplevel\|ls-files' origin/main -- skills/fleet/scripts/'*.test.mjs' \
    | grep -vE 'muted-git-guard-sweep|unattended-git-sweep|worktree-listing-sweep'
origin/main:skills/fleet/scripts/ci-state.test.mjs
origin/main:skills/fleet/scripts/ci-vacuous-green.test.mjs
origin/main:skills/fleet/scripts/instruments.test.mjs
origin/main:skills/fleet/scripts/no-undo-audit.test.mjs
origin/main:skills/fleet/scripts/refuter-scratch-prose.test.mjs
origin/main:skills/fleet/scripts/release-ticket.test.mjs
origin/main:skills/fleet/scripts/review-pr-refuter-scratch.test.mjs
```

None binds a root at module scope: their occurrences are comments, prose-assertion regexes, or fixture repositories they build themselves. `release-ticket.test.mjs` is the measured trap — a file-wide conjunction returns it, but its module-scope `execFileSync` calls are `command -v git` / `command -v awk`, so it is excluded by reading, not by the pattern. All of them run clean in the same extraction the measurements below use, which is the settling evidence rather than the grep:

```
$ env -u FORCE_COLOR node --test skills/fleet/scripts/{ci-state,ci-vacuous-green,instruments,no-undo-audit,refuter-scratch-prose,release-ticket,review-pr-refuter-scratch}.test.mjs
tests 318   pass 318   fail 0   skipped 0      EXIT=0
```

The scan would still miss a file reaching `rev-parse --show-toplevel` through a helper import, or binding it at module scope via `await`. Neither shape exists here today; neither is ruled out by the pattern.

## The change

`skills/fleet/scripts/repo-root.mjs` — one shape for all of them:

- `repoRoot(cwd)` answers the working tree, or `null` where there is genuinely none. **Every other way `git rev-parse` can fail throws**, because those leave the question unanswered rather than answering it "no" — see the next section.
- `skipWithoutRepo(root, subject)` turns that one condition into a node:test `{ skip: <reason> }` value, in the vocabulary already used for the darwin-only and gojq-gated tests. `subject` names what did not run, in the calling file's own words, so the reason cannot overclaim over a file where other tests ran.
- `trackedShellScripts(root)` keeps the list coming from `git ls-files`, and **rejects a root `repoRoot` did not answer**: `execFileSync` reads `cwd: null` as "inherit the calling process's directory", so absorbing a null root would answer some unrelated repository's scripts with nothing to say which.

**The conditions stay strictly apart.** A root that *answers* never skips, however empty its match list — an empty list is the wrong repository or a broken glob, and it must keep reaching each file's non-vacuity guard and failing there. Only a working tree that is genuinely absent skips.

The premise is unchanged: no directory walk, no untracked file in any sweep.

**Gated per test, not per file.** The fixture tests in `muted-git-guard-sweep.test.mjs` run against literal strings and its behaviour tests build their own repositories, so they keep running in an extraction rather than being refused alongside the sweeps. That file's reads of what ships also moved off module scope, so a tracked script that cannot be read now fails the test that needed it rather than taking the file down as one synthetic entry — the same defect this ticket is about, one layer in. Measured at this commit, in a tree whose root resolves and one of whose tracked scripts is `chmod 000`:

| `muted-git-guard-sweep.test.mjs` | result |
|---|---|
| reads at module scope (as this PR stood before this commit) | `tests 1   fail 1` — the whole file, at line 1 |
| reads deferred into the tests | `tests 13   pass 8   fail 5` — the tests that needed the read, by name | This refuses less than the acceptance criterion's literal "each file reports its tests as skipped", and satisfies what it is for: zero failures, no synthetic line-1 entry.

## "Cannot answer" is not "there is none"

The reason a probe that could not look must never read as an answer is the whole of #1149, so the skip is bought by exactly one condition and never by a git failure. `repoRoot` returns `null` only where git reports a discovery walk that found nothing **and** no `.git` is reachable at or above `cwd`; everything else throws, which is the loud module-load failure the pre-#1149 code produced for these.

The filesystem check is not redundant with git's message. Measured, git 2.50.1 (Apple Git-155): a `chmod 000` `.git` makes the walk step over it and print the **same** `fatal: not a git repository (or any of the parent directories): .git` as an extraction that genuinely has none — so stderr alone cannot separate that pair, and nothing but the filesystem can. `GIT_CEILING_DIRECTORIES` is honoured with git's own semantics (decline to step *up* into a ceiling entry), so a fixture that scopes git's search scopes this too.

The message match is pinned to `LC_ALL=C` on the spawn, the same pin and the same reason as `json.sh`'s. Apple's git ships no message catalogues and reads identically under `de_DE.UTF-8` and `fr_FR.UTF-8` (measured), but a git built with NLS does not, and an unmatched message would turn the extraction case from a skip into a throw.

Measured at this commit, running the two sweep files whose every test is gated (`unattended-git-sweep`, `worktree-listing-sweep`) against a full extraction of the tree placed in each fault. "before" is this same tree with `repoRoot` reverted to its bare `catch { return null }`:

| condition | before | now |
|---|---|---|
| `git archive` extraction — no `.git` anywhere | `skipped 7`, EXIT=0 | `skipped 7`, EXIT=0 |
| `.git` present but unreadable (`chmod 000`) | `skipped 7`, EXIT=0 | `fail 2`, EXIT=1 |
| `.git` naming a `gitdir:` that is not there | `skipped 7`, EXIT=0 | `fail 2`, EXIT=1 |
| a bare repository (`must be run in a work tree`) | `skipped 7`, EXIT=0 | `fail 2`, EXIT=1 |
| `git` absent from `PATH` | `skipped 7`, EXIT=0 | `fail 2`, EXIT=1 |

(The loud rows report one failure per file rather than per test: `repoRoot` throws while the module is evaluating, which is the pre-#1149 diagnostic — correct here, because nothing about the tree has been established and no test's subject is reachable.)

Each of the loud rows carries `git could not answer whether <dir> is inside a working tree, which is not the same as there being none:` and git's own stderr. The extraction row is the one that must NOT change, and does not.

## What a skip can wrongly hide, and the pin against it

A skip is an absence of coverage. Nothing inside a skipped file can pin the condition that skipped it, so `repo-root.test.mjs` pins both answers in one process, without spawning a sub-suite:

- `null` outside a repository — fixture built with `GIT_CEILING_DIRECTORIES` set to the **parent**, because git starts its search in the working directory and only declines to chdir *up* into a ceiling entry; naming the directory itself still lets the walk reach a repository above it.
- the root, and `skipWithoutRepo(root, …) === false`, inside one.
- **a repository whose tracked-script list is empty is still NOT a skip** — the exact conflation this ticket forbids.
- **every git failure that is not an absent working tree throws** — the unreadable `.git`, the stale worktree pointer, the bare repository, and no `git` binary at all. The unreadable case is its own test and declines under `uid 0`, following this repo's own convention, because `chmod 000` is no barrier to root; it first asserts git gave the indistinguishable message, so it cannot pass over a fixture that is not the fault it means to be.
- `trackedShellScripts` refuses a null root rather than answering an empty list.
- and the guard does not fire in this checkout.

Fixtures scrub `GIT_DIR`/`GIT_WORK_TREE` and assert the `.git` they created, not `git init`'s exit status — under an inherited `GIT_DIR`, `git init <dir>` exits 0 having created nothing there.

## Each sweep now defends its own list

The skip design leans on each file's non-vacuity guard to tell "wrong repository / broken glob" (must FAIL) from "no working tree" (may skip). Only `muted-git-guard-sweep.test.mjs` carried one; the other two never had one, on `origin/main` or here:

```
$ git grep -nE 'SHELL_SCRIPTS|shellScripts|\.length|assert\.ok' origin/main -- skills/fleet/scripts/unattended-git-sweep.test.mjs
59:const SHELL_SCRIPTS = execFileSync("git", ["ls-files", "*.sh"], { cwd: REPO, encoding: "utf8" })
95:  for (const f of SHELL_SCRIPTS) {
```

(`worktree-listing-sweep.test.mjs`'s only `length` assertion on `origin/main` is over the hardcoded `ALLOWED` array, not over the glob result.) A pre-existing gap, not a regression here — but this PR is what makes those guards load-bearing, so both files get one, phrased against the subjects each already names.

`worktree-listing-sweep.test.mjs`'s `shellScripts()` wrapper is gone with it: the list is a module `const`, matching its two siblings, and read once instead of per test.

## Measured, at this commit

Normal checkout, nothing skipped:

```
tests 1686   pass 1686   fail 0   skipped 0      EXIT=0
```

No-`.git` extraction (`git archive HEAD | tar -x -C <scratch>`; scratch confirmed outside any repository — `git rev-parse --show-toplevel` → `fatal: not a git repository`, exit 128):

```
$ env -u FORCE_COLOR node --test skills/fleet/scripts/{muted-git-guard-sweep,unattended-git-sweep,worktree-listing-sweep,repo-root}.test.mjs
tests 28   pass 17   fail 0   skipped 11      EXIT=0
```

No occurrence of `Command failed: git rev-parse --show-toplevel`, and none of `:1:1`, in that output. Each skipped entry carries its own reason:

```
﹣ the sweep sees the scripts it is supposed to police # no ambient git working tree at or above
  this file, so the sweep over what ships did not run and nothing about it was checked — a
  `git archive` extraction has no `.git`. Run it against a real checkout or worktree.
```

## Mutation testing

Each added test kills its own mutant, and a change it should not catch stays green. Mutants applied to copies under a scratch directory, never to the checkout; each substitution printed before the run.

| Mutant | Result |
|---|---|
| `repoRoot` reverted to the bare `catch { return null }` | `fail 2` — exactly the two loudness pins; the null/empty-list pins stay green |
| a comment in `repo-root.mjs` reworded (should NOT be caught) | `tests 8   pass 8   fail 0` |
| `trackedShellScripts` glob `*.sh` → `*.shx` | `fail 2` — exactly the two new non-vacuity guards |
| an **untracked** `.sh` added to the tree (should NOT be caught) | `tests 7   pass 7   fail 0` |

## What this newly does

Resolution moved off module scope for the absent-tree case, so a missing working tree now surfaces at test-declaration time as a skip rather than at import time as a throw. Consequence worth naming for anyone reading a snapshot baseline: a `git archive` measurement of this repo now reports skips where it reported failures, and discovers the full test count instead of losing it to synthetic entries. A reported skip on this suite remains a snapshot artifact, not a lost test — and a skip is now reachable **only** from that artifact, never from a git that could not answer.

Sizing: heavy

Closes #1149
Refs #578, #1056, #1145


feigi:
Review findings applied at `38c03d5`. Deferrals filed as **#1234** (`wontfix`, closed record — `PR #1232 review: the suggestion band, checked`), which carries every checked-and-declined entry with its dimension and a re-open trigger.

Applied: the two `survived` correctness/silent-failure findings (one defect — `repoRoot` conflating every git failure with an absent working tree), the `survived` tests finding (non-vacuity guards for `unattended-git-sweep` and `worktree-listing-sweep`), the `refuted` types finding (1-1 split, overturned — `trackedShellScripts` now rejects a root `repoRoot` did not answer), and four in-scope `suggestion` entries.

Suite at `38c03d5`: `tests 1686 pass 1686 fail 0 skipped 0`, EXIT=0.

Issue #1083, opened 2026-08-30
review-pr.js: verify-* refuter scratch dirs are never cleaned, and the default root is per-PR — a re-review reads the previous run's mutants


feigi:
Dimension: `silent-failure` (review of PR #1082).

`review-pr.js` never removes the refuter scratch directories it creates, and the
default scratch root is deterministic per PR, so a second review of the same PR
hands a *different* finding the byte-identical directory holding the previous
run's mutants.

Two lines carry it:

    workflows/review-pr.js:358   const scratch = A.scratch || `/tmp/review-pr-${pr}`;
    workflows/review-pr.js:559   rm -rf ${scratch}/snapshot

`snapshot` is the only path wiped. The refuter paths under `${scratch}/verify-*`
are created and never cleaned.

    settled by: grep -nF 'rm -rf ${scratch}' workflows/review-pr.js
                grep -nF 'A.scratch ||' workflows/review-pr.js

Corroborated operationally: on 2026-08-30, before a review run, the fleet
controller found every shared `verify-*` path still populated with four earlier
reviews' leftovers — `verify-types`, `verify-silent-failure`, `verify-comments`
and `verify-correctness` all held entries no current run had written (observed
counts 16, 14, 4 and 3). It passed that run a dedicated scratch root to work
around it.

## Why this is filed rather than fixed in PR #1082

This was raised against PR #1082 and **refuted there, correctly**: the finding
read the prompt's "That directory is yours alone" and the AC-3 pin's "isolated
across runs" as newly-false claims. They are not. The prompt's sentence carries a
colon clause — *"yours alone: every other refuter of this dimension, on this
finding and on the others, is given a different one"* — which scopes the
guarantee to within the run, and never claimed cross-run isolation. PR #1082
introduced no false claim, so the remedy did not belong in it.

The refutation is recorded here so nobody re-litigates whether #1082 caused this.
The mechanical facts above are independent of it and were verified directly.

Not #496 (which the ledger surfaces on vocabulary overlap): #496 is one scratch
dir shared by every refuter of a dimension *within* a run, fixed by PR #1082.
This is the same directory tree surviving *between* runs.

## Remedy

`rm -rf ${scratch}/verify-*` beside the existing `rm -rf ${scratch}/snapshot`.
Filed `ready-for-agent` on that basis; if wiping `verify-*` would destroy
evidence a later reader wants, the alternative is to key the root on the run
rather than on the PR.

Deferred from PR #1082 review.

PR #1233, opened 2026-09-03
fix(review-pr): give each review run its own snapshot destination

### LINKED ISSUE #1129: review-pr: every snapshot lands on one fixed path, so a second review silently replaces the tree a live fix-applier is still citing

feigi:
Closes #1129

Sizing: heavy

## What was wrong

Every artefact a review wrote was addressed by the caller's `scratch` argument plus a fixed literal, so two runs sharing one scratch root resolved to one absolute path. The fleet passes the session's scratchpad for every PR it reviews, so that is the normal case rather than a race: the snapshot block wiped its destination before extracting, and the second review therefore deleted and replaced the tree the first review's fix-applier was still citing by absolute path. Nothing failed — the path still existed, still held a plausible checkout of this repo, and answered a read with another PR's code.

## What the review caught, and why the first attempt could never have run

The first attempt minted the per-run segment in the workflow script, from `Date.now()` and `Math.random()`. The Workflow sandbox does not merely discourage those — it replaces both with functions that **throw**, so the derivation killed `review-pr.js` at its own top level, before a single `agent()` dispatched. Read out of the runtime rather than off the documentation, since this repo's assertions about the sandbox are the thing #538 exists to stop anyone trusting unmeasured:

```
$ strings -a ~/.local/share/claude/versions/2.1.259 \
  | grep -o 'Math.random = function random() { throw new Error(RANDOM_ERR) }'
Math.random = function random() { throw new Error(RANDOM_ERR) }

$ strings -a ~/.local/share/claude/versions/2.1.259 \
  | grep -o 'RealDate.now = function now() { throw new Error(NOW_ERR) }'
RealDate.now = function now() { throw new Error(NOW_ERR) }

$ strings -a ~/.local/share/claude/versions/2.1.259 \
  | grep -o 'Math.random() is unavailable in workflow scripts (breaks resume)[^"]*'
Math.random() is unavailable in workflow scripts (breaks resume). For N independent samples, include the index in the agent label or prompt.
```

That prelude is installed into the workflow VM context before the script runs. The ban protects the resume this script itself promises its caller — `resumeFor` tells a fix-applier that a relaunch replays the unchanged prefix of `agent()` calls — and a prompt carrying a fresh token every run has no unchanged prefix left to replay. So the token was not merely a resume hazard, as the review framed it; it was fatal, and the resume promise was its second casualty.

There is also no harness-side run id to seed from instead. Enumerating the globals the sandbox injects turns up `log`, `phase`, `console`, `budget`, `setTimeout`, `clearTimeout`, `agent`, `parallel`, `pipeline`, `workflow`, `args` and the ECMAScript builtins — nothing that names the run.

## What changed

The per-run segment moves to the only process this script can legally vary: the snapshot agent's shell. `mktemp -d`, under a prefix the script owns, is a stronger guarantee than the token it replaces — it **creates** the directory it names, so it cannot hand two runs one path even in the same millisecond, and it fails rather than returning a name that already exists.

The wipe goes with it. A freshly minted run root has never held anything, so the destination cannot exist before `mkdir -p` creates it and `tar -x` has nothing to merge into. `rm -rf` was the one command in that block that could reach outside the run, and it is retired rather than re-bounded.

The script keeps the one segment it can own, and nothing else spells a path off the caller's argument:

```
$ node --input-type=module -e 'import {readFileSync} from "node:fs";
  import {stripComments} from "./skills/fleet/scripts/strip-comments.mjs";
  console.log(stripComments(readFileSync("workflows/review-pr.js","utf8"))
    .split("\n").filter(l=>l.includes("${scratch}/")).join("\n"));'
const runRootParent = `${scratch}/pr${pr}`;
```

## Checks, not prose

The root comes back on a **required** `runRoot` field and is verified. Before this, the invariant every consumer depends on lived only as a paragraph asking the snapshot agent to copy `SNAPSHOT_DEST` verbatim and not reconstruct it — and a refuter reverted that paragraph and ran all 18 test files that read `review-pr.js` with nothing turning red. Prose an agent is asked to obey is not an invariant.

`snapshotMissing` now refuses a run root outside the caller's own prefix, a snapshot path outside that root, and a path that carries the prefix but climbs out of it with `..`. `usableDiff` rebuilds the diff path from the checked root rather than reading the reported one, so `diffPath` decides only whether the capture succeeded — which is what it actually reports.

Both halves are pinned, because neither substitutes for the other: a refusal is not a working review, and prose is not a guarantee. Deleting the four guards reds four named tests; reverting the paragraph reds the pin that holds it, while a pure reflow of the same sentence stays green.

## Measured, not reasoned

The mint, executed twice — two runs of one PR at one head, sharing one scratch root, which is the case a per-PR path does not fix:

```
$ sh -c '<the block's mkdir/mktemp/echo lines, rendered>'   # twice
SNAPSHOT_RUN_ROOT=<scratch>/pr1129/run-i8NfcKav
SNAPSHOT_DEST=<scratch>/pr1129/run-i8NfcKav/snapshot-466da31
SNAPSHOT_RUN_ROOT=<scratch>/pr1129/run-OqVw8xK3
SNAPSHOT_DEST=<scratch>/pr1129/run-OqVw8xK3/snapshot-466da31

$ find <scratch> -maxdepth 3
<scratch>/pr1129
<scratch>/pr1129/run-i8NfcKav
<scratch>/pr1129/run-i8NfcKav/snapshot-466da31
<scratch>/pr1129/run-OqVw8xK3
<scratch>/pr1129/run-OqVw8xK3/snapshot-466da31
```

Both trees stand. The test does this same thing rather than asserting it, for the reason the executed derivation it replaces gives: a presence pin over `mktemp` would stay green on a block that had stopped varying. It lifts only the lines up to `SNAPSHOT_RUN_ROOT` — and the block prints the root ahead of the sha precisely so those lines stand alone without `git`, because the suite is **run from a `git archive` extraction** during a review and that extraction is not a repository. Verified there:

```
$ git archive HEAD | tar -x -C <snap> && cd <snap> && git rev-parse --show-toplevel
fatal: not a git repository (or any of the parent directories): .git
$ env -u FORCE_COLOR node --test skills/fleet/scripts/review-pr-{snapshot-path,reads,testcmd,refuter-scratch}.test.mjs
ℹ tests 70 / ℹ pass 70 / ℹ fail 0 / ℹ skipped 0
```

## Two pins loosened where they rejected correct changes

The `SNAP=` sequence needle was an unquoted-only literal substring, so quoting the right-hand side of a bash assignment — a no-op, and what a shellcheck pass suggests — red it with a message claiming a #1129 collision regression that had not happened. The symlink regex pinned one arbitrary quote placement across two POSIX-identical spellings. A pin that reds on correct changes teaches its next reader to weaken it, so both now match the shape that decides behaviour. Equivalence measured, not assumed:

```
$ sh -c 'RUN="/a b/run-x"; SHA=abc123; A=$RUN/snapshot-$SHA; B="$RUN/snapshot-$SHA"; [ "$A" = "$B" ] && echo IDENTICAL'
IDENTICAL
$ sh -c 'SNAP="/a b/snapshot-abc"; f(){ printf "[%s]" "$@"; echo; }; f "$SNAP/node_modules"; f "$SNAP"/node_modules'
[/a b/snapshot-abc/node_modules]
[/a b/snapshot-abc/node_modules]
```

## Tests

Every mutant below run against all four review-pr suites, each on its own copy of a `git archive` snapshot of the shipping commit:

| mutant | expect | reds |
|---|---|---|
| `Date.now()`/`Math.random()` reintroduced | RED | the clock/RNG pin, plus both mint tests |
| `SNAP=` unquoted — behaviour-preserving | GREEN | nothing |
| `SNAP=` back onto the bare `scratch` | RED | the sequence test and the artefact-path test |
| symlink quote moved — behaviour-preserving | GREEN | nothing |
| symlink back onto the bare `scratch` | RED | the artefact-path test and the symlink pin |
| `mktemp -d` replaced by a fixed name | RED | both mint tests and the sequence test |
| the four run-root guards deleted | RED | the four guard tests, and only those |
| the `rev-parse` refusal clause dropped | RED | the named-refusal test |
| a comment naming `Date.now` reworded — behaviour-preserving | GREEN | nothing |
| the report paragraph reverted to pre-#1129 wording | RED | the prose pin, and only that |
| the report paragraph reflowed — behaviour-preserving | GREEN | nothing |
| the symlink's `[ -n "$SNAP" ]` guard removed | RED | the symlink pin, and only that |
| the symlink's two `[ ]` tests swapped — behaviour-preserving | GREEN | nothing |

```
ℹ tests 1692
ℹ pass 1692
ℹ fail 0
ℹ skipped 0
```

## The ceiling, and what it costs

The empty-scratch guard still runs first (`sh -c '[ -n "" ] || { echo SNAPSHOT_SCRATCH_UNSET; exit 1; }'` prints the marker and exits 1). What it buys is now a **named** refusal rather than blast-radius containment: with the wipe gone, the worst an unguarded empty `scratch` can do is create directories at `/`, and both commands fail there for any user who is not root. Neither `mktemp` nor `rev-parse` failing is left unread — each contributes a path component, and an unread failure would shorten the destination to a path another run could also produce, which is this ticket's own collision arriving by a different route.

The symlink is guarded on `[ -n "$SNAP" ]` for the same reason the first line tests `scratch`: it is the one command in the block whose target no earlier line created, so it is the only one that would still act on an empty `$SNAP` — writing at `/node_modules`. With the wipe gone that was the last way anything in the block could reach outside the run's own root.

Every run still leaves its own tree instead of overwriting one, and nothing removes it. That growth is #1083's, which owns cleanup for these same directories, and it is deliberately not absorbed: a wipe on return is the trade this ticket refused, since it forecloses the post-hoc inspection a fix-applier reading a finding's path depends on. #1083's other half — a re-review reading the previous run's mutants — is closed here as a side effect, since the refuter directories hang off this root too.

## Out of scope, untouched

Findings still carry absolute snapshot paths rather than repo-relative paths plus a head sha; the snapshot is still not removed on return; #1149 (the snapshot is a `git archive` extraction and therefore not a git repository) is left standing and is not made harder — nothing here changes what the snapshot contains, only where it sits.


feigi:
Review findings adjudicated at `1c3d146`.

**Filed:** #1236 — `PR #1233 review: the suggestion band, checked` (`wontfix`, closed record). It carries the four refutations with the evidence that settled each, the three below-the-claim-bar entries with a re-open trigger apiece, and what was applied.

**No open deferrals.** Every in-scope finding was either applied or measured not to be a defect; nothing was left unchecked.

**The one that outranked the rest:** the per-run token was drawn from `Date.now()`/`Math.random()`, which the Workflow sandbox replaces with functions that *throw*. This PR as previously pushed would have killed `review-pr.js` at its own top level before a single `agent()` dispatched — a defect no CI job on this repo can see, since nothing here executes a workflow script. Read out of the runtime binary rather than off the docs. The mint moved to the snapshot agent's shell.

PR #1300, opened 2026-09-08
test: add coverage for two fail-closed guards (diff-stats.mjs, no-undo-audit.sh)

### LINKED ISSUE #770: diff-stats.mjs:167 and no-undo-audit.sh:530 — two fail-closed guards with zero test coverage (severity corrected: the dramatic failure scenario was refuted)
Deferred from PR #767's review. Two independent findings, same shape, same file-adjacent area — filed together because the remedy is one harness pattern applied twice. Dimension: **tests**.

**Read the severity section before scoping this.** The originating finding stated a dramatic failure scenario that was **refuted by measurement**. The coverage gap is real; the catastrophe is not.

## Guard 1 — `skills/fleet/scripts/diff-stats.mjs:167`

```js
if (!Array.isArray(info.files)) die("gh returned no files array")
```

Zero coverage. Measured two ways, both in isolated copies of the **whole** `skills/fleet/scripts/` directory with an unmutated baseline first:

- Deleting the guard line: `node --test skills/fleet/scripts/diff-stats.test.mjs` → 13/13 pass, unchanged. A wider run (`select-dimensions`, `review-pr-testcmd`, `review-pr-reads`, `review-pr-snapshot-path`, `review-pr-citation-prose`, `review-pr-unrun`, `review-path-default`) → 105/105 pass, unchanged.
- Replacing it with the exact regression it was added to remove (`info.files = Array.isArray(info.files) ? info.files : []`): full fleet suite **1042/1042 pass, 0 fail**.

`grep` for `no files array` across the tree hits only the source line — no test names it. `diff-stats.test.mjs`'s 13 tests exercise `computeStats`/`classify` and CLI `--pr` argument handling only, never a malformed `gh` payload.

## Guard 2 — `skills/fleet/scripts/no-undo-audit.sh:530`

```sh
at_risk=$(... | awk '!seen[$1]++') || die "awk failed deduplicating..."
```

This statement is kept **deliberately separate** from the `git log`/`xargs` pipe above it — PR #767's touched comment says so explicitly — precisely so an awk-side fault gets diagnosed by name rather than swallowed into the pipe's status. Nothing forces that statement to fail and checks the message, so the separation it documents is unpinned.

## Severity — corrected, and this is the part not to skip

The originating finding claimed guard 1's absence lets `diff-stats.mjs` **exit 0 with a fabricated measurement** indistinguishable from a real empty PR. Two refuters split 1-1, and the split is settled by **method**, not by counting:

- One showed only that no test goes red under the mutation. That proves the guard is **untested**; it says nothing about whether the scenario is reachable.
- The other **ran the scenario end-to-end** with a fake `gh` on `PATH` emitting the claim's exact payload `{"changedFiles":1}`, and found it does **not** silently produce a fabricated measurement.

Running the claimed scenario is strictly the stronger method, so: **the coverage gap is real, the stated failure scenario is false.** Whoever takes this should establish what each guard actually prevents before writing a test that asserts a consequence nobody has reproduced — writing a test around the refuted scenario would pin a fiction.

## Suggested shape

`diff-stats.test.mjs` already has a stub-`gh` CLI harness (see the `CLI: a well-formed --pr value is accepted` case). Reuse it: have `gh` emit JSON with no `files` key, assert exit 2 and the `gh returned no files array` stderr. Guard 2 wants the same treatment via a `PATH` shim that makes `awk` fail.

**Each added test must kill its own mutant** — delete the guard, confirm that test and only that test reds — **and stay green under a change it should not catch.** This repo has shipped fixtures that pass with their fault injection removed entirely (#759), so the discrimination check is not optional.

## Redundancy check

`ledger.mjs check` → `verdict: tracker-hit`, `tracker.ok: true`. Hits reviewed and none covers these guards: **#739** (`review-pr.js` tests-only profile shadowing the silent-failure floor), **#625** (`reap.sh:219,245` worktree status probes sending git's diagnosis to `/dev/null` — adjacent in spirit, different guards and a different defect: those *discard a diagnosis*, these *are untested*), **#517** (CLOSED). Highest ledger near-miss 0.33 (#769, the line-distance sweep — unrelated).

feigi:
## Summary

Adds the two missing tests deferred from PR #767's review (#770). Both guards were re-verified live at their current (post-hoist) paths before writing anything:

- `scripts/diff-stats.mjs:172` — `if (!Array.isArray(info.files)) die("gh returned no files array");`
- `scripts/no-undo-audit.sh:677` — `at_risk=$(... | awk '!seen[$1]++') || die "awk failed deduplicating..."`

**Severity, per the ticket's own correction:** the originating finding's dramatic failure scenario (a fabricated measurement indistinguishable from a real empty PR) was refuted by running it end-to-end — it does not happen. The coverage gap is real; the catastrophe is not. Tests below pin what each guard actually does (refuse and name the cause), not the refuted scenario.

- `test(diff-stats)`: stubs `gh` to return `{"changedFiles":1}` with no `files` key, asserts exit 2, empty stdout, and the `gh returned no files array` stderr message. Verified the mutant (guard deleted) is killed — it crashes downstream with a different, unnamed failure instead.
- `test(no-undo-audit)`: shims `awk` on PATH to fail only the dedupe call, selected on the at-risk list flowing **through** the call rather than on the program text handed to it — the way the file's existing `withBrokenEscaper` selects — so the earlier stash-counting `awk` is untouched and a behaviour-preserving rewrite of the dedupe expression cannot silently disarm the injection. The shim also records that it fired, and the test asserts it, so a rewrite routing the dedupe away from `awk` altogether reds as "the fault never fired" instead of masquerading as a guard regression. Asserts exit 2, empty stdout, the `awk failed deduplicating the at-risk commits` message, and that the sibling guard's message (`listing commits ... git log or xargs`) is NOT what fires. Verified by hand under both `/bin/sh` and `/bin/dash` (CI's `sh`) that deleting the `|| die` leaves a bare `set -eu` abort — exit 13 (awk's own status), no named message — which the assertions correctly fail against.

## Class coverage

The ticket names exactly these two guards ("two independent findings... filed together because the remedy is one harness pattern applied twice") and scopes itself to them explicitly ("Out of scope: Nothing else is named"). No other guard is named in passing, so no other guard is in scope here.

**Accept-side (false-positive) coverage**, per guard:
- Guard 1: already covered by the existing `CLI: a well-formed --pr value is accepted...` test (well-formed `files` array, exit 0).
- Guard 2: already covered by the existing conflict-repo tests (`a plain conflicting path names the commits at risk`, the xargs-split dedupe test, etc.) — all exercise the dedupe awk succeeding and producing the correct `atRisk`.

No new accept-side tests were needed; both are pre-existing and stayed green.

**This change's own blast radius:** test-only, no production code touched. The two new tests were each verified to kill their own mutant (guard deleted) and confirmed against the existing suite (89 no-undo-audit tests, 14 diff-stats tests, all green) that nothing else regressed.

## Review findings applied

The `fleet:review-pr` fan-out returned one `survived` finding against this diff, on `scripts/no-undo-audit.test.mjs`'s `withFailingDedupeAwk`: the shim keyed its fault injection on the **literal** awk program source `!seen[$1]++`, so a behaviour-preserving rewrite to `{if(!seen[$1]++)print}` — the `|| die` guard itself untouched — fell through to the real `awk`, disarmed the injection, and left the test failing for an unrelated reason while reading as a guard regression. Fixed in `d39da08` by the content selection described above.

Discrimination matrix, `scripts/no-undo-audit.test.mjs` alone, each in its own `git archive` snapshot:

| tree | mutation | tests | fail | red test |
|---|---|---|---|---|
| baseline | none | 89 | 0 | — |
| guard removed | the `\|\| die` continuation deleted | 89 | 1 | this test, sole red |
| awk rewritten | `!seen[$1]++` → `{if(!seen[$1]++)print}`, guard intact | 89 | 0 | — (the shape that red the old fixture) |

Full suite at `d39da08`: `node --test scripts/*.test.mjs` → **1869 tests, 1869 pass, 0 fail, 0 skipped**.

## Review findings deferred

- **#1307** (`ready-for-agent`, dimension `correctness`) — `scripts/diff-stats.mjs:172` refuses a body with no `files` array but crashes with a `TypeError` at that line, exit 1, on a `null` body. Deferred as a **source behaviour change**: #770 scopes this PR to these two guards' *test coverage* and names no guard or behaviour change, and it does not claim the guard is fully covered.
- **#1308** (`wontfix`, CLOSED) — the suggestion band's durable record. One in-scope suggestion (`scripts/diff-stats.test.mjs:200`, strengthen the new fixture to `{"files":{},"changedFiles":1}`) was checked by a refuter and **refuted on measurement**: the suggested fixture is outcome-identical to the shipped one under all three conditions, buying zero discrimination.

Sizing: light

Closes #770


feigi:
Review deferrals filed: **#1307** (`ready-for-agent` — `diff-stats.mjs:172`, a null `gh` body crashes with a TypeError at the guard instead of the named exit-2 refusal; deferred as a source behaviour change outside #770's test-coverage scope) and **#1308** (`wontfix`, CLOSED — the suggestion band's durable record; the one in-scope suggestion was refuted on measurement). The `survived` finding was applied in `d39da08`.

PR #1301, opened 2026-09-08
fix(#769): settle the bare "the line above/below" distances and widen the gate to reach them

### LINKED ISSUE #769: fleet comments: seven bare "the line above/below" distances — an eleventh form no published grep reaches, and it falsifies the last of #765's scope
Found by `impl-765` while sweeping #765's ten sites, and verified by the controller at `9afea29`. Filed rather than folded into PR #767: #765 was ruled at ten sites and widening a third time on the same PR is a worse trade than a well-specified follow-up. Dimension: **comments**.

## The form

A distance of **exactly one line, with no numeral word at all** — so no numeral-word alternation can ever reach it, however many synonyms it lists.

```
git grep -rniE 'the line (above|below|before|after)' -- skills/ workflows/ .github/ hooks/
```

Seven sites, all verified live at `9afea29`:

| site | text |
|---|---|
| `skills/fleet/scripts/verify-sha.test.mjs:202` | `echoed on the line after this guard` |
| `skills/fleet/scripts/claim-ticket.test.mjs:307` | `` `$root` sits on the line above `` |
| `skills/fleet/scripts/release-ticket.test.mjs:1098` | `the line after it collided` |
| `skills/fleet/scripts/review-path-default.test.mjs:363` | `the same reason the line above is` |
| `skills/fleet/scripts/no-undo-audit.sh:360` | `the line above degrades` *(non-test)* |
| `skills/fleet/scripts/reap.sh:136` | `git's status was already taken on the line above` *(non-test)* |
| `.github/scripts/rerun-rebase-check.sh:70` | `captured into $JOBS_JSON on the line above` *(`.github/`)* |

## Why it is a separate ticket and not more of #765

**The published key cannot reach it — measured, not asserted.** #765's corrected enumeration key is *direction words + numeral words*. Piping that grep's output through a filter for this form returns **0**: no overlap at all. It is a third key, not a wider alternation of the second.

```
$ git grep -rnE '(~?[0-9]+|a few|several|a couple( of)?|one|two|three|four|five|six|some|many) lines? (earlier|later|above|below|apart|up|down)' 9afea29 -- skills/ workflows/ .github/ \
    | grep -cE 'the line (above|below|before|after)'
0
```

## It falsifies the last surviving half of #765's scope claim

#765's body claimed the class was confined to test files. A [correction on that ticket](https://github.com/feigi/claude-config/issues/765#issuecomment-5372406224) already recorded two **non-test** sites in `skills/`. This form adds a third scope hole: `.github/scripts/rerun-rebase-check.sh:70`. So the class spans test files, `skills/` sources, **and** `.github/` — all three, and each was discovered only after a grep that could not see the form returned zero.

That is the same shape three times now: *a probe that could not look returns what a probe that looked and found nothing returns.* The lesson for whoever takes this is to run all three keys and still not treat the union as certified.

## Remedy

The one #701 / PR #713 / PR #720 / PR #762 / PR #767 established: **replace the distance with the named construct**, carrying the file when constructs live in different files. Never swap a distance for a `file:line`, which rots identically.

**Per-site settling commands are mandatory — do not write one shared reason.** On PR #767, **five of ten** distances turned out factually wrong rather than merely stale (off-by-one counts, one wrong in *direction*, one nineteen lines out), and two were unmeasurable because the referent had been deleted. A shared replacement would have been false at most sites. Expect the same here.

## Two traps PR #767 hit, worth carrying

1. **A whole-file source pin can be satisfied by your comment.** `diff-stats.test.mjs:147` pins `/computeStats\(info\.files, info\.changedFiles\)/` over the entire `diff-stats.mjs` source *including comments* (unlike `arg.test.mjs`, which strips them). Naming a construct with its argument list there would have let the **comment** satisfy the pin, so the code could be deleted with the suite green. Name constructs bare, and check every reader of a file you edit for whether it strips comments.
2. **A stale quote hides inside a distance.** One site quoted a two-argument `readRules(...)` call that has taken three arguments since `snap` was added. Removing the distance is the moment that surfaces — re-pinning the quote would have preserved a second rotting claim.

## Explicitly out of class — do not "fix" these

- **Count + direction over named constructs** (~15 sites): `the two cases above`, `the three tests below`, `the four tests below pin one branch each`. These count *semantic units*, not lines — they survive reflow and line insertion, and rot only if the constructs reorder. Same reasoning that clears `quiet-payload-prose.test.mjs:139`.
- **Ordinal + line forms** (`first line`, `last line`, `a fourth line`): descriptions of *output data*, not pointers to source constructs.
- `json.test.mjs:5` (`the same ~40 lines were copy-pasted`): a size, not a position.
- `docs/metrics/tier-outcomes.tsv` rows: dated past-tense measurements. Rewriting them falsifies the record.

## Redundancy check

`ledger.mjs check` → `verdict: clean`, `tracker.ok: true`, zero tracker hits. Top near-miss is #765 itself (0.67), which this deliberately follows rather than duplicates — #765 owns the numeral+word-form keys and is being closed by PR #767 at ten sites.

feigi:
Sizing: heavy

#765's gate required a numeral word before `lines`. A distance of exactly one
line carries no numeral at all, so that pattern could never reach it however
many synonyms the alternation listed — a separate key, which is why this is a
separate ticket.

## The class, enumerated

Both sweeps were run over the current tree and unioned, and the union is the
site list:

1. **Line-oriented** — `the (line|lines) (above|below|before|after|up|down|earlier|later)`
   repo-wide.
2. **Newline-tolerant** — comment continuations collapsed (`\n\s*(#|//)\s*` → space)
   before matching, over `*.sh`, `*.mjs`, `*.js`, `*.md`, `*.yml`, to catch the
   straddling form PR #767's review found.

The newline-tolerant pass added nothing the line-oriented pass had not already
found, and no plural, ordinal or other-determiner variant exists in the tree.
Ten sites, all covered. Both sweeps return empty on this branch.

**Deliberately left:** the forms the ticket puts out of class — counts over
named units (`the two cases above`), ordinal-plus-line forms describing output
data, size descriptions, and the dated measurement rows under `docs/metrics/`.
The gate is built so none of them can trip it, asserted rather than assumed.

## Settling, per site

Every replacement was derived against the code it names and verified by
reading that code, never from a shared template. **Four of the nine real
pointers were factually wrong, not merely stale** — the failure mode PR #767
hit five times:

| Site | What it claimed | What is actually there |
|---|---|---|
| `rerun-rebase-check.sh` | `$JOBS_JSON` captured on the preceding line | captured by the `gh api` guard opening the loop body, seven lines up |
| `no-undo-audit.sh` (first) | the preceding line degrades git's `bad object` | the `refs/stash is not absent` message, assigned in an earlier branch |
| `no-undo-audit.sh` (second) | the following line asserts what the reflog contains | the `-s "$stash_reflog"` branch, past two intervening statements |
| `reap.sh` | git's status taken on the preceding line | taken by the `git cherry` probe, eight lines up |
| `verify-sha.test.mjs` (fetch guard) | `tip =` echoed on the line after this guard | the rev-parse guard sits between them — and it *passes* off the tracking ref, which is precisely why the trace's absence still proves fatality |
| `verify-sha.test.mjs` (cat-file guard) | each neighbouring guard has a marker on the following line | true of the rev-parse guard only; the merge-base guard's markers sit inside its branches |

The other three were accurate and are named anyway — a correct distance rots on
the next insertion.

`release-ticket.test.mjs` is the one site that never pointed at source: it
describes the plan's **output**, where the next line of output collided onto a
truncated one. Out of class as a source pointer, reworded only so it stops
reading as one.

**Stale quotes checked** (PR #767's second trap): both quoted strings carried
inside these comments — the test name `a healthy run stays quiet` and the
message `still names entries no ref points at` — still exist verbatim. Neither
was rotting behind the distance.

## The gate

Extended in place rather than duplicated, as the ticket directs — same file,
second arm on the pattern, plus the walk widened to `.github/scripts/` and
`workflows/`, whose sites sat outside the previous gate's directory scope
entirely.

Discrimination proven in both directions, and the red was watched before the
settling commit landed:

- **Red** on the plain, trailing-`#`, trailing-`//`, and line-wrapped spellings
  in both comment styles, on the plural, on the `before`/`after` wording, and on
  the counted `before`/`after` form.
- **Green** on a count with no direction word, a count over named units, an
  ordinal over output data, a size, a bare direction word with no `line`, a
  temporal direction word, and a `//` inside a URL.
- **Self-scan** — the widened pattern is still invisible to this file's own
  rationale, so the gate cannot trip on its own documentation (PR #767's first
  trap). Fixture text is assembled from joined tokens so no live instance ever
  sits on a line of the gate.
- The walk's reach is asserted **by name**, not by a tally, so landing another
  CI helper does not red it — and no live count is written into any comment. It
  is asserted against the files `sweep` actually opened; see the review round
  below for why a re-derived listing was not enough.

## Review round

Two `important` findings survived adversarial verification, and three
suggestion-band findings were checked here rather than left unchecked. Every
change below is proven by mutation, not by reasoning: red on the exact planted
evasion, green on the header's documented out-of-class list.

**The widened gate had the same scoping gap #769 exists to close.** `DIRS`
covered `scripts/` and `.github/scripts/` only, so `workflows/review-pr.js` —
the tree's only `.js` file, and the sole reason `MARKER` carries `".js"` — sat
outside the walk. Adding it turned up a live in-class site on the first run.
`skills/`, `commands/` and `agents/` stay out of reach and the header now says
so and why: Markdown has no comment marker to stitch blocks from, so covering it
means scanning whole files, which is a different gate.

**The counted arm had no `before`/`after`**, while `BARE` had just gained them,
so `N lines before` walked straight through the arm built to catch counted
distances. `WHERE` is now a union with `BARE` rather than a second hand-kept
list, so the two cannot drift apart again — measured clean across all three
swept directories, because `COUNT` pins a numeral immediately before `lines` and
bare temporal `before`/`after` prose never reaches that arm.

Two more sites, both **factually wrong rather than merely stale**, surfaced and
were settled onto named constructs:

| Site | What it claimed | What is actually there |
|---|---|---|
| `review-pr.js` (dimension resolution) | the call site is 500 lines down | the `const dimensions = …` call site is at 1088, 619 lines away |
| `reap.sh` (`git_probe` header) | the dirty test is the very next line after the probe | `[ -n "$gp_out" ]` is 17 lines later (468 → 485, `gp_cut_short` between); second pair 811 → 822 |

**Two pins in the gate itself were vacuous**, both proven so by mutation:

- `the walk reaches the CI helper scripts too` re-derived the listing from
  `DIRS` instead of observing `sweep`, so a sweep skipping `.github/scripts/`
  entirely stayed green — the file count moves 136 → 134, far above the `> 50`
  vacuity floor, so no count-based threshold on this shape could pin it.
  `sweep` now returns the files it opened and the test asserts against those.
- The out-of-class table asserted every source under **both** markers while each
  source carries one, so six of twelve assertions could not fail: under a
  match-everything `IDIOM` they stayed green. Each case now carries its own
  marker, as the red table already did. The URL case keeps `//` — it is the
  file's only guard on the leading `\s` in `commentBlocks`' `trail` regex, and
  it reds under a mutation that drops that `\s`.

Two comment blocks this branch had rewrapped left a four-word orphan and a
two-word tail; reflowed.

**Deferred, not dropped.** A proposal to widen `BARE` to tolerate an intervening
adverb was **refuted by running it**: it does not match its own cited site (two
intervening words, not one), it reds the tree with five hits, two of them
grammatical (temporal `before`), and it breaks the header's own ordinal
out-of-class rule. Record in #1312, filed closed. A live compound-numeral escape
(`two hundred lines above`, `select-dimensions.test.mjs:507`) is #765's class,
not this ticket's, and is filed open as #1311.

Full suite: 1870 tests, 1870 pass, 0 fail.

Closes #769


feigi:
Review deferrals filed:

- **#1311** (open, `ready-for-agent`) — line-distance gate: compound numerals (`two hundred lines above`) escape `COUNT`. Live escape at `scripts/select-dimensions.test.mjs:507`, confirmed by running the gate's own assembled `IDIOM`. Deferred as **#765's class** (a counted distance with a numeral-vocabulary gap), not #769's bare no-numeral form; remedy left open, and the body says so.
- **#1312** (closed, `wontfix`) — `PR #1301 review: the suggestion band, checked`. Durable record of the one suggestion-band finding that was checked and **refuted** (widening `BARE` to tolerate an intervening adverb: does not reach its own cited site, adds five hits including two grammatical false positives, and breaks the header's ordinal out-of-class rule — all measured), plus one entry below the claim bar with its re-open trigger. Filed closed on purpose: `ledger.mjs check` reads the tracker at `--state all`, so a future run re-deriving either lands on the evidence instead of re-litigating it.

Everything else from the review was applied in 72d75a9 and is described in the PR body's **Review round** section.

Issue #924, opened 2026-08-25
reap.sh:168 — git cherry takes $base UNQUALIFIED, so a tag named origin/main authorizes deleting an unmerged [gone] branch


feigi:
Surfaced by the `correctness` dimension during the **PR #914** review (issue #634) and verified **2-0** — both refuters reproduced end-to-end against the post-fix snapshot, and neither could refute it.

## The defect

`skills/fleet/scripts/reap.sh`:

```sh
base=${BASE_REF:-origin/main}          # ~:79
...
git cherry "$base" "refs/heads/$b"     # ~:168
```

PR #914 qualified the **branch** side (`refs/heads/$b`) precisely because a bare name is ambiguous as a rev and git resolves an ambiguous rev by preferring `refs/tags/`. **The base side was not qualified.** A tag literally named `origin/main` therefore wins over `refs/remotes/origin/main`, and the merge probe answers about the tag's commit — while `git branch -D` deletes the branch.

## Measured, post-fix `reap.sh`, exit 0, commit lost

Fixture: `feature/solo` holding a commit that exists nowhere else, pushed then deleted on origin so it reads `[gone]`; plus `git tag origin/main <that same tip>`.

```
warning: refname 'origin/main' is ambiguous.
rev-parse origin/main        => a39fa5d…   (the TAG)
refs/remotes/origin/main     => 37c4029…   (the real base)
    REAPED feature/solo
{"applied":true,"reaped":["feature/solo"],"kept":[]}
BRANCH-DELETED-unmerged-commit-lost
```

**Control, same repo, same run:** `feature/solo2` — unmerged `[gone]`, with the tag *not* reaching its tip — is `KEEP … unmerged commits` both with and without the tag present. So the deletion is caused by the base-side shadowing, not by the fixture generally.

## The existing guard cannot see it, and `--quiet` is why

`reap.sh` already validates the base near `:84` with `git rev-parse --verify --quiet "$base"`. Measured on the same fixture:

```
git rev-parse --verify --quiet origin/main  2>&1 >/dev/null   ->  (nothing)   rc=0
git rev-parse --verify        origin/main  2>&1 >/dev/null   ->  warning: refname 'origin/main' is ambiguous.   rc=0
git rev-parse --symbolic-full-name origin/main               ->  stdout EMPTY, stderr "error: … is ambiguous", rc=0
```

`--quiet` suppresses the one signal that would have surfaced it, and the rc is 0 either way. So the run is silent.

[...]

PR #1320, opened 2026-09-08
fix(fleet): point a claimed branch's upstream at its own ref, not origin/main

### LINKED ISSUE #760: claim-ticket.sh: a fresh worktree's branch tracks origin/main, so a bare `git push` silently no-ops and `@{u}` is not a guard
Surfaced by `fix-pr-756` during PR #756, and verified by the controller against both live worktrees of this run.

## The defect

`claim-ticket.sh` creates each worktree with

```
git worktree add .worktrees/<N>-<slug> -b <type>/<N>-<slug> origin/main
```

`-b … origin/main` sets the new branch's **upstream to `origin/main`**, not to its own (not-yet-existing) remote ref. Consequences, both measured this run:

1. **A bare `git push` pushes nothing and does not fail usefully.** In `.worktrees/628-withshim-migration` it printed only the `branch.autoSetupMerge` advice; `git status -sb` still read `[ahead 3]` and `git ls-remote` still showed the old SHA. It took an explicit refspec — `git push origin HEAD:refs/heads/simplify/628-withshim-migration` — to land the commits.
2. **`git rev-parse HEAD @{u}` is not a guard against this**, which is the sharp part: `@{u}` resolves to `origin/main`, so the natural "did my push land?" check compares the local head against **main** and answers a different question entirely.

Measured on this run's two worktrees:

```
$ git -C .worktrees/628-withshim-migration rev-parse --abbrev-ref '@{u}'
origin/main
$ git -C .worktrees/633-gitshim-extraction rev-parse --abbrev-ref '@{u}'
origin/simplify/633-gitshim-extraction
```

Both were created by the same `claim-ticket.sh` invocation shape. They differ only because `impl-633` happened to push with `-u`, which rewrote its upstream; `impl-628` could not, because a concurrent github.com SSH outage forced it to push via an explicit HTTPS URL, and that path does not set upstream.

## Why it matters to the fleet specifically

A member's contract is to **report a pushed SHA**, and the controller dispatches a finisher against that SHA. A silent no-op push produces a report naming a commit that does not exist on the remote. `fix-pr-756` caught it only because it independently ran `git ls-remote` rather than trusting the push's exit code — the report says so explicitly. Nothing in the dispatch prompt requires that check today.

The failure is also **order-dependent and therefore intermittent**: a member that pushes with `-u` once never sees it again, which is exactly how it stayed invisible until an unrelated SSH outage removed the accident that was masking it.

## Remedies worth weighing (not settled — pick one)

- Have `claim-ticket.sh` set the branch's upstream to its own ref at claim time, so `@{u}` means what every caller assumes.
- Leave the config alone and require the explicit refspec form in the dispatch prompts, plus a `git ls-remote` confirmation before any member reports a SHA.
- Both.

## Scope searched

`git -C <wt> rev-parse --abbrev-ref '@{u}'` on both of this run's worktrees; `claim-ticket.sh`'s `worktree add` line. `ledger.mjs check` -> not in this run's filed list; five tracker hits reviewed (#453, #607, #532, #525, #427) and none concerns push tracking or upstream configuration.

feigi:
## Summary

`claim-ticket.sh` creates each worktree with `git worktree add "$wt" -b "$branch" origin/main`. The `-b … origin/main` form leaves the new branch tracking `origin/main`, so `@{u}` **resolves — to main**. Every "did my push land?" check a member might reach for (`git rev-parse HEAD @{u}`, `git status -sb`) then answers a question about `main` and reads healthy no matter what the push did. That is #760's defect.

Fixed by passing `--no-track` to `git worktree add`, so a claim carries **no upstream at all** until its first push. `@{u}` then fails loudly instead of lying, and the real upstream arrives with the documented first push (`git push --force-with-lease -u origin HEAD`, `skills/next-ticket/SKILL.md` step 7) — after which the branch can read `[gone]`, which is exactly when `reap.sh` should see it.

### What this PR corrected about its own first cut

The first cut pre-seeded `branch.<b>.remote=origin` / `branch.<b>.merge=refs/heads/<b>` at claim time. Review proved that makes things **worse**: git computes `%(upstream:track)` against the REMOTE ref, and an upstream naming a ref that has **never existed** reads `[gone]` — not "no upstream". So a freshly claimed, never-pushed branch read `[gone]` immediately and `reap.sh --apply` deleted its worktree and branch before any work was done in it. Reproduced end to end against a real bare origin.

Two false claims in that cut are also corrected (they appeared verbatim in both `claim-ticket.sh` and the test file): a bare `git push` does **not** "silently no-op". Under git's default `push.default=simple` it fails loudly (exit 128, "upstream branch … does not match"); under `push.default=upstream` it *actively pushes the branch's commits onto `origin/main`*. Neither is silent. (The silent no-op #760 observed came from a push through an explicit URL during an SSH outage, which sets no upstream either way.)

### Second file: `release-ticket.sh`

`--no-track` breaks the release path, and this PR fixes that too. `git branch -d` measures merged-ness against **HEAD and the branch's upstream**; with no upstream it falls back to local `HEAD` alone and **refuses a pristine claim whenever local `main` is behind `origin/main`** — half-releasing it: worktree deleted, branch stranded, `in-progress` still on the ticket.

This was the review's one disputed finding (two refuters, contradictory results). Settled empirically here — the finding is **real**, and the ordering is what the contradicting refuter missed: `origin/main` must advance **before** the claim. Advance it after, and the branch's base still equals local `HEAD`, `git branch -d` finds it trivially merged, and the bug hides. Measured on a real bare-origin fixture:

| claim's upstream | local `main` behind `origin/main` | `git branch -d` |
|---|---|---|
| `origin/main` (pre-PR) | yes | rc 0 — "merged to `refs/remotes/origin/main`, but not yet merged to HEAD" |
| `refs/heads/<b>` (first cut) | yes | **rc 1 — "not fully merged"** |
| none (`--no-track`, this PR) | yes | **rc 1 — "not fully merged"** |

So the delete becomes `git branch -D`, authorized by the `ahead` and `git cherry` guards **already in the file** — both measured against a remote-tracking ref, immune to local staleness, both blocking at exit 1 before any delete. (The ref they measure with is qualified to `refs/remotes/origin/main` in the fifth round below; `$base` keeps the readable spelling in every message.)

### Third round: what `-D` took away that `-d` had been providing

Guards that run *before* the delete are not the same thing as a delete-time refusal, and the second review round found two places where that gap is load-bearing. Both are fixed here, both reproduced end to end against a real bare origin and mutation-verified in both directions (neither fix closes the other's hole):

1. **A mid-release commit was destroyed silently.** The precondition block runs before the `gh issue view` that sits between it and the first delete. A commit landing in the claim's worktree across that call leaves the worktree **clean**, so `git worktree remove`'s own second opinion allows the removal, and the commit itself reached `git branch -D` measured by nothing — destroyed at **exit 0** with `"released":true` and an empty `blockers` list. `-d` used to refuse it ("not fully merged"). The `ahead` count is now recomputed against `$base` immediately before the delete: near delete-time, unlike the precondition block, but against the remote-tracking ref, so it does not reintroduce the local-HEAD staleness `-d` fails on. It is a separate `git` invocation from the delete, so it narrows that window rather than closing it as `-d` did — see the fourth-round note below. `git cherry` is deliberately **not** recounted — a commit landing in that window is ahead of `$base` by construction. Measured: mutant deletes at exit 0 `"released":true`, fixed halts at exit 2 `PARTIALLY RELEASED — … gained 1 commit(s) since the checks` with the branch and its commit intact.
2. **`BASE_REF` naming the claim's own branch made both guards vacuous.** `origin/<branch>` *is* a remote-tracking ref, so it passed the accept-list the file already had — and then `ahead` reads 0 and `git cherry` empty whatever the branch carries. A stale remote-tracking ref left by a pushed-then-deleted branch is how one exists locally with no branch on `origin`, so the pushed-branch check does not fire either. `-d` was the backstop that made this harmless; `-D` removed it. Refused by name next to the accept-list, which is what the comment there already claimed the accept-list delivered. Measured: mutant destroys a commit that exists nowhere else at exit 0 `"released":true`; the delete-time recount above does **not** close this, because it measures against the same vacuous `$base`.

`reap.sh` authorizes its own `[gone]` deletes with `git cherry` **alone** — an earlier draft of this PR's comment called it "the same pairing", which it is not; corrected.

The `the script carries no escape hatch` test is **narrowed, not dropped**: `--force` is still banned flat; `-D` is allowed at exactly one call site and only while both authorizing guards and the delete-time recount are present. A second `-D`, or this one with any of them removed, still fails it.

Its detection was also widened. It counted `-D` only in its command-substitution form `$(git branch -D `, so an **additional, unauthorized** `git branch -D` added elsewhere as a bare statement scored zero against that count and walked through. It now counts the bare command in whatever syntax surrounds it, separating a call from a mention by the print-wrapper that heads the command the `-D` belongs to, with the `$(…)` regex kept as a secondary check on the one authorized call's form. Measured: the old assertion passes the bare-statement mutant, the new one fails it, and both accept the clean tree.

`skills/run-team/references/reaping.md`'s "`-d` here, `-D` in reap — opposite calls, both correct" paragraph was made false by this change and is rewritten.

### Fourth round: the widened detection had its own evasion, and two comments overstated

- **The wrapper test was prefix-wide, so it could be evaded.** It excluded a line from the force-delete count whenever `echo`/`printf`/`halt`/`die`/`block` appeared **anywhere** before the `-D` in that line, not only as the wrapper the `-D` actually belongs to. So a real, unguarded second delete chained after a print — `echo "cleanup" >/dev/null && git branch -D "$stray"` — kept the whole line out of the count. Measured with that exact line inserted after the authorized delete: the old filter counts 1 (green, evaded), the fixed one counts 2 (red). The prefix is cut at the last separator and the wrapper must head what remains, which still excuses the dry-run plan's own `[ … ] && echo "    would: git branch -D …"` mention — clean tree stays green. (The separator set gains `|`, and the whole match moves onto a normalized line, in the fifth round below.)
- **The recount's comment claimed more than the recount does.** It said the recount "replaces" `-d`'s delete-time refusal. It does not: it is its own `git` invocation, so a commit landing in the gap between the count and the `git branch -D` below is still force-deleted at exit 0. The comment now says it *narrows* that window, names the compare-and-swap that would close it (`git update-ref -d refs/heads/<b> <tip>` against the SHA just counted), and records why that is **not** taken — `update-ref -d` carries no equivalent of `-D`'s refusal on a branch checked out in a registered worktree, so it trades this window for a different gap in a script whose whole domain is worktrees. Filed as follow-up rather than fixed here.
- **`reaping.md` still named only two authorizers.** Its parallel paragraph — rewritten by this same PR — listed `ahead` and `git cherry` while `release-ticket.sh`'s own comment and `skills/run-team/SKILL.md` (also both rewritten here) name the delete-time recount as a third. All three now agree.
- **The `-D`-not-`-d` argument was stated three times in one file.** The header's copy is cut to the fact plus a pointer at the delete site, which carries the argument once.

### Fifth round: git's own ref-ambiguity defeated all three guards at once

- **A local ref shadowing `origin/main` made every guard vacuous simultaneously.** `origin/main` is a **shorthand**, and git resolves a shorthand through its own disambiguation order (gitrevisions), in which `refs/tags/<name>` outranks `refs/remotes/<name>`. So a local **tag** named `origin/main` pointing at the claim's own tip answers every measurement against `$base`: `ahead` 0, `git cherry` empty, **and** the fourth-round delete-time recount 0 as well — all three vacuous at once, and `-D` refuses nothing.

  Neither guard already in the file can see it. The accept-list passes it because it *is* a remote-tracking spelling; the round-three self-reference guard string-matches the literal `BASE_REF` value, and here that value is the legitimate **default**, so there is no suspicious string to catch.

  It is a regression, not a pre-existing hole: `-d` refused this ("not fully merged"). Measured on a real bare-origin fixture — claim a ticket, commit unpushed work in the worktree, `git tag origin/main refs/heads/fix/42-slug`. `git rev-parse origin/main` then resolves to the **tag** (the branch's own tip, carrying the unpushed commit) under git's own `refname 'origin/main' is ambiguous` warning, which nothing here reads. Pre-fix: **exit 0, `"released":true`, `"blockers":[]`, the branch and its unpushed commit destroyed.** Reconstructing the pre-PR code with `-D` → `-d` as the only change: **exit 2, "not fully merged", branch survives** — which isolates `-d` as the backstop this class lost.

  Fixed by qualifying the ref that is **measured** with, not by another name check: `base_rev` is `refs/remotes/$base`, where nothing is left to disambiguate, and it is used at the four sites that actually measure — the `rev-parse --verify` resolve check, the `ahead` count, `git cherry`, and the delete-time recount. `$base` is unchanged and keeps every user-facing message on the readable spelling (`1 commit(s) ahead of origin/main`). Verified against that same fixture: **exit 1, both guards fire, branch survives.** Pinned by a new behavioural test (`a local ref shadowing origin/main cannot make the commit guards vacuous`) which asserts **both** blockers, so a half-applied fix reds it, plus two source-text pins that no guard measures against the bare `"$base"` again.

- **The escape-hatch detector was defeated a third way — by requoting.** `git branch "-D" "$x"` is the same command the shell runs, and `/\bgit branch -D\b/` does not match it at all, so the line was never even collected. Measured with a real unguarded second delete in that spelling: the count stayed **1**, test green. Enumerating spellings is what lost the previous two rounds, so the detector now matches a **normalized** line instead — strip what the shell strips (quote characters, runs of whitespace) — and `|` joins the separator set so a delete reached through a pipe is no longer excused by a wrapper heading a different command. Mutation-verified both ways: clean tree stays at **1**, while `"-D"`, `'-D'`, `-"D"`, `git  branch   -D`, a piped delete and an `&&`-chained one each count **2**. Normalization is only for *finding* the call — the audited-form assertion still reads the raw source line, so the one authorized delete must stay written exactly as it is.

- **The recount pin asserted existence without the comparison.** It matched only the `n=$(git rev-list --count …)` assignment, which a recount nothing reads would also satisfy — the same vacuous guard with a line of evidence in front of it. The pin now matches the `[ "$n" -eq 0 ] || halt` refusal in the same expression.

`reap.sh` and `worktree-audit.sh` take a `BASE_REF`-like input with **no accept-list at all** and are exposed to this same ambiguity class. Both were verified rather than assumed and both stay out of this diff (out of #760's scope, and this diff is already large): `reap.sh` was already filed as **#924** — its own source comment names the hole — and is worse-placed, authorizing its `-D` on `git cherry "$base"` **alone**; `worktree-audit.sh` is newly measured and filed as **#1329** (it emits `"ahead":0,"dirty":0` for a worktree carrying one unpushed commit, and that record is what a controller reads to decide whether replacing a member would destroy work). The design-level observation behind all three — one enforcement point where there used to be two, resting partly on a prose-only convention — is filed as **#1330**.

## Verification

Full lifecycle driven end to end against a real bare origin, in the new `scripts/claim-lifecycle.test.mjs`:

- fresh claim → `@{u}` fails with `no upstream configured`; `%(upstream:track)` is empty, **not** `[gone]`; `reap.sh --apply` returns `{"reaped":[]}` and the branch and worktree survive
- `push -u` → `branch.<b>.merge` becomes `refs/heads/<b>`; merge + delete remote ref + `fetch --prune` → track reads `[gone]` and `reap.sh --apply` reaps it
- `release-ticket.sh --apply` with local `main` deliberately stale → `released: true`, no blockers, branch gone
- control: a claim carrying a commit still blocks (`commit(s) ahead`), so `-D` is not deleting unconditionally

Each mutation-verified — baseline green, and red under all three of: `--no-track` removed (pre-PR baseline), the first cut's pre-seeded config, and `-D` reverted to `-d`.

The old regression test is **deleted, not kept**: it was vacuous. `claim-ticket.test.mjs`'s fixture fabricates `refs/remotes/origin/main` with `update-ref` and configures no `origin` remote, so git sets no upstream from it regardless — measured, the pin stayed green with `--no-track` removed. That is the coverage gap that let the first cut through review, so the replacement exercises consequences (reap-ability, release-ability) rather than config values.

Push paths audited for `-u`, since the fix depends on it: `skills/next-ticket/SKILL.md` step 7 uses `git push --force-with-lease -u origin HEAD`, and its documented no-lease fallback is `git push -u origin HEAD`. Both carry `-u`; `commands/run-merge-bot.md` rebases server-side and pushes nothing. No gap to file.

Two new behavioural cases cover the third round, each mutation-verified red under its own fix's removal and green under the other's:

- `a commit landing in the check-then-act window is not deleted` — the `gh` stub commits into the claim's worktree during `gh issue view`
- `BASE_REF must not name the claim's own branch` — a fabricated stale `refs/remotes/origin/<branch>` makes both guards read vacuous, confirmed in the fixture before the run

`docs/specs/2026-07-22-run-team-agent-fleet-design.md`'s Phase-1 snippet gains `--no-track` (it was made stale by this PR), and `skills/run-team/SKILL.md` — which still told the controller the release deletes with `-d` — now names `-D`, its authorization, and which preconditions are and are not recomputed at the delete.

Full suite: `node --test scripts/*.test.mjs` → **tests 1880, pass 1880, fail 0**.

Sizing: light

Closes #760


feigi:
Second review round applied in 21569eb. Six `important` findings, all six applied — the two safety ones reproduced against a real bare origin and mutation-verified in both directions (neither fix closes the other's hole):

- a commit landing across the `gh issue view` between the checks and the delete was destroyed by `git branch -D` at **exit 0** with `"released":true,"blockers":[]` — the mutant does exactly that, the fix halts at exit 2 with the branch and its commit intact
- `BASE_REF=origin/<the claim's own branch>` passes the remote-tracking accept-list and makes both commit guards vacuous — the mutant destroys a commit that exists nowhere else at exit 0; the delete-time recount does **not** close it, since it measures against the same vacuous $base

Plus: the escape-hatch test's `$(…)`-anchored count missed an added bare-statement `git branch -D` (measured: old assertion passes the mutant, new one fails it); `skills/run-team/SKILL.md` still said `-d`; the script's comment wrongly called reap.sh's guard "the same pairing" (reap uses `git cherry` alone); the Phase-1 design-spec snippet gained `--no-track`.

Suggestion band and the two simplify-only items: checked and recorded closed in #1324. `net_kill_tree`'s ETIMEDOUT did not reproduce — two full runs, 1879 pass / 0 fail.

Full suite: `node --test scripts/*.test.mjs` → tests 1879, pass 1879, fail 0. PR body reconciled.

feigi:
Round 3 deferrals filed: **#1325** (`ready-for-agent` — the residual check-then-act window between the delete-time recount and `git branch -D`; route (b) taken in `d67561d`, route (a)'s tradeoff recorded there) and a round-3 append to the closed record issue **#1324** (`wontfix`) carrying the two below-the-claim-bar entries. Both round-3 `important` findings were applied, not deferred.

feigi:
Round 4 fix-applier — deferrals filed.

| # | state | label | what |
|---|---|---|---|
| [#1329](https://github.com/feigi/fleet-plugin/issues/1329) | open | `ready-for-agent` | `worktree-audit.sh:182` — same unqualified-`$base` ambiguity; **measured** exposed (reports `"ahead":0,"dirty":0` over an unpushed commit) |
| [#1330](https://github.com/feigi/fleet-plugin/issues/1330) | open | `needs-triage` | design-level: the delete invariant now rests on ONE enforcement point, and part of its premise (`first push carries -u`) is prose-only |
| [#924](https://github.com/feigi/fleet-plugin/issues/924#issuecomment-5597416762) | open (comment) | — | `reap.sh` was already filed for this class; independently confirmed, and the `$base_rev` fix pattern added for reuse |
| [#1324](https://github.com/feigi/fleet-plugin/issues/1324#issuecomment-5597413111) | closed (comment) | `wontfix` | round-4 below-the-claim-bar entries (`claim-lifecycle.test.mjs` simplify-only items), plus the two suggestion-band items taken rather than deferred |

Applied in a03cd02, both verified against the exact scenarios the review used. Full suite `node --test scripts/*.test.mjs` → tests 1880, pass 1880, fail 0.

PR #1322, opened 2026-09-08
docs(run-team): make a CI-watcher API outage a latched event, not silence

### LINKED ISSUE #784: phase-3 CI watcher reports an API outage as silence — "no event" is indistinguishable from "not green yet"

feigi:
Closes #784.

Sizing: light

Split out of #262, which **stays open** — the four budget-governance questions (who burns the budget, cadence, backoff policy, a cheaper behind-count probe, duplicate-monitor detection) are untouched here. This is #262's one decided half.

## The defect

The natural phase-3 watcher shape is:

```sh
st=$(node scripts/ci-state.mjs --pr "$pr" 2>/dev/null) || continue
```

`ci-state.mjs` fails closed correctly on stderr, but the watcher discards it — so a rate-limited probe emits **nothing** and "no event" becomes indistinguishable from "not green yet". Every open PR stalls unlabelled while the run reads as merely quiet. Observed live twice.

## The change

`skills/run-team/SKILL.md`'s phase-3 "Own the CI waits" block, plus one test that runs the snippet it adds. A controller reads and follows this prose to build its own watcher each run; `ci-state.mjs` itself is unchanged.

The guidance now says to check the unmetered core REST budget *before* polling and to make an outage a transition event:

- **Every probe guarded, not just the ones with a verdict field.** The block makes *three* API calls, and the third — `gh pr list` — is the one that looks like plumbing rather than a probe. A non-quota failure there (secondary/abuse limit, 5xx, network blip, expired token) never lowers `.resources.core.remaining`, so the budget gate clears and an unguarded `for pr in $(gh pr list …)` iterates zero times over the empty command substitution: no error, no latch, no line. Measured on the pre-fix block over three ticks with a healthy budget — output was `SLEEP 120` three times and nothing else, at exit 0, which is precisely the silence-reads-as-green this section exists to remove.
- **Latched, at the level each cause lives at.** One line entering a degraded state, one leaving it, never one per tick — an unlatched line at a 120s poll gets the Monitor auto-stopped for volume during even a two-minute outage, reintroducing the same blindness by another route. The budget and the open-PR list get one global flag each, because both fail at the account level and hit every PR at once; the payload latch is keyed **by PR**, because that cause is per-PR and a shared scalar for those is an unlatched line wearing a latch's clothes.
- **One gate, not a gate plus a caveat.** An empty, garbage, *or* self-named `verdict: "rate-limited"` payload is an outage, and `jq -e` exits non-zero on unparseable input and on a false result alike, so one test covers all of them — including empty stdin, which produces no result at all and which `-e` reports as exit **4** (measured). The `[ -z "$st" ] ||` pre-check the block used to carry in front of the gate was therefore the second gate this heading rules out, testing what the gate already tests; it is gone, and reintroducing it reddens five tests. The gate is `jq -e '.verdict and .verdict != "rate-limited"'`, and the `.verdict and` half is load-bearing rather than belt-and-braces: inequality alone never tests that a verdict *exists*, so `{}`, `null` and `{"verdict":null}` each evaluate `null != "rate-limited"` → true and **cleared** the gate (measured), handing an empty or error object downstream as a CI reading — against this same block's prose claim that the gate tests the verdict rather than mere parseability.
- **Both loops iterate an inline `$(printf '%s\n' "$var")`, never a bare `$var`.** zsh word-splits a command substitution's *result* but not a bare parameter expansion, so `for pr in $prs` runs ONE iteration there with every PR number glued into a single value, while `sh` and `bash` split it correctly — measured in all three. The `blind`-latch removal loop had the same bare form, where the consequence is worse than a skipped iteration: the recovered PR never leaves the list, so the block re-emits RECOVERED on every later tick — the per-tick volume the latch exists to prevent.
- **"Silence is NOT green"** is in the emitted line itself, not just the surrounding prose.
- **Recovery emits on the first good pass** even when the CI state is unchanged, so a reader can tell a recovered watcher from a dead one.
- **The block states its loop shape** — one outer tick loop, an inner pass over open PRs — and carries **all** pacing in the tick loop's single tail `sleep`. A sleep in the per-PR pass multiplies by the open-PR count; a degraded branch that `continue`s the outer loop skips the tail sleep and busy-spins probe pairs during exactly the outage it is reporting. All three degraded branches fall through to the one sleep.
- **Judge `ci-state` on its payload, never its exit code** — `not-green` is an ordinary, frequent state that exits non-zero; a watcher treating any non-zero exit as an outage misfires constantly.
- **Explicitly does not tell a controller to back off and wait.** These outages are short (rolling window; one measured reset came 24 seconds after `remaining: 0`), and pausing members for a fixed interval stalls the fleet longer than the outage would have. Report it, keep local work moving (git, tests, mutation runs are unaffected), re-probe.
- **Names the second-order effect:** while the budget is exhausted, `ledger.mjs check` reads `verdict: "unverified"` (ledger read, tracker not). Correct behaviour — but `unverified` exits **0**, the same as `clean`, so a member gating on the exit code files blind in exactly this window. Read `verdict`, not the exit code, under a flagged outage.

## Verification

The snippet is prescriptive shell a controller copies verbatim, so `scripts/watcher-degraded-block.test.mjs` **runs** it rather than matching strings in it: it lifts the fenced block out of SKILL.md, stubs `gh`, `ci-state.mjs` and `sleep` on PATH, and drives it through 15 cases. Every probe latches at the level its cause lives at; each latch is one-shot in both directions; one sleep per tick whatever went wrong; an ordinary `not-green` never trips the outage path; and the per-PR loop splits per PR under zsh as well as sh.

Both stubs take per-tick sequences, which is what makes a *sustained* outage and a *recovery* distinguishable from a single bad tick — a static reading can observe neither. For `ci-state.mjs` the sequence is also **per PR**: `$SEQ.<pr>` overrides `$SEQ` for that PR and keeps its own counter. Without that, no scenario can hold two PRs in *different* states across more than one tick, which is the only shape the per-PR `blind` latch's bookkeeping is observable in at all — and it is why three separate latch mutations (wholesale wipe on recovery, overwrite instead of append, and unanchored `case` matching) each survived the suite green until this round.

The multi-PR case asserts the **ordered** event sequence rather than DEGRADED/RECOVERED counts. Counting is not enough: the unanchored-latch mutant only *delays* a PR's line to the tick a sibling's latch clears, so it lands on exactly the real block's two-DEGRADED, one-RECOVERED totals and reads as green. Its two PRs are numbered **#4 and #42** so that the `case " $blind " in *" $pr "*` space anchoring is exercised rather than assumed.

Pacing is checked on the healthy path too, not only under the two outages. Both outage scenarios structurally route *around* the block's `: # your normal handling` line — a budget outage never enters the per-PR pass, and a payload outage `continue`s before reaching it — so a `sleep` planted on that one line turned 8 open PRs into 9 pauses per tick with the whole suite still green.

Mutation-verified — every row below re-measured against the final test file this round, each mutation applied to a scratch copy of the committed block and the worktree never written:

| mutation | reddens |
| --- | --- |
| the list probe's ERR detection defeated (`\|\| prs=ERR` → `\|\| :`) | the two list-probe tests, only |
| budget DEGRADED guard `[ -z "$budget_out" ]` → `true` | the budget one-shot test, only |
| the `budget_out=` reset deleted from the RECOVERED branch | the budget-recovery test, only |
| the gate's `.verdict and` presence half removed | the no-verdict test, only |
| the gate reduced to bare `jq -e '.'` (parseability only) | the no-verdict *and* rate-limited tests |
| the per-PR `blind` latch replaced by one shared scalar | the three per-PR latch tests |
| the `blind` latch wiped wholesale on any recovery (`blind=$keep` → `blind=`) | the multi-PR sibling test, only |
| the `blind` latch overwritten instead of appended (`blind="$blind $pr"` → `blind="$pr"`) | the multi-PR sibling test, only |
| the latch test's space anchoring dropped (`*" $pr "*` → `*"$pr"*`) | the multi-PR sibling test, only |
| `sleep 120` added to the per-PR degraded branch | both pacing tests |
| `sleep 120` added to the `: # your normal handling` line | the pacing test, only |
| `[ -z "$st" ] \|\|` reintroduced in front of the payload gate | five tests |
| the non-numeric budget guard (`case "$rl" in …`) deleted | the non-numeric test, only |
| `for pr in $(printf …)` → bare `for pr in $prs` | the zsh per-PR split test, only |
| `for b in $(printf …)` → bare `for b in $blind` | the zsh blind-latch test, only |
| a comment reworded, no behaviour change | nothing — all 15 stay green |

The block was also run verbatim, outside the test file, against the exact scenario the list guard exists for — `rate_limit` returning 5000 and `gh pr list` exiting 1 — under `sh`, `bash` and `zsh`: each emits one latched DEGRADED line and three sleeps across three ticks, where the pre-fix block emitted three sleeps and nothing else.

Also settled at commit time:

- `gh api rate_limit --jq '.resources.core.remaining'` → a plain integer at exit 0 (`4994` at the first commit, `4999` re-checked at this one). Field path confirmed current, and the endpoint is itself unmetered.
- The fenced block parses under `sh -n`, `bash -n` *and* `zsh -n`, extracted verbatim from the committed SKILL.md rather than retyped.
- `node --test scripts/*.test.mjs` → **tests 1888, pass 1888, fail 0**.

## Out of scope

The budget itself (#262). Throttling anything — measured: the monitor is not the main REST burner, review fan-out and refuters are, so throttling the monitor alone does not fix exhaustion. The merge bots' own `until` loops, which still hit a bare 403 (separate).

The in-progress-`conclusion: ""` hazard is deliberately not repeated — it is already covered at length in this same block, and this ticket's example does not key on `conclusion`.

Review deferrals: #1326 (closed record — the suggestion band, checked).


feigi:
Second review round, deferrals filed: **#1326** — `PR #1322 review: the suggestion band, checked` (`wontfix`, filed closed per `docs/adr/0002-filing-second-bar-worth-a-claim.md`).

No open issues filed: both `important` findings (the unguarded `gh pr list` probe; the untested budget-latch one-shot semantics) survived 2/2 and are applied in 25ec5a4, and both `suggestion`-band entries were checked by running them, held, and are applied too. The one entry below the claim bar — the block's external GitHub-API facts having no corroborating record in this repo — is recorded in #1326 with a re-open trigger.

Dedupe: `ledger.mjs check` exit 3, tracker hits were the per-PR record issues for #1320/#1304/#1301/#1300/#1245; none for #1322.

PR #1323, opened 2026-09-08
test(claim-ticket): pin the emitted runner's divergence-walk anchor

### LINKED ISSUE #797: claim-ticket runner: the divergence walk's $shared anchor is declared load-bearing and pinned by nothing — substituting $root reportedly survives the suite
Surfaced during the PR #787 review round (issue #230) by a refuter dispatched for an unrelated finding, and reported to the controller as "surfaced, not filed — your call". Filing it because an implementer's "worth a ticket" promise has no owner otherwise.

**Not a review deferral** — no specialist generated it, so it carries no `dimension`.

## The gap

The vendored-exclusion guard in the runner emitted by `claim-ticket.sh` walks up to the ancestor shared between the runner's own location and the argument, then matches only the diverging remainder:

```sh
shared=$root
while [ -n "$shared" ]; do
  case "$resolved" in "$shared"/* | "$shared") break ;; esac
  shared=${shared%/*}
done
case "/${arg##/*}/ /${resolved#"$shared"}/" in
```

**Substituting `$root` for `$shared` in `${resolved#"$shared"}` reportedly survives the entire suite.** Measured by the reporting member at 1049 tests (the PR's pre-fix count); the suite is 1051 post-fix, so **re-measure before acting** rather than trusting that number.

## Why it matters

The guard's own comment declares that anchor load-bearing, in the same breath as describing what goes wrong without it — that anchoring at the worktree root instead is "the opposite error", and that a symlink to a vendored tree elsewhere on disk then runs green, which is #186's class.

So the tree simultaneously asserts the anchor is load-bearing and fails to notice when it is removed. That is the same shape as two findings already confirmed in this area — the spelling term surviving deletion (fixed in PR #787) and, on the sibling PR #788, two `|| die` guards shipping with no discriminating test. A later simplifier reading `$shared` and `$root` as interchangeable has nothing that would stop them.

## Why it was left out of PR #787's scope

Correctly. The divergence walk is **pre-existing** — PR #787 touches only the fallback line and the spelling term — so guarding it would have widened a PR already carrying a correctness fix plus two test clusters. The member checked the diff before deciding, and reported rather than filing unasked.

## What a fix has to respect

- **The runner is generated**, emitted by `claim-ticket.sh`'s heredoc. Any test must exercise an *emitted* runner, not a checked-in copy — the copies under `.worktrees/*/agent-test` are decoys frozen at claim time.
- **`$root` and `$shared` differ only when the argument diverges above the runner's own directory**, so the fixture has to place them apart deliberately; a fixture where they coincide pins nothing.
- **Build the symlinked ancestor explicitly rather than borrowing `$TMPDIR`'s.** macOS supplies one for free and Linux does not, so a fixture relying on it discriminates on one platform and silently stops discriminating on the other. PR #787's own acceptance test was written this way for exactly that reason.
- The pin must red on the `$root` substitution **and** stay green under a behaviour-preserving edit, or it proves failure rather than discrimination.

## Acceptance

- [ ] Substituting `$root` for `$shared` in the remainder match reds at least one test
- [ ] That test stays green under a behaviour-preserving change to the same expression
- [ ] The fixture places `$root` and `$shared` at genuinely different paths, and builds its own symlinked ancestor
- [ ] The test drives an emitted runner, not a checked-in one

Filed under the repriced bar from #239 — the gap is confirmed by measurement (with the re-measure caveat above), and the remedy shape is open.

## Out of scope

- The fallback anchoring and the spelling term — fixed in PR #787
- #353, #401, #424, #428, #429 — related runner tickets

feigi:
Closes #797

Sizing: light

## What was unpinned

The vendored-exclusion guard in the runner **emitted** by `scripts/claim-ticket.sh` walks up to the ancestor shared between the runner's own location and the argument, then matches only the diverging remainder:

```sh
shared=$root
while [ -n "$shared" ]; do
  case "$resolved" in "$shared"/* | "$shared") break ;; esac
  shared=${shared%/*}
done
case "/${arg##/*}/ /${resolved#"$shared"}/" in
```

The guard's own comment declares `$shared` load-bearing and names what goes wrong without it. Nothing noticed when it was silently removed.

Re-measured at the current suite size, not the ticket's stale 1049:

| Mutation | Suite before | Suite after |
|---|---|---|
| `${resolved#"$shared"}` → `${resolved#"$root"}` (`claim-ticket.sh:616`) | **1873 pass / 0 fail** | reds 1 |
| `${fresolved#"$fshared"}` → `${fresolved#"$froot"}` (`claim-ticket.sh:762`, the file arm's copy of the same walk) | **1873 pass / 0 fail** | reds 1 |

The second row is the enumeration result: `$shared` has exactly one consumer, but the file arm runs a second, structurally identical walk under `$fshared`, and it was unpinned by the same measurement. One fixture holds both.

## Why every existing row was blind to it

They all name an argument that resolves **under** the worktree, where `$root` and `$shared` strip the same text and the walk does no work. The new fixture diverges above it — an ordinary directory and file living outside the worktree but under the `node_modules` ancestor it shares with the runner. `$root` is then no prefix of the resolution, `#` strips nothing, the ancestor's own `node_modules` stays in the text being matched, and a directory that is not vendored is refused as vendored. That is the "opposite error" the guard's comment names, reached from its false-refusal side.

## Fixture notes

- Drives an **emitted** runner via the existing `apply()` helper (real claim, real worktree, real `agent-test`), not a checked-in copy.
- Two rows, one per guard arm: a relative directory symlink and a relative file symlink, both pointing outside the worktree but under the shared `node_modules` ancestor. The divergence is built from the *placement* — a sibling of the repo under that ancestor — so it does not depend on `$TMPDIR` being symlinked and discriminates identically on macOS and Linux.
- **No absolute spelling of either arm**, and that is a measured decision rather than an omission. `$resolved` comes from `pwd -P`, which erases the spelling, so an absolute directory argument reaches the `case` with `$root`, `$resolved` and `$shared` byte-identical to the relative row's; it can differ only in the spelling term `${arg##/*}`, which the test directly above already pins as its control leg and which the unreadable-directory test below pins again through a symlinked ancestor. Nine mutations of this guard were run against a fixture carrying an absolute row: none reds the absolute row while leaving the relative row and those pre-existing tests green. For the file arm the reason is different — an absolute `$arg` carrying a `node_modules` segment is exempted from the file arm's resolved check by design (#401). Both reasons are recorded in the test's comment so the row is not re-added.
- Vendored legs are not repeated — the two tests directly above already pin that outside-and-genuinely-vendored still refuses.

## Discrimination, both directions

Re-measured against the final fixture:

- `${resolved#"$shared"}` → `${resolved#"$root"}` reds this test and **only** this test — `claim-ticket.test.mjs` goes 93 pass / 0 fail → 92 pass / 1 fail.
- A behaviour-preserving rewrite of the same expression stays green: swapping the two halves of the `case` subject to `"/${resolved#"$shared"}/ /${arg##/*}/"` — 93/93.

## Verification

`node --test scripts/*.test.mjs` → **1874 pass / 0 fail** on this branch (1873 before). Test-only diff; `scripts/claim-ticket.sh` is untouched.


feigi:
Suggestion band checked and adjudicated (both in scope for #797, one refuter each, biased to refuse):

- **Refuted** — `scripts/claim-ticket.sh:613`, "the walk's exact-match arm `| "$shared"` is unpinned; comment it as defensive or delete it". The arm is **live code**, so both proposed remedies would have made the tree worse. Record: #1328 (closed, `wontfix`).
- **Survived, applied in `239aaf0`** — `scripts/claim-ticket.test.mjs:456`, "the absolute `outside` row adds no measured discriminant". Nine mutations; none is red with the row and green without it.

Residual filed open: #1327 — the exact-match arm really is pinned by nothing, and it is not dead code.

PR #1366, opened 2026-09-09
docs(801): fix stale stub-surface claim in reaping.md

### LINKED ISSUE #801: reaping.md:50 says release-ticket.test.mjs stubs "only `gh`" — git, sed and awk PATH shims predate that claim, and PR #788 added two more
Deferred from PR #788 review (issue #243). Dimension: **comments**. Severity: **suggestion**.

## The finding

`skills/fleet/skills/run-team/references/reaping.md:50`

> Check: `node --test skills/fleet/scripts/release-ticket.test.mjs` — real throwaway repos, **only `gh` stubbed**, one case per precondition proving it blocks on its own.

That is false, and was already false before PR #788. `release-ticket.test.mjs` puts three other binaries on the fixture PATH:

| shim | site | added by |
|---|---|---|
| `git` | `gitShim`, ~`:232` | pre-existing |
| `sed` | ~`:366` | pre-existing |
| `awk` | ~`:2329` (`{c++}`) | pre-existing, issue #395 |
| `awk` x2 | the `#243` LOOKUP cases | PR #788 |
| `awk` x2 more | the `main_branch` / `main_wt` LOOKUP cases | PR #788 review fixes |

So the gap the sentence describes has widened from three shims to seven, across two PRs.

Each shim falls through to the real binary for every invocation it is not aimed at, which is why "real throwaway repos" stays true — it is specifically the "only `gh`" clause that is wrong.

## Scope

**Fixed in PR #788:** the same claim in `release-ticket.test.mjs`'s own header, because that PR is inside that file and is adding to the gap. It now reads:

> `gh` is stubbed on PATH for every case. `git`, `sed` and `awk` are stubbed on PATH too, but only by the cases that simulate a specific failure, and each of those shims falls through to the real binary for every invocation it is not aimed at.

**Filed rather than fixed:** this copy, because `reaping.md` is a file #243's PR has no business reaching into.

## Remedy

Bring `:50` into line with the header above — or, since the point the sentence is actually making is "no mocked git history", say that and drop the binary inventory, which is what keeps going stale. A sweep for the same "only `gh` stubbed" phrasing elsewhere under `skills/fleet/` would be worth doing in the same pass.

Deferred from PR #788 review

feigi:
## Summary

`plugin/skills/run-team/references/reaping.md`'s Check line said `release-ticket.test.mjs` has "only `gh` stubbed" — false, and false before PR #788 too (`git`, `sed`, `awk` shims already existed; #788 added more `awk` shims). PR #788 already fixed the identical claim in `release-ticket.test.mjs`'s own header, but had no reason to touch `reaping.md` since that file is outside #243's scope. Filed as #801 rather than fixed in #788.

Fix: drop the binary inventory (the thing that keeps going stale) and state the property the line is actually making, matching the wording the test file's own header already uses (git history is real, never mocked).

Settling commands, run at the commit this PR ships:
- `git grep -F -ln "only \`gh\` stubbed" -- plugin/skills` on `origin/main` before this change -> exactly one hit, `plugin/skills/run-team/references/reaping.md` -- no other carrier of the phrase (or near-miss spellings checked: "just \`gh\`", "gh alone", "solely \`gh\`", "gh only", "only gh stubbed") under the skills tree, so the ticket's "sweep elsewhere" half has nothing further to fix.
- `node --test plugin/scripts/*.test.mjs` -> tests 2125, pass 2125, fail 0 (foreground run). `reaping-prose.test.mjs` asserts other lines of this file but none touch the Check line this PR edits.

Sizing: light

Closes #801


feigi:
Fix-applier pass on the review of this PR. One finding, suggestion band (`refutersDispatched: 0` — the band's policy budget, so it arrived unchecked).

Judged **in scope** of #801 (it argues about `reaping.md:66`, the exact line this PR changed, and about which of #801's own two offered remedy branches was taken), so one refuter was dispatched against it. **Refuted, measured:** all nine `git` shims in `release-ticket.test.mjs` leave `log --all`, `for-each-ref`, the object-set hash and `.git/worktrees` byte-identical; two of them alter `git worktree list` **stdout** only, which is listing output derived from the real registry, not history. Suite baseline green: `env -u FORCE_COLOR node --test scripts/release-ticket.test.mjs` -> exit 0, `tests 128 pass 128 fail 0`.

Nothing applied — HEAD unchanged at `9ac0faf`, no new commit.

Deferrals filed: #1368 (`wontfix`, closed — the refutation record).

Issue #491, opened 2026-08-15
ci-state-prose.test.mjs: same-paragraph relocation still escapes the paragraph-bounded slice — the residual PR #487 measured and did not close


feigi:
Filed by the fleet controller from the PR #487 review (issue #198). **Not a defect introduced by that PR** — it is the measured residual left after PR #487 tightened the slice, recorded so the evidence is not re-derived.

## What #487 fixed, and what is left

The prose pins in `skills/fleet/scripts/ci-state-prose.test.mjs` originally sliced from a start anchor to the **next section marker**, so a pinned sentence could be gutted and the suite stayed green if the phrases reappeared anywhere before that anchor. PR #487 bounded the slice at the **paragraph**:

```js
const para = rest.indexOf("\n\n");
return startAnchor + rest.slice(0, para !== -1 && para < end ? para : end);
```

Measured by an adversarial refuter on a snapshot of `e3b14f2`, and again by the fix-applier on the composed tree:

- gutting the sentence and relocating the phrases into an **inserted following paragraph** — green before the fix, **red after**. Closed.
- gutting the sentence and restating the phrases as a **second sentence in the same paragraph** — **still green, 2/2**, with the fix applied.

So the escape hatch is narrowed, not closed. A pin whose purpose is anti-drift can still be satisfied by text that no longer makes the claim, provided the replacement sits in the same paragraph.

## Why it was not closed in #487

Two reasons, both deliberate:

1. **The remedy is unmeasured.** The original finding offered a second option — pin one contiguous literal span, e.g. `` /reads as `pending` in an aggregating checks summary, never here: `ci-state.mjs` names it in the reason/ `` — so a restatement elsewhere cannot satisfy it. Nobody ran it. Shipping an unmeasured remedy on a two-sentence documentation ticket was out of scope.
2. **The ceiling is already disclosed in the file.** `ci-state-prose.test.mjs:10-12` states verbatim: *"THE CEILING: these prove the clauses are PRESENT in the smallest slice that can hold them. They cannot prove a sentence added beside one does not negate it."* Same-paragraph relocation is a near neighbour of that disclosed limit.

## Why `needs-triage` rather than `ready-for-agent`

The approach is genuinely open, and the choice is not local to this file:

[...]

PR #1367, opened 2026-09-09
test(823): pin run-team SKILL.md's staleness verdict block

### LINKED ISSUE #823: the new run-team SKILL.md staleness block has no prose-pin test

feigi:
PR #813 shipped run-team phase 0's liveness prose unpinned. Settled at the
shipped commit `5fab5bf`:

```
git grep -lF 'could not check' origin/main -- 'plugin/scripts/*.test.mjs'
  → no output, exit 1
```

`staleness-qualifier-prose.test.mjs` is the near neighbour and pins #196's
finisher-dispatch twins, a different clause. `staleness.test.mjs` reads no
SKILL.md text (settled: it contains neither `SKILL.md` nor `readFileSync`).

Sizing: light

The ticket names three pieces of #813's prose and sets a floor of two. All
three are covered, split into the rules a reader must obey to use the probe:

| Pinned | |
|---|---|
| usage block | the invocation names `--path`, the flag `staleness.mjs` actually accepts |
| usage block | `--gone`/`--present` glosses bound to the direction each means |
| usage block | exactly one direction per invocation |
| exit-code table | each row's verdict bound to what it does to supply, pinned per row so a first failure cannot mask the rest |
| phase-0 instruction | a `could not check` survivor arrives annotated as one |

**Deliberately left unpinned** — the same block's rationale, which has grown
since #813: probe placement in phase 0, the `--` end-of-options separator #240
needs, the fold-into-a-neighbour hazard behind the third value, the
`origin/main`-only read, the absent-and-never-changed positive control, and the
refusal to run a reproduction. Rewriting reasoning should not be a red. The
`origin/main`-only paragraph also names `claim-ticket.sh`; no slice reaches it,
and the diff touches neither that file nor `docs/specs/` (settled: `git diff
--name-only origin/main...HEAD | grep -E 'claim-ticket|docs/specs'` → exit 1).

## Mutation-tested both ways

Reds measured on **token-preserving** semantic flips, never on vocabulary
deletion — each keeps every pinned word and flips only meaning, and each reds
its own assertion and no other:

| Mutant | Result |
|---|---|
| exit 0 and exit 2 rows swap their supply actions | 4 pass / 2 fail — the two swapped rows |
| `--gone` and `--present` glosses swapped | 5 pass / 1 fail — the directions pin |
| annotate instruction bound to a different verdict | 5 pass / 1 fail — the annotate pin |
| invocation drifts to a flag the script rejects (`--path` → `--file`) | 6 pass / 1 fail — the invocation pin |

Greens measured on benign reflow: both pinned paragraphs rewrapped at 60, 80,
120 and 400 columns (`textwrap`, `break_on_hyphens=False`) — 6 pass / 0 fail at
every width. `phrase()` joins words on `\s+`, so a rewrap is a no-op.

**The slice anchor is what anchors.** With every pinned clause gutted in place
and one stray line carrying each original wording appended at the end of the
file: sliced, 0 pass / 6 fail. The identical assertions matched over the whole
document instead of the slice, on that same document: 6 pass / 0 fail. The pass
is bought entirely by the decoy — the bound is the pin.

## What this edit newly does

A new file in `plugin/scripts/` joins the glob every suite runs under, and the
slice anchors become constraints on future edits to that SKILL.md section: the
table header row, and the sentences opening the usage and survivor-annotation
paragraphs. A moved anchor throws with a message saying to re-anchor rather than
widen, instead of silently matching the whole file; so does a DUPLICATED one,
which would otherwise bind the pin to the wrong copy.

`paragraph()` is added to `prose-pin.mjs`, the shared module, rather than copied
a fifth time — `ci-state-prose`, `liveness-rationale-prose`, `quiet-payload-prose`
and `staleness-qualifier-prose` each hand-rolled the same search-to-blank-line
slicer (settled: `grep -lF 'indexOf("\n\n")' plugin/scripts/*.test.mjs` → those
four). Migrating them is out of scope: each is an already-merged,
mutation-tested deliverable of its own ticket, and those four carry both
defects this PR's review found in `paragraph()` — filed as #1372. Its guards
are unit-tested in `prose-pin.test.mjs`, covering the decoy the blank-line
bound defeats, the decoy a blank line CARRYING WHITESPACE defeats, the reflow
a literal `indexOf` anchor would have reddened, the moved-anchor throw and the
duplicated-anchor throw.

## Declared ceilings

- These are positive pins: they see a clause deleted or reworded, never a
  contradiction added beside it. A sentence appended after "Give exactly one"
  permitting both flags leaves this suite green.
- The reflow ceiling is hyphen-breaking, as `staleness-qualifier-prose.test.mjs`
  records for its own pins. It does not bite here, because no clause pinned in
  this file carries a hyphenated token — a fact about the current wording, so a
  reword introducing one inherits the ceiling.
- The bold markers around `**could not check**` are pinned literally. Loosening
  them to tolerate a seam is the `\*{0,2}` trap; a deliberate un-bolding reds
  and should be re-anchored.

## Applied from this PR's own review

Three findings landed in `2cd5cc5` after the review; all three are
mutation-verified against the real SKILL.md rather than a fixture.

- **The slice bound was defeated by a blank line carrying whitespace.**
  `rest.indexOf("\n\n")` does not see `\n   \n`, which is what an editor
  writes when it keeps a list item's indent. Reproduced: with the exit-2
  verdict row gutted in place, one space on the blank line ending the table
  and the original wording restated below, the old bound reported 6 pass / 0
  fail — fully green with the real rule gone. Now `rest.search(/\n[ \t]*\n/)`.
- **The anchor took the first match with no uniqueness assert.** An
  un-gutter'd restatement of the anchored block placed ABOVE the real one
  bound the pin to the copy while the real clause was gutted — also measured
  green. `paragraph()` now asserts exactly one match, the standard
  `markedLine` already holds its own marker to.
- **The invocation's `--path` was pinned by nothing** (above), so the doc
  could drift to a flag the script rejects with no red; a reader following
  the drifted doc gets exit 2, which this table then reads as could-not-check.

The two decoy fixtures in `prose-pin.test.mjs` were re-derived for the
uniqueness assert: they repeated the ANCHOR, where the decoy that matters is a
copy of the pinned WORDING. They now duplicate only the wording, and the
duplicated-anchor case gets its own throw test. The pre-existing clean-`\n\n`
decoy test passes both before and after the bound fix, so it could not see that
regression — the whitespace-bearing decoy test is what reds when the bound is
reverted, and it reds alone.

The remaining six suggestion-band findings were each checked and none survived;
recorded closed as #1373.

Full suite: 2138 pass, 0 fail.

Closes #823


feigi:
Review deferrals filed, pushed as `2cd5cc5a9e875bda78713071dcc87beda25727b6`.

| Issue | State | Label | What |
|---|---|---|---|
| #1372 | open | `ready-for-agent` | The four hand-rolled paragraph slicers (`ci-state-prose:42`, `liveness-rationale-prose:60`, `quiet-payload-prose:144`, `staleness-qualifier-prose:99`) carry both defects fixed here in `paragraph()` — literal `indexOf("\n\n")` bound and first-match anchor. Residual surfaced by the tests dimension, out of #823's scope by this PR's own recorded decision. Related but distinct: #491. |
| #1373 | closed | `wontfix` | Refutation record — the six suggestion-band findings, each checked, none survived. |

Applied: three findings (slice bound, anchor uniqueness, the invocation pin), all mutation-verified against the real `SKILL.md`. Full suite `tests 2138 / pass 2138 / fail 0`.

PR #1375, opened 2026-09-10
docs(1371): correct pkill victim signature in run-team SKILL.md

### LINKED ISSUE #1371: run-team SKILL.md states the pattern-kill victim signature as `cancelled`/exit-144 — a SIGKILLed node --test leaves NO summary block, so the stated marker fails toward clean
run-team SKILL.md's pattern-kill paragraph states the victim signature as `cancelled`/exit-144. Measured this run: a `pkill -9`'d `node --test` leaves **no summary block at all**, and the stated signature fails toward "clean".

## Where the claim lives

`plugin/skills/run-team/SKILL.md`, the "A pattern kill reaches siblings" bullet:

> The victim sees a `cancelled`/exit-144 it cannot attribute, on a run it did not abort

## What was measured

2026-09-09, during a `/fleet-ctl:run-team 2 5` wave. A review specialist ran `pkill -9 -f "node --test"` twice, machine-wide, at roughly 20:25-20:32Z.

`impl-1073` had a bare `node --test` run finish at **20:29:27Z**, inside that window. Its output file contains **zero** reporter lines:

```
grep -acE '^ℹ|^TAP|^ok|^not ok' <output>   ->  0
```

No `ℹ tests` line, no `cancelled` line, no TAP body. The member initially misdiagnosed this as the 120s Bash tool timeout, and corrected itself on the timing plus the emptiness: a tool timeout on a ~384s run would still have left partial stdout, because the reporter streams as it goes.

`cancelled` implies the test harness lived long enough to report a cancellation. `SIGKILL` means it did not.

## Why the stated signature fails in the dangerous direction

A member told to look for `cancelled` searches a killed run's output, finds no `cancelled` line, and reads that as "not cancelled, therefore not killed". Absence of the marker reads as absence of the fault. That is a false clear on exactly the runs the rule exists to catch — and a killed **baseline** compared against a clean mutant is a wrong measurement that looks like a real result.

The controller relayed this wording verbatim into three dispatch prompts this run before it was falsified.

## The discriminator that does work

**Presence of the `ℹ tests/pass/fail/cancelled` summary block**, not its contents.

Every surviving run in this wave carried a complete block; the one casualty carried none. Two independent corroborations from the same window:

- `fix-pr-1366`'s refuter baseline: `exit 0, tests 128 pass 128 fail 0 cancelled 0` — a complete block, so not killed.
- `impl-1073`'s leg B: `tests 127 / pass 127 / fail 0 / cancelled 0` — the file's full count minus exactly the one case it had removed. A process that died would not arrive at 127.

The count is a second, independent discriminator: a killed run cannot report a full count.

`impl-804` reported the same shape from the other side — `rc=137`, zero bytes of output, empty log — which is the same "no block" signature seen through the exit code.

## Suggested direction (remedy open)

Correct the bullet's stated signature. Roughly: the victim sees an output file with **no summary block at all** (SIGKILL), or `rc=137`; `cancelled` appears only when the harness survived to report, so it is the wrong thing to search for. Whether to also state the positive test ("check the `ℹ tests` line is present, and that its count matches what you expected to run") is a judgement for whoever takes it.

Worth checking in the same pass whether any other fleet prose states the same signature.

## Why this is filed rather than fixed mid-run

The mid-run tooling-fix rule exists because deferring loses the evidence that found the defect. The evidence is fully captured here, so nothing is lost by filing. Against that, editing `SKILL.md` in the main checkout while four PRs are open and a merge wave is live means mutating a tracked instrument mid-wave and re-pinning under it; the proportionate call was to capture the measurement durably instead. No work was lost to the misdiagnosis — the member had already discarded the killed run and re-run it before reporting.

feigi:
## Summary

`plugin/skills/run-team/SKILL.md`'s "A pattern kill reaches siblings" bullet stated the pkill victim's signature as a `cancelled`/exit-144 it cannot attribute. That was wrong. The first correction in this PR replaced it with a *different* wrong claim — that a SIGKILLed run leaves no summary block, therefore `cancelled` is "the wrong thing to search for", and that a complete green block is "kill-proof by construction". Review falsified both. The bullet has now been **re-derived whole** from measurement rather than patched, because the falsified sentences overlapped.

There is not one victim signature. There are four, and one of them is a clean green.

| what was signalled | what the runner leaves | rc |
|---|---|---|
| `pkill -f <pattern>` (SIGTERM — pkill's **default**) | `Interrupted while running:` then a **complete** block with `cancelled` non-zero | 1 |
| `pkill -9 -f <pattern>` (SIGKILL) | no summary block at all | 137 |
| a kill reaching only a **child** test process | complete block, `fail` non-zero, `cancelled 0` | 1 |
| a kill reaching a process a **test itself spawns** | complete `pass N / fail 0 / cancelled 0` at the expected N, over a corrupted run | 0 |

Each reproduced in this session against a fixture spawned here and killed by its own PID (node v26.8.1/darwin). The fourth is the sharpest and is why "kill-proof by construction" had to go: `spawnSync` reports a SIGKILLed child as `status null`, so a test asserting only that the process it spawned exited non-zero — `assert.notEqual(r.status, 0)`, a form used across this suite in files that spawn the shell scripts they cover — passes vacuously.

```
$ node -e 'const {spawnSync}=require("child_process");
  const r=spawnSync(process.execPath,["-e","process.kill(process.pid,\"SIGKILL\")"]);
  console.log(r.status, r.signal)'
null SIGKILL
```

So the guarantee the bullet now states is the narrow one that survives: **a complete summary block proves the runner survived, not that the run was uncontaminated.**

## What else changed in the bullet

- **`cancelled` is restored as a signature**, split by signal. It is exactly the right thing to search for in `pkill`'s default SIGTERM case, and two of the three kill-the-node-process signatures exit `rc=1`, not 137.
- **argv is stated as an unreliable partition, not as no partition.** Measured with `ps -Ao pid,ppid,args -ww` over two fixtures spawned here: an absolute-path invocation puts a member-unique string in argv; a repo-relative one is byte-identical across members. (Without `-ww`, macOS `ps` truncates and the truncation reads as an absence.)
- **The cited command and path are corrected.** `ls scripts` → `No such file or directory`; the suite is `plugin/scripts/*.test.mjs` (124 files at this commit). Members are briefed on `./agent-test <file-or-dir>` and nothing else (`sed -n '455,462p'` of the same file), and that runner ends in `exec $testcmd "$@"` (`plugin/scripts/claim-ticket.sh`), so node always receives the file list the caller's shell already expanded — never a literal glob.
- **"from the same checkout" is dropped.** This document's own model is per-member worktrees (`sed -n '90,95p'` and `sed -n '455,462p'`), and `agent-test` derives its ports from its own directory name, which only works because each member holds a different checkout. The reason a pattern kill reaches siblings is the machine-wide process table plus filenames identical in every worktree — not a shared checkout.
- **No counts in the prose.** The vacuous-pass assertion shape is stated as a property, not as a number of occurrences, so it cannot rot on the next commit — which is the failure mode this whole bullet is being corrected for.

The operational rule is unchanged and was never in question: **do not pattern-kill; clear only your own leftovers, by PID.**

## Scope note

#1371 generalised from two `pkill -9` incidents to all pattern kills. The measurement does not support that generalisation — `pkill`'s default signal produces the opposite signature from `pkill -9`. The bullet now says what is true rather than what the ticket assumed; the ticket's operational conclusion is unaffected.

## Sibling safety

`plugin/scripts/prose-pin.mjs` and the four `*-prose.test.mjs` files are live in PR #1376. This diff touches no file under `plugin/scripts/`. Three of those pins slice this same `SKILL.md` via anchors `**Both halves of that command are load-bearing.**`, `**Own the CI waits.**`, `` Gate on the `check` job `` and `**The fix-applier pushes and exits` — all outside the bullet this diff rewrites. Re-verified post-edit at the shipping commit: `grep -cF` returns `1` for each.

## Suite

`./agent-test plugin/scripts` → `tests 2138`, `pass 2138`, `fail 0`, `cancelled 0`.

Diff is confined to `plugin/skills/run-team/SKILL.md`.

Sizing: light

Closes #1371


feigi:
**Review deferrals filed:** #1378 — `docs and test headers cite `node --test scripts/` — a path this repo does not have`, `ready-for-agent`, dimension `comments`. It is finding M3 widened: the bullet's own occurrence is fixed here, the other 31 files (`grep -rlF -- 'node --test scripts/' plugin/ | wc -l` → 32) are out of #1371's scope and partly under `plugin/scripts/`, which this diff must not touch while #1376 is live.

All six review findings are applied in this PR, none refuted. No `wontfix` record issue for this review.

PR #1370, opened 2026-09-09
fix(804): bound claim-ticket.sh's arity and guard the --apply slot

feigi:
Closes #804

Sizing: light

`claim-ticket.sh` was the last of the four scripts taking `--apply` to carry #250's demotion. Its guard was an arity **lower bound alone**, with no value check on the flag slot.

## The defect, re-derived against `origin/main` before implementing

Both directions measured on `origin/main`'s script against a fresh fixture repo, with a stubbed `gh` so nothing could reach the live tracker.

**1. A mistyped flag was silently demoted to "not `--apply`".**

```
$ sh claim-ticket.sh 5 slug fix
$ sh claim-ticket.sh 5 slug fix --aply
```

Both rc=0 with `"applied":false`, and `cmp` reported stdout **and** stderr byte-identical. Nothing anywhere said the flag was not understood.

**2. Trailing arguments were dropped and the run entered APPLY mode.**

```
$ sh claim-ticket.sh 5 slug fix --apply extra --whatever
```

Reached `gh issue edit 5 --add-label in-progress`, then died `could not label issue 5` — rc=2 from the stubbed `gh` refusing, **not** from an argument refusal. Args 5 and 6 were dropped without a word and the run mutated.

## The remedy

The shape both already-guarded siblings use, lifted unchanged: an `[ $# -ge 3 ] && [ $# -le 4 ]` arity bound plus a `case "${4:-}"` accepting `''|--apply` and refusing everything else at exit 2.

## Enumerate and declare — the whole class

The ticket names four scripts taking the flag. Established for myself rather than taken on trust:

| Script | Slot | State on `origin/main` | This PR |
|---|---|---|---|
| `claim-ticket.sh` claim path | `$4` | **unguarded** — arity lower bound only, no value check | **covered** |
| `claim-ticket.sh --write-runner` | arity | **unguarded** — `[ $# -ge 2 ]`, no upper bound | **covered** |
| `release-ticket.sh` | `$4` | already guarded, refuses at rc=2 | left |
| `drop-merged-label.sh` | `$2` | already guarded, refuses at rc=2 | left |
| `reap.sh` | arity + value | **already guarded** | left |

**A correction to the phase-0 probe.** `git show origin/main:plugin/scripts/reap.sh | grep -c 'unknown argument'` returns 0, but `reap.sh` is *not* unguarded — it refuses the same typo under its own wording:

```
$ git show origin/main:plugin/scripts/reap.sh | grep -F "unrecognised argument '"
[...]

Issue #608, opened 2026-08-18
REFUTED: keep the full relocate() rationale in all three suites, and no byte-identity test over the copies


feigi:
Deferred from PR #601 review (issue #178). Dimensions: `simplify` and `types`. Location: `skills/fleet/scripts/reap.test.mjs:90`, `release-ticket.test.mjs:109`, `worktree-audit.test.mjs:73`.

Filed as a **record, not as work** — `wontfix`. This covers **two** findings from the same review that share one subject: the duplicated `relocate()` docstring (`simplify`) and the absence of a byte-identity test over the three `relocate()` copies (`types`). Both were refuted; recorded together so a future `code-simplifier` or `type-design-analyzer` pass does not re-derive either.

## Finding A — "shorten two of the three docstrings to a pointer" (`simplify`)

CLAIM: the 11-line rationale above `relocate()` is repeated verbatim in three suites, so the explanation drifts unguarded while the code it explains is only 4 lines — the repo's convention for a duplicated definition is to state the reason once and name the file that holds it.

SUGGESTED_FIX: keep the full rationale on `release-ticket.test.mjs` and shorten the other two to the `reap.sh` form (`Same fixture and same reason as release-ticket.test.mjs's own relocate(); named there, not by line number, per #129`).

### Refutation — measured

**The cited precedent says the opposite of what the finding needs.** `reap.sh:20-26` keeps a **full standalone rationale** and *adds* the pointer as its closing clause; the finding quotes only line 25 — mid-sentence — and presents that clause as the whole treatment:

```
# Is $1 established ABSENT, or merely a path this script cannot stat? A bare
# `[ -e ]` failure is both — an unreadable parent fails it identically to a
# directory that was actually removed — and only the second is nothing to
# protect. Walk up to the nearest ancestor that exists and require THAT to be
# searchable: only then is "not there" a measurement, not a guess. Same shape
# and same reason as release-ticket.sh's own `gone()` (not shared code — the
# callers differ in nothing else); named there, not by line number, per #129.
gone() {
```

`worktree-audit.sh:31-38` is the same shape, with the rationale *retailored* to its own call site (`[ -d ]` not `[ -e ]`). The two shell copies are **not** verbatim — itself evidence the repo writes per-copy prose rather than deferring it. The suggested fix would delete the rationale and keep only the pointer: the inverse of the precedent.

[...]

PR #1376, opened 2026-09-10
fix(1372): migrate the hand-rolled paragraph slicers onto prose-pin's paragraph()

### LINKED ISSUE #1372: prose pins: the four hand-rolled paragraph slicers carry both defects #823 just fixed in `paragraph()`
Filed by the fix-applier from the PR #1367 review (issue #823). **Not a defect introduced by that PR** — it is the residual its tests-dimension specialist surfaced and explicitly scoped out, recorded so the evidence is not re-derived.

**dimension:** tests (residual surfaced while checking finding T1)

## The defect

PR #1367 fixed two independent false greens in `plugin/scripts/prose-pin.mjs`'s exported `paragraph()`. Four already-merged pin files hand-roll the *same* slicer locally and carry *both* defects by construction:

- `plugin/scripts/ci-state-prose.test.mjs:42` — `const para = rest.indexOf("\n\n");`
- `plugin/scripts/liveness-rationale-prose.test.mjs:60` — `const end = rest.indexOf("\n\n");`
- `plugin/scripts/quiet-payload-prose.test.mjs:144` — `const end = rest.indexOf("\n\n");`
- `plugin/scripts/staleness-qualifier-prose.test.mjs:99` — `const end = rest.indexOf("\n\n");`

**Defect 1 — the literal bound.** `indexOf("\n\n")` does not see a blank line carrying whitespace. An editor keeping a list item's indent writes `\n   \n`; the slice then widens silently into the next paragraph and a decoy there buys the green the bound exists to deny. Measured on `paragraph()` against the real `plugin/skills/run-team/SKILL.md`: with the exit-2 verdict row gutted in place, one space on the blank line ending the table, and the original wording restated below it, the literal bound reported `tests 7 / pass 7 / fail 0` — fully green with the real rule gone. `search(/\n[ \t]*\n/)` reds exactly that pin.

**Defect 2 — first-match anchor.** Each of the four locates its anchor with a plain search and takes the FIRST match with no uniqueness assert, so an un-gutter'd restatement of the anchored block placed ABOVE the real one binds the pin to the copy while the real rule is gutted. Measured the same way: restating the usage block above the real one and gutting the real clause reported `tests 7 / pass 7 / fail 0`. An `assert.equal(hits.length, 1, ...)` — the standard `markedLine` already holds its own marker to — reds it.

Both are **latent, not live**: the repo has zero whitespace-only lines in any `*.md` today, and the anchors in these four files are unique today. Nothing enforces either condition — there is no `.editorconfig`, no prettier or markdownlint config, and no CI check for trailing whitespace — and every paragraph pinned so far is indented list content, which is exactly where an editor produces `\n   \n`.

## Why it was deferred rather than fixed in #1367

Out of #823's scope, and deliberately so before this review existed. `plugin/scripts/prose-pin.mjs`'s own header comment records the decision verbatim:

> Extracted here rather than copied a fifth time: `ci-state-prose`, `liveness-rationale-prose`, `quiet-payload-prose` and `staleness-qualifier-prose` each hand-rolled this same search-to-blank-line slicer locally. Those four are already-merged, mutation-tested deliverables of their own tickets, so migrating them is deliberately left out of #823 — this export exists so the next pin does not become a fifth copy.

Migrating them touches four other tickets' reviewed deliverables and would invalidate their mutation testing without a matching re-review pass. The tests specialist stated the same scope limit itself: it checked only the three anchors #1367 uses, in one SKILL.md, and did not audit these four.

## The remedy is not open

`paragraph()` now exists and is the intended destination — that is why it was extracted. Migrating each of the four onto it removes the copy and closes both defects at once. The open question is only re-verification, not design.

## Acceptance criteria

- [ ] All four files import `paragraph` from `./prose-pin.mjs` and the local slicers are deleted
- [ ] Each file's existing mutation record still holds — re-run the token-preserving semantic mutations each file documents, confirming each reds its own assertion and no other
- [ ] For each file, the whitespace-bearing-blank-line decoy reds, and the duplicate-anchor restatement reds
- [ ] Each file's anchors are confirmed to occur exactly once in its target document, since `paragraph()` now throws when they do not
- [ ] Full suite green (`./agent-test plugin/scripts`; it reported `tests 2138 / pass 2138 / fail 0` at the time of filing)

## Related, not duplicate

**#491** — same family, different mechanism. It records that *same-paragraph relocation* still escapes a paragraph-bounded slice: gutting a sentence and restating its phrases as a second sentence in the same paragraph stays green even with a correct bound. Neither defect above closes that one, and closing that one would not close these. Whoever picks up either should read both.

## Provenance

Deferred from PR #1367 review (issue #823), head `5fab5bf9904a3ae0301e2287804fa2bf782be8e4`; fix pushed as `2cd5cc5a9e875bda78713071dcc87beda25727b6`.

feigi:
Four pin files hand-rolled the same search-to-blank-line slicer that #823 extracted as `paragraph()`, and every copy carried both false greens #1367 fixed there. Each now imports the shared bound; the local copies are gone.

**Sizing: heavy** — four files plus the shared module, and per-file re-verification of anchor multiplicity and slice equivalence was the real unknown. Entered at `writing-plans` as an unattended fleet member (`brainstorming` is the maintainer-present step).

## Enumerate and declare

Censused **by shape** — an anchor search followed by a blank-line bound and a slice — never by the helper's name, since a name-based census is what missed these originally.

```
git grep -nF 'indexOf("\n\n")' -- 'plugin/scripts/*.mjs'          # the literal bound
git grep -nE 'split\(/\\n\\s\*\\n/\)|search\(/\\n' -- 'plugin/scripts/*.mjs'   # regex blank-line bounds
```

| file | verdict |
|---|---|
| `ci-state-prose.test.mjs` | migrated |
| `liveness-rationale-prose.test.mjs` | migrated |
| `quiet-payload-prose.test.mjs` | migrated (document sites; source site takes `anchorAt`) |
| `staleness-qualifier-prose.test.mjs` | migrated |
| `review-pr-citation-prose.test.mjs` | **left, with reason** |
| `finisher-name-prose.test.mjs` | **left, with reason** |

The census found **two class members the ticket does not name**, both carrying defect 1 by construction:

- `review-pr-citation-prose.test.mjs` — `section.indexOf("\n\n", at)` after a first-match anchor. **Left deliberately**: it also asserts `notEqual(end, -1)`, so a missing blank line *reds* there. `paragraph()` has no such assert — it returns the remainder. Migrating would trade a literal-bound false green for a **new** silent widening to the end of the enclosing section, and would drop a bespoke diagnostic far richer than `paragraph()`'s generic one. That is a net loss, not a fix; it needs a variant that throws on a missing blank line.
- `finisher-name-prose.test.mjs` — `between(doc, "Names follow", "\n\n", label)`, i.e. defect 1 reached through the shared `between()`. **Left deliberately**: it is a `between` call, not a hand-rolled paragraph slicer, and its own doc-comment records `between` as a reasoned choice. Migrating means rewriting another ticket's reviewed rationale. Its anchor is unique (`Names follow` occurs 1× in `skills/run-team/SKILL.md`), so it is migratable when someone picks it up.

Both are now filed, as #1377. Neither is live today — no tracked `.md` carries a whitespace-only line:

```
git grep -lE '^[ \t]+$' -- '*.md'      # -> no output, exit 1
```

The three `\n\s*\n` splitters (`tracker-block-copy`, `watcher-key`, `no-ci-gate`) are **not** class members: `\s` already covers a whitespace-bearing blank line, so they carry neither defect.

## What this migration can wrongly REFUSE

`paragraph()` throws when an anchor is not unique, so migrating a file whose anchor is duplicated turns a passing pin red. **Every anchor was checked before migrating. All occur exactly once, so none needed narrowing and nothing this PR guards changed.**

`paragraph()`/`anchorAt()` throw on a non-unique anchor, so a green run of the migrated files *is* that check:

```
node --test plugin/scripts/prose-pin.test.mjs plugin/scripts/ci-state-prose.test.mjs \
  plugin/scripts/liveness-rationale-prose.test.mjs plugin/scripts/quiet-payload-prose.test.mjs \
  plugin/scripts/staleness-qualifier-prose.test.mjs
# -> tests 31 / pass 31 / fail 0
```

## What the edit newly does

Nothing to what the migrated pins match — measured, not assumed. **Every migrated slice is byte-identical to the hand-rolled one it replaces**, so no assertion matches different text and no pin passes for a new reason. The bound change is latent by the whitespace-only-line check above; the uniqueness assert is latent because every anchor is unique.

`ci-state-prose`'s copy also took a second bound, the opening of the next paragraph, and used the blank line only as a tightener. In both documents the blank line already falls where that opening begins, so the slice WIDTH is unchanged:

```
node -e 'const fs=require("fs");for(const[d,a,e]of[["plugin/commands/review-and-fix.md","Its job-presence check is what catches the case above","\n\n**A repo with no workflow files"],["plugin/commands/run-merge-bot.md","That presence requirement is what catches the case above","**Re-query at the moment you merge"]]){const t=fs.readFileSync(d,"utf8"),r=t.slice(t.indexOf(a)+a.length);console.log(d,r.search(/\n[ \t]*\n/)<=r.indexOf(e))}'
# -> both true (they coincide in review-and-fix.md, whose end anchor opens on that very blank line)
```

Width was not what that second bound bought. Its **assert that the opening was reachable** was — and dropping that was a real loss of coverage. See **Review fix**.

## Why `anchorAt`

`quiet-payload-prose`'s source site cannot use `paragraph()`: in `ci-state.mjs` it anchors on a declaration and takes the `//` block **above** it, bounded by code at both ends (#695). It would otherwise have kept a first-match anchor or copied the assert — a fifth copy of the thing being removed. The uniqueness half is now `anchorAt()`, which `paragraph()` also calls, so that site closes too. Behaviour and messages are unchanged. `prose-pin.test.mjs` gains one case for `anchorAt`'s offset contract (see **Review fix** below) and is otherwise untouched.

Mutations applied to **copies** under a scratch tree, never the worktree. Each ran the migrated file **and** its `origin/main` predecessor over the same mutated tree, so every row is before-and-after.

| file | gut the clause | reflow 60/100 cols, no word changed | defect 1: gut + `\n   \n` + decoy below | defect 2: un-gutted copy above |
|---|---|---|---|---|
| `ci-state-prose` | new RED, old RED | see note | **old GREEN → new RED** | **old GREEN → new RED** |
| `liveness-rationale-prose` | new RED, old RED | both green | **old GREEN → new RED** | **old GREEN → new RED** |
| `quiet-payload-prose` | new RED, old RED | both green | **old GREEN → new RED** | **old GREEN → new RED** |
| `staleness-qualifier-prose` | new RED, old RED | both green | **old GREEN → new RED** | **old GREEN → new RED** |

In every gut and every defect row, the other three migrated pin files stayed green — the mutation reds its own pin and no other. One documented exception: the defect-2 mutation on `review-and-fix.md` duplicates step 6, and `quiet-payload-prose` anchors on that same block in that same document, so its uniqueness assert fires too. That is the guard working on a document that genuinely now has a duplicated anchor, not a spurious red.

**`ci-state-prose` reflow — stated, not hidden.** Its assertions are literal-space regexes (`/reads as \`pending\` in an aggregating checks summary/`), so a rewrap reds them. **Pre-existing and unchanged**: the same assertion reds identically on `origin/main`. The migration strictly *improves* reflow tolerance — at 60 cols `origin/main` additionally reds with `start anchor ... moved`, because its literal `indexOf` anchor broke on the wrap, while the migrated `phrase()` anchor survives. Two reds before, one after. Making those assertions reflow-safe is a separate change to what the pin guards and is not in this PR.

**`quiet-payload-prose` defect 1 needed the whole sentence removed**, not a partial gut. Its CLAUSE bound (`/`--quiet`[^.]*?(drops[^.]*)/`) takes the *first* `--quiet … drops` sentence in the slice, so a partially-gutted real sentence still matches first and reds regardless of the decoy — the clause bound pre-empts the paragraph bound. Removing the sentence outright lets the decoy be reachable, and then `origin/main` goes green (`tests 5 / fail 0`) while the migrated file reds (`tests 5 / fail 1`). Worth recording: for that site the paragraph bound is the second line of defence, not the first.

## Review fix

The review found that dropping `ci-state-prose`'s end anchor lost real coverage, and that the comment justifying the drop was wrong about *why*. Both corrected in `dfde972`.

**The defect.** Gut the pinned clause leaving its anchor, append the gutted text verbatim to the paragraph the end anchor used to open on, and delete the blank line between the two. `paragraph()` then slices through the merged paragraph and the decoy satisfies every assertion in the file — **green with the clause gutted**, on either document. Control, same gut and same decoy but the blank line **kept**: red, as it should be. So the paragraph merge, not the decoy, is what buys the green.

**The fix.** The end anchor comes back as its own check rather than as a bound, with the separating blank line inside the literal it asserts — which is the part `paragraph()` cannot see. Re-running the merge mutation against the fix reds that check and only that check, on either document; an unrelated paragraph inserted between the two does **not** red it; deleting the blank line alone does.

**The rationale.** The comment claiming the end anchor "bounded nothing it was asked to bound" was right about slice width and wrong about detection — width was never what the assert bought. Corrected in place, beside the corrected mechanism.

**Scope of what the shared bound closes.** The copied "shared bound, not a local copy" comment read as though the paragraph bound were now robust. Narrowed to name what it does close — a whitespace-bearing blank line and a duplicate anchor — and to say that a blank line **deleted outright** is a hole predating it. That hole is measured at `liveness-rationale-prose` and both `staleness-qualifier-prose` sites, **before and after this PR alike**, so it is not a regression here; the evidence is recorded on #1377, which already owns the `paragraph()` variant work.

**`anchorAt` coverage.** Three review dimensions noted the new export had no direct test, exercised only through consumers. One offset-contract case added to `prose-pin.test.mjs`: it reds when the export returns the match end, and stays green when `paragraph`'s blank-line bound is what breaks, so it pins the offset rather than the bound.

Suggestion-band findings that were checked and not applied are recorded on #1379 (closed).

## Suite

```
./agent-test plugin/scripts
# -> tests 2143 / pass 2143 / fail 0
```

2138 at filing time, plus the three that landed with #1370 in the rebase, plus the two the review fix adds.

## Boundary

**This diff does not touch `plugin/skills/run-team/SKILL.md`** — three of the four files pin it, and `impl-1371` is editing its Guards section concurrently.

```
git diff --stat origin/main...HEAD
# -> plugin/scripts/{ci-state,liveness-rationale,quiet-payload,staleness-qualifier}-prose.test.mjs,
#    plugin/scripts/prose-pin.mjs, plugin/scripts/prose-pin.test.mjs
#    — no document, no SKILL.md
```

Closes #1372


feigi:
Review deferrals filed, one line per destination:

- **#1377** (open, `ready-for-agent`) — commented, not filed new. `ledger.mjs check` scored it 0.50 against the subject, and it already owns the `paragraph()` variant work. The comment adds the measured **merged-paragraph** false green: deleting the blank line between two paragraphs merges the second into the slice, which item 1's "throws on a missing blank line" variant does not catch, because there is still a blank line — just a paragraph further down. Measured at `ci-state-prose`, `liveness-rationale-prose` and both `staleness-qualifier-prose` sites, before and after this PR alike.
- **#1379** (closed, `wontfix`) — the suggestion band, checked. Four entries under **Below the claim bar**: the `ci-state-prose` thunks, the rationale comment copied into four files, `anchorAt`'s `.source` re-flagging, and `liveness-rationale-prose`'s per-call `readFileSync`. Three are out of #1372's scope (pre-existing lines the migration relocated or left alone — each entry carries its `git show dcda49e:…` settling command). The fourth, the copied rationale, is in scope and was declined on precedent: #608 already refused that shape, and the only alternative is the cross-file pointer that refusal rejected.

Applied to the tree in `dfde972` rather than deferred: the dropped end anchor (correctness + tests + a silent-failure re-run), its false rationale comment, the over-claiming "shared bound" comment, and `anchorAt`'s missing offset test. Details in the **Review fix** section of the PR body.

Issue #951, opened 2026-08-27
board-cli.test.mjs's #363 gate is darwin-only and the test job is ubuntu-latest, so it never executes on CI — and #363 is closed, so it has no open home


feigi:
Deferred from PR #947 review. **Dimension: correctness.** Severity as reported: `suggestion`.

## The gap

`skills/fleet/scripts/board-cli.test.mjs:217` gates a test on `IS_DARWIN`:

```
{ skip: IS_DARWIN ? false : "async pipe-write loss (#363) is darwin-only — a green run here is not coverage" }
```

`.github/workflows/ci.yml`'s only test-running job (`check`) runs on `ubuntu-latest`. So this gate takes the skip branch on every CI run, and a skipped gate reports under `fail 0` exactly like one that ran — which is #337's premise verbatim, applied to a gate #337's own PR does not touch.

## Evidence

`grep -rn "skip:" skills/fleet/scripts/*.test.mjs | grep -v "skip: false"` over the tree returns exactly two conditional module-level gates: this one, and `candidates.test.mjs`'s gojq gate, which PR #947 fixes. `grep -rn "\.skip(" skills/fleet/scripts/*.test.mjs` shows the remainder are root-euid and python3 runtime skips, which are not CI-shaped. `python3 -c 'yaml.safe_load(...)'` over `ci.yml` confirms `check`/`ubuntu-latest` is the only job that runs tests.

## Why it has no home yet

The finding's own `suggested_fix` said "No change in this PR — it is #363's". **#363 is CLOSED** (`board.mjs: a single child emitting >64 KiB into a piped stderr may swallow die()'s refusal — the console.error half of the #246/#328 family`), so deferring this to #363 would strand it. Hence this issue.

## Recorded context

The controller separately accepted this exact gap when ruling on #337's escalation: a suite-wide `skipped 0` is unreachable on CI **by construction**, because this gate is platform-conditional and the job that runs the suite is not darwin. The acceptance criterion was therefore met on the per-file count for `candidates.test.mjs` instead, which goes from `skipped 2` to `skipped 0`. PR #947's body records the same reasoning.

Related but not a duplicate: #298 (CLOSED) also concerns skips and `candidates.test.mjs`, but is about two tests going red under concurrent suite runs, not about a platform gate that never executes on CI.

## Not a #947 change

[...]

PR #1568, opened 2026-09-17
test(scripts): executed byte-count coverage for die()/verdict()'s short-write retry loop

### LINKED ISSUE #1548: Add a deterministic byte-count test for die()/verdict()'s short-write resumption
Deferred from PR #1523 review (ticket #889).

**Dimension:** tests

**Finding:** `arg.mjs`'s `die()` short-write/EAGAIN-resume loop is pinned only by a source-shape regex (`candidates.test.mjs:1288`); nothing in the suite exercises it end to end and asserts the delivered byte count. `staleness.mjs`'s `verdict()` has the same gap on its own pin. A deterministic executed test is buildable at low cost: force fd 2 non-blocking with a `console.error("")` call inside a child process, then call `makeDie("probe")("x".repeat(200_000))` with a draining reader, and assert the reader received the full expected byte count (200,010 with the loop, 65,537 without it, on darwin — zero-variance over repeated runs, no pipe-stall harness or python dependency needed).

PR #1523 added two regression tests (`node --test --test-name-pattern "retry cap"` in `arg.test.mjs`/`staleness.test.mjs`) that pin the *bounded-time-under-permanent-stall* behavior added by that PR's retry cap. This is a different, complementary gap: a deterministic, always-reachable assertion that `die()`/`verdict()` actually deliver the **full** message across a genuine short write, independent of the shape-regex pin.

**Suggested remedy:** Add a test to `arg.test.mjs` beside the existing EBADF test (which already spawns a child copy of the module) that spawns a child forcing fd 2 non-blocking via `console.error("")`, calls `makeDie("probe")` with a message long enough to force at least one short write, and asserts the full byte count reaches a draining reader. Mirror it for `staleness.mjs`'s `verdict()` against fd 1.

Not applied in PR #1523: it is a net-new test-infrastructure addition (a scratch draining-parent harness) rather than a fix for a confirmed defect, and PR #1523 already closed the more severe gap (the retry cap having no regression coverage at all). Filed here so the byte-count coverage isn't lost.

feigi:
Closes #1548.

## What
Adds one deterministic, executed test each to `arg.test.mjs` (`die()`) and `staleness.test.mjs` (`verdict()`) that asserts the full expected byte count is delivered across a genuine short write, closing the gap the existing source-shape regex pin (`candidates.test.mjs`'s regex over `arg.mjs`'s own source, read through that file's `ARG_MODULE`) and the retry-cap timing test (#889) both leave open: neither can tell a working retry loop from a mutant that collapses it to one bare `writeSync` call.

## How
No python3/fcntl dependency, unlike the neighbouring retry-cap tests: initialising Node's own stream object for fd 2/1 is what puts a pipe fd into `O_NONBLOCK` — the same mechanism `ci-state.mjs`'s own `vlog` relies on. Once the fd is non-blocking, a payload past one pipe buffer short-writes under `spawnSync`'s own draining, exercising the retry loop for real.

- `die()`: drives `makeDie("probe")` with a 200,000-char message inside a child copy of `arg.mjs` (same idiom as the neighbouring EBADF/retry-cap tests) and asserts stderr is exactly the newline `console.error` printed, the filler that landed, then `die()`'s own message whole. The message travels as file content, so no argv limit applies.
- `verdict()`: `staleness.mjs` exports nothing importable, so a copy is run through a wrapper that prepares fd 1; a `--gone` needle is echoed verbatim into the JSON payload's own `needle` field, and the captured payload is round-tripped through `JSON.parse`, which throws on the truncated shape rather than pinning the `why` prose.

### The preconditions are not assumed
None of the preconditions this discrimination rests on is visible in the delivered bytes, so each is either asserted or measured and reported:

- **Fixture size.** A `PIPE_BUFFER_BYTES` guard, the way `ci-state.test.mjs` guards its own pipe fixtures. The guard covers the size of the write under test, not the bytes that arrived, so a truncated delivery names the defect instead of blaming the fixture (the trap `ci-state.test.mjs` documents above its own stderr-completeness pin).
- **Argv admissibility.** The needle reaches the script as one argv item, and Linux caps a single element at `MAX_ARG_STRLEN` (32 pages — 131072 bytes on the 4 KiB-page kernel CI runs), refusing the whole spawn with `E2BIG` past it, while darwin caps only the ~1 MiB total. So the needle has an upper bound it must respect, asserted, alongside the `PIPE_BUFFER_BYTES` lower bound above. `ledger.test.mjs` already carries this measurement and that constant name; this file reuses both rather than spelling it a second way. `git log -S` takes the same needle as its own argv item, so the cap binds the grandchild too. Clearing the lower bound turns out to be necessary but *not sufficient* to force a short write — see below.
- **Spawn actually happened.** A refused `execve` returns no `output`, so `r.stdout`/`r.stderr` are `undefined` and any diagnostic formatting them throws a `TypeError` over the top of the real cause. `r.error` is asserted first and names `E2BIG`.
- **That a single write is short at all**, which is the one that is *not* portable — and not for the reason it first appears. What one non-blocking `write()` to a pipe transfers is not the pipe's capacity: `spawnSync`'s reader drains concurrently and Linux's `pipe_write` keeps filling slots as the reader frees them, returning only once it finds the pipe full. So the ceiling is a **race against the reader**. Measured on CI: 146176 bytes in one call. `board-cli.test.mjs`'s own 100-run ubuntu measurement independently reports the same 146176 for one `console.error` write and a *range* of 146239–182783 for the looped build — the spread is the race. darwin instead returns at one buffer and short-writes reliably, which is exactly why this reads as portable from a darwin machine and is not.

  The variability is survivable; the magnitude is not. `146176 > 131072` (`ARG_STRLEN_MAX`), and the payload is grown through the needle, which is one argv item — so on that runner **no needle both outgrows a single write and survives `execve`**. The two constraints have no overlap, and no fixed size fixes it. The wrapper therefore forces the flag through the stream handle rather than inferring it from `console.log`, measures the ceiling, and the test dispositions on whether a usable size *exists*:

  | observed | meaning | outcome |
  | --- | --- | --- |
  | first write short | the loop resumed for real | delivery assertions are the kill — full rigor |
  | ceiling below `ARG_STRLEN_MAX` | a larger needle would short-write and still spawn | **red**, naming the size that would work |
  | ceiling at or above it | no admissible needle reaches the loop here | recorded via `t.diagnostic` with the numbers (#1578) |

  Only the middle row is fixable by editing this file, so it is the only one worth a red. The probe that reveals the ceiling is 2 MiB — past any capacity a kernel grants unprivileged, and an order of magnitude past the largest transfer this repo has measured — and it only runs when the first write was not short, so platforms that short-write pay nothing. The needle is 120_000, leaving margin under the cap, so every platform whose ceiling falls below that gets the assertion rather than the diagnostic. Asserting the last row is what reddened CI three times against a correct build; reporting it keeps the coverage hole visible, since a silent one is how #1548 came to exist. No platform gate and no skip: #951 removed this suite's last one after measuring that it hid real coverage rather than protecting a vacuous green. `arg.test.mjs` keeps its hard assertion, because fd 2 short-writes on both platforms by measurement.

### The `git log -S` branch is asserted too
`staleness.mjs` answers `unknown` at the same exit code with the needle echoed back whether the walk ran and found nothing or the walk failed outright, so "exercises the gone not-found path, not a stub" needed pinning rather than documenting. Only the empty-walk answer carries `found` and `bytes`; both are asserted, with `bytes` derived from the fixture's own `origin/main` blob rather than written as a literal.

Both fixture dirs are reaped through `t.after(rmSync)` the way `staleness.test.mjs` reaps its own repos. The `die()` fixture embeds the message, so it is the largest dir `arg.test.mjs` creates — measured at 232 KB against 200 KB for the next largest, which is why it gets the cleanup its unreaped neighbours in that file do not.

## Verification
Mutation-verified in scratch copies, every failure naming its own cause and no `TypeError` in any cell:

| mutation | result |
| --- | --- |
| unmutated control, both files | green |
| retry loop collapsed to a bare `writeSync` | reds the delivery assertion (`verdict()`'s as `JSON.parse` "Unterminated string in JSON") |
| collapsed loop **+ fixture shrunk** | reds the size guard — this combination passed before this review |
| fixture shrunk on honest code | reds the size guard |
| collapsed loop **+ uninitialised stream** (fd 2) | reds the first-write guard — this combination passed before this review |
| uninitialised stream on honest code (fd 2) | reds the first-write guard |
| `git log -S` forced to fail | reds the `found` assertion (actual `undefined`, expected `false`) |
| needle pushed back over `ARG_STRLEN_MAX` | reds the argv guard by name |
| spawn genuinely refused | reds the `r.error` guard reporting `E2BIG`, not a `TypeError` |
| **ceiling at or above the cap, fd blocking** (stream init and forcing both disabled) | green **+ diagnostic** naming the blocking fd — and green with the loop collapsed too, which is the honest outcome: that mutant is not discriminable there |
| **ceiling at or above the cap, fd non-blocking** — the real CI shape, reached locally by lowering the cap under darwin's own ceiling | green **+ diagnostic**, correctly naming the drain race rather than a blocking fd |
| **ceiling below the cap** | reds, reporting that a needle over 65536 would short-write and still spawn |
| `MAX_EAGAIN_RETRIES` lowered; trailing comment on each loop's resume line; filler byte changed | all green |

Temp residue measured before and after: pre-fix, each run left one fixture dir of each kind behind; post-fix, a run that creates and passes both leaves none.

Full suite, CI's own command: `node --test plugin/scripts/*.test.mjs` → `tests 2577 / pass 2577 / fail 0 / cancelled 0 / skipped 0 / todo 0`.

**Not verified here:** the Linux leg is not runnable from this machine, so the dispositions above were exercised by disabling or re-pointing the mechanism locally rather than on Linux. What Linux does is measured from the CI runs themselves — fd 2 short-writes, fd 1 does not, and its one-write ceiling is 146176 — rather than assumed, and it is corroborated independently by this repo's own 100-run ubuntu measurement in `board-cli.test.mjs`.

**Known residual, tracked in #1578:** on that runner `verdict()`'s retry loop is not executed and the collapsed-loop mutant is not caught there, and that is now known not to be fixable by sizing the fixture. The loop's executed coverage rests on the darwin leg, with fd 2's Linux coverage in `arg.test.mjs` as the cross-platform half. #1578 records the routes that remain.


feigi:
Review fixes applied. Deferral filing for this review: **#1575** — `PR #1568 review: the suggestion band, checked`, labelled `wontfix`, closed (the durable record; `check` reads the tracker at `--state all`).

What this run did with the review's output:

- **Applied** — the three `survived` findings: the missing fixture-size guard in both new tests, the unasserted `O_NONBLOCK` short-write precondition in both, and the comment that credited `candidates.mjs` with a source-shape pin that lives in `candidates.test.mjs` and matches `arg.mjs`.
- **Applied** — the two in-scope `suggestion`-band findings, after checking them: the `git log -S` branch discrimination (`found`/`bytes`) and the fixture temp-dir cleanup.
- **Recorded, no open work** — the three `suggestion`-band entries whose own claim is that the code is already correct (collapsed-loop discrimination, benign-mutation stability, whole-suite integration), the comments entry about measurement provenance (independently re-measured in this run), and one residual from the fixture-size remedy: its preference for a shared `PIPE_BUFFER_BYTES` over a per-file declaration. All five are in #1575 with `file:line`, dimension, claim and a re-open trigger.

No finding was deferred as unchecked, and no open issue was filed.

Verification is in the PR body and the fix commit's message: a mutation matrix in which every guard added here reds for its own cause, four benign mutations stay green, and `node --test` from the worktree reports `tests 2577 / pass 2577 / fail 0`.

Issue #1329, opened 2026-09-09
worktree-audit.sh:182 — `ahead` measures against the unqualified `$base`, so a shadowing tag reports a worktree carrying work as `ahead:0`


feigi:
Deferred from PR #1320 review (round 4). Dimension: `code-reviewer`.

`scripts/worktree-audit.sh:88` takes `base=${BASE_REF:-origin/main}` with **no accept-list at all**, and `:182` measures with it unqualified:

```sh
ahead=$(git -C "$wt" rev-list --count "$base"..HEAD 2>/dev/null)
```

`origin/main` is a shorthand. git resolves a shorthand through its own disambiguation order (gitrevisions), where `refs/tags/<name>` outranks `refs/remotes/<name>` — so a local tag named `origin/main` answers the measurement instead of the real upstream. Same class as #924 (reap.sh) and as the one PR #1320 just closed in `release-ticket.sh`.

**Measured** (git 2.50.1, darwin, real bare-origin fixture: claimed branch off `refs/remotes/origin/main`, one unpushed commit in the worktree, then `git tag origin/main refs/heads/fix/42-slug`):

```
ahead against refs/remotes/origin/main : 1
ahead against origin/main              : 0
worktree-audit.sh record               : {"branch":"fix/42-slug","ahead":0,"dirty":0,"dirtyFiles":[],"readable":true}
```

**Why it matters.** This script is not destructive — it has no `-D`, no `worktree remove`, no `--apply`. But its record is what a fleet controller reads to decide whether replacing a member would REDO work or DESTROY it (the script`s own comment at :172 says exactly that). `ahead:0, dirty:0` is the record for "this worktree carries nothing"; here it is emitted over a commit that exists nowhere else. The wrong answer is handed to a caller that acts destructively, rather than acted on here.

**Remedy** (the pattern PR #1320 landed in `release-ticket.sh`): canonicalize the ref that is MEASURED with, keep `$base` for user-facing text.

```sh
case "$base" in
  refs/remotes/*) base_rev=$base;;
  *) base_rev="refs/remotes/$base";;
esac
```

then use `$base_rev` at `:90` (`git rev-parse --verify`) and `:182` (the `rev-list --count`). Note this script has no accept-list either, so `BASE_REF=refs/heads/whatever` is currently accepted too — decide whether to add `release-ticket.sh`s accept-list here as well, or whether an audit-only script wants to stay permissive.

[...]

Issue #1249, opened 2026-09-05
PR #1242 review: the suggestion band, checked


feigi:
Adjudication of the `suggestion`-band findings from the review of PR #1242 (fix/727-worktree-guard-eacces-ancestor). The workflow gives this band 0 refuters by policy, so each in-scope finding was checked here with one dispatched refuter, biased to refuse, verifying by running something rather than reasoning. Findings that survived and were in-scope were applied directly in PR #1242 (commit 18a92b4): the ancestor-refusal wording fix, the missing worktree.sh guard test, and the redundant table-row removal.

One finding was out of this ticket's scope (#727 is specifically about claim-ticket.sh's worktree guard) and is below the "worth a claim" bar (its own claim is that correct code could be de-duplicated — a redundancy/simplification claim, never its own open issue per the filing rule).

## Below the claim bar

**file:** `skills/fleet/scripts/claim-ticket.sh:90`
**dimension:** simplify
**claim:** `claim-ticket.sh:90` hand-inlines `[ -e "$wt" ] || [ -L "$wt" ]`, which `release-ticket.sh:653` already ships as `occupied()`. The reviewer's suggested_fix is to move `occupied()` into `worktree.sh` beside `gone()` and delete the `release-ticket.sh` copy.
**verification on record:** the reviewer ran this refactor in a scratch copy and confirmed 309/309 + 26/26 tests green, shellcheck clean — the recipe is ready to reuse, not just a proposal.
**out of scope for PR #1242:** touches `release-ticket.sh`, which #727's stated scope (claim-ticket.sh's worktree guard) has no business editing.
**re-open trigger:** a future review returning this same finding as `survived` again, or a maintainer independently hitting the duplication, promotes it to an open issue — cite this record and the reviewer's already-verified refactor recipe (309/309 + 26/26 green, shellcheck clean) so it isn't re-derived from scratch.

feigi:
Below-the-claim-bar record for PR #1242's suggestion-band review; see body.

PR #1569, opened 2026-09-17
fix(no-undo-audit): canonicalize the origin/main shorthand for measurement (#1565)

### LINKED ISSUE #1565: no-undo-audit.sh measures ahead/behind against bare origin/main shorthand, same tag-shadowing class as #1329
Deferred from PR #1562 review (dimension: simplify, out-of-dimension heads-up).

`plugin/scripts/no-undo-audit.sh` still measures against the bare `${BASE_REF:-origin/main}` shorthand at 4 call sites, the same tag-shadowing class #1329 (fixed in PR #1562 for worktree-audit.sh) and #1320 (fixed for release-ticket.sh) describe:

- line 168: `base=${BASE_REF:-origin/main}`
- line 298: `git rev-parse --verify "$base"`
- line 638: `git merge-tree ... "$base" ...`
- line 722, 746: `git merge-base`/`git log ... "$base" ...`

A local ref (tag or branch) literally named `origin/main` would shadow the real remote-tracking branch per git's ref disambiguation order, giving a wrong verdict from a script whose whole job is proving a merge is safe/non-destructive.

**Suggested remedy**: apply the same qualification pattern as PR #1562/#1320 (canonicalize to `refs/remotes/$base` for measurement, keep `$base` unqualified for user-facing text), and decide whether to add an accept-list matching the sibling scripts.

Not applied in PR #1562 — that PR's ticket (#1329) scoped to worktree-audit.sh only.

feigi:
Fixes #1565.

`no-undo-audit.sh` measured ahead/behind against the bare `${BASE_REF:-origin/main}` shorthand at four call sites: `git rev-parse --verify "$base"` (existence check), `git merge-tree --write-tree --name-only -z "$base" "origin/$branch"` (the conflict probe), and `git merge-base "$base" "origin/$branch"` / `git log --oneline "$fork".."$base"` (the at-risk-commit listing). git resolves a shorthand through gitrevisions' disambiguation order (`refs/<name>`, `refs/tags/<name>`, `refs/heads/<name>`, `refs/remotes/<name>`, …), where `refs/remotes/origin/main` comes *last*. A local ref literally named `origin/main` — a tag or branch created for an unrelated reason — outranks the real remote-tracking branch, so every measurement against the shorthand answers about the wrong commit: a real add/add conflict against main's tip can read as no conflict at all, and main's real at-risk commit drops out of the list. git prints `warning: refname 'origin/main' is ambiguous.` to stderr at rc 0, where this script reads nothing, so the wrong answer arrives beside a warning nobody consumes rather than with nothing at all.

The same class is fixed in `release-ticket.sh:264` and `reap.sh:198`, which is where the accept-list and the qualify step are cited from — by `file:line`, the way `reap.sh:140` cites `release-ticket.sh`. `worktree-audit.sh` does **not** carry this fix: it still reads `BASE_REF` into a bare `$base` (`worktree-audit.sh:117`) and feeds it to `rev-parse --verify` (`:119`) and `rev-list --count` (`:211`) with no accept-list and no qualification. That half is tracked as #1329, which is closed with nothing in the tree to show for it.

### Fix

- Restrict `BASE_REF` to an `origin/*|refs/remotes/*` accept-list, then qualify to a new `$base_rev = refs/remotes/$base` (pass through unchanged if already qualified) and measure against `$base_rev` at all four sites. The two diagnostic trace lines that echo the command as it is about to run (`git merge-tree …`, `git log --oneline $fork..$base_rev …`) use `$base_rev` too, matching this script's own existing rule that those lines represent the command AS RUN.
- **Refuse a `BASE_REF` that names the audited branch.** The accept-list alone does not deliver a sound base: `origin/$branch` *is* a remote-tracking ref and clears it, and every measurement then asks whether the branch conflicts with itself — `conflicts: []`, `atRisk: []` at exit 0 on a worktree that really does add/add-conflict with the real `origin/main`. Reproduced on a bare-origin fixture, where the same tree and commit report `conflicts: ["conflict.txt"]` with one at-risk main commit when `BASE_REF` is unset. This carries `release-ticket.sh:239`'s second guard for the reason `release-ticket.sh:233` gives; `reap.sh:161` records why *it* can omit the guard — its `fetch --prune` drops the stale refs that make the same spelling vacuous — and this script fetches nothing. Both spellings of the branch are refused, and both are pinned.
- **`die` text names the two spellings the accept-list takes** rather than calling them "a remote-tracking ref". Measured: `refs/remotes/upstream/main` is accepted and `upstream/main` is refused, so the generic phrasing told a second remote's shorthand it was not something it is. The accept-list itself is unchanged — the message now states what the arm enforces.
- **Where the operator's spelling and the qualified ref can disagree, the `die` names both.** The rev-parse guard checks `$base_rev` while its message named `$base`, so a local ref shadowing the shorthand reported `origin/foo does not resolve` for an `origin/foo` that resolves by hand — the true fault (no `refs/remotes/…` ref by that name) misreported as the shorthand being unresolvable. The `merge-base` die had the identical mismatch. `$base` still stands alone everywhere the two cannot diverge, where the shorthand is what an operator expects to read.
- **Scope decision: accept-list included, not skipped.** `no-undo-audit.sh`'s own header says the merge bot leans on this audit's exit status to authorize a REBASE, so a wrong `BASE_REF` here does not just make a report wrong, it can feed a false "clean" into a decision that replays unpushed work. That is the blast radius `release-ticket.sh` restricted `BASE_REF` for, so this fix restricts it here too rather than leaving it as permissive as an audit-only reader could afford. Documented as a comment at the fix site.

### Docs

- The design-spec row (`docs/specs/2026-07-23-fleet-plugin-design.md`) and this script's derived exit-2 cause census (`no-undo-audit.test.mjs`'s `CAUSES`/`EXIT2_ENUMERATION`) carry both new `BASE_REF` refusals.
- `plugin/commands/run-merge-bot.md`'s exit-2 enumeration — the operator-facing doc for this script's only documented invocation — carries them too. It had gone stale against the row with the suite green, because nothing derived that list from anything.

### Test

Added a regression test for the shadowing class: a local tag literally named `origin/main`, pointing at the fork point rather than main's real tip, on a fixture where main's real tip carries an add/add conflict with the branch. Before the fix, the merge-tree probe reads no conflict and the at-risk list drops the real commit; after, both measure correctly against `refs/remotes/origin/main`. No equivalent test exists in `worktree-audit.test.mjs` to model it on — that script has neither the fix nor a test for the class.

Added `both operator-facing docs name every BASE_REF refusal this script can reach (#1565)`, which closes the gap that let the command doc drift: the clause list is asserted closed against the script's own `die` messages, so a third `BASE_REF` refusal reds until it is listed and named in both docs. Mutation-verified on a `git archive` copy — dropping the audited-branch clause from `run-merge-bot.md` turned that test red and only that test; adding a third `BASE_REF` `die` bound to an existing clause turned it red while the pre-existing row census stayed green, which is exactly why nothing caught this class before; and rewording a clause the pin does not claim left all four green.

Both audited-branch spellings are pinned as exit-2 preconditions. Two fixtures that audit their own checkout now name its feature branch rather than `main`: `BASE_REF` defaults to `origin/main`, so `main` there would be refused by the new guard before either shape reached the linkage question those tests exist to ask.

Also updated: two pre-existing `BASE_REF` fixtures whose values no longer clear the accept-list — the blob-dereference test (now uses a directly-written `refs/remotes/origin/blobtag` ref rather than a bare tag, to keep exercising "resolves to the wrong type" rather than "wrong shape"), and the "BASE_REF does not resolve" precondition case (now `origin/no-such-ref`, plus a case for the shape refusal itself).

Mutation-tested: reverting `$base_rev` back to `$base` at all four measurement sites in a scratch copy reds the shadowing test (`conflicts: []` instead of `['conflict.txt']`); restored and confirmed green.

Full suite: `node --test plugin/scripts/*.test.mjs` — 2577/2577 passing.

### Review deferrals

- #1576 — the **branch** side of every measurement is still the bare `origin/$branch` shorthand, so a tag by that name reproduces the identical false safe this PR removes for the base. Pre-existing and outside #1565's `BASE_REF` scope; reproduced against this branch's head.
- #1577 — the suggestion band, checked: what was applied without a separate issue, what sits below the claim bar, and the confirmations.


feigi:
## Review fixes applied — `c6f8c64`

Adjudication of the PR #1569 review. Every finding was applied, satisfied by another finding's fix, filed, or
recorded; none dropped.

### Applied (8)

**Survived findings (5) — all applied.**

| finding | dimension | what landed |
| --- | --- | --- |
| `no-undo-audit.sh:170` (critical) | comments | the false "worktree-audit.sh already carries this accept-list" claim is gone from all three comments and the test comment |
| `no-undo-audit.sh:169` | correctness | the accept-list is cited where it lives — `release-ticket.sh:229`/`reap.sh:156`, by `file:line` |
| `no-undo-audit.sh:184` | correctness | `release-ticket.sh`'s own-branch guard is carried; both spellings refused and pinned |
| `no-undo-audit.sh:340` | silent-failure | the rev-parse die names both spellings, so a shadowed shorthand is not reported as unresolvable |
| `run-merge-bot.md:290` | correctness | the operator doc's exit-2 enumeration carries both `BASE_REF` refusals, and a new pin asserts it does |

**Suggestion band, in scope and verified before applying (3).** `no-undo-audit.sh:186` (the accept-list die
asserted something false of its own accepted input), `:194` (the "silently, at rc 0" claim — git prints
`warning: refname ... is ambiguous.`), `:765` (the `merge-base` die carried the same mismatch as `:340`).

### Satisfied by an applied fix, not filed (2)

`no-undo-audit.sh:340`'s correctness restatement of the silent-failure defect, and the `#1320` ticket-number
attribution at `:170`. Both are closed by fixes above. Recorded in #1577 rather than filed, per the
in-scope-rediscovery rule.

Worth flagging: **#1320 is a merged PR titled "fix(fleet): point a claimed branch's upstream at its own ref, not
origin/main"** — a different fix. The accept-list citation was a misattribution, not merely unverifiable, so the
comment now cites `file:line` and the PR body no longer names #1320 at all.

### Deferred and filed (1)

- **#1576** (`ready-for-agent`) — the **branch** side of every measurement is still the bare `origin/$branch`
  shorthand. Out of #1565's scope and pre-existing, but reproduced against this branch's head: with a tag named
  `origin/fix/1-thing`, the audit reports `conflicts: []`, `atRisk: []` at rc 0 on the same fixture that reports
  the conflict and one at-risk commit without it.

### Recorded closed (12)

- **#1577** (`wontfix`, closed) — the suggestion band, checked. Carries the three dedup/simplify angles as one
  `Below the claim bar` entry with a re-open trigger (grouped on the controller's ruling; the ledger names #1249
  as the precedent for this shape), and the nine confirmations that claim no defect.

### Verification

- `node --test plugin/scripts/*.test.mjs` from the worktree: **2577 tests, 2577 pass, 0 fail** (baseline 2576;
  the added test is the new doc pin). Run before the commit, on the tree the commit contains.
- The own-branch false clean was reproduced pre-fix and confirmed refused post-fix, on a bare-origin add/add
  fixture.
- The added pin was mutation-verified on a `git archive` copy — never in the worktree, restored with `cp` and
  confirmed byte-identical. Dropping the audited-branch clause from `run-merge-bot.md` reds that test and only
  that test; adding a third `BASE_REF` `die` bound to an existing clause reds it while the pre-existing row
  census stays green (which is why this class went uncaught); rewording a clause the pin does not claim leaves
  every one of the four doc pins green.

### One thing a human should look at

`#1329` — "worktree-audit.sh:182 — `ahead` measures against the unqualified `$base`" — is **closed**, while
`worktree-audit.sh` still reads `BASE_REF` into a bare `$base` and feeds it to `rev-parse --verify` and
`rev-list --count` with no accept-list and no qualify step. The comment in this PR now says exactly that.
Whether #1329 should be re-opened is a triage call, not one this review pass made.

### Not done here

No refuter could be dispatched against the suggestion band: agent spawning was refused for the fix-applier
session (`Cannot spawn 'task'. Allowed: none`). The controller verified the three in-scope entries by running
commands directly and ruled the shaping entries out of scope; every applied claim was re-derived against the
tree before it was committed. #1576 and #1577 both say so in their bodies, so a resumed run can still put a
refuter on them rather than reading them as already attacked.

Issue #1573, opened 2026-09-18
merge-bot latency is queue depth x CI cycle, not poll interval - 48 min median measured


feigi:
Split out of #164, which settled the merge *gate*. This is the merge *latency* it exposed. Gated behind ADR 0007's guard 1 — filed now so the measurements do not evaporate, `onhold` until that guard fires.

## The complaint

PRs sit for a long time after they are signed off. That is real, and the first hypothesis — that `merge-bot` dispatch is poll-driven and a `ready-to-merge` webhook would fix it — is **wrong**. Measured 2026-09-18.

## Measurements

`ready-to-merge` label → `mergedAt`, last 25 merged PRs (timeline API per PR):

- **median 48 min**, range 8–95 min
- `ci.yml` success duration: **5–6 min** (8 runs at 5, 9 at 6)
- merge spacing inside one wave (#1527→#1530): 18:50:02, 18:57:01, 19:03:42, 19:10:07 — **~6.7 min apart**
- controller poll interval: `sleep 120` — **2 min**

So the wait decomposes as **queue depth × one serialized CI cycle**: 48 ÷ 6.7 ≈ 7 PRs deep, which matches the 7-PR run #164's own comment describes. Polling accounts for **≤4%** of the median. A label webhook is worth at most 2 minutes of 48, and `run-team/SKILL.md` already permits an edge-triggered *wake* — the prohibition at `## …Reconcile, do not wait for an event` is on edge-triggered *decisions* (`re-derive the deficit on every tick, whatever woke you`). So the wake is legal and also not the bottleneck.

## The actual cause

`merge-bot` is `target: 1` (`plugin/scripts/fleet-tick.mjs`), so merges are strictly serialized, and each one pays a full CI cycle. The levers are **parallelism** or **batching**, not notification.

## Why it is not simply "raise the cap"

ADR 0007 set `strict_required_status_checks_policy: true`. That is free at cap 1 — the bot is the only merger, so the candidate it just rebased is always at `main`'s tip. It is **not** free above cap 1: two bots rebase onto tip `T`, one merges producing `T'`, the other must rebase and re-run CI, so the ruleset reimposes the serialization with wasted CI on top.

[...]

Issue #68, opened 2026-07-30
review-and-fix mandates a CI gate in a repo with no CI — ci-state.mjs exits 2 and the standalone reviewer has no path to a verdict


feigi:
## Problem

`skills/fleet/commands/review-and-fix.md` spends its steps 4 and 6 on binding green to a run and gating the `ready-to-merge` label on the `check` job:

> **Standalone (no controller): watch** the run until the diff-validating jobs are green (bind via `ci-state.mjs` below)

and

> Gate it on the diff-validating **`check`** job, **not** on `ci-state --quiet` exit 0

`feigi/claude-config` has no `.github/workflows/`. PR #57 adds that fact to this very file ("`feigi/claude-config` has no compose file, no `globalSetup`, no vitest, and no CI"), so the document now states, twenty lines apart, both that CI must gate the label and that there is no CI.

## Reproduction

```
$ ~/.claude/skills/fleet/scripts/ci-state.mjs --pr 57
$ gh pr view 57 --json headRefName,headRefOid,state,mergeStateStatus
    branch=fix/43-reviewer-workflow-unreachable head=0edcda0 state=OPEN mergeState=CLEAN
ci-state: cannot read .github/workflows/ci.yml: ENOENT: no such file or directory, open '.github/workflows/ci.yml'
$ echo $?
2
```

```
$ gh pr checks 57
no checks reported on the 'fix/43-reviewer-workflow-unreachable' branch
```

## Why it matters

It **fails closed** — exit 2, loud — so this is not the silent-green class. The cost is that the mandated step has no success path in this repo: a standalone reviewer told to "watch the run until green" gets exit 2 forever, and the step-6 label gate reads per-job state from a payload that never exists. The current escape is ad-hoc — every reviewer independently decides that "no CI" means "skip the gate", which is exactly the kind of unstated judgment the file's own currency notes were written to remove.

## Why this is `needs-triage`, not `ready-for-agent`

Defect verified; *what to build* is not decided. Three approaches with real trade-offs:

1. **Add a CI workflow** — `node --test skills/fleet/scripts/*.test.mjs` on PR. Makes every existing instruction true as written and nothing in the file changes. Introduces CI cost and a required-check policy the repo has so far done without. Interacts with #55 (`agent-test` untracked, a fresh checkout cannot run the suite).
[...]

PR #1574, opened 2026-09-18
ci(ruleset): ratify main's merge gate, delete the refresh workflow (#164)

### LINKED ISSUE #164: make rebase-check a required status check on main — nothing enforces it today
Deferred from PR #157 review. This is the follow-up named in the PR body, filed so it does not evaporate with the PR.

`rebase-check` now exists and runs on every PR, but **nothing enforces it**:

```
$ gh api repos/feigi/claude-config/branches/main/protection
{"message":"Branch not protected","status":"404"}
```

Until `main` has branch protection with `rebase-check` as a required status check, the job is advisory: a red `rebase-check` does not block a merge, and the entire `rebase-check-refresh.yml` workflow — which exists to flip stale greens back to red when main advances — has no gate to feed.

This also affects how the fleet reads the board: #157's review found the refresh workflow's own header comment asserting "the branch protection still blocks the merge", which was false. That comment is corrected in #157, but the underlying gap is this issue.

The *what* is decided; the *how* is not, and it is partly outside an agent's reach:

- setting branch protection needs repo-admin scope, which the agent token may not carry
- the ruleset needs choices this review did not settle: required reviewers, whether `check` is also required, linear-history enforcement, and whether admins are exempt
- making `rebase-check` required while it is still advisory-new could strand open PRs immediately

Related: #111 and #68 both describe the fleet stranding PRs because this repo produced no CI signal. #157 supplies the signal; this issue supplies the enforcement.

feigi:
Closes #164. Splits #1573 out for the latency half.

## What #164 turned out to be

All three of its premises were stale, measured 2026-09-18:

| #164 says | measured |
|---|---|
| nothing enforces `rebase-check` | ruleset `main` (id 20119969) has required it since **2026-07-31**; three versions, none since |
| `…/branches/main/protection` → 404 proves no protection | that is the **classic** endpoint; it 404s for a repo governed by a **ruleset**. Read `…/rules/branches/main` |
| needs repo-admin scope the agent token may not carry | `rulesets`, the ruleset object, and its version history are all admin-only reads and all succeeded |
| the false header comment "is corrected in #157" | it was **not**; still on `main`, still citing the wrong endpoint, still forward-referencing #164 |

So the enforcement existed and the *record* did not — which is the one done-when item nobody had done.

## The gate is now a tracked file with an applier that proves the write

`.github/rulesets/main.json` holds the intended state; `.github/scripts/apply-ruleset.sh` resolves the ruleset **by name** (a recreated ruleset keeps its name, gets a new id), PUTs the file, then re-reads and refuses unless the live object matches.

That readback is not decoration: **two UI edits silently failed to save** while
this was being settled, detectable only because the ruleset's `updated_at` had
not moved. Measured idempotent against the spec as first written — apply, then
re-apply:

```
$ .github/scripts/apply-ruleset.sh
apply-ruleset: applied and verified — feigi/fleet-plugin ruleset 20119969 matches .github/rulesets/main.json
$ .github/scripts/apply-ruleset.sh
apply-ruleset: feigi/fleet-plugin ruleset 20119969 already matches .github/rulesets/main.json — nothing to apply
```

The spec has since gained four required contexts (below), which the live object
does not carry yet, so the next run of the applier is a real write rather than a
no-op. That divergence is exactly what the readback exists to make visible.

**ADR 0007** ratifies every field with its reason. Two are forced by code, not preference: `allowed_merge_methods: ["merge"]` and no linear-history rule, because `prove-merge.sh:166` dies `has no second parent — not a merge commit` and exits 2, and exit 2 means "could not evaluate", not "false" — so a squash leaves the merge bot with no proof at all.

Applied live with this change: `strict_required_status_checks_policy: true`, `required_review_thread_resolution: true`, and `require_extra_approval_for_unattributed_changes: false` — that last one was `true` against a **0**-approval floor, so most PRs needed nobody while some agent-authored PRs silently needed one human, which is the undiagnosable stall #111 and #68 describe.

## `rebase-check-refresh.yml` is deleted

`strict: true` refuses a behind PR at the merge button regardless of check colour, so the stale green the refresh existed to flip is inert. It failed **12 of its last 40 runs** on an unfixable race — `POST /actions/jobs/{id}/rerun` is refused with 403 *"The workflow run containing this job is already running"*, and its error text then blamed permissions or rate limits. At 33 merges/day that window is open almost continuously.

Gone: the workflow, `.github/scripts/rerun-rebase-check.sh`, its test, an `actions: write` grant, a 100-PR pagination cap — 1105 lines, measured off this PR's own diff: 315 of workflow YAML, 126 of shell, 664 of test. Four dependents re-pointed:

- `line-distance-prose.test.mjs` sentinel → `apply-ruleset.sh` (same directory, still a real file)
- `ledger.mjs` / `ledger.test.mjs` `GH_REPO` citation → `release-label.yml`, which exports it identically
- `unattended-git-sweep.test.mjs` gh-caller list → `apply-ruleset.sh`
- `ci-state.test.mjs` synthetic sibling fixture → `release-label.yml`
- `ci.yml`'s load-bearing-job-name NOTE → the coupling it described no longer exists

`rebase-check` stays required: it also asserts **no merge commits in `base..HEAD`**, which no ruleset field expresses, and it prints the actionable `git rebase` line a merge-button refusal does not.

## Two false claims corrected

- the refresh header's *"main is NOT a protected branch today … so nothing blocks the merge"* — deleted with the file
- `release.yml:41`'s *"branch protection bypassed?"* — `validate-release-label` was not a required context, so it had never gated the label. It is required now, and safe to require because it runs `if: always()` and never skips. `auto-label-bots` deliberately stays unrequired: it is `if:`-gated on bots and a required skipping check strands a PR forever
- `run-team/SKILL.md`'s `## Fix the tooling mid-run` now says a mid-run fix cannot be pushed to `main` directly, with the measured rejection

## Native merge queue: considered, rejected

The fleet's own `merge-bot` is the queue of record. GitHub's would need `merge_group` triggers no workflow here has; `validate-release-label` runs on `pull_request_target` and has no merge-group path; and entering the queue needs green required checks, so a behind PR with a red `rebase-check` could never be queued — the native queue and requiring `rebase-check` are mutually exclusive. Reasoning recorded in ADR 0007.

## Applied from this PR's own review (61b385e)

`apply-ruleset.sh` is the merge gate's only write path and had no executable
coverage; every defect below was reproduced under a stub `gh` before the fix and
is now pinned by `plugin/scripts/apply-ruleset.test.mjs`.

- **Resolve by name *and* `source_type == "Repository"`.** The listing carries
  inherited org/enterprise rulesets, which share the name space. A same-named
  parent made the applier refuse an unambiguous repo object; with only the
  parent present it aimed the PUT at an id this repo's token cannot write.
- **Guard every read with the script's own `die`.** The post-PUT readback was an
  unguarded substitution, so the one state that matters — write landed, cannot
  be proven — exited on `gh`'s raw status with nothing naming the unverified
  write. It now says so, in the script's own voice.
- **Stop guessing the cause.** A 401, a 422 and a 503 were each reported as
  "this needs repo-admin scope" or "run this inside the checkout", pointing an
  operator debugging the gate at the wrong fix.
- **Compare `rules` order-insensitively.** A server-side reorder turned a
  correct gate into a permanent exit 2. A genuine drift is still refused, at
  both top level and inside a rule's parameters.
- **Page the listing; pass the ruleset name to jq as data.** A quote in the name
  raised a jq syntax error that surfaced as a permissions failure. The filter
  also no longer runs under `gh --jq` (gojq), only under system `jq`.
- **Require `validate-claude`, `smoke-omp`, `npm-name-gate` and
  `install-and-smoke`.** ADR 0007's own test for leaving a context out is
  whether it can *skip* and strand the PR. Only `auto-label-bots` can: it is the
  sole `if:`-gated job among them, and it stays unrequired. Leaving the other
  four out contradicted both the ADR's "`check` is the only thing between a red
  tree and `main`" and `ci.yml:269`, which already calls `validate-claude`
  "Blocking, every PR".
- **Three factual corrections in ADR 0007.** Guard 2 described GitHub-release
  creation as a future event when `release.yml` already cuts it; the
  `held-behind-#<lower>` hold rule lives in `run-merge-bot.md`, not
  `pr-overlap.mjs`; the deletion is 1105 lines, not 315.

## Verification

- full suite **2582/2582 pass, 0 fail** (`node --test` from the worktree, at 61b385e)
- `shellcheck -x -S warning .github/scripts/apply-ruleset.sh` clean (CI's exact invocation)
- `jq empty .github/rulesets/main.json` valid (CI's exact invocation)
- live ruleset re-read: `strict: true`, `threads: true`, contexts `["rebase-check","check","validate-release-label"]`, `bypass_actors: []`, `allowed_merge_methods: ["merge"]` — the live object still carries the three contexts it always has; the four this PR adds to the spec land when the applier is next run

This PR is itself the evidence for the gate: it faces every context the live
ruleset requires, and a behind branch is now refused at the button rather than
by a check that might be stale.


feigi:
Review fixes pushed as 61b385e.

**Apply/defer split** — 8 `survived` findings, all applied; 15 `suggestion`-band findings, 7 applied, 8 recorded.

| band | count | disposition |
|---|---|---|
| survived | 8 | applied in 61b385e |
| suggestion, in scope, checked and survived | 7 | applied in 61b385e |
| suggestion, checked, nothing to apply | 8 | #1579 (closed, `wontfix`) |

No finding deferred as unchecked, and none deferred for a crashed refuter — every `unverified` entry was `refutersDispatched: 0`, the policy budget for the band, not a crash. Agent spawning was disabled for this fix-applier, so the one-refuter-per-in-scope-suggestion step was served by my own reproductions under `/tmp/pr1574-scratch/pr1574/`, each run synchronously with a positive control against a stub `gh` on PATH and the real script. #1579 records which, and how.

**Deferral issues filed:** #1579 only — the suggestion-band record, closed `wontfix` per the refutation-record rule. Nothing from this review is open work.

**Tests:** `node --test` from the worktree at 61b385e — `tests 2582 / pass 2582 / fail 0 / cancelled 0 / skipped 0 / todo 0`. 2562 before, plus the 20 in the new `plugin/scripts/apply-ruleset.test.mjs`. CI's own `shellcheck -x -S warning` and `jq empty` invocations over the tracked file sets are clean.

**One live action left for a human:** the spec now declares four required contexts the live ruleset does not carry. `.github/scripts/apply-ruleset.sh` has not been run against the live repo by this review — running it is a real write to the merge gate, which is the maintainer's call, not a reviewer's.

PR #1580, opened 2026-09-18
fix(scripts): require a salvaged ci-state payload to parse before it counts as a verdict

### LINKED ISSUE #875: board.mjs runCiState: the exit-code gate accepts a truncated or signal-killed read as a CI verdict, discarding the prevCi carry-forward
**Dimension:** silent-failure (folds a second silent-failure finding from the same review) · **Deferred from PR #868 review**

`skills/fleet/scripts/board.mjs:125`

## Scope — read this first

The finding as originally claimed blamed `maxBuffer`. **That trigger is refuted and measured unreachable**, and this issue does not rest on it. A residual survives the refutation and is reachable today.

**Refuted half (do not re-file):** the child here is `ci-state.mjs --pr N --quiet`, whose verdict payload is one line of JSON — `--quiet` drops `jobs` and `missing`, and verbose tracing goes to stderr, which `maxBuffer` accounts separately. A refuter measured two structural ceilings well below the default cap: `ci-state.mjs`'s own uncapped `gh` read dies first (`ENOBUFS` at ~5000 jobs), and a realistic worst case (256 jobs, 100-char names) is 33,786 bytes. Reaching 64 MiB of stdout is not reachable in practice.

## The residual — the gate mishandles any non-2 exit, not just a size kill

```js
} catch (e) {
  const out = e.stdout ? e.stdout.toString() : "";
  if (e.status !== 2 && out.trim()) return out;
```

`e.status !== 2` is true for a signal kill (`status` is `null`) and for any nonzero exit that is not 2. If `e.stdout` carries any non-whitespace bytes, those bytes are returned **as a CI verdict**. `mapCi` then fails to parse them and answers `"unknown"`, and because the return was non-null, `gather()`'s `prevCi` carry-forward — which only a `null` return reaches — is skipped. A PR's last-known-good CI state is overwritten.

That is precisely the #262 regression the comment above the function says the exit-code gate exists to prevent.

## It is reachable today — measured

`ci-state.mjs` writes its verdict with `console.log(JSON.stringify(payload))` followed by `process.exit(...)`. On a pipe, the queued remainder is discarded at one pipe buffer. Driving the real payload path with a stub shaped like `ci-state.mjs`:

```
{ "threw": true, "status": 1, "signal": null,
  "stdoutBytes": 65536, "gateReturnsThisAsVerdict": true, "parses": false }
```

Exactly 65536 bytes of truncated, unparseable JSON, at exit 1, accepted by the gate as a verdict. No signal, no `maxBuffer`, nothing #868 touches. The upstream half of this loop is filed separately — see the companion issue on `ci-state.mjs` truncating its own verdict payload.

## Suggested remedy (open — not settled, never run)

The gate needs to separate "the child answered" from "the child was cut off". Candidates, unranked and unverified: require the salvaged stdout to parse before returning it; or treat any `status` that is `null` or otherwise indicates abnormal termination the way exit 2 is treated. Either way the fallback must reach `null` so the `prevCi` carry-forward fires.

## Why deferred

The gate predates PR #868 — that PR only adds `READ_OPTS` to this call. Fixing it changes CI-verdict semantics for the whole board, which is not #807's buffer fix.

feigi:
Closes #875

Sizing: heavy — the ticket's own remedy section is "open — not settled, never run", offering two unranked, unverified candidates, and picking between them changes CI-verdict semantics for the whole board. That ambiguity in *what to build* is the heavy-row signal; file count (2 source files + 1 sibling comment) would have said light on its own. Run at `next-ticket` step 6, before any code was written and before this PR was opened.

## Re-derived against `origin/main` first

Both claims hold. The gate had drifted from the ticket's `:125` to `plugin/scripts/board.mjs:248` (`if (e.status !== 2 && out.trim()) return out;`), and `gather()`'s carry-forward is still reached only by a `null` return (`ci[p.number] = out === null ? (prevCi.get(p.number) ?? "unknown") : mapCi(out, p.number)`). The refuted `maxBuffer` half is not re-filed and nothing here rests on it. The companion upstream issue — `ci-state.mjs` truncating its own payload — is separately **already fixed** on main: `emit()` now loops on `writeSync`'s return. That narrows the producer's door but does not close it (`emit()` still abandons a non-EAGAIN short write, keeping the exit code intact), and it does not touch the consumer residual this ticket is about.

## Measured before choosing a remedy

Four stubs shaped like `ci-state.mjs`, driven through the real `execFileSync` call:

| stub | status | signal | today's gate returns it as a verdict | parses |
|---|---|---|---|---|
| exit 1, cut mid-JSON | 1 | null | yes | no |
| killed mid-payload | null | SIGKILL | yes | no |
| exit 1, complete not-green verdict | 1 | null | yes | yes |
| exit 1, complete no-ci verdict (`status: null`) | 1 | null | yes | yes |

The gate accepts all four. `parses` separates them exactly — which decides between the ticket's two candidates:

- **Chosen: the salvaged bytes must parse.** Catches both cut-off rows, keeps both real verdicts.
- **Rejected: treat abnormal termination like exit 2.** It misses the ticket's own measured case (exit 1, *no signal*), and tightened far enough to catch it — demanding a usable `status` — it starts refusing row 4: `no-ci` is a real verdict that shares exit 1 (this call never passes `--declare-no-ci`) and its `status` field is null. That would resurrect a stale red for a PR with no CI run behind it.

So both halves now hold: a non-2 exit **and** bytes that parse. Wholeness is the only question asked in the gate; shape stays `mapCi`'s, which is total over every payload that parses.

## The bug class — covered vs. deliberately left

From the ticket's prose, then the mechanism:

- signal kill, `status === null`, partial payload — **covered** (test + own diagnostic naming the signal)
- exit 1 with a write cut mid-JSON — **covered** (the ticket's measured case)
- exit 1 with a non-JSON line ahead of the payload — **covered** (same predicate; it is the test's vehicle)
- any other nonzero non-2 exit carrying non-whitespace stdout — **covered** (the predicate is exit-code-agnostic)
- `maxBuffer`/ENOBUFS kill — **covered, and now named.** Reached the carry-forward incidentally from the start (it arrives as `status: null` + truncated stdout), but the review reproduced it against this diff and the diagnostic misattributed it: `execFileSync` throws `status: null`, `signal: "SIGTERM"`, `code: "ENOBUFS"`, `message: "spawnSync node ENOBUFS"` when a child overruns `READ_OPTS.maxBuffer`, so a signal-first `how` billed this file's own 64 MB cap to an outside killer. Corrected in 231b4c1 — see the follow-up section.
- spawn failure (ENOENT): `e.stdout` undefined → already `null` today, not in the class
- exit 2 with a payload (#262) — unchanged, still `null` by the existing arm
- **left, deliberately:** exit 0 with unparseable stdout. `runCiState()` returns a successful read's stdout unconditionally and `mapCi` answers "unknown" with a warn, no carry-forward. That is the other side of the seam, pinned by its own tests and by #605's empty-payload row; moving it would rewrite `mapCi`'s contract, not this gate's. Now tracked on its own as #1593 rather than only commented, so the surviving half of the class has a ticket.
- **left, deliberately:** the producer-side short-write abandon in `ci-state.mjs`'s `emit()` — the companion issue's territory.

The whole risk of a stricter gate is that it eats a real verdict and hands the board a stale value. Three accept-side rows guard it, and all three were green before the fix and stayed green after:

- `exit 1 is a verdict, not a failed read` (pre-existing) — complete not-green payload, prev `green`, must read `red`.
- **new:** `a complete no-ci verdict at exit 1 is an answer, not a failed read` — `status: null`, prev `red`, must read `unknown`. This is the row that fails under the rejected remedy.
- `a rate-limited ci-state — exit 2 WITH a payload` (pre-existing) — the #262 arm must keep carrying forward.

## What my edit newly does

It newly returns `null` where it used to return bytes, so it newly *reaches* the carry-forward. Two of the ticket's own criteria could break on that restructuring, and both are tested:

1. **"exit 1 must still beat the previous value"** — the inverted first condition (`e.status === 2 || !out.trim()`) is the old predicate negated; the parse only gates what survives it. Pinned by the two accept-side rows above.
2. **#605's wiring pin** — `mapCi`'s warn keys on a PR number `mapCi` has no other use for, so the argument exists only if `gather()` passes it, and the only end-to-end driver was an unparseable payload at **exit 1** — a payload that now legitimately takes the carry-forward and never reaches `mapCi`. Left alone, that contract would have quietly lost its only driver. Moved to **exit 0**, the one arm that still hands `mapCi` an unparseable payload, so the pin keeps a live and reachable vehicle. Its `ci === "unknown"` assertion is kept there, where it is still true, rather than re-pinned to the new behaviour.

**Could the suite even see the mode?** Yes, but only through `gatherCi` in `board.test.mjs`, which injects `scriptDir` and drives the real `runCiState()` out of process — `runCiState` is not exported, so no in-process test can reach it. Verified the hard way: the four new rows were written first and the two refuse-side ones went red (`pass 93 / fail 2`) against unmodified `board.mjs`, then green after the fix.

## Diagnostics

The refusal gets its own stderr line rather than reusing `ci-state --pr N failed`: "failed" reads as *the child never answered*, where an operator looking at a carried-forward value needs to know a payload **did** arrive and was thrown away, with the parse error and what became of the child. After 231b4c1 that last part reads `e.code` first — the three-way `ci-state.mjs` already uses at its own child reads — falling back to the signal and then the exit status, so node's own `maxBuffer` enforcement names itself instead of posing as an external kill; `status` is null on a real signal, so "exit null" would name nothing actionable and the signal arm stays ahead of the exit arm. The line also routes through this file's warn-once gate on its own `ci-salvage` channel, keyed on the PR, so a truncation whose cause persists costs one line per PR rather than one per `serve` tick.

Two prose claims elsewhere described the gate as it was and are retired in the second commit: `board.test.mjs`'s "at any exit but 2, non-empty stdout is a real verdict" (now names the exit-0 arm) and `board-prev-shape.test.mjs`'s "the ONLY way into the carry-forward arm" (no longer true; the comment now says what its stub is actually chosen for).

## Verification

- `node plugin/scripts/board.test.mjs` — 95 pass / 0 fail (93/2 before the fix, the two new refuse-side rows red)
- `node --test plugin/scripts/board-prev-shape.test.mjs plugin/scripts/board-tryparse-null.test.mjs` — 21 pass / 0 fail
- `node --test` (full, after rebase onto `origin/main`) — **2589 pass / 0 fail**

## Review follow-up (231b4c1)

The automated review of this PR returned two confirmed defects in the salvage arm and a `suggestion` band. Both defects are fixed in 231b4c1, on top of the two commits above; the band was checked and its outcome recorded.

Applied:

- The refusal line went out through a bare `console.error`, where the sibling unparseable-payload class in `mapCi` has been warn-once gated per PR since it was written. `serve()` re-gathers on a timer in one process, so a truncation whose cause persists cost a line on every tick. Now gated, on its own `ci-salvage` channel: for any one payload the two arms are mutually exclusive — bytes refused in `runCiState` return `null` and never reach `mapCi` — so a shared channel would buy nothing and would let whichever arm a PR happened to take silence the other for the rest of the run.
- `how` was signal-first, which billed `READ_OPTS.maxBuffer` to an outside killer. Now `e.code ?? (e.signal ? ... : ...)`, matching `plugin/scripts/ci-state.mjs:155` and `:659`.
- A salvage-arm comment recorded that the pre-#875 gate "accepted all four rows in this block". Re-running that gate against this file's own stubs, the rows it accepted are the not-green and no-ci verdicts plus the two cut-off ones — and not-green is defined outside the labelled block, while the exit-0 constant that *is* inside it never reaches the gate, since the gate lives in the catch and a child that exits 0 never throws. The comment now names the rows.

New pin: `gather: a refused salvage payload warns ONCE per PR across ticks, and a second PR is not masked`. It drives `gather()` across ticks for two PRs in one process, the only shape where a per-process line and a per-tick line differ; `gatherCi` grew `prs` and `ticks` for it, both defaulting to what every other arm already needs. Mutation-verified against the committed tree: reverting the gate to `console.error` made it the only failure in `board.test.mjs` + `board-prev-shape.test.mjs` (`pass 113 / fail 1`, observing one line per PR per tick), keying the gate on its channel alone with an empty key made it the only failure again (observing the second PR masked), and renaming the channel — behaviour-preserving — left the run at `pass 114 / fail 0`.

Deferred, and filed: #1593 (the exit-0 arm, `ready-for-agent`) and #1594 (the `suggestion` band's refutations and below-the-bar entries, `wontfix`, closed as the durable record).

The `maxBuffer` ENOBUFS path is proven by a standalone probe rather than a permanent test: `READ_OPTS` is a module-level constant with no injection seam, so pinning it in the suite would mean pushing more than 64 MB through a child on every run — a real cost for a diagnostic string. The signal arm of the same expression stays pinned by the existing SIGKILL row, which passes only because `e.code` is absent there.

Verification at 231b4c1, from the worktree:

- `node --test plugin/scripts/board.test.mjs` — 96 pass / 0 fail
- `node --test` (full) — **2590 pass / 0 fail**, exit 0


feigi:
## Review findings: apply/defer split

Pushed **231b4c1**. Both `survived` findings applied; the `suggestion` band checked and split.

The band is budgeted no refuters by policy, so every entry arrived unchecked. Agent spawning was unavailable in the fix-applier session, and the controller authorized doing the refutation directly under the same discipline — run something, biased to refuse. Every claim restated below was settled by a command against this tree, not read off a finding.

### Applied — 231b4c1

| finding | file:line | dimension | how it was settled |
|---|---|---|---|
| Refusal line not warn-once gated, so it floods per `serve` tick | `board.mjs:299` | `correctness` | `survived`, two refuters; routed through `warnOnce("ci-salvage", pr, …)` and pinned by a new across-ticks test |
| Diagnostic drops `e.code`, so an ENOBUFS `maxBuffer` overrun reads as `killed by SIGTERM` | `board.mjs:298` | `silent-failure` | `survived`, two refuters; re-measured `status: null` / `signal: SIGTERM` / `code: ENOBUFS` / 65536 bytes buffered / parse fails |
| Same `e.code` defect, restated in the `suggestion` band | `board.mjs:298` | `correctness` | same defect as the row above, already double-refuted; landed with it |
| Salvage-arm comment's `this block` antecedent does not hold the rows it names | `board.test.mjs:214` | `comments` | refutation attempted and **failed** — see below; reworded to name the rows |

The comment refutation that failed: the attempt was that the labelled block does hold four payload constants, so no reword is needed. Re-running the pre-#875 gate against this file's own stubs kills it — the rows the gate accepted are `NOT_GREEN_EXIT_1` and `NO_CI_EXIT_1` (both parse) and `UNPARSEABLE_EXIT_1` and `KILLED_MID_PAYLOAD` (neither parses), while `UNPARSEABLE_EXIT_0`, the constant inside the labelled block, never reaches the gate at all (`threw: false`; the gate lives in the catch). So the sentence's rows span a row defined outside that block, and the block's own membership cannot satisfy "the two real verdicts".

### Deferred and filed

- **#1593** — `ready-for-agent`, open. `board.mjs:266`, dimension `correctness`: the exit-0 arm still discards an unparseable payload, replacing the PR's last-known CI with `unknown`. Refused as out of scope — #875's residual is written against the catch block — so it was never checked against a remedy, which makes it real deferred work rather than a refutation. The defect is confirmed by this tree's own passing pin at `board.test.mjs:327`; the remedy is open between refusing in `runCiState` and fixing `emit()`'s short-write abandon, and the issue says so.
- **#1594** — `wontfix`, closed as the durable record. Carries the band's refutations with their measurements, and its below-the-claim-bar entries with a re-open trigger each. Closed rather than open because refutations left open read as queued agent work; `ledger.mjs check` reads the tracker at `--state all`, so a later run re-deriving one of these lands on the evidence instead of re-litigating it.

Refuted by measurement, recorded in #1594:

- Composing `TRUNCATED_WRITE` into `KILLED_MID_PAYLOAD` (`board.test.mjs:280`, `simplify`) — they are not two spellings of one fixture. Run and read: one puts a `warning:` line ahead of the JSON (57 bytes, preamble present), the other writes a bare truncated object (26 bytes, no preamble). The composition makes them byte-identical, deleting the bare shape from the file — and silently, since the targeted run stayed at `tests 1 / pass 1 / fail 0` either way. Both `board.mjs` and `board.test.mjs` name the preamble as its own route into `mapCi`.
- "This PR had to rewrite both rationale copies in the same commit" (`board.mjs:346`, `simplify`) — refuted as stated. `git log -L` attributes the `board.mjs` copy to `b75d19e` and the `board.test.mjs` copy to `3bcc603`; the evidence behind the claim was the aggregate PR diff, which cannot tell one commit from two. The redundancy it also observes is real and is recorded below the bar, measured: bag overlap 0.73, longest verbatim shared run 23 words, texts not identical.

Below the claim bar, recorded in #1594 with re-open triggers: the validate-only parse that `mapCi` then repeats (`board.mjs:296`, `silent-failure`), the `string | null` return that collapses a hard failure and a garbled salvage (`board.mjs:264`, `types`), and the rationale-copy overlap above. Three further band entries claimed no defect at all — the gate's own correctness, the new tests' discrimination, and the repo being untyped JavaScript — and are recorded so the band is accounted for end to end.

### Test result

`node --test` from the worktree at 231b4c1: **`tests 2590`, `pass 2590`, `fail 0`, `skipped 0`**, exit 0. `node --test plugin/scripts/board.test.mjs`: 96 pass / 0 fail.

The new pin is mutation-verified against the committed tree, each injection proved to have landed before its result was read: reverting the gate to `console.error` made it the only failure (`pass 113 / fail 1`, observing one line per PR per tick); keying the gate on its channel alone made it the only failure again (observing the second PR masked); renaming the channel — behaviour-preserving — left the run at `pass 114 / fail 0`, so it discriminates rather than merely fails.

Issue #1349, opened 2026-09-09
Port the PR review to omp: shared core, two shims, fleet-owned specialists (per #1303)


feigi:
Part of #1292. Ruled on #1303.

**Type:** task (AFK)

## Question

Port the PR review to omp: one shared core, two thin host shims, and six fleet-owned specialist definitions that replace the vendored `pr-review-toolkit:*` agents on **both** harnesses.

## What the ruling fixed

- **Controller-run on both harnesses.** Claude: `Workflow({name:"fleetctl:review-pr", …})` as today. omp: `eval` loading the core from the Install root through the Resolver (`fleet-run --path workflows/review-pr.js` or equivalent) — never a literal path. Members do not run it on either harness.
- **Shared core + shims.** Core: schemas (`:28-124`), phase logic, every pure function. Claude shim: the Workflow sandbox (its globals; `Date`/`Math.random`/`import` restrictions at `:389-430`, `:912-916`). omp shim: `agent()` with `.wait()`, `phase`, `log`, `args`.
- **Re-run, no replay cache.** `resumeFor` reports which findings' refuters all crashed; on Claude it still points at `resumeFromRunId`; on omp it says re-run. Reopen only on a measured cost trigger.
- **No per-call tier.** No `agent()` call passes `model`/`effort` (measured: silently dropped on omp). Every dispatch names a fleet-owned definition carrying `effort:` + `thinking-level:` (#1302) and a bare-alias `model:`. An audit test asserts the absence of per-call tier in the core.
- **Fleet-owned specialists.** `agents/fleet-review-correctness`, `-silent-failure`, `-tests`, `-comments`, `-types`, `-simplify` (`.agent.md`), plus `fleet-review-snapshot` (haiku tier) and `fleet-review-verifier` (low thinking) for the other two call sites. Prompts move in from the vendored definitions; tiers per the current per-dimension `model` choices (`:154-185`). Names `fleet-`-prefixed: collision-free on omp's bare-name registry, `fleetctl:fleet-review-…` on Claude.

## Acceptance criteria

- [ ] The same core file runs under both shims against one fixture PR and produces findings validated by the same schemas — shown once per harness.
- [ ] On omp, each specialist dispatch resolves to its declared family and `thinking-level` (session file / job record), for all eight definitions — the #1298 layer-2 check passes.
- [ ] `grep -n 'pr-review-toolkit' workflows/ skills/ commands/` returns nothing; the six vendored names are never dispatched.
[...]

PR #1183, opened 2026-09-02
docs(specs): replace stale file:line citations with quoted-phrase anchors

feigi:
## Summary

Two bare `file:line` citations in `docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md` had drifted from the code they point at:

- `review-pr.js:284-286` (cited for the "tests 0" convention) now lands mid the "BOUND EVERY READ" paragraph — the convention itself now lives at `review-pr.js:1061` (`grep -n "'tests 0' is a FAILED run, not a pass" workflows/review-pr.js`).
- `run-team/SKILL.md:382` (cited for the finisher's duty summary) now lands in the tail of the test-path expansion paragraph — the quoted duty text now lives at `SKILL.md:1409` (`grep -n "a fresh small agent, not the fix-applier resumed" skills/fleet/skills/run-team/SKILL.md`).

Per the issue's ruling, the fix is a stable anchor, not a corrected number — a `:NNN` in a design doc rots on the next edit to the cited file.

**The two hunks do not get the same treatment, because their quoted text is not equally usable as an anchor.**

- Finisher hunk: the doc's quoted duty text is byte-identical to the source, so the quote itself is the anchor and the parenthetical says so — `` (`run-team/SKILL.md`, same quoted words) ``.

  `grep -c -F 'a fresh small agent, not the fix-applier resumed' skills/fleet/skills/run-team/SKILL.md` → `1`

- "tests 0" hunk: the doc quotes the rule with markdown backticks (`` `tests 0` ``) while the source quotes it with single quotes (`'tests 0'`), so the doc's own bold sentence matches **nothing** in the file it points at and cannot serve as the anchor.

  `grep -c -F '`tests 0` is a FAILED run, not a pass' workflows/review-pr.js` → `0`, exit 1 (also `0` after flattening line wraps)

  So the parenthetical names the rule by a fragment whose bytes **are** shared with the source, and resolves on its own — `` (`review-pr.js`, the "is a FAILED run, not a pass" rule) ``.

  `grep -c -F 'is a FAILED run, not a pass' workflows/review-pr.js` → `1` (`workflows/review-pr.js:1061`, the only `FAILED run` site in the file)

Writing `same quoted words` on the second hunk too would have been a false claim in a `class=correction` diff, which is why the two parentheticals read differently.

Sizing: light

## Scope

[...]

Issue #1130, opened 2026-09-01
review-pr.js line citations in two docs rotted by PR #1126's +10 readRules insert — both notes' substance still holds, only the pointers


feigi:
Deferred from PR #1126 review (issue #532). Dimensions: **comments** (both halves).

Two line-number citations, in two documents PR #1126 does not touch, both rotted by exactly the +10 lines that PR's `readRules` hunk (`@@ -258,6 +258,16 @@`) inserts above them. Same class, same cause, filed once.

## The two citations

1. `docs/specs/2026-08-27-fleet-member-outcomes-instrumentation-design.md:521` — "Open since the #211 spec; `review-pr.js:364-370` still carries the UNVERIFIED note". Measured at PR head `30e4548`: lines 364-370 of `workflows/review-pr.js` are the `const scratch` / `explicitDimensions` / `verifiers` / `snapshotModel` defaults block. `grep -n UNVERIFIED workflows/review-pr.js` finds the note at **375**, in a comment block spanning 373-379.

2. `.out-of-scope/prose-rule-as-derivation.md:35` — "Re-measured when this record was written: `workflows/review-pr.js:404` ... now carry[ies] it only as an explicit negation". Measured: `grep -n "NOT because those four alone face refuters" workflows/review-pr.js` returns **414**.

Reverse-applying the PR's `review-pr.js` hunks in an isolated scratch tree puts the UNVERIFIED block back at 363-369, confirming the +10 shift is this diff's.

## What is NOT wrong

The **substance** of both notes still holds. The UNVERIFIED comment is still there and still unresolved; the retired prose rule still survives only as an explicit negation. Only the pointers rot. That is why the review's own refuter for (1) voted refuted, and why this is filed rather than applied.

**This diff worsened an existing drift rather than creating it**: citation (1) was already off by one before PR #1126 — pre-PR the block sat at 363-369, not the cited 364-370.

## Remedy is open

Two options, and the choice is not settled here:
- Touch up the numbers (372-379 and 414), which rots again on the next line-shifting PR; or
- Switch both to the content-based citation style `.out-of-scope/prose-rule-as-derivation.md` itself recommends one paragraph up ("**Cited by content, not by line**: both blocks have already moved once"), and which `skills/fleet/scripts/review-pr-citation-prose.test.mjs` enforces for `review-pr.js`'s own citations after #317 measured two of them rotting silently.

The second is the repo's stated direction but costs a pin per citation. Whoever takes this decides.

[...]

PR #1581, opened 2026-09-18
docs(specs): re-anchor the 2026-08-06 spec's nine review-pr.js citations

### LINKED ISSUE #1188: docs/specs/2026-08-06 review-pr spec: bare review-pr.js:NNN citations remain, at least one stale — #516 excludes .md and #117's sweep never reached this doc
Deferred from PR #1183 review. `dimension: correctness`.

`docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md` — the bare `review-pr.js:NNN` citations PR #1183 did **not** touch.

## Why this is deferred rather than fixed in PR #1183

PR #1183 closes #654, which rules on exactly two citations by name. It re-anchored those two. Every other bare `file:line` citation in the same doc is untouched and out of that ticket's scope.

## The defect class is live, not hypothetical

The remaining sites, at the PR head:

```
$ git show HEAD:docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md \
    | grep -noE 'review-pr\.js:[0-9]+(-[0-9]+)?'
77:review-pr.js:169
116:review-pr.js:31
117:review-pr.js:304
122:review-pr.js:207
171:review-pr.js:106
194:review-pr.js:148
317:review-pr.js:138
348:review-pr.js:175
439:review-pr.js:145
```

**At least one is stale.** Doc line 122 cites `review-pr.js:207` for `git archive HEAD`. That construct is not there:

```
$ grep -n 'git archive HEAD' workflows/review-pr.js
240:// `git diff` (code-reviewer.md:21). The snapshot is `git archive HEAD | tar -x`
303:The snapshot IS the source of truth: 'git archive HEAD', byte-identical to

$ sed -n '205,209p' workflows/review-pr.js
// perfectly good diff.
//
// A MISSING prHead is deliberately not disqualifying: `gh pr view` can fail
// while `gh pr diff` succeeded, and dropping a good diff over an absent
// cross-check would let missing input narrow coverage — the inversion the
```

`:207` is the prHead comment — real, plausible prose about the same subsystem, which is the quietly-misleading shape rather than the obviously-broken one.

**At least one is accurate.** Doc line 116 cites `review-pr.js:31` for the `dimension` requirement, and line 31 sits inside the construct it names:

```
$ grep -n 'FINDINGS_SCHEMA' workflows/review-pr.js | head -1
28:const FINDINGS_SCHEMA = {
```

**So the set is mixed, and a batch shift would break the citations that are currently right.** Resolve each against the text it names — the same lesson #317 recorded as "not a uniform off-by-four". The rest of the list above is **unmeasured**: this issue asserts the property (the set contains both stale and accurate members), not a tally of which.

## Why no existing ticket owns this

- **#516 (OPEN)** explicitly excludes it. Its own scope note: *"the 20 covers `skills/fleet/**` only … It misses citations in `.md` files entirely (specs and skills prose), which #117 swept separately."* So #516 rules `.md` out of scope on the grounds that a separate ticket handled it.
- **#117 (CLOSED)** is that separate sweep — *"the remaining bare `:NNN` citations are unverified, and no cross-`docs/specs` sweep has been run"*. It is closed, and `review-pr.js:207` is still stale in `docs/specs/` today.

**The gap between those two is this ticket:** #516 hands `.md` to #117, #117 is closed, and this doc was not reached.

- **#317 (CLOSED)** is a different subject — `review-and-fix.md:49` citations inside `.js`/`.mjs`.

## The open question this inherits

#654 ruled that the fix for a rotting citation is a stable anchor, not a corrected number. Whether these specs should track code line numbers **at all** is the same unsettled question #516 is asking in its own scope, and #516 has not been answered. A triage pass should settle the approach before an implementer picks this up:

- re-anchor every site on a quoted fragment + construct, or
- re-point only the sites measurement shows wrong (#129's policy), or
- rule that a dated design spec is a historical record and its citations are not maintained at all.

Note nothing enforces the convention mechanically — `.github/workflows/ci.yml` runs syntax checks and `node --test`, no doc lint — so this rots silently either way.

`needs-triage`: the defect is confirmed at `:207`, but the remedy inherits #654's and #516's unanswered scope question, so the approach is what needs deciding first.

Deferred from PR #1183 review.

feigi:
Closes #1188

Sizing: light — the signal is the sizing table's light row in full: the Agent Brief names the one document, the exact remedy form (quoted-phrase anchors, `grep -c -F` → 1), and the two PR #1183 hunks to match, so no design choice is open and the change lands in one doc plus one existing pin file. Run before the first edit, and it is the `## Agent Brief` comment that closed the approach question the raw issue body left open (`needs-triage` was on the remedy, not on the defect). Category is bug, so `systematic-debugging` ran first as the table requires on either row: all eleven numbers were resolved against the current tree before anything was rewritten.

## The defect, measured

#654/PR #1183 ruled on two of this document's citations — one into `review-pr.js`, one into `run-team/SKILL.md` — and re-anchored exactly those. The nine remaining `review-pr.js:NNN` refs were left bare. Read against the current tree (which is how this document's citations actually get read, whatever the front matter says about resolving them at `8a84402`), old number → what sits on that line today in `plugin/workflows/review-pr.js`:

| cited | for | lands on today |
|---|---|---|
| `:31` | `FINDINGS_SCHEMA` requires `dimension` | its `required` array's own comment — **still accurate** |
| `:106` | the `simplify` prompt | the sibling `evidence` field's "What was RUN" description |
| `:138` | `testCmd`'s default | the `agentType` threading comment |
| `:145` | the `args.dimensions` override path | the `fleet-review-*` fork comment |
| `:148` | beside `verifierEffort` | same comment, "this file's PREVIOUS per-call override (removed below)" |
| `:169` | `selectDimensions` | `DEFAULT_DIMENSIONS`' `silent-failure` `agentType` |
| `:175` | the empty-profile widening | `DEFAULT_DIMENSIONS`' `tests` `agentType` |
| `:207` | `git archive HEAD` | the prHead/headRefOid cross-check |
| `:261` | the override path | a `readRules` comment on why a diff was dropped |
| `:304` | `dimension` on findings | a `readRules` comment on the 0/absent diff-line count |
| `:331` | `dimension` on findings | `skew`, inside `readRules`' rejected-diff template |

`:31` is still right, so the batch shift the issue warned against would have broken it. `:148` is the quietly-misleading shape this class is named for: it lands on a comment that really does discuss a removed per-call model override — just not the construct cited.

## The fix

Each site is resolved against the construct it NAMES, in PR #1183's form: `` (`review-pr.js`, "fragment") ``. Every anchor verified `grep -c -F <fragment> plugin/workflows/review-pr.js` → 1, and every one kept contiguous on a single markdown line so a reader can copy it out of the raw file and grep it — 4323514's lesson, where a parenthetical quoted its rule in backticks while the source used single quotes, so the doc's own phrase matched nothing in the file it pointed at.

Two cited constructs changed shape, so the sentence moved with the anchor rather than merely re-pointing:

- **`testCmd`** is no longer defaulted at all (the ticket's own acceptance criterion). `resolveTestCmd` derives it from the repo under review and refuses the review when it cannot — "never fall back to a guess" (#142) — so "already defaulted" was false about the mechanism, not just the line.
- **`verifierEffort`/`specialistModel`** are gone outright (#1349). The citation now points at the comment that records the removal, "are GONE, not renamed", so the pointer resolves even though what it named does not exist. The dated claim above it stands as the record of `8a84402`.

## The regression pin

The issue notes nothing enforces this convention mechanically. `review-pr-inbound-citation-prose.test.mjs` exists because #1130 measured this exact class, and its own ceiling note read "a third document citing review-pr.js by line is not caught here" — this is that document: the ban covers it, and every anchor is asserted quoted by the doc, contiguous on one line, and occurring exactly once in the source. The ceiling note now names the documents it covers as a class rather than counting them, since this PR had to bump that count to write it.

Mutations run, each red alone and green on revert:

| mutation | result |
|---|---|
| pre-fix document restored | ban red (9 hits) **and** anchors red |
| one anchor re-wrapped across two lines | contiguity red, flattened-quote assertion still green — the two are independent |
| `never fall back to a guess` reworded in `review-pr.js` | uniqueness red, 0 occurrences |
| ban pattern given a `$` so it matches nothing | the control red **while the corpus-driven ban stayed green** — the reason the control exists |

## Scope

Out, deliberately, and pinned as acceptable rather than silently folded in: every `<file>:NNN` citation this same document keeps into a file other than `review-pr.js`, and the bare `:NNN` self-references in its Edits changelist, which record what was edited at the base commit rather than pointing a reader at the current tree. The front-matter `git show "8a84402:<path>"` paragraph stays — those out-of-scope citations still need it, and each of them was measured to resolve at `8a84402` under its pre-rename path. That the paragraph now claims a set the `review-pr.js` anchors have left, with no note saying so, is filed as #1595. #516's repo-wide question is untouched; this only continues the #654/PR #1183 precedent for this document.

Named as a class above rather than listed, because the list #1188's out-of-scope section wrote is no longer the set. Measured at this head:

```
$ grep -noE '[A-Za-z0-9_/.-]+\.(md|mjs|js|json|yml|sh):[0-9]+(-[0-9]+)?' \
    docs/specs/2026-08-06-review-pr-cost-and-apply-policy-design.md | grep -v 'review-pr\.js:'
108:review-and-fix.md:7    109:review-and-fix.md:10    138:review-and-fix.md:8
207:diff-stats.mjs:76-82   321:run-team/SKILL.md:350   349:run-team/SKILL.md:287
355:run-team/SKILL.md:337-338                          411:review-path-default.test.mjs:87
```

Against #1188's list, `run-team/SKILL.md:350-355` and `diff-stats.mjs:105` are not citations this document carries in that form, and `review-path-default.test.mjs:87` is one it does and the list omits. Which is the reason the pin's ceiling note names the class instead: any enumeration of this set, here or in a comment, is a partial inventory a reader takes for the whole.

## Review fixes applied (`182e619`)

The automated review returned two `survived` findings and six in the `suggestion` band. Spawning was disabled for the fix-applier, so the controller authorized it to run the suggestion-band refutations itself, biased to refuse; the full record is #1596 (closed) and #1595 (open).

**The anchor set covered only the citations this ticket converted.** The same document carries two `review-pr.js` citations that never had a line number and so were never in #1188's sweep — the `"is a FAILED run, not a pass"` rule and the comment above `DEFAULT_DIMENSIONS`. Being content-anchored already is not being checked: rewording either one out of its target left the pin file at pass 7 fail 0, so the rot this PR exists to stop was still reachable in the document the PR was written for.

**The obvious fix does not close it, and running it is what showed that.** Adding the FAILED-run rule as an anchor row left that mutation *green*: the document carries the phrase twice — once as the citation into `review-pr.js` (line 358), once restated in its own voice on the next line — and a row is satisfied by its fragment wherever the document happens to carry it. So the two halves are now held apart. The row holds the source half, the fragment occurring exactly once in `review-pr.js`. A new test holds the citing half, the form the document spells the citation in, occurring exactly once. `DEFAULT_DIMENSIONS`'s citation quotes no fragment a row could compare at all, so `commentAbove` holds what it does resolve through: the declaration's uniqueness and the comment's presence.

Mutations, on a scratch export, each injection verified as landed and reverted by clean copy before the next:

| mutation | before | after |
|---|---|---|
| the FAILED-run **citation** reworded (line 358) | green | **red** — citation occurs 0 times |
| the `DEFAULT_DIMENSIONS` citation repointed | green | **red** — citation occurs 0 times |
| `is a FAILED run, not a pass` reworded in the source | — | **red** — occurs 0 times, not once |
| the comment above `const DEFAULT_DIMENSIONS = [` deleted | — | **red** — "nothing but code sits above" |
| the citation form duplicated in the doc | — | **red** — occurs 2 times, not once |
| the doc's **restatement** reworded, citation intact | — | green — the citation still resolves |
| that comment reworded but still present | — | green — negative control |
| an unrelated comment added to the source | — | green — negative control |

**Two comment corrections.** The landing-spot example named `test_run`'s description, and no converted number lands there — `:106` is the sibling `evidence` field's description, and `test_run`'s own sits at a line this document never cited. The table above carried the same slip and is corrected with it. Separately, the ceiling note enumerated the sibling `<file>:NNN` citations the document keeps, naming three of the eight it carries, so a partial inventory read as the whole of what is uncovered; it names the class now.

## Tests

`node --test plugin/scripts/*.test.mjs` → tests 2589 / pass 2589 / fail 0 at `182e619`, run from the worktree. Nothing pinned the old citation text. The pre-review head measured 2588 / 2588 / 0 on the same command; the one added test is the citing-side pin described above.


feigi:
Review deferrals filed (review-and-fix step 5), dedupe-checked through `ledger.mjs check` — both came back `found: false` with only closed near-misses, and the rows were read rather than trusted to the exit code:

- **#1595** — OPEN, `ready-for-agent`. The front-matter `8a84402`-versus-HEAD resolution split. Confirmed defect, remedy open, out of this ticket's scope by #1188's own acceptance criterion ("No other prose in the document is altered beyond the nine citations and the `testCmd` sentence"). Includes the measurement the review could not make — every remaining line-numbered citation does resolve at `8a84402` under its pre-rename path, so the paragraph is load-bearing and needs scoping, not deleting.
- **#1596** — CLOSED `wontfix`. The suggestion band, checked: two entries applied in `182e619` with their mutation evidence, one refuted with measurements, four verification records that allege no defect, and one residual below the claim bar with a re-open trigger. Also records a correction to the review's own arithmetic — the diff removed nine `review-pr.js:NNN` citations, not eleven; `:331` and `:261` are bare continuation refs on the following line.

Nothing was dropped: two `survived` + six `suggestion` = three applied, one refuted, one deferred open, three recorded.

Issue #1578, opened 2026-09-18
staleness.test.mjs: verdict()'s retry loop is not exercised on Linux CI — one write there outruns the largest needle execve accepts


feigi:
## What

`plugin/scripts/staleness.test.mjs`, the `verdict() resumes from a genuine short write and delivers the full payload` test (#1548, PR #1568).

That test exists to prove `verdict()`'s `writeSync` retry loop is executed — that a mutant collapsing the loop to one bare `writeSync(1, buf)` is caught. It can only do that if the first write to fd 1 comes back SHORT, which requires fd 1 to be non-blocking. **On this repo's Linux CI runner, fd 1 does not short-write**, so on CI that test most likely passes while reporting — through `t.diagnostic` — that it did not exercise the loop at all.

The test is green and correct. The gap is that on CI it is not currently pinning what it was written to pin, and the loop's executed coverage rests on the darwin leg alone.

## Measured, not inferred

Across two CI runs (`35323757791` and `35325624714`, the latter on `25cc16c`):

- `die()`'s companion fixture in `arg.test.mjs`, which does the same thing on **fd 2** via `console.error("")`, short-writes and its hard first-write assertion PASSES on Linux.
- The `verdict()` fixture on **fd 1** via `console.log("")` does not: `fd 1 took all 100000 bytes in one write ... (first write returned 100000)`.
- darwin short-writes both.

So this is fd-1-specific on Linux, not a platform-wide property of stdio pipes — a roomier pipe or a non-pipe stdout would have hit fd 2 identically, and fd 2 passed both times. That rules out pipe capacity and "stdout isn't a pipe here" as causes.

## Current state in the tree

`63a0c81` does two things, neither of which closes the hole:

1. The wrapper **forces** the flag through the stream handle (`process.stdout._handle.setBlocking(false)`) instead of inferring it from `console.log`'s side effect, and records whether that call was available.
2. When the first write is not short, the wrapper issues one further write past any pipe capacity a kernel grants unprivileged (2 MiB — Linux's `fs.pipe-max-size` ceiling defaults to 1 MiB, so a non-blocking pipe cannot swallow it) and the test dispositions the result:
   - first write short → the loop resumed for real; the delivery assertions are the kill.
   - one call, 2 MiB probe short → the fd IS non-blocking and the pipe merely holds more than the payload, i.e. the fixture is too small → **red**, naming that.
[...]

PR #1682, opened 2026-09-21
PROBE (do not merge) — #1578 Linux mutation check

feigi:
Throwaway verification probe for #1578. Deliberately collapses writeAll()'s retry loop to one bare writeSync to confirm PR #1681's fixture RED**s on ubuntu-24.04, not just darwin. Will be closed and the branch deleted as soon as the check job reports. Do not merge, do not label.


feigi:
Probe complete. Confirmed on ubuntu-24.04 (run 35657414996): with writeAll()'s retry loop collapsed, #1681's fixture reds with `verdict() delivered 4096 bytes — the 4096-byte window and nothing after it — so the retry loop did not resume`. That is #1578's coverage hole closed on Linux, measured rather than reasoned. Throwaway branch, never for merge.

